PR #8, opened 2016-03-09
[WIP] unify with cifetch

runcom:
carry https://github.com/projectatomic/skopeo/pull/2

TODO:
- add smoke tests under integration/ for the layer command
- unify cifetch - layers command (take into account a possible v1)
- **decide if we're going to support v2 only**
- do not vendor docker,engine-api from master


giuseppe:
I have done some refactoring here:

https://github.com/projectatomic/skopeo/pull/9

PR #13, opened 2016-03-19
Add a new "test-style" Makefile target

mtrmac:
I am not certain how the various `test-`* commands (with `test-local` to come) are best to be exposed and used, but at least having `Makefile` targets is better than fishing out the commands from my shell history.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -372,16 +372,16 @@ func getDefaultConfigDir(confPath string) string {
 	return filepath.Join(homedir.Get(), confPath)
 }
 
-type DockerAuthConfigObsolete struct {
+type dockerAuthConfigObsolete struct {
 	Auth string `json:"auth"`
 }
 
-type DockerAuthConfig struct {
+type dockerAuthConfig struct {
 	Auth string `json:"auth,omitempty"`
 }
 
-type DockerConfigFile struct {
-	AuthConfigs map[string]DockerAuthConfig `json:"auths"`
+type dockerConfigFile struct {
+	AuthConfigs map[string]dockerAuthConfig `json:"auths"`
 }
 
 func decodeDockerAuth(s string) (string, string, error) {
@@ -408,7 +408,7 @@ func getAuth(hostname string) (string, string, error) {
 		if err != nil {
 			return "", "", err
 		}
-		var dockerAuth DockerConfigFile
+		var dockerAuth dockerConfigFile
 		if err := json.Unmarshal(j, &dockerAuth); err != nil {
 			return "", "", err
 		}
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -51,6 +51,17 @@ man:
 shell: build-container
 	$(DOCKER_RUN_DOCKER) bash
 
+test-style: test-fmt test-vet test-lint
+
+test-fmt:
+	test -z "$$(gofmt -s -l . | grep -v ^vendor/ | tee /dev/stderr)"
```
runcom: While I like these, isn't it better to call the corresponding shell files in the container? Like for test-integration?

mtrmac:
Oops, _now_ I have noticed `make validate`.

I’d still prefer having a `Makefile` target which runs locally without a Docker container and `sudo`, but this might need to be reworked to avoid this duplication.

`make test-lint` does warn about more things than `hack/make.sh validate-lint`, so this work at least has not been completely useless.

Anyway, I suppose this is now more of a tracked RFE than a mergeable PR.

Makefile
mtrmac: See more in the top-level conversation; I didn't notice the `validate` target.

Still, the difference between ~7 seconds for `sudo make validate` (and _much_ more for an initial uncached run) and ~1.5 seconds for a non-container run is pretty noticeable.

Or perhaps we should have IDE configuration running the checks locally during development, and `Makefile` targets for release compilation and end users? But then which IDE?

(I would _so_ love to have all of this figured out by some build system specialist instead of reinventing/cargo-culting for every project again.)

runcom: Honestly the idea here is the same as the docker upstream repo which in more than a year worked flawlessly for me (apart from the first not coached run). My vim ide is already configured to run at least lint and this is per developer so it's not up for this repository. 
If it's better for you at least, we can add those local makefile targets.

runcom:
Btw something is wrong with the shell scripts cause it isn't caching what you fixed here. Will investigate.

Makefile
mtrmac: Nah, personal quirks have no place in a shared project.

I'll primarily figure out the cause of the `golint` results difference, and perhaps submit a VS Code configuration (which I would be actually using all the time, instead of the Makefile targets).

runcom:
Also, I think you can run the validate script locally with just `./hack/make.sh validate` IIRC

Makefile
runcom: I found the differences, will fix this. Thx Miloslav

runcom:
@mtrmac can you edit this PR and leave just the golint fixes? I'll merge it afterwards

runcom:
@mtrmac I've included lint errors fixes in #15, feel free to either fix this PR or create an issue to track what we discussed here

mtrmac:
@runcom Thanks; I have dropped the `Makefile` changes (in favor of #19), and I will close this after #15 is merged just because that way it is easier for me to verify that the changes have been included.

PR #15, opened 2016-03-21
drop docker/ code

runcom:


rhatdan:
Love removing that much code.

LGTM.

runcom:
I'll leave this open while I fix everything but yes dropping alot of code

runcom:
Pretty much done with this.

The major change is that we don't use docker/docker code and the image manifest we print has now well defined values (take a look at the manifest type)

@mtrmac would you have a fast look at this?

runcom:
@rhatdan also

integration/check_test.go
```diff
@@ -84,7 +84,7 @@ func (s *SkopeoSuite) TestVersion(c *check.C) {
 func (s *SkopeoSuite) TestCanAuthToPrivateRegistryV2WithoutDockerCfg(c *check.C) {
 	out, err := exec.Command(skopeoBinary, "--docker-cfg=''", "--username="+s.regV2WithAuth.username, "--password="+s.regV2WithAuth.password, "inspect", fmt.Sprintf("docker://%s/busybox:latest", s.regV2WithAuth.url)).CombinedOutput()
 	c.Assert(err, check.NotNil, check.Commentf(string(out)))
-	wanted := "Error: image busybox not found"
+	wanted := "Invalid status code returned when fetching manifest 401"
 	if !strings.Contains(string(out), wanted) {
```
rhatdan: Shouldn't this just be a constant defined externally?

runcom: not sure about this, those error strings aren't part of the API and are just returned in one place. I don't want to use typed errors in tests cause otherwise if I have a bug in the error type I won't catch it if using the error type itself

rhatdan: Fine, I was saying that if you change this again in the future you need to change 4 lines.  As upposed to changing it one place in the test tool.  I don't think this should be external to the test module.

runcom: Oh gotcha, will do it

runcom:
This is good to go :) and lots of code removed!

PR #16, opened 2016-03-21
Support for in-process command testing

mtrmac:
Instead of each command handler terminating the process, let it return
an error object, and terminate the process at the top level.

See https://github.com/mtrmac/skopeo/commit/3ccea6cdeaf7e375c50a907b7a709da2f184a5ae for usage.

This will need to be extended to also capture stdout/stderr.

### DIFF
--- a/cmd/skopeo/inspect.go
+++ b/cmd/skopeo/inspect.go
@@ -43,7 +43,7 @@ var inspectCmd = cli.Command{
 			return err
 		}
 		if c.Bool("raw") {
-			fmt.Println(string(rawManifest))
+			fmt.Fprintln(c.App.Writer, string(rawManifest))
 			return nil
 		}
 		imgInspect, err := img.Inspect()
@@ -77,7 +77,7 @@ var inspectCmd = cli.Command{
 		if err != nil {
 			return err
 		}
-		fmt.Println(string(out))
+		fmt.Fprintln(c.App.Writer, string(out))
 		return nil
 	},
 }
--- a/cmd/skopeo/main.go
+++ b/cmd/skopeo/main.go
@@ -17,7 +17,8 @@ const (
 	usage = `interact with registries`
 )
 
-func main() {
+// createApp returns a cli.App to be run or tested.
+func createApp() *cli.App {
 	app := cli.NewApp()
 	app.Name = "skopeo"
 	if gitCommit != "" {
@@ -67,6 +68,11 @@ func main() {
 		standaloneSignCmd,
 		standaloneVerifyCmd,
 	}
+	return app
+}
+
+func main() {
+	app := createApp()
 	if err := app.Run(os.Args); err != nil {
 		logrus.Fatal(err)
 	}
--- a/cmd/skopeo/main_test.go
+++ b/cmd/skopeo/main_test.go
@@ -1,0 +1,14 @@
+package main
+
+import "bytes"
+
+// runSkopeo creates an app object and runs it with args, with an implied first "skopeo".
+// Returns output intended for stdout and the returned error, if any.
+func runSkopeo(args ...string) (string, error) {
+	app := createApp()
+	stdout := bytes.Buffer{}
+	app.Writer = &stdout
+	args = append([]string{"skopeo"}, args...)
+	err := app.Run(args)
+	return stdout.String(), err
+}
--- a/cmd/skopeo/signing.go
+++ b/cmd/skopeo/signing.go
@@ -70,7 +70,6 @@ func standaloneVerify(context *cli.Context) error {
 
 	mech, err := signature.NewGPGSigningMechanism()
 	if err != nil {
-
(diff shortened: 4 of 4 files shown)


mtrmac:
From an IRC discussion:

- ATM it is easy enough to split the handlers (e.g. `standaloneSign`) into a `cli.*` glue and an unit-testable subroutine, without wrapping `cli` the way this PR does.
- Ultimately most commands will need to use the integration testing framework because of external dependencies (which is independent on depending on `cli.*`).
- Will defer more discussion about merging and this approach after writing more code and tests with such external dependencies, when the needs will be better understood.

mtrmac:
Updated to capture stdout, and separate the infrastructure commit from the commit migrating the existing `layers` and `inspect` commands.

(Merging decision still deferred on getting more experience.)

runcom:
Still unsure about all this, I'm testing out https://github.com/sstephenson/bats to run these kind of tests you're adding (that are, to me, integration tests). With bats it should be really easy and fast to run tests even without a dev container (I really don't want a wrapper around cli atm)

mtrmac:
I am not sure either.

To me, quick and non-destructive things like https://github.com/projectatomic/skopeo/pull/26/files#diff-c778645be79a73d9345e93b6e8e3575aR59 belong in the unit test concept, and having proper Go tests and coverage computed and the like is nice; and testing this necessarily requires some kind redirection of `os.Stdout`, `logrus.Fatal`, `os.Exit` and the like.

ATM I just want to track the code with the detailed discussion and trade-off description, better than leaving an anonymous branch in my fork.

runcom:
> To me, quick and non-destructive things like https://github.com/projectatomic/skopeo/pull/26/files#diff-c778645be79a73d9345e93b6e8e3575aR59 belong in the unit test concept

sorry but I do not agree this is a unit test in the proper sense of unit testing. running a cmd can be seen as a unit test but, even if it doesn't use external stuff, it's an integration test to me. Just because it's wrapped in a function and you're testing input and output doesn't mean it's unit test concept.
That test could be basically run directly from the command line. I feel those wrapper are redunant, even if they speed up "integration cli", there should be a better way not to clutter each command, I found bats for instance.

runcom:
https://github.com/projectatomic/skopeo/pull/26/files#diff-c778645be79a73d9345e93b6e8e3575aR30

just the comment here makes me feel this is better suited for an integration test suite.

I don't want to bike-shed long for this. I do not like those wrapper around the commands (maybe just because I'm my own experience with golang and *cli I've never had to do such things, see runc and docker). I don't think those other projects lost something by not unit-test their cmd - indeed I think their test coverage is great (both unit and integration), except for speed and test containers which I understand it's an issue for you.

We can merge this as is, but I'll look forward to find a fastest and better way to run those instead of wrapping them

runcom:
https://github.com/projectatomic/skopeo/pull/16/files#diff-c6c318da062a6b9637745a480d10434eR68
Having such a function in main is also weird. Again, I don't see any added value in wrapping those instead of working out to find a good integration-test framework which is fast and stable. What I said above apply though.

mtrmac:
> except for speed and test containers which I understand it's an issue for you.

… and coverage _reporting_. You’re right that integration tests can cover this easily enough, but out-of-process tests won’t allow me to see that I haven't tested something.

Let’s not bikeshed about this now :-) I will gain a bit more experience with the more difficult cases and it is pretty likely I will change my mind.

> https://github.com/projectatomic/skopeo/pull/16/files#diff-c6c318da062a6b9637745a480d10434eR68
Having such a function in main is also weird.

Oh, right, that belongs in a `*_test` file of course. Will fix this, thanks for pointing this out.

runcom:
> … and coverage reporting. You’re right that integration tests can cover this easily enough, but out-of-process tests won’t allow me to see that I haven't tested something.

Coverage is easily accomplished by unit test functions used in the main commands, isn't it? cmd line flags, arguments and such are also covered when using unit test against functions called in cmds

mtrmac:
And that is exactly what this patch set allows: unit testing the handling of `cli.Context.Args()`. It seems to me we are really only disagreeing in how the `cli.Context` mock or almost-mock will be attached.

(I will grant you that `app.Wrap` is really ugly, but that is because `cli` is not helping with handling of errors or the redirection at all. https://github.com/spf13/cobra already has the command handlers returning errors and a “direct output here” member of the context — but then it requires similar ugly dances to make option values member variables instead of globals, so it is a wash.)

runcom:
I don't understand why testing cli at all when you can just unit test functions used with different input. I don't understand what coverage gives you more than just unit testing the logic inside core functions. Cli is just glue and, maybe my mistake, I don't get why unit test it sorry :(

mtrmac:
With the recent urfave/cli update, this is trivially small (basically split `main()` into `createApp` and running the app), and hopefully quite noncontroversial. Please review.

runcom:
lgtm

PR #17, opened 2016-03-21
Use contents of local checkout instead of last commit for validation

mtrmac:
Validating only committed files is not useful in the natural
>  $test_everything_passes; commit; push

workflow; the failures will not be caught locally, only by Travis later (and only if PRs are used instead of direct commits to master).

So, use the working directory state instead of last commit for validations; and remove misleading comments in checks which already use the working directory state.

### DIFF
--- a/hack/make/.validate
+++ b/hack/make/.validate
@@ -21,9 +21,7 @@ if [ -z "$VALIDATE_UPSTREAM" ]; then
 	VALIDATE_COMMIT_DIFF="$VALIDATE_UPSTREAM...$VALIDATE_HEAD"
 	
 	validate_diff() {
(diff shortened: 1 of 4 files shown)


mtrmac:
#18 will test that this works as expected.

mtrmac:
… confirmed, #18 did include one of the modified files and failed:
```
docker run --rm -i  -t "skopeo-dev:HEAD" hack/make.sh validate-git-marks validate-gofmt validate-lint validate-vet
---> Making bundle: validate-git-marks
There is trace of conflict(s) in the following files :
 - cmd/skopeo/main.go
```

runcom:
awesome, thax @mtrmac, merging

PR #19, opened 2016-03-21
Add infrastructure for running unit tests

mtrmac:
There haven’t been any unit tests in the code base so far; after adding them,

- Just running them is more difficult than `go test .`, so `Makefile` targets are useful.
- A full test run (Docker,Travis,non-Docker) now means validate+unit+integration tests, and it is easy to forget one or more; so, this adds new all-in-one `Makefile` targets.
- Travis should run the unit tests; by using the all-in-one target we get this for free.
- IDEs should be able to run the tests; this adds a demo for VS Code

The VS Code aspect is honestly a bit dubious; I’ll gladly drop it if you think it does not belong upstream.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,8 @@ man:
 shell: build-container
 	$(DOCKER_RUN_DOCKER) bash
 
+check: validate test-unit test-integration
+
 test-integration: build-container
 	$(DOCKER_RUN_DOCKER) hack/make.sh test-integration
 
@@ -60,6 +62,12 @@ test-unit: build-container
 
 validate: build-container
 	$(DOCKER_RUN_DOCKER) hack/make.sh validate-git-marks validate-gofmt validate-lint validate-vet
+
+# This target is only intended for development, e.g. executing it from an IDE. Use (make test) for CI or pre-release testing. 
+test-all-local: validate-local test-unit-local
+
+validate-local:
+	hack/make.sh validate-git-marks validate-gofmt validate-lint validate-vet
 
 test-unit-local:
 	go test $$(go list -e ./... | grep -v '^github\.com/projectatomic/skopeo/\(integration\|vendor/.*\)$$')
--- a/Makefile
+++ b/Makefile
@@ -54,5 +54,12 @@ shell: build-container
 test-integration: build-container
 	$(DOCKER_RUN_DOCKER) hack/make.sh test-integration
 
+test-unit: build-container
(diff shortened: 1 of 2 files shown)


runcom:
> The VS Code aspect is honestly a bit dubious; I’ll gladly drop it if you think it does not belong upstream.

honestly I don't believe, as you, this belongs upstream, I'll merge after you drop it
Thanks @mtrmac

mtrmac:
VS Code commit dropped.

Makefile
```diff
@@ -20,6 +20,8 @@ ifeq ($(INTERACTIVE), 1)
 	DOCKER_FLAGS += -t
 endif
 DOCKER_RUN_DOCKER := $(DOCKER_FLAGS) "$(DOCKER_IMAGE)"
+# Use =, not :=, so that this is not expanded when we are not running unit tests.
+UNIT_TEST_PACKAGES = $(shell find -name '*_test.go' | sed 's,/[^/]*$$,,' | sort -u | grep -v '^\./\(integration\|vendor\)$$')
```
runcom: should `go list ./...` take care of this filtering?

mtrmac: - `go list ./...` does not automatically filter `integration` and `vendor` subdirectories; are you saying that should somehow happen?
- `go list` lists all packages; this only finds packages which contain tests. I am unsure about which is better; `go list` causes `? $package_name [no test files]` lines to be output, arguably both useful and useless. Right now the `find -name '*.test.go'` variant means `$(UNIT_TEST_PACKAGES)` is empty and `make test-unit` does not even start the container, but this difference will go away with the first added unit test.

runcom: > go list ./... does not automatically filter integration and vendor subdirectories; are you saying that should somehow happen?

`go list ./... | grep -v ^vendor/|integration/` (or something like this)

> go list lists all packages; this only finds packages which contain tests. I am unsure about which is better; go list causes ? $package_name [no test files] lines to be output, arguably both useful and useless. Right now the find -name '*.test.go' variant means $(UNIT_TEST_PACKAGES) is empty and make test-unit does not even start the container, but this difference will go away with the first added unit test.

I'm ok with the "no tests found" line (cause somehow helps figuring instantly when running tests  coverage)

mtrmac: Getting `go list` to work was surprisingly non-trivial (e.g. it doesn't work within `sudo` when `GOPATH` is not set), but it allowed to drop the "empty package list" case and generally simplified the `Makefile`.

Thanks!

runcom: unfortunately anything dealing with sudo and env variables is fine with go :(

runcom:
LGTM

PR #20, opened 2016-03-22
Fix hack/.vendor-helper.sh for main package move.

mtrmac:
Otherwise the `clean` step of `hack/vendor.sh` would drop most `.go` files from `vendor/` as unused.

Also commits refreshed versions of a few of the vendored packages.


runcom:
Great catch! Thanks!

PR #23, opened 2016-03-23
[WIP] integration tests setup

runcom:


rhatdan:
If this is going to be run by a privileged job, which it looks like it will, it should use /run not /tmp

runcom:
ack

PR #24, opened 2016-03-23
Do not clean test-only dependencies from vendor packages

mtrmac:
Instead of only checking dependencies of the `"main"` packages, include also test dependencies of all subpackages of the project, and their transitive dependencies.


mtrmac:
This is necessary for using `testify`, which depends on other packages, only in `*_test.go` files.

runcom:
:+1:

PR #26, opened 2016-03-23
Signing unit tests

mtrmac:
This demonstrates command handler unit tests based on #16 .

Note: depends on #16, do not merge as is.

### DIFF
--- a/cmd/skopeo/fixtures/.gitignore
+++ b/cmd/skopeo/fixtures/.gitignore
@@ -1,0 +1,4 @@
+/*.gpg~
+/.gpg-v21-migrated
+/private-keys-v1.d
+/random_seed
--- a/cmd/skopeo/fixtures/corrupt.signature
+++ b/cmd/skopeo/fixtures/corrupt.signature
@@ -1,0 +1,1 @@
+../../../signature/fixtures/corrupt.signature
--- a/cmd/skopeo/fixtures/image.manifest.json
+++ b/cmd/skopeo/fixtures/image.manifest.json
@@ -1,0 +1,1 @@
+../../../signature/fixtures/image.manifest.json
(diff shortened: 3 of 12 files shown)


mtrmac:
Updated.

I have sort-of-dealt with the ugly `../../signature/fixtures` paths by instead using similar symlinks.  That seems better to me than hard copies, it shows the conceptual relationship of these tests.

Re: the philosophy of the thing, I am thinking of this primarily as unit testing the command-line parsing, error handling etc, which would not be tested otherwise.

I could perhaps be persuaded to take out the substantial integration-like tests (like `TestStandaloneSign` verifying that the contents of the output file really are a valid signature). Though, with the Atomic Registry integration and the corresponding longer time needed to run the in-container integration tests, I think having these locally and quickly run tests is even more valuable. The 34 milliseconds spend running these tests is well worth the peace of mind and quick turnaround time in my view.

mtrmac:
(Still depends on #16, I will rebase as necessary.)

runcom:
> I could perhaps be persuaded to take out the substantial integration-like tests (like TestStandaloneSign verifying that the contents of the output file really are a valid signature). Though, with the Atomic Registry integration and the corresponding longer time needed to run the in-container integration tests, I think having these locally and quickly run tests is even more valuable. The 34 milliseconds spend running these tests is well worth the peace of mind and quick turnaround time in my view.

I'm ok with this at this point - I just want to make sure we don't have two same tests in unit AND integration..

mtrmac:
> I just want to make sure we don't have two same tests in unit AND integration..

We don't:

 - `integration.TestSignVerifySmoke` generates fresh keys, and verifies that `standalone-verify` accepts the output of `standalone-smoke`.
 - These newly added tests use keys in fixtures, validate `standalone-sign` and `standalone-verify` completely independently using fixture data, and the argument handling unit tests have no counterpart the integration test.

The only thing the two have is common is reliance on `fixtures/image.manifest.json` instead of generating a manifest from scratch.

runcom:
#16 merged, will merge this after rebase

fine with me then, maybe we'll get rid of that only common manifest in the future

PR #25, opened 2016-03-23
Signing

mtrmac:
Note: This depends on #24; ATM #24 is included via a merge in this PR, I will rebase this branch after #24 is merged.

PR #28, opened 2016-03-24
remove cmd/ subdir

runcom:


version.go
```diff
@@ -1,7 +0,0 @@
-package skopeo
-
-// Version is a version of thils build.
-const Version = "0.1.12-dev"
```
mtrmac: The `skopeo/signature` package from #25 embeds the `Version` into signatures (to allow future code to workaround past mistakes); that won't work when the `Version` is in the `main` package.

Basically, either the commands are in a subdirectory, or we need a `skopeo/version` subpackage importable by both the commands and `skopeo/signature`.

I don’t care all that much, I just wanted to highlight this tradeoff.

runcom: > to allow future code to workaround past mistakes

could you explain? why signature embeds Version?

runcom: I see, could you instead define an hash or something (even 0x01) in the signature type instead of relying on skopeo global version? are there any downsides?

mtrmac: Assume that either the current version of the code has a bug which makes the signature invalid but still with a valid structure (say, computing a hash over UCS-4 representation instead of a raw `[]bytes`, using a non-canonical Docker reference, or incorrectly stripping a trailing `'\n'` _if there is one_). In the last example it may very well be a bug we notice only after a few months/years, with both correct and invalid signatures in the wild, both sharing the same formal version number.

In such cases, having the precise version number of the creator allows us to deploy targeted workarounds (when verifying a signature made by a known buggy version, compute the hash in the same way the buggy version does).

mtrmac: > I see, could you instead define an hash or something (even 0x01) in the signature type instead of relying on skopeo global version? are there any downsides?

There already is a format identifier (`critical.type`), but that has a different function.

(I do hope that we will never have to deploy workarounds based on the creator version (`optional.creator`), but I think it is prudent to have that option.)

runcom: alright, I think I'll do a version pkg then

runcom:
@mtrmac I've rebased and created a `version` pkg so signature can still continue to use it, merging

PR #29, opened 2016-03-30
Add ImageSource and ImageDestination abstractions

mtrmac:
`docker.go` is currently doing everything related to Docker images. To better support the Docker Registry, Atomic Registry, `docker save,load}` tarballs, and local directories (for debugging) in the future, introduce `ImageSource` and `ImageDestination` as abstractions of dumb storage which does not modify its contents.

Currently the only implementations are the existing ones, Docker Registry `dockerImageSource` and a local directory (`dirImageSource`, `dirImageDestination`). The `dockerImage` struct remains, as a layer on top of `dockerImageSource`; eventually this should probably be restructured further (e.g. `dockerImage.Layers` should probably mostly happen in the top-level `layers.go`, so that we don’t hard-code creation of random-named directories in the current directory).

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -10,7 +10,6 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
-	"path"
 	"path/filepath"
 	"regexp"
 	"strings"
@@ -415,11 +414,12 @@ func (i *dockerImage) Layers(layers ...string) error {
 	if err != nil {
 		return err
 	}
+	dest := NewDirImageDestination(tmpDir)
 	data, err := json.Marshal(m)
 	if err != nil {
 		return err
 	}
-	if err := ioutil.WriteFile(path.Join(tmpDir, "manifest.json"), data, 0644); err != nil {
+	if err := dest.PutManifest(data); err != nil {
 		return err
 	}
 	if len(layers) == 0 {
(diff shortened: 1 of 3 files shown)


docker.go
```diff
@@ -189,7 +182,61 @@ func sanitize(s string) string {
...
 }
 
-func (i *dockerImage) makeRequest(method, url string, auth bool, headers map[string]string) (*http.Response, error) {
+type dockerImageSource struct {
+	ref             reference.Named
+	tag             string
+	registry        string
+	username        string
+	password        string
+	WWWAuthenticate string // Obtained by s.ping(); FIXME: why in GetManifest?
```
runcom: will fix this, though, this whole interaction should go to vbatts/docker-utils one day

types/types.go
```diff
@@ -23,6 +24,20 @@ type Repository interface {
...
+	GetLayer(digest string) (io.ReadCloser, error)
+	GetSignatures() ([][]byte, error)
+}
+
+// ImageDestination is a service, possibly remote (= slow), to store components of a single image.
+type ImageDestination interface {
+	PutManifest([]byte) error
+	PutLayer(digest string, stream io.ReadCloser) error
+	PutSignatures(signatures [][]byte) error
+}
```
runcom: I'm not really fond to those interfaces (and their naming). Why a single interface/object can't do both (get|put)?

docker.go
mtrmac: I have a tentative fix in https://github.com/mtrmac/skopeo/commit/950fe6c637d62eb05d9f5d8802ce92e6f587fda8 (together with the rest of #30 ).

types/types.go
mtrmac: I think the interface separation is conceptually clearer, and may make a difference if we ever implement direct `docker {save,load}` streaming, where, once the child process is started, it is impossible to reverse the direction of the operation. (`docker {save,load}` might be needed as a component of `atomic {push,pull}`).

Of course a single object can implement both interfaces, so having two different interfaces does not require us to have two different objects for the two directions (though that would require more thought about what exactly the “a single image” is and what the semantics of overwrite are — we can’t overwrite a digest-identified image, so would a single `ImageDestination` refer to a tag? If so, can the contents mutate over time if someone else reassigns the tag?).

At the moment, this is basically a skeleton of an idea: the way the non-`dockerImageSource` code in `dockerImage` is in an awkward limbo clearly indicates the organization of the code needs more thought.

I care little about the naming, what would you prefer?

mtrmac: (I realize `docker save` streaming wouldn’t work with this interface, because we as the receiver can't dictate the order of layers coming from the stream; something to worry about later.)

mtrmac:
This is now ready for review, as the first component of Atomic Registry support.

runcom:
LGTM, merging but we need to remember https://github.com/projectatomic/skopeo/pull/29#discussion_r58041301

PR #30, opened 2016-03-30
Cleanups

mtrmac:
Do not merge now, as is this depends on and includes #29, filed mostly as a reminder to myself.

The cleanups are equally applicable to the non-#29 code, but the commits wouldn’t apply; after we deal with #29 I will rebase this manually one way or the other.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -477,15 +477,16 @@ func newDockerImageSource(img, certPath string, tlsVerify bool) (*dockerImageSou
 		return nil, err
 	}
 	var tr *http.Transport
-	if certPath != "" {
+	if certPath != "" || !tlsVerify {
 		tlsc := &tls.Config{}
 
-		cert, err := tls.LoadX509KeyPair(filepath.Join(certPath, "cert.pem"), filepath.Join(certPath, "key.pem"))
-		if err != nil {
-			return nil, fmt.Errorf("Error loading x509 key pair: %s", err)
+		if certPath != "" {
+			cert, err := tls.LoadX509KeyPair(filepath.Join(certPath, "cert.pem"), filepath.Join(certPath, "key.pem"))
+			if err != nil {
+				return nil, fmt.Errorf("Error loading x509 key pair: %s", err)
+			}
+			tlsc.Certificates = append(tlsc.Certificates, cert)
 		}
-
-		tlsc.Certificates = append(tlsc.Certificates, cert)
 		tlsc.InsecureSkipVerify = !tlsVerify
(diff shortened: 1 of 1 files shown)


mtrmac:
The `.ping()` move is fine as is right now; eventually `GetManifest` might need to call it explicitly again to get the API version—but even if that happened, we would want `makeRequest` to call it implicitly, and to cache the results so that it is called at most once, whatever higher-level methods are called.

mtrmac:
This is now ready for review, as a second component of Atomic Registry support. It is built on top of #29 (i.e. start with the “Remove a redundant dockerImageSource.makeRequest parameter” commit), I will be rebasing as necessary.

This PR is a set of fixes to docker.go, most of them necessary for the OpenShift client.

runcom:
should you rebase this now?

mtrmac:
Yes, rebased.

docker.go
```diff
@@ -473,15 +477,16 @@ func newDockerImageSource(img, certPath string, tlsVerify bool) (*dockerImageSou
 		return nil, err
 	}
 	var tr *http.Transport
-	if certPath != "" {
```
runcom: why did you refactor here? I don't like the  OR here - it was far more readable before this - I don't think this is fixing anything - if so could you please revert just this?

runcom:
I'll merge after my comment above

docker.go
mtrmac: This is not refactor, this a bug fix (https://github.com/projectatomic/skopeo/pull/30/commits/9b73007ba196e432b2a6009c5a7bbf675ce10d62 ). We need to go down this `if` if `!tlsVerify` to create a `tls.Config{InsecureSkipVerify: true}`, even if we do not have any certificates; without that, `!tlsVerify` is ignored.

mtrmac: (to create a `tls.Config` and a `http.Transport` which uses it, to be precise.)

docker.go
```diff
@@ -477,15 +477,16 @@ func newDockerImageSource(img, certPath string, tlsVerify bool) (*dockerImageSou
...
+	if certPath != "" || tlsVerify == false {
```
runcom: alright, gotcha - you could remove `==` and just use `!tlsVerify`

mtrmac: Fixed.

runcom:
merging

runcom:
could you rebase it? (not sure what's happened :?)

mtrmac:
Rebased (I think the merge of #38 happened).

PR #31, opened 2016-04-05
Vendor fixes

mtrmac:


runcom:
What is this exactly doing?

mtrmac:
See the individual commit messages. (I could be persuaded that the “Allow keeping vendor subdirectories” commit should not be merged without a user.)

PR #32, opened 2016-04-09
Test command output before eror status in signing integration tests

mtrmac:
If executing the command fails, the output to stderr is more useful than
the exit code, so test the output first.

### DIFF
--- a/integration/signing_test.go
+++ b/integration/signing_test.go
@@ -107,11 +107,11 @@ func (s *SigningSuite) TestSignVerifySmoke(c *check.C) {
 	defer os.Remove(sigOutput.Name())
 	out, err := exec.Command(skopeoBinary, "standalone-sign", "-o", sigOutput.Name(),
 		manifestPath, dockerReference, s.fingerprint).CombinedOutput()
-	c.Assert(err, check.IsNil)
+	c.Assert(err, check.IsNil, check.Commentf("%s", out))
 	c.Assert(string(out), check.Equals, "")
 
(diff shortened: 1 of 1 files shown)


integration/signing_test.go
```diff
@@ -107,11 +107,11 @@ func (s *SigningSuite) TestSignVerifySmoke(c *check.C) {
 	defer os.Remove(sigOutput.Name())
 	out, err := exec.Command(skopeoBinary, "standalone-sign", "-o", sigOutput.Name(),
 		manifestPath, dockerReference, s.fingerprint).CombinedOutput()
-	c.Assert(err, check.IsNil)
```
runcom: You could add as third parameter `check.Commentf(out)` so that if it fails you'll get the output also

runcom:
It's weird to check the output first but that's simply me so I'm fine with this if you are

integration/signing_test.go
mtrmac: Right, that’s better. Thanks!

integration/signing_test.go
```diff
@@ -107,11 +107,11 @@ func (s *SigningSuite) TestSignVerifySmoke(c *check.C) {
...
+	c.Assert(err, check.IsNil, check.Commentf("%s", out))
```
runcom: Misleading but no need for %s :)

mtrmac: `"%s"` is actually needed; otherwise `'%'` characters in `out` would be interpreted as format string metacharacters. Also, `out` is a `[]byte`, not a `string`.

PR #33, opened 2016-04-13
Add a workaround for a glibc bug when -lphtread precedes -lgpgme

mtrmac:
See detailed explanation inside the added file.

If you want to reproduce this, add an use of the `net` package somewhere early in the `main` package, and `sudo make check` (ensuring that your `fedora` image is Fedora 23).

### DIFF
--- a/cgo_pthread_ordering_workaround.go
+++ b/cgo_pthread_ordering_workaround.go
@@ -1,0 +1,32 @@
+package main
+
+/*
+This is a pretty horrible workaround.  Due to a glibc bug
+https://bugzilla.redhat.com/show_bug.cgi?id=1326903 , we must ensure we link
+with -lgpgme before -lpthread.  Such arguments come from various packages
+using cgo, and the ordering of these arguments is, with current (go tool link),
+dependent on the order in which the cgo-using packages are found in a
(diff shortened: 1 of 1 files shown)


runcom:
This sounds fine to me, the only stuff I don't like is the C linkage to compile. I would love to have a build tag to enable the whole gpg think, what do you think? Is it that bad to provide 2 kind of the same binaries with different functionalities? (I don't think so)

mtrmac:
Um, that C linkage is the whole point of this workaround?!

I may miss the wider context, why is this undesirable?

To make this workaround optional with a build tag, all of the `signature` subpackage would have to be optional as well, and then I’m not sure what we would do about the CLI missing commands (and eventually essential commands if, as I hope, verification is enabled by default on every pull).

rhatdan:
This can be dropped when glibc is fixed correct?

LGTM

runcom:
LGTM

runcom:
We'll need to edit skopeo.spec in Fedora dist-git once we cut a new release of skopeo with this feature (and ensure deps are added in the spec and everything is fine)

mtrmac:
> This can be dropped when glibc is fixed correct?

Yes.

PR #35, opened 2016-04-20
Update man page

jwhonce:
Resolves https://github.com/projectatomic/skopeo/issues/12

* Fill out man page
* Remove TODO's from go code regarding man page
* Additional information on building instructions

Signed-off-by: Jhon Honce <jhonce@redhat.com>


jwhonce:
@runcom @mtrmac From reading the source this is my best guess. :)

Additionally, @runcom I could do a much better job of formatting the COMMANDS section if I just used the man macros rather than markdown.  If that is okay with you, please let me know and I'll convert to nroff.

runcom:
> Additionally, @runcom I could do a much better job of formatting the COMMANDS section if I just used the man macros rather than markdown

I know this but I went with go-md2man since I didn't know man syntax :( I guess it's fine for me if you want to convert it (I'll learn the man syntax)

runcom:
seems like this needs a rebase also

jwhonce:
@runcom rebased and reformatted.

mtrmac:
The sign-* descriptions look good. I would, rather, like for documentation of these commands to come with a disclaimer “this is primarily for debugging; it should not be your normal operations workflow” (at least after we actually provide better tooling). I can prepare a PR after this one is merged.

runcom:
do we need to remove go-md2man from the Makefile man target also right?

jwhonce:
@runcom I removed the whole man target in the Makefile as it's not needed.

man1/skopeo.1
```diff
@@ -0,0 +1,99 @@
...
+.B key-fingerprint
+key identity to use for verification
+.br
+.B signature
+Path to file containing signature
+.TP
+.B help
+show help for \fBskopeo\fR
+.SH AUTHORS
+Antonio Murdaca <runcom@redhat.com>
```
runcom: @jwhonce might want to add your name here :angel:

runcom:
looks so fine to me :) thx Jhon (I'll make a note to update the spec files in RHEL/Fedora for this change)

PR #37, opened 2016-04-21
v2s1 manifest followup

mtrmac:
This is an update to #34 :

- Fix references to the manifest format: v2s1, not v1s1
- Move the code to a separate `dockerutils` package: the OpenShift client pushing data (or, possibly, a Docker registry push client) will need it, so the `signature` package is not appropriate.

 Similarly, the OpenShift client will probably like to reuse the layer enumeration code from `docker.go`, and it seems cleaner to put such transport-independent utilities into a separate package (and leave `docker.go` for the Docker Registry transport only).

Alternatively, we could just vendor in more of `docker/distribution` and perhaps `docker/docker` instead of having our own implementations, but I got the impression (perhaps mistaken?) that this would be unwelcome.


runcom:
Vendoring docker stuff was how skopeo was implemented but given the fast pace they're developing and fixing their libraries it wasn't feasible to me at all. This is how docker-utils from Vincent Batts is also implemented. The docker libraries pull in so much code also. 

This LGTM

runcom:
What I said applies to just the main interaction with registries. If you feel their libraries makes a better job at the stuff you're working on then I think I could leave with it. Otherwise I'd like to keep the vendor code base small and do things on our own.

runcom:
Needs a rebase then I'll merge

mtrmac:
Rebased.

PR #43, opened 2016-05-02
Add Atomic Registry support for push and pull, and a new “copy” command

mtrmac:
This is on top of #29 and #30, which are perhaps better reviewed separately; I will be rebasing as necessary. (To review this set individually, start with the “Make dockerutils.GuessManifestMIMEType public” commit.)

This:
 - Adds an `ImageDestination` (push) support to the Docker client, for untagged images only for now.
 - Adds support for push and pull to the Atomic Registry (OpenShift imagestreams).
 - Adds a `skopeo copy` command, as a generalized push/pull implementation, intended as a backend for the `atomic` command, e.g.

 ```
skopeo copy atomic:mitr/busybox:latest dir:t-down # pull
skopeo copy dir:t-up atomic:mitr/busybox:latest # push
```
  The `dir:` image source/destination is compatible with `skopeo layers`, and is intended as an integration point for possibly having push/pull clients written in atomic (Python) and skopeo (Go) interoperate.
 - Adds `skopeo copy --sign-by=$key_id` , allowing signing on push. (Even without this command, signatures can be manually created for local images using the `dir:` format.)


mtrmac:
Note that `skopeo copy --sign-by=` is not usable right now because the Atomic Registry client in this PR can not store signatures.

#44 demonstrates how this could be done using annotations.

runcom:
you can finally rebase this one!

mtrmac:
Rebased. (BTW does this thematically-split-PR model work for you or is it just annoying hassle?)

runcom:
It's perfect and awesome - if it's OK for you I'd  continue on this model

runcom:
merging - except the 10k dependencies :(

PR #44, opened 2016-05-02
DO NOT MERGE: Storing signatures in OpenShift annotations

mtrmac:
This is a proof-of-concept only for storing signatures in the Atomic Registry, using annotations on Image objects.

DO NOT MERGE this, signature storage will instead be explicitly supported per https://github.com/openshift/origin/pull/8371 .

(This is on top of, and includes all of, #43 ; only the last commit is new.)

### DIFF
--- a/openshift/openshift.go
+++ b/openshift/openshift.go
@@ -2,6 +2,7 @@ package openshift
 
 import (
 	"bytes"
+	"encoding/base64"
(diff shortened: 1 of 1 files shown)


mtrmac:
Moved to https://github.com/containers/image/pull/11 .

PR #45, opened 2016-05-06
Rerun hack/vendor.sh to pick up gpgme changes

mtrmac:
See https://github.com/proglottis/gpgme/pull/7 for the full discussion.

PR #46, opened 2016-05-07
WIP: Add support for (make check CHECK_BASE_IMAGES='fedora centos')

mtrmac:
This is intended for use in Travis, modifying the `make check` command to the above (but running this locally would of course also be possible).  The default behavior remains the same, uses the `fedora` image.

At the moment, this does not work because in `centos`, the `python-pip` package is not available (perhaps in EPEL?).

(Also note that the `(yum install)` does not fail when one of the packages is missing.)

Anyone, feel free to take this over (adding conditionals to `Dockerfile`s? Maintaining a `Dockerfile.fedora` and `Dockerfile.centos` separately?) if you have the time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,7 +74,7 @@ jobs:
       services:
       - docker
       script: 
-      - make vendor && ./hack/tree_status.sh && make local-cross && make check
+      - make vendor && ./hack/tree_status.sh && make local-cross && make check CHECK_BASE_IMAGES='fedora centos'
 
     # Run 3 image-build-push tasks in parallel for linux/amd64, linux/s390x and linux/ppc64le platforms (for upstream and stable)
     - stage: image-build-push
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 FROM fedora
 
-RUN dnf -y update && dnf install -y make git golang golang-github-cpuguy83-md2man \
+RUN if (. /etc/os-release ; [ "$ID" != centos ]); then dnf=dnf; md2man=golang-github-cpuguy83-md2man; else dnf=yum; md2man=golang-github-cpuguy83-go-md2man; $dnf install -y epel-release; fi \
+	&& $dnf -y update && $dnf install -y make git golang $md2man \
 	# storage deps
 	btrfs-progs-devel \
 	device-mapper-devel \
@@ -14,7 +15,7 @@ RUN dnf -y update && dnf install -y make git golang golang-github-cpuguy83-md2ma
         bats jq podman runc \
 	golint \
 	openssl \
-	&& dnf clean all
+	&& $dnf clean all
 
 # Install two versions of the registry. The first is an older version that
 # only supports schema1 manifests. The second is a newer version that supports
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,16 @@ ifeq ($(GOOS), linux)
 endif
 
 GIT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD 2>/dev/null)
-IMAGE := skopeo-dev$(if $(GIT_BRANCH),:$(GIT_BRANCH))
+# Default base container image on which to run tests
+DEFAULT_BASE := fedora
+# Base image for most commands, overridable by command-line arguments
+BASE_IMAGE := fedora
+# Base images (one or more) used for (make check), overridable by command-line arguments
+CHECK_BASE_IMAGES := $(BASE_IMAGE)
+# Name for our containers: skopeo-dev[-base-image][:branch]
+define image_name
+skopeo-dev$(if $(filter-out $(DEFAULT_BASE),$1),-$1)$(if $(GIT_BRANCH),:$(GIT_BRANCH))
+endef
 # set env like gobuildtag?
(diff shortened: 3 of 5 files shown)


cgwalters:
FYI, my work in CentOS CI on PR testing is now triggering for skopeo.

It chains into rpmdistro-gitoverlay (i.e. it's building the way we're shipping).  See
https://github.com/cgwalters/sig-atomic-buildscripts/blob/ebd950bf686b54cfdca49dfca2a305a4363cc2cd/centos-ci/jjb/sig-atomic-testing.yml#L190

I'll post more about this when it's farther along.

It might be worth looking at how rpmdistro-gitoverlay works, as I'd like it to be a convenient tool for *local* development too.  (It's a bit away from that now, but it also has fundamental architectural advantages over `Dockerfile` per repo)

mtrmac:
OK, this now seems to work (but, man, the compile-test-debug-edit cycle with Dockerfiles is long); except that it correctly diagnoses #89 and breaks the build.  So, trying with #91 to see whether that would be enough.

mtrmac:
Fixed up, everything seems to work per #91. Note that on CentOS we don't run golint because it does not build against Go 1.4.2.

I will keep rebasing this and #91 from time to time, so that we have a chance to know about CentOS failures before Colin finds out ☺

mtrmac:
As of https://rhn.redhat.com/errata/RHSA-2016-1538.html / https://lists.centos.org/pipermail/centos-announce/2016-August/022005.html , CentOS has golang 1.6.3, so this is no longer broken — OTOH it needs to be simplified to drop the 1.4. conditionals.

rhatdan:
Any progress on getting this mergable?

mtrmac:
We still need to figure out who will host, and keep rebuilding, the test base images, and then separate the test image build and the running of the tests in the `Makefile`.

Worst case we can just find a suc^Wvolunteer to rebuild them every Monday manually and push to a personal repo on the internet… I’m afraid it might come to that.  The Docker Hub image rebuild functionality might be suitable but those builders don’t have enough memory to build OpenShift.

cgwalters:
We have automation tooling available in [CentOS CI](https://wiki.centos.org/QaWiki/CI) as well as [internally](https://url.corp.redhat.com/9eca818).  There's also an effort to do a Travis-like system for the latter [here](https://github.com/jlebon/redhat-ci) that we own and can extend (for example, to declaratively define container dependencies).

rhatdan:

mtrmac:
Yes, I am keeping this branch alive. The question of hosting / rebuilding base images for the tests hasn’t been resolved so far; Colin has posted some helpful pointers, but I haven’t had time to look into them.

rhatdan:
@mtrmac Still want this?

mtrmac:
Yes, I’d still like this to happen, and I still haven’t looked into who could host and rebuild the base images.

In the mean time, I keep rebasing the branch and testing it locally, which gives at least a delayed check that we didn’t break CentOS.

rhatdan:
@mtrmac You still want to keep this?  Please update.

mtrmac:
> @mtrmac You still want to keep this? Please update.

Yes, no change to the desire nor to the benefit of keeping this around. I’ll rebase this along with all my other branches in a brach sometime.

rhatdan:
@mtrmac This needs a rebase?  Is it still important? Can we get it merged permanently?

mtrmac:
Per https://github.com/containers/skopeo/issues/47:
> I would hope that this gets addressed along with the conversation about using Travis CI in general, and maybe consolidating the CI designs across containers/*. Any help would of course be welcome.
>
> (Really the biggest win for skopeo CI would be not building the OpenShift binary on every CI run.)
>
> (I can’t remember any instance of CentOS-specific breakage since #45 , so this is not all that urgent.)

I will rebase this $later.

rhatdan:
This might be the oldest PR we have, should we close it?

mtrmac:
Nah, if we ever add CentOS to the CI, the work in this PR (even in its current partially-broken state) will be useful.

rhatdan:
3.5 years since this has been updated?  Should we close?

mtrmac:
Last updated less than a month ago, actually.

I’d love for this to be integrated in some sort of CI (and for the CI not to spend >15 minutes building OpenShift 1.5 every single time), but this is still better than nothing.

cevich:
IMHO this PR should indeed be closed, it's no-longer relevant nor applicable given the amount of changes on the CI side. Downstream consumers need to have their own environment-independent test-suites.  If those can reasonably be shared with upstream (i.e. like how we have a separate system-tests package for podman) that's fine.  But it's unworkable to hold-back upstream development based on downstream (i.e. necessarily "old") environment requirements.

mtrmac:
> But it's unworkable to hold-back upstream development based on downstream (i.e. necessarily "old") environment requirements.

(c/image is, unlike Podman, simple and environment-independent enough that adding compatibility upstream seemed easier than maintaining old stable branches; the major issue was just noticing that we broke something, hence this PR.

OTOH that was much more of an issue when still targeting RHEL 7, and that’s no longer much of a concern.)

---

> IMHO this PR should indeed be closed, it's no-longer relevant nor applicable given the amount of changes on the CI side.

Given the recent changes, and that this originally targeted RHEL 7, I agree; especially with the recent changes to build test base images separately, imposing RHEL 7 targets on that work doesn’t make sense any more.

PR #49, opened 2016-05-09
*: split docker.go for future pkg creation

runcom:
cc @mtrmac 

PTAL, this splits that ugly `docker.go` - I'll follow up with another PR to move to have a library for this whole docker push/pull but for now it's a lot better to have those files now


utils.go
```diff
@@ -57,3 +58,30 @@ func parseImageDestination(c *cli.Context, name string) (types.ImageDestination,
...
 	return nil, fmt.Errorf("Unrecognized image reference %s", name)
 }
+
+func parseDockerImage(img, certPath string, tlsVerify bool) (types.Image, error) {
+	s, err := newDockerImageSource(img, certPath, tlsVerify)
+	if err != nil {
+		return nil, err
+	}
+	return &dockerImage{src: s}, nil
+}
```
mtrmac: I think this really belongs into `docker_image.go`; it is a constructor for `dockerImage`.

mtrmac: … and it has nothing to do with the CLI like the other functions in here.

utils.go
```diff
@@ -57,3 +58,30 @@ func parseImageDestination(c *cli.Context, name string) (types.ImageDestination,
...
+
+// parseDockerImageName converts a string into a reference and tag value.
+func parseDockerImageName(img string) (reference.Named, string, error) {
```
mtrmac: Similarly, this would probably be better in `docker_client.go` - it is an internal helper shared between the `newDockerImageSource` and `NewDockerImageDestination` constructors, i.e. any user of `dockerImageSource` will need this function as well; so pulling `docker_*.go` into a separate subpackage would drag in the top-level package for `utils.go`.

mtrmac:
This is definitely a great cleanup. Just two nits.

utils.go
runcom: Right

utils.go
runcom: You meant `docker_image.go`?

runcom: I moved it to `docker_image.go` though - as I think `docker_client.go` was a typo, let me know if I'm wrong

mtrmac: I did mean `docker_client.go`; `NewDockerImage{Source,Destination}` each calls both `parseDockerImageName` and `newDockerClient`. (We could even combine the two? Probably not, that seems uglier.)

The layering is `docker_image` → `docker_image_{source,destination}` → `docker_client`, putting it in `docker_image` causes a reference in the other direction.

Ultimately, if all of `docker_*` goes into the separate package, this does not matter that much of course :smile:

runcom: oh that makes sense, moving it there then

runcom: so what if we hide this parse login directly into `newDockerClient` (so it calls `parseDockerImageName` and `newDockerClient` just accepts a `string`?

runcom: nah, it's ugly, moving `parseDockerImageName` to an new `docker_utils.go` just for now as I don't want it under `docker_client.go` :(

mtrmac: Even better. Thanks for bearing with me! Will merge after CI.

PR #51, opened 2016-05-10
*: move pkg main into cmd/skopeo/

runcom:
Following go best practices - main package has been moved under `cmd/skopeo` and the top level directory now will just contain *libraries* (aka importable) pkgs

@mtrmac PTAL


mtrmac:
I thought the point was to separate the library pieces (~ the `types.Image*` implementations) from the CLI; so why move _everything_ to `cmd/skopeo`?

runcom:
@mtrmac this is just another piece of moving the libraries pieces to their own subdirs - this justs move them there for now, then docker_* will be moved back to the root dir under subdirs

mtrmac:
To me that’s a counterintuitive way to go about it, but it does work, of course. Merging.

PR #52, opened 2016-05-11
move libraries to pkgs

runcom:
@mtrmac PTAL (I've also added some TODO)

This is just another small steps in defining a good API for sub pkgs (which could be later used by other pkgs, i.e. k8s)
Eventually I'd love to remove `docker/docker_image.go` and integrate relevand code either in img src or img dest. (@mtrmac please confirm this is your view also :smile: )
The major change (other then moving files around) is that I've strongly types `GetManifest()` to return an interface (with some methods just added) so it will work across implementation.

Again, this is still a WIP but I would like to have this merged and tested as we going forward with this (and also avoid having to review 10k LOC PRs :) :) :))

This code compiles right but I'd love if you could test it out (specifically for openshift related stuff and signing)

### DIFF
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -1,28 +1,10 @@
 package main
 
 import (
-	"encoding/json"
-
 	"github.com/Sirupsen/logrus"
 	"github.com/codegangsta/cli"
 	"github.com/projectatomic/skopeo/signature"
 )
-
-// FIXME: Also handle schema2, and put this elsewhere:
-// docker.go contains similar code, more sophisticated
-// (at the very least the deduplication should be reused from there).
-func manifestLayers(manifest []byte) ([]string, error) {
-	man := manifestSchema1{}
-	if err := json.Unmarshal(manifest, &man); err != nil {
-		return nil, err
-	}
-
-	layers := []string{}
-	for _, layer := range man.FSLayers {
-		layers = append(layers, layer.BlobSum)
-	}
-	return layers, nil
-}
 
 func copyHandler(context *cli.Context) {
 	if len(context.Args()) != 2 {
@@ -45,11 +27,7 @@ func copyHandler(context *cli.Context) {
 		logrus.Fatalf("Error reading manifest: %s", err.Error())
 	}
 
-	layers, err := manifestLayers(manifest)
-	if err != nil {
-		logrus.Fatalf("Error parsing manifest: %s", err.Error())
-	}
-	for _, layer := range layers {
+	for _, layer := range manifest.Layers() {
 		stream, err := src.GetLayer(layer)
 		if err != nil {
 			logrus.Fatalf("Error reading layer %s: %s", layer, err.Error())
@@ -75,7 +53,7 @@ func copyHandler(context *cli.Context) {
 			logrus.Fatalf("Error determining canonical Docker reference: %s", err.Error())
 		}
 
-		newSig, err := signature.SignDockerManifest(manifest, dockerReference, mech, signBy)
+		newSig, err := signature.SignDockerManifest(manifest.Raw(), dockerReference, mech, signBy)
 		if err != nil {
 			logrus.Fatalf("Error creating signature: %s", err.Error())
 		}
@@ -87,7 +65,7 @@ func copyHandler(context *cli.Context) {
 	}
 
 	// FIXME: We need to call PutManifest after PutLayer and PutSignatures. This seems ugly; move to a "set properties" + "commit" model?
-	if err := dest.PutManifest(manifest); err != nil {
+	if err := dest.PutManifest(manifest.Raw()); err != nil {
 		logrus.Fatalf("Error writing manifest: %s", err.Error())
 	}
 }
--- a/directory/directory.go
+++ b/directory/directory.go
@@ -1,4 +1,4 @@
-package main
+package directory
 
 import (
 	"fmt"
@@ -77,13 +77,15 @@ func NewDirImageSource(dir string) types.ImageSource {
 	return &dirImageSource{dir}
 }
 
-func (s *dirImageSource) GetManifest() ([]byte, string, error) {
+func (s *dirImageSource) GetManifest() (types.ImageManifest, string, error) {
 	manifest, err := ioutil.ReadFile(manifestPath(s.dir))
 	if err != nil {
 		return nil, "", err
 	}
+	// TODO(runcom): we should hide this type from the pkg API
+	m := &types.DockerImageManifest{RawManifest: manifest}
 
-	return manifest, "", nil // FIXME? unverifiedCanonicalDigest value - really primarily used by dockerImage
+	return m, "", nil // FIXME? unverifiedCanonicalDigest value - really primarily used by dockerImage
 }
 
 func (s *dirImageSource) GetLayer(digest string) (io.ReadCloser, error) {
--- a/docker/docker_image.go
+++ b/docker/docker_image.go
@@ -2,20 +2,13 @@ package docker
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
-	"regexp"
 	"strings"
-	"time"
 
 	"github.com/projectatomic/skopeo/directory"
 	"github.com/projectatomic/skopeo/types"
-)
-
-var (
-	validHex = regexp.MustCompile(`^([a-f0-9]{64})$`)
 )
 
 type dockerImage struct {
@@ -83,57 +76,9 @@ type config struct {
 	Labels map[string]string
 }
 
-type v1Image struct {
-	// Config is the configuration of the container received from the client
-	Config *config `json:"config,omitempty"`
-	// DockerVersion specifies version on which image is built
-	DockerVersion string `json:"docker_version,omitempty"`
-	// Created timestamp when image was created
-	Created time.Time `json:"created"`
-	// Architecture is the hardware that the image is build and runs on
-	Architecture string `json:"architecture,omitempty"`
-	// OS is the operating system used to build and run the image
-	OS string `json:"os,omitempty"`
-}
-
 // TODO(runcom)
 func (i *dockerImage) DockerTar() ([]byte, error) {
 	return nil, nil
-}
-
-// will support v1 one day...
-type manifest interface {
-	String() string
-	GetLayers() []string
-}
-
-type manifestSchema1 struct {
-	Name     string
-	Tag      string
-	FSLayers []struct {
-		BlobSum string `json:"blobSum"`
-	} `json:"fsLayers"`
-	History []struct {
-		V1Compatibility string `json:"v1Compatibility"`
-	} `json:"history"`
-	// TODO(runcom) verify the downloaded manifest
-	//Signature []byte `json:"signature"`
-}
-
-func (m *manifestSchema1) GetLayers() []string {
-	layers := make([]string, len(m.FSLayers))
-	for i, layer := range m.FSLayers {
-		layers[i] = layer.BlobSum
-	}
-	return layers
-}
-
-func (m *manifestSchema1) String() string {
-	return fmt.Sprintf("%s-%s", sanitize(m.Name), sanitize(m.Tag))
-}
-
-func sanitize(s string) string {
-	return strings.Replace(s, "/", "-", -1)
 }
 
 func (i *dockerImage) retrieveRawManifest() error {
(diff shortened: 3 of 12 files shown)


directory/directory.go
```diff
@@ -77,13 +77,15 @@ func NewDirImageSource(dir string) types.ImageSource {
 	return &dirImageSource{dir}
 }
 
-func (s *dirImageSource) GetManifest() ([]byte, string, error) {
+func (s *dirImageSource) GetManifest() (types.ImageManifest, string, error) {
 	manifest, err := ioutil.ReadFile(manifestPath(s.dir))
 	if err != nil {
 		return nil, "", err
 	}
+	// TODO(runcom): we should hide this type from the pkg API
```
runcom: This will be eventually just an anonymous struct with methods from the interface to be complain with the return interface arg (or at least I'd love to do so)

docker/docker_image_src.go
```diff
@@ -64,7 +65,36 @@ func (s *dockerImageSource) GetManifest() (manifest []byte, unverifiedCanonicalD
...
+	tags, err := s.getTags()
+	if err != nil {
+		return nil, "", err
+	}
+	manifest, err = makeImageManifest(s.ref.FullName(), manblob, unverifiedCanonicalDigest, tags)
+	return
+}
+
+func (s *dockerImageSource) getTags() ([]string, error) {
+	// FIXME? Breaking the abstraction.
```
runcom: I don't remember why this is here (@mtrmac ?)

docker/docker_image.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	imgManifest, err := makeImageManifest(i.src.ref.FullName(), i.rawManifest, i.digest, tags)
+	if err != nil {
+		return nil, err
+	}
+	return imgManifest, nil
+}
+
+func (i *dockerImage) getTags() ([]string, error) {
+	// FIXME? Breaking the abstraction.
```
runcom: I don't remember why this is here (@mtrmac ?)

docker/docker_image_src.go
```diff
@@ -1,69 +1,92 @@
...
+		return nil, "", err
 	}
-	return imgManifest, nil
+	manifest, err = makeImageManifest(s.ref.FullName(), manblob, unverifiedCanonicalDigest, tags)
+	return
 }
 
-func (i *dockerImage) getTags() ([]string, error) {
+func (s *dockerImageSource) getTags() ([]string, error) {
 	// FIXME? Breaking the abstraction.
```
runcom: I don't remember why this is here (@mtrmac ?)

mtrmac: "breaking the abstraction” refers to `dockerImage` directly using members of `i.src` = `dockerImageSource`. That would not be applicable any longer if the method were moved to `dockerImageSource`.

runcom: thx, I've removed it

mtrmac:
I am not sure about the direction of this; I was thinking of `ImageSource`/`ImageDestination` as fairly dumb storage layers, with manifest really being a (blob, MIME content type) pair only. You can get a manifest blob from any source you like, they are all parsed in the same way. 

Basically we need to very carefully think about which attributes are common to all storage mechanisms, and which are specific to Docker Registry / OCI bundles / ostree subtrees / … .

(It might well turn out that the dumb storage abstraction is simply unworkable. But if so, we should not be using the “manifest” term referring to v2s[12] config+layer lists to mean “any container image in any format whatsoever”.)

runcom:
@mtrmac I don't still know if Image[Source|Destination] will be good. However this PR is simply moving files around (and just adapting some code).
I believe after we have moved everything to pkgs we can start concentrating at defining common things across the stuff we need to support.

skopeo was born to just fetch and display manifest from registries at the beginning. However, I think it would be good trying to abstract a common template for manifests (as originally thought upstream also).

> (It might well turn out that the dumb storage abstraction is simply unworkable. But if so, we should not be using the “manifest” term referring to v2s[12] config+layer lists to mean “any container image in any format whatsoever”.)

when referring to *manifest* I'm referring to `types.ImageManifest` which is an interface though (code probably doesn't refer to this everywhere)

runcom:
I'm thinking for instance, that the storage layers you added (source|destination) can be very well suited *inside* a top level object exposed in the API (thinking out loud)

cmd/skopeo/copy.go
```diff
@@ -41,17 +24,14 @@ func copyHandler(context *cli.Context) {
 	}
 	signBy := context.String("sign-by")
 
+	// TODO(runcom): unused digest??? Miloslav?
```
mtrmac: (Strictly speaking it is printed to stdout, so not unused…)

The digest is there only because `dockerImage.retrieveRawManifest` was using it to eventually store it into a `DockerImageManifest`, and `skopeo inspect` is printing it to stdout. I did not look further into whether that should happen at all (though why print the unverified digest instead computing a fresh one, now that I look at it?) and simply preserved the original data flow.

runcom:
> I am not sure about the direction of this; I was thinking of ImageSource/ImageDestination as fairly dumb storage layers, with manifest really being a (blob, MIME content type) pair only. You can get a manifest blob from any source you like, they are all parsed in the same way.

I understand your concerns here though, let me rework this again so that those 2 interfaces just give out a blob

mtrmac:
> However this PR is simply moving files around (and just adapting some code).

I would agree with a simple move in a jiffy. The “adapting some code” here is changing the philosopy, and that will require more thought. You may well be right but I’d like to think about it a bit more.

mtrmac:
> when referring to manifest I'm referring to `types.ImageManifest` which is an interface though (code probably doesn't refer to this everywhere

This is probably the root of the disconnect; I am thinking about the Docker Registry manifest concept, which does not know that much about containers or tags or anything.

mtrmac:
> I'm thinking for instance, that the storage layers you added (source|destination) can be very well suited *inside* a top level object exposed in the API (thinking out loud)

Yes, I am thinking along the same lines. If we separate the distribution aspect (the `ImageSource` and `ImageDestination` are really `DockerV2ImageSource` and `DockerV2ImageDestination` now) and the contents aspect, we could have a
```go
func ParseDockerImage(types.ImageSource) (types.Image, error)
```
which would internally download the manifest blob, and then convert it into something much more useful to actual end users (e.g. list of layers, labels, configuration extracted from the manifest into nice fields, a method to convert that into a runc setup, whatever).

(Yes, that is quite similar to the existing `docker_image` top layer; but the distribution of code between `dockerImage` and `dockerImageSource`, and especially between `dockerImage` and its CLI callers, might be different than it is now — the `ImageDestination` use in `dockerImage` really should not be there.)

(`DockerV2`)`Image`{`Source`,`Destination`} would not be used by callers who want to _understand_ the image; it would be only for those who want to copy it and make sure that they don’t modify it (in particular that they don’t break signatures). Though it is unclear that the various transport methods have enough in common that they can be abstracted like this. (e.g. IIRC some versions of the `docker save` tarballs don’t preserve manifests?)

rhatdan:
One other thing while you are doing this rework could you remove "docker" wherever possible and use oci, if applicable.  Since we will be pulling and pushing oci images as well as docker images.

cmd/skopeo/copy.go
mtrmac: Actually I don’t think there is a good reason for `copy` to write this to stdout at all. https://github.com/projectatomic/skopeo/pull/53 .

directory/directory.go
mtrmac: The `ImageManifest` returned by `dirImageSource` needs to understand the manifest contents just as well as the `dockerImageSource` implementation.

I.e. I think the “download a manifest” and “parse and understand a manifest” code should be separate, with the `ImageSource` abstraction serving as the border.

docker/docker_image.go
```diff
@@ -40,20 +43,14 @@ func (i *dockerImage) RawManifest(version string) ([]byte, error) {
 }
 
 func (i *dockerImage) Manifest() (types.ImageManifest, error) {
-	// TODO(runcom): unused version param for now, default to docker v2-1
-	m, err := i.getSchema1Manifest()
-	if err != nil {
+	if err := i.retrieveRawManifest(); err != nil {
```
mtrmac: BTW (independent of the rest):

If all `retrieveRawManifest` does is store the values of `i.src.GetManifest`, wouldn’t it be simpler to call it directly here, and remove the `retrieveRawManifest` function entirely?

I suppose we might want caching… but if so, perhaps it should be somehow universal. Make it a contract of `ImageSource` that data is cached, or add a `CachingImageSource`?

runcom: you're totally right - I just overlooked thinking about this but it does smell a lot - thanks for the suggestion

types/types.go
```diff
@@ -56,12 +56,15 @@ type Image interface {
...
 // TODO(runcom)
 type ImageManifest interface {
+	Raw() []byte
+	Layers() []string
 	String() string
 }
 
 // DockerImageManifest is a set of metadata describing Docker images and their manifest.json files.
 // Note that this is not exactly manifest.json, e.g. some fields have been added.
 type DockerImageManifest struct {
```
mtrmac: Quite a few fields have been added; most worrisome to me are those (e.g. `Name`, `Tag`, `RepoTags`) which can’t be parsed from the manifest and are not a property of the manifest. It works, of course, but it feels unclean.

types/types.go
```diff
@@ -71,7 +74,17 @@ type DockerImageManifest struct {
...
+
+// Raw returns the raw manifest as received from the registry
+func (m *DockerImageManifest) Raw() []byte {
+	return m.RawManifest
+}
+
+// Layers returns a list of the FS layers in the manifest
+func (m *DockerImageManifest) Layers() []string {
+	return m.FSLayers
 }
```
mtrmac: Why provide these methods, when the underlying struct members are public as well? Having either one public seems sufficient.

runcom: I'm thinking that OCI manifest for instance will have different members (or a completely new struct) so by having an interface (those methods satisfy the ImageManifest interface above) we could support other manifest struct

docker/docker_image_src.go
```diff
@@ -65,11 +73,37 @@ func (s *dockerImageSource) GetManifest() (manifest types.ImageManifest, unverif
 	if res.StatusCode != http.StatusOK {
 		return nil, "", errFetchManifest{res.StatusCode, manblob}
 	}
+	// TODO(remove, already set in manifest, below): Miloslav?
```
mtrmac: Well, the commit keeps the variable but removes its only user :) Do either, but not both. Or perhaps reconsider why `skopeo inspect` even wants to print that value, I don’t know the context.

types/types.go
runcom: agree - they're there since the beginning and are only used when pretty-printing to the cli (after an inspect command) - though - those fields are inside this struct, most of the time I'd like to deal with image manifests struct via the interface above

docker/docker_image_src.go
runcom: I haven't well understood that the digest we get could not be the *right* one so in the past I embedded it to display it to the user - however my current understanding is that this digest is not to be trusted so I don't think it's needed to be exposed in skopeo - correct?

docker/docker_image_src.go
```diff
@@ -65,11 +73,37 @@ func (s *dockerImageSource) GetManifest() (manifest types.ImageManifest, unverif
...
+	if err != nil {
+		return nil, "", err
+	}
+	manifest, err = makeImageManifest(s.ref.FullName(), manblob, res.Header.Get("Docker-Content-Digest"), tags)
 	return
 }
 
+func (s *dockerImageSource) getTags() ([]string, error) {
+	url := fmt.Sprintf(tagsURL, s.ref.RemoteName())
+	res, err := s.c.makeRequest("GET", url, nil, nil)
```
mtrmac: I don’t understand this. AFAICT the list of tags in a repository (`rhel7/rhel` → `7.0`, `7.0.1`, `7.2`, …) has nothing to do with the single digest or tag reference to which this `dockerImageSource` points (`rhel7/rhel:latest`). Why is the list of all tags a property of the image, and why does `skopeo inspect` of a single image list all of the unrelated tags?

It may well make sense for `skopeo inspect` to provide the information, but the `ImageSource` abstraction is not suited for that; that would be a better fit for an `ImageRepository` of some kind (which in turn could not be used to talk about `save`/`load` streams transfered from stdin/to stdout).

Perhaps a cleaner thing to do would be to lift this all the way to `inspect.go`, and give `inspect.go` an API to get the underlying `dockerClient`? I don’t know, I don’t feel like I understand the requirements.

types/types.go
mtrmac: Then make the struct members private, and have everyone use the methods? That does run into `docker inspect` unmarshaling the struct to stdout right now.

types/types.go
mtrmac: It seems to me that `inspect` should define its own struct (with whatever fields it wants, coming from whatever mutlitude of sources needed) separate from this library interface, then.

Having `Labels`, which come from the config stored inside the manifest, is very useful and completely fine; having `RepoTags`, which have very little to do with the manifest, is very strange, and even the `Name` and `Tag` (no longer present in v2s2) have unclear properties. When an image which calls itself `library/busybox:latest` is being pulled from `private.mitr/l33t/busybox:$timestamp and pushed to `library/busybox:theNextGeneration`, does it make sense to talk about “the” name and tag? (OTOH it still makes perfect sense to talk about “the” labels and layers.)

types/types.go
runcom: > Then make the struct members private, and have everyone use the methods

yes was next step actually :P

docker/docker_image_src.go
mtrmac: > however my current understanding is that this digest is not to be trusted so I don't think it's needed to be exposed in skopeo - correct?

The model in presence of signature assumes the registry to be malicious: it can serve a completely different manifest from the one which has been signed, and it can lie about the digest (e.g. when asked for a manifest with a specific digest, it can return a manifest with a different digest value; and whatever manifest it returns, it can return a non-matching value in the HTTP header).

_If_ we don’t trust the registry and want to verify signatures, when verifying signatures we _must_ recompute the digest from the blob we have actually received.

OTOH if we are displaying information from the registry without doing any signature verification, we are ~by definition trusting the registry and using the value would be fine.

Still, there seems to be no reason to show the possibly untrusted value to the user: `skopeo inspect` can recompute a gurananteed-correct value using `dockerutils.ManifestDigest` and we can simply ignore the HTTP header.

AFAICT the only thing the HTTP header could be useful for is protection against data corruption with non-TLS HTTP with only the 16-bit TCP checksums, or against using HTTP incorrectly, causing the data to be corrupted during transfer; this would benefit from the uploader and downloader verifying the same digest (regardless of whether higher layers know about any signatures). We don't currently do this; and even if we did, the digest in the HTTP header would be used only by the `ImageSource` / `ImageDestination` transport and there isn’t really a reason to provide it to callers.

types/types.go
mtrmac: > yes was next step actually :P

:) Never mind me then.

mtrmac:
> One other thing while you are doing this rework could you remove "docker" wherever possible and use oci, if applicable. Since we will be pulling and pushing oci images as well as docker images.

At the moment, this implements Docker Registry (for which there is no corresponding OCI specification) and Registry V2 “Schema 1” manifests (likewise; OCI standardizes “Schema 2”).

The top-level types are named generically, e.g. `ImageSource` (though they do assume a lot about a Docker-like structure at the moment).

Other than that, throughout the code we refer to

 - “Docker manifest digest”, the `$algorithm:$value` hash of a Docker manifest. That one might be made more generic, OTOH OCI defines it by referring to the Docker Registry API doc…
 - “Docker reference”, the `[hostname/]namespace/repo` string format. Many parties may use strings with a similar form, so saying “Docker reference” as opposed to generic “URL-like string” makes sense to me. OTOH I _am_ uncertain about this, especially when using the “Docker reference” term in signatures to refer to OpenShift images (which can't be pulled using a plain `docker pull` with the string being signed). I would welcome independent review whether we should be using the “docker reference” format and naming for OpenShift.

 (Note: We do separate Docker and OpenShift in the UI, `docker://host/namespace/repo:tag` vs `atomic:host/namespace/imagestream:tag`; but both look the same, without the `docker://` / `atomic:` prefix, in signatures.)

mtrmac:
> Other than that, throughout the code we refer to

… and then there is the `dockerutils` subpackage, which is almost, but not completely, not specific to Docker.

runcom:
closing this in favor of what @mtrmac is doing (though the conversation was interesting)

PR #53, opened 2016-05-11
Don‘t write the mainfest digest on stdout in (skopeo copy)

mtrmac:
The `dir:` source type does not return the value, the value is
untrusted/not validated, and it is not at all clear why we should print
it in the first place.


runcom:
:+1:

PR #56, opened 2016-05-16
Move directory, docker and openshift from cmd/skopeo to their own subpackages

mtrmac:
Does not change behavior.  This is a straightforward move and update of package references, except for:

- Adding a duplicate definition of `manifestSchema1` to `cmd/skopeo/copy.go`.  This will need to be cleaned up later, for now preferring to make no design changes in this commit.
- Renaming `parseDockerImage` to `NewDockerImage`, to both make it public and consistent with common golang conventions.

### DIFF
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -24,6 +24,20 @@ func manifestLayers(manifest []byte) ([]string, error) {
 	return layers, nil
 }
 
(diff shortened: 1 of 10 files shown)


mtrmac:
Submitting this as a separate PR so that we have a hopefully uncontroversial base to further discuss the API on.

runcom:
Awesome, this is basically like my previous pr but doesn't change any logic (iiuc)

PR #57, opened 2016-05-16
API for signatures

mtrmac:
This adds methods useful for validating signatures to `types.Image`, namely the ability to retrieve signatures, and the supposedly signed identity, and renames `RawManifest` for consistency.

The implementation remains Docker-specific (in particular, not usable for OpenShift), but having the type and API will allow the policy evaluation code to compile, which is on my critical path now.

See comments in the individual commits for details.

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -47,9 +47,12 @@ type ImageDestination interface {
 // Image is a Docker image in a repository.
 type Image interface {
 	// ref to repository?
+	// GetManifest is like ImageSource.GetManifest, but the result is cached; it is OK to call this however often you need.
+	GetManifest() ([]byte, error)
+	// GetSignatures is like ImageSource.GetSignatures, but the result is cached; it is OK to call this however often you need.
+	GetSignatures() ([][]byte, error)
 	Layers(layers ...string) error // configure download directory? Call it DownloadLayers?
 	Manifest() (ImageManifest, error)
-	RawManifest(version string) ([]byte, error)
 	DockerTar() ([]byte, error) // ??? also, configure output directory
 }
 
--- a/types/types.go
+++ b/types/types.go
@@ -30,6 +30,10 @@ type Repository interface {
 
 // ImageSource is a service, possibly remote (= slow), to download components of a single image.
 type ImageSource interface {
+	// GetIntendedDockerReference returns the full, unambiguous, Docker reference for this image, _as specified by the user_
+	// (not as the image itself, or its underlying storage, claims).  This can be used e.g. to determine which public keys are trusted for this image.
+	// May be "" if unknown.
+	GetIntendedDockerReference() string
 	GetManifest() (manifest []byte, unverifiedCanonicalDigest string, err error)
 	GetLayer(digest string) (io.ReadCloser, error)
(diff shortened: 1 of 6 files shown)


types/types.go
```diff
@@ -47,9 +47,12 @@ type ImageDestination interface {
 // Image is a Docker image in a repository.
 type Image interface {
 	// ref to repository?
+	// GetManifest is like ImageSource.GetManifest, but the result is cached; it is OK to call this however often you need.
+	GetManifest() ([]byte, error)
+	// GetSignatures is like ImageSource.GetSignatures, but the result is cached; it is OK to call this however often you need.
+	GetSignatures() ([][]byte, error)
```
runcom: nit, but I don't like Get(er) or setter in Go - calling `image.Signatures` is pretty straightforward and clean WDYT?

runcom: alright, I've seen the comment in commit message - yet I prefer not to use Get|Set - is there anything specific that needs this Get|Set prefix?

runcom: This is valid also for Put (which recall me of HTTP for instance) - is it ok to use something like `UploadManifest`? `SendManifest`? I like it much more

runcom:
apart from my naming conventions nit - this LGTM

mtrmac:
See #58 for more context.

types/types.go
mtrmac: For now the immediate barrier is that there already is a `Manifest` method. But #58 proposes renaming it to `Inspect` anyway…

Dropping the `Get` prefix is perfectly fine with me.

As for `Put`, `Upload`/`Send` are a bit weird with `dirImageDestination`; and generally `ImageDestination` may have ordering dependencies, e.g. see how `openshiftImageDestination` in https://github.com/mtrmac/skopeo/tree/openshift-annotation-sigs only stores signatures locally in `PutSignatures` and uploads them only in `PutManifest`. `Upload`/`Send` could be misleading in such a case — but honestly the whole `ImageSource`/`ImageDestination` will need to be rethought if we ever support non-seekable tarball streams.

runcom: > For now the immediate barrier is that there already is a Manifest method. But #58 proposes renaming it to Inspect anyway…

this is indeed awesome and what I like

runcom: I'm fine merging this as-is given we could remove at least `Get` (https://github.com/projectatomic/skopeo/pull/57#discussion_r63395075)

runcom: > but honestly the whole ImageSource/ImageDestination will need to be rethought if we ever support non-seekable tarball streams.

agree, maybe they could be embedded directly in images and are available only as a mean of developing new internal source|destination exposed by types.Image directy

mtrmac: I do think that the source/destination distinction is useful; but for sources, this PR already goes a long way towards most users using a `types.Image` and not caring about `ImageSource` at all (basically unless you need layer contents, in which case tarball streams and ordering etc. are really important, `Image` gives you the same information and the caching and parsing makes it the preferred interface.).

runcom:
Merging, will defer https://github.com/projectatomic/skopeo/pull/57#discussion_r63392704 to #58

PR #58, opened 2016-05-16
General image API cleanups

mtrmac:
This is a follow-up to #57, I will rebase as needed.

This somewhat cleans up `types.Image` towards becoming the primary inspection endpoint for images:

 - We can now change the API without breaking `skopeo inspect` output format
 - `Image.Manifest` turns into `Image.Inspect`, which parses the manifest, config and layers only; its other former functions are split into new methods.
 - Added comments on how the API types are expected to be used.

This does not yet completely clean up the API (e.g. moving parts of `Layers` to `cmd/skopeo`).

Nor does this make `Image` independent of the underlying Docker implementation yet: (`InspectInfo.Name` returns the parsed Docker ref, and `GetRepositoryTags` needs a Docker client). Non-Docker replacements for this mostly depend on what values we define `skopeo inspect` to return for non-Docker sources.

See comments in the individual commits for details.

### DIFF
--- a/cmd/skopeo/inspect.go
+++ b/cmd/skopeo/inspect.go
@@ -3,10 +3,25 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"time"
 
 	"github.com/Sirupsen/logrus"
 	"github.com/codegangsta/cli"
 )
+
+// inspectOutput is the output format of (skopeo inspect), primarily so that we can format it with a simple json.MarshalIndent.
+type inspectOutput struct {
+	Name          string
+	Tag           string
+	Digest        string
+	RepoTags      []string
+	Created       time.Time
+	DockerVersion string
+	Labels        map[string]string
+	Architecture  string
+	Os            string
+	Layers        []string
+}
 
 var inspectCmd = cli.Command{
 	Name:  "inspect",
@@ -34,7 +49,19 @@ var inspectCmd = cli.Command{
 		if err != nil {
 			logrus.Fatal(err)
 		}
-		out, err := json.MarshalIndent(imgInspect, "", "    ")
+		outputData := inspectOutput{
+			Name:          imgInspect.Name,
+			Tag:           imgInspect.Tag,
+			Digest:        imgInspect.Digest,
+			RepoTags:      imgInspect.RepoTags,
+			Created:       imgInspect.Created,
+			DockerVersion: imgInspect.DockerVersion,
+			Labels:        imgInspect.Labels,
+			Architecture:  imgInspect.Architecture,
+			Os:            imgInspect.Os,
+			Layers:        imgInspect.Layers,
+		}
+		out, err := json.MarshalIndent(outputData, "", "    ")
 		if err != nil {
 			logrus.Fatal(err)
 		}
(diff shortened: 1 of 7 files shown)


mtrmac:
Rebased on top of the removal of `Get` prefixes.

cmd/skopeo/inspect.go
```diff
@@ -22,19 +38,39 @@ var inspectCmd = cli.Command{
...
 		imgInspect, err := img.Inspect()
 		if err != nil {
 			logrus.Fatal(err)
 		}
-		out, err := json.MarshalIndent(imgInspect, "", "    ")
+		manifestDigest, err := dockerutils.ManifestDigest(rawManifest)
+		if err != nil {
+			logrus.Fatalf("Error computing manifest digest: %s", err.Error())
+		}
+		repoTags, err := img.GetRepositoryTags()
```
runcom: I agree with you this does smell here...

docker/docker_image.go
```diff
@@ -62,7 +65,7 @@ func (i *dockerImage) Signatures() ([][]byte, error) {
 	return i.cachedSignatures, nil
 }
 
-func (i *dockerImage) Inspect() (types.ImageManifest, error) {
+func (i *dockerImage) Inspect() (*types.ImageInspectInfo, error) {
```
runcom: No need for pointer here since this is an interface

cmd/skopeo/inspect.go
mtrmac: This improves the interface at least a little. How to do it truly cleanly and in a generic way, without breaking existing users (if any?) is not clear to me yet; I was thinking

```go
if di, ok := img.(DockerImage); ok {
   tags = di.GetDockerRepositoryTags()
} else {
  tags = nil;
}
```
as another small step towards kicking this out of `types.Image`. Eventually; this is not really hurting me :)

docker/docker_image.go
mtrmac: The original `types.ImageManifest` was an interface. `types.ImageInspectInfo` (formerly `types.DockerImageManifest`) is a struct.

runcom: Aww gotcha sorry for the noise

cmd/skopeo/inspect.go
runcom: That type assertion sounds fine to me and it's clean (tags can be omitemtpy when/if marshaling also)

mtrmac: We will eventually need to generalize `dockerImage` to work with `openshiftImageSource`, at that point the type assertion will make sense, because we could test the `tags = nil` path.

Or do you want it done immediately so that the `GetRepositoryTags` method is never added to `types.Image`?

runcom: I'm fine doing it later (hopefully we would freeze the API in the coming weeks so we are still more time to think more about this :) 

We can leave a comment though just in case we forgot about doing this larer

runcom:
LGTM (given https://github.com/projectatomic/skopeo/pull/58#discussion_r63417813)

cmd/skopeo/inspect.go
mtrmac: Added a comment pointing to this thread, within the “Add comments on the use of the API and the general direction” commit.

PR #60, opened 2016-05-17
move dockerutils under docker

runcom:
also remove fixtures pkg as it would clutter godoc (there's not need
to have a .go files with fixtures)


runcom:
@mtrmac PTAL

mtrmac:
I was a bit fond of keeping the fixtures info in the fixtures directory :sob: but I can see the rationale.

Perhaps move `fixtures/info.go` to `fixtures_info_test.go`, and keep it separate from the other `_test.go` files, as they may be added over time? (See how the various `_test.go` files share fixtures in `skopeo/signature`.)

(I was also considering just adding this all to `skopeo/docker` or perhaps naming this `skopeo/docker/utils`, but neither seems better than what you do here.)

runcom:
> (I was also considering just adding this all to skopeo/docker or perhaps naming this skopeo/docker/utils, but neither seems better than what you do here.)

This is fine for me! much cleaner to have `docker/utils`

> Perhaps move fixtures/info.go to fixtures_info_test.go, and keep it separate from the other _test.go files, as they may be added over time? (See how the various _test.go files share fixtures in skopeo/signature.)

moved to `fixtures_info_test.go`

mtrmac:
> > (I was also considering just adding this all to skopeo/docker or perhaps naming this skopeo/docker/utils, but neither seems better than what you do here.)

> This is fine for me! much cleaner to have `docker/utils`

The trouble is that then all users would either have to use the vague `utils.`… naming, or they would end up importing under a `dockerutils` or similar rename anyway. So I do prefer `docker/dockerutils`, actually.

Anyway, this is trivial and not worth reworking the code over and over. Use whichever you like.

runcom:
> The trouble is that then all users would either have to use the vague utils

right, anyhow, this is how golang works also and I like `docker/utils` (which of course will be renamed to `dockerutils` lol)

PR #61, opened 2016-05-17
signature: remove pkg fixtures

runcom:
@mtrmac PTAL I somehow missed this one also

### DIFF
--- a/integration/signing_test.go
+++ b/integration/signing_test.go
@@ -9,7 +9,6 @@ import (
 	"strings"
 
 	"github.com/go-check/check"
-	"github.com/projectatomic/skopeo/signature/fixtures"
 )
 
 const (
@@ -97,7 +96,7 @@ func (s *SigningSuite) TearDownTest(c *check.C) {
 }
 
 func (s *SigningSuite) TestSignVerifySmoke(c *check.C) {
-	manifestPath := "../signature/fixtures/image.manifest.json"
+	manifestPath := "fixtures/image.manifest.json"
 	dockerReference := "testing/smoketest"
 
 	sigOutput, err := ioutil.TempFile("", "sig")
@@ -111,5 +110,5 @@ func (s *SigningSuite) TestSignVerifySmoke(c *check.C) {
 	out, err = exec.Command(skopeoBinary, "standalone-verify", manifestPath,
 		dockerReference, s.fingerprint, sigOutput.Name()).CombinedOutput()
 	c.Assert(err, check.IsNil, check.Commentf("%s", out))
-	c.Assert(string(out), check.Equals, "Signature verified, digest "+fixtures.TestImageManifestDigest+"\n")
+	c.Assert(string(out), check.Equals, "Signature verified, digest "+TestImageManifestDigest+"\n")
 }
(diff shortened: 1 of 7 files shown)


integration/signing_test.go
```diff
@@ -99,8 +98,10 @@ func (s *SigningSuite) TearDownTest(c *check.C) {
 }
 
 func (s *SigningSuite) TestSignVerifySmoke(c *check.C) {
+	// ugly to me :(
 	manifestPath := "../signature/fixtures/image.manifest.json"
 	dockerReference := "testing/smoketest"
+	testImageManifestDigest := "sha256:20bf21ed457b390829cdbeec8795a7bea1626991fda603e0d01b4e7f60427e55"
```
mtrmac: Duplicating a hard-coded value…

integration/signing_test.go
```diff
@@ -113,5 +114,5 @@ func (s *SigningSuite) TestSignVerifySmoke(c *check.C) {
 	out, err = exec.Command(skopeoBinary, "standalone-verify", manifestPath,
 		dockerReference, s.fingerprint, sigOutput.Name()).CombinedOutput()
 	c.Assert(err, check.IsNil, check.Commentf("%s", out))
-	c.Assert(string(out), check.Equals, "Signature verified, digest "+fixtures.TestImageManifestDigest+"\n")
+	c.Assert(string(out), check.Equals, "Signature verified, digest "+testImageManifestDigest+"\n")
```
mtrmac: … so _that’s_ why `fixtures` was a separate package. I have quite forgotten about this.

Wouldn’t it be cleaner to just copy&paste the relevant fixtures to the `integration` package then?

mtrmac:
You’re right the `integration` references to `signature/fixtures` have been ugly… this is not quite helping.

Copy&pasting the `signature` fixtures into `integration` would obviously work, and we probably can live with the duplication (a little cross-package drift might even be beneficial in helping test older versions of the data).

Alternatively, a separate `signature-fixtures` subpackage shared between all test users? That would be cleaner than `"../signature"` but there would still be no way to hide this from `godoc` I guess.

runcom:
> Copy&pasting the signature fixtures into integration would obviously work, and we probably can live with the duplication (a little cross-package drift might even be beneficial in helping test older versions of the data).

That's the point, I don't think it won't hurt anyone by embedding the stuff from `../signature` in that integration test (which is smoke test and should continue to work or break when stuff change). I think that's the only stop where `signature` folder and/or pkg is read/imported so I wouldn't mind copy/pasting the manifest & digest directly in that test. We can change it later if we really need to.

runcom:
@mtrmac are you fine with https://github.com/projectatomic/skopeo/pull/61#issuecomment-220407437?

mtrmac:
>> (a little cross-package drift might even be beneficial in helping test older versions of the data).
> That's the point,

I am a bit confused, you seem to agree with me here but then you seem to be suggesting keeping the `../signature` reference; to be clear, I was thinking about something like https://github.com/mtrmac/skopeo/tree/signature-fixtures .

runcom:
> to be clear, I was thinking about something like https://github.com/mtrmac/skopeo/tree/signature-fixtures .

That's indeed what I was thinking (sorry for the confusion)

runcom:
@mtrmac updated

PR #62, opened 2016-05-18
[WIP] adapt code and interfaces to MIME types

runcom:
This is a WIP to support more MIME types when fetching manifests - in particular this aims to provide support for manifest list (from Docker) -> see https://integratedcode.us/2016/04/22/a-step-towards-multi-platform-docker-images/ (I personally want this here in skopeo :))

@mtrmac PTAL

- First commit just adapts code to ask/use/retrieve MIME types (and breaks `inspect` command :smile: till we do the point below)
- TODO: choose v2s1 to maintain backcompatibility -> this means when asking for a manifest we should also provide the mime type we would like to have back (falling back to a default if anything's wrong) **OR** we can do the manifestlist->v2s1 conversion when needed (cmd/skopeo/inspect)
- TODO: command line

### DIFF
--- a/docker/docker_image.go
+++ b/docker/docker_image.go
@@ -19,9 +19,10 @@ var (
 )
 
 type dockerImage struct {
-	src              *dockerImageSource
-	cachedManifest   []byte   // Private cache for Manifest(); nil if not yet known.
-	cachedSignatures [][]byte // Private cache for Signatures(); nil if not yet known.
+	src                    *dockerImageSource
+	cachedManifest         []byte // Private cache for Manifest(); nil if not yet known.
+	cachedManifestMIMEType string
+	cachedSignatures       [][]byte // Private cache for Signatures(); nil if not yet known.
 }
 
 // NewDockerImage returns a new Image interface type after setting up
@@ -42,15 +43,16 @@ func (i *dockerImage) IntendedDockerReference() string {
 }
 
 // Manifest is like ImageSource.GetManifest, but the result is cached; it is OK to call this however often you need.
-func (i *dockerImage) Manifest() ([]byte, error) {
+func (i *dockerImage) Manifest() ([]byte, string, error) {
 	if i.cachedManifest == nil {
-		m, err := i.src.GetManifest()
+		m, mt, err := i.src.GetManifest()
 		if err != nil {
-			return nil, err
+			return nil, "", err
 		}
 		i.cachedManifest = m
+		i.cachedManifestMIMEType = mt
 	}
-	return i.cachedManifest, nil
+	return i.cachedManifest, i.cachedManifestMIMEType, nil
 }
 
 // Signatures is like ImageSource.GetSignatures, but the result is cached; it is OK to call this however often you need.
@@ -172,7 +174,7 @@ func sanitize(s string) string {
 }
 
 func (i *dockerImage) getSchema1Manifest() (manifest, error) {
-	manblob, err := i.Manifest()
+	manblob, _, err := i.Manifest()
 	if err != nil {
 		return nil, err
 	}
--- a/docker/docker_image.go
+++ b/docker/docker_image.go
@@ -7,10 +7,12 @@ import (
 	"io/ioutil"
 	"net/http"
 	"regexp"
+	"runtime"
 	"strings"
 	"time"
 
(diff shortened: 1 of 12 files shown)


runcom:
this is really important though as if you _get_ the manifest from `docker://docker.io/estesp/busybox`  you get a manifest list. We now have to understand _how_ to expose this. We can (as Docker does) detect os and architecture an print the one skopeo is running on **by default**. Then provide a new option to print just a specific os/architecture. How does this sound? Also, where the actual `switch` on the mime type should go? image_src? or directly into image?

runcom:
The second commit does a `switch` to understand if dealing with a manifest list to still provide a v2s1 struct.

@mtrmac what's needed signing side to support this? I'd like to first introduce this w/o breaking anything and then provide flags to select which manifest in a list to select.

runcom:
Also does signing works with v2s2? Right now we don't fetch this manifest media type at all (with this change we could however)

docker/docker_image.go
```diff
@@ -172,10 +186,39 @@ func sanitize(s string) string {
...
+				digest = m.Digest
+				break
+			}
+		}
+		if digest == "" {
+			return nil, errors.New("no supported platform found in manifest list")
+		}
+		url := fmt.Sprintf(manifestURL, i.src.ref.RemoteName(), digest)
+		// just Accept v2s1 by not providing Accept headers here
+		res, err := i.src.c.makeRequest("GET", url, nil, nil)
```
runcom: this is ugly done here...

runcom:
closing in favor or #67

docker/docker_image.go
mtrmac: I think it would be clean to add a `desiredMIMEType` (with `""` meaning “default/best available”) to `ImageSource.GetManifest`.

mtrmac:
> We now have to understand _how_ to expose this. We can (as Docker does) detect os and architecture an print the one skopeo is running on **by default**.
Yes.

> Then provide a new option to print just a specific os/architecture. How does this sound?

We were talking about encoding this into the URL-like string (`docker-with-arch:arm64,busybox:latest` or something) earlier; but you were right, this should perhaps be an option, or env variable instead; otherwise we would have to define such syntax for every one of the string formats (`docker`/`atomic`/`dir`).

> Also, where the actual `switch` on the mime type should go? image_src? or directly into image?

Ouch. It is kind of weird for `ImageSource` to actually be a source for a set of images (“but not a repository!”), but overall it seems cleanest:

- Having the MIME type switch and handling in `dockerImage`, near the only code parsing manifests and understanding their contents, is nice.
- Having a single `ImageSource` represent all the archs simultaneously does feel ugly, but it also naturally lends itself into a supporting an “copy all archs of an image from a source to destination” operation, which is sure to come up eventually — and without having to worry the typical “just give me the default arch” user about manifest lists at all: Who needs the default image, will just call `GetManifest`; who needs all of them, will call an arch-aware API of `ImageSource`.

mtrmac:
> The second commit does a switch to understand if dealing with a manifest list to still provide a v2s1 struct.

(Perhaps such things could be rearranged to support the new format first, and start pulling it second; then nothing would break `git bisect`.)

> @mtrmac what's needed signing side to support this? I'd like to first introduce this w/o breaking anything and then provide flags to select which manifest in a list to select.

For v2s2 non-lists, nothing needs to happen: we just hash the manifest blob.

For manifest lists, AFAICT there is nothing to guarantee the different archs of an image have anything in common (layers, files, labels, UIDs, entrypoints, nothing), so it seems reasonable that each one would be signed separately. Per the above discussion about arch-aware and arch-unaware API, and `Image` but not `ImageSource` parsing manifest lists, for now we can define `Image.Manifest` to return the non-list manifest for the default arch, and then the policy evaluation I am working on, which refers to `types.Image`, should transparently work with the default arch. (This would definitely benefit from testing though!)

The existing `skopeo standalone-sign` and `skopeo standalone-verify` commands work with manifests as blobs, so the responsibility for working with lists would be up to the user.

`skopeo copy` will, however, need to know about manifests lists: right now it only copies the default manifest (= the list) from the source; neither pushing nor signing that manifest would work. So, after all, the “given a default manifest, get me a non-list manifest for the selected arch” functionality might need to be a publicly available utility, not internal to `dockerImage`, we’l see.

PR #65, opened 2016-05-18
provide a way to pass multi values-signle key headers

runcom:
There's currently no way to pass more values for a single header, this fixes it

@mtrmac PTAL

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -78,7 +78,7 @@ func newDockerClient(refHostname, certPath string, tlsVerify bool) (*dockerClien
 	}, nil
 }
 
-func (c *dockerClient) makeRequest(method, url string, headers map[string]string, stream io.Reader) (*http.Response, error) {
+func (c *dockerClient) makeRequest(method, url string, headers map[string][]string, stream io.Reader) (*http.Response, error) {
 	if c.scheme == "" {
 		pr, err := c.ping()
 		if err != nil {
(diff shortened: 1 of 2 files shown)


docker/docker_client.go
```diff
@@ -78,7 +78,7 @@ func newDockerClient(refHostname, certPath string, tlsVerify bool) (*dockerClien
 	}, nil
 }
 
-func (c *dockerClient) makeRequest(method, url string, headers map[string]string, stream io.Reader) (*http.Response, error) {
+func (c *dockerClient) makeRequest(method, url string, headers map[string][]string, stream io.Reader) (*http.Response, error) {
```
mtrmac: Would this work, and be cleaner, to use the existing `http.Header` typedef? Especially looking at `docker_image_dest` having to copy this type declaration.

mtrmac:
LGTM. One possible cleanup, but if you don’t agree or don’t see the value (or if it doesn’t work :)) I’d be happy to merge as is.

docker/docker_client.go
runcom: I firstly used http.Header but later changed just because I wanted to keep this simple (wrt types) because we might change for instance the `http` pkg (hard we do this lol) and don't want to change that all over the place - so a map[string][]string is generic enough :smile:

mtrmac: This function returns `http.Response`, so we are not transparently replacing `http`. Works for me either way.

PR #67, opened 2016-05-19
Mimetypes

runcom:
These are some changes from https://github.com/projectatomic/skopeo/pull/62 that don't break skopeo in any way but lay the ground for future work (#62-like)

This depends on #61 and #65 (I'll rebase after those are merged)

@mtrmac PTAL

### DIFF
--- a/integration/fixtures/image.manifest.json
+++ b/integration/fixtures/image.manifest.json
@@ -1,0 +1,26 @@
+{
+    "schemaVersion": 2,
+    "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
+    "config": {
+        "mediaType": "application/vnd.docker.container.image.v1+json",
+        "size": 7023,
+        "digest": "sha256:b5b2b2c507a0944348e0303114d8d93aaaa081732b86451d9bce1f432a537bc7"
+    },
+    "layers": [
+        {
+            "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
+            "size": 32654,
+            "digest": "sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f"
+        },
+        {
+            "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
+            "size": 16724,
+            "digest": "sha256:3c3a4604a545cdc127456d94e421cd355bca5b528f4a9c1905b15da2eb4a4c6b"
+        },
+        {
+            "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
+            "size": 73109,
+            "digest": "sha256:ec4b8955958665577945c89419d1af06b5f7636b4ac3da7f12184802ad867736"
+        }
+    ]
+}
--- a/integration/fixtures_info_test.go
+++ b/integration/fixtures_info_test.go
@@ -1,0 +1,6 @@
+package main
+
+const (
+	// TestImageManifestDigest is the Docker manifest digest of "fixtures/image.manifest.json"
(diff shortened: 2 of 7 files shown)


docker/utils/manifest.go
```diff
@@ -10,11 +10,22 @@ import (
...
 // FIXME: Should we just use docker/distribution and docker/docker implementations directly?
 
+// GetManifestMIMETypes returns a slice of MIME types we currently support
+func GetManifestMIMETypes() []string {
+	return []string{
+		V2Schema1MIMEType,
+		V2Schema2MIMEType,
+		V2ListMIMEType,
+	}
+}
```
mtrmac: Could just be a public `var`? OTOH that’s perhaps too risky.

Drop the `Get` prefix though :)

docker/utils/manifest.go
```diff
@@ -10,11 +10,22 @@ import (
...
-	// DockerV2Schema1MIMEType MIME type represents Docker manifest schema 1
-	DockerV2Schema1MIMEType = "application/vnd.docker.distribution.manifest.v1+json"
-	// DockerV2Schema2MIMEType MIME type represents Docker manifest schema 2
-	DockerV2Schema2MIMEType = "application/vnd.docker.distribution.manifest.v2+json"
+	// V2Schema1MIMEType MIME type represents Docker manifest schema 1
+	V2Schema1MIMEType = "application/vnd.docker.distribution.manifest.v1+json"
+	// V2Schema2MIMEType MIME type represents Docker manifest schema 2
+	V2Schema2MIMEType = "application/vnd.docker.distribution.manifest.v2+json"
+	// V2ListMIMEType MIME type represents Docker *fat* manifest aka manifest list
+	V2ListMIMEType = "application/vnd.docker.distribution.manifest.list.v2+json"
```
mtrmac: Please keep the `Docker` naming here.

runcom: I will, I forgot to put it back

docker/utils/manifest.go
runcom: > Drop the Get prefix though :)

ouch

cmd/skopeo/copy.go
```diff
@@ -54,7 +54,7 @@ func copyHandler(context *cli.Context) {
 	}
 	signBy := context.String("sign-by")
 
-	manifest, err := src.GetManifest()
+	manifest, _, err := src.GetManifest()
```
mtrmac: (This is the right thing to do now.

Eventually we might need to do conversions, i.e. turn the core of `copy` into a function available in the library (which would be called for `pull` operation by k8s).  But that’s for much later, if ever.)

types/types.go
```diff
@@ -37,8 +37,7 @@ type ImageSource interface {
 	// May be "" if unknown.
 	IntendedDockerReference() string
 	// GetManifest returns the image's manifest.  It may use a remote (= slow) service.
-	// FIXME? This should also return a MIME type if known, to differentiate between schema versions.
-	GetManifest() ([]byte, error)
+	GetManifest() ([]byte, string, error)
```
mtrmac: Please update the comment also to say what the second parameter is, and that "" means “unknown”.

mtrmac:
I can’t see any dependency on #65 FWIW.

runcom:
> I can’t see any dependency on #65 FWIW.

I had included the commit here - so now it's merged and I've rebased.

Also, I've fixed what you requested - after #61 is merged I will rebase here and we could merge this also if we want

mtrmac:
Thanks, LGTM. Will merge after we agree on what we mean in #61 :)

runcom:
I'll reopen once I re-fix this

PR #71, opened 2016-05-24
remove logging from libraries and cleanup

runcom:
Logging inside libraries/pkgs is ugly - callers should log instead.
Also, move `ManifestMIMETypes` directly into `docker_img_src` where it will be later used (and un-export it, no need to have this in the API, the const(s) suffice)

@mtrmac PTAL

### DIFF
--- a/docker/docker_image_dest.go
+++ b/docker/docker_image_dest.go
@@ -7,7 +7,6 @@ import (
 	"io/ioutil"
 	"net/http"
 
-	"github.com/Sirupsen/logrus"
 	"github.com/projectatomic/skopeo/docker/utils"
 	"github.com/projectatomic/skopeo/reference"
 	"github.com/projectatomic/skopeo/types"
@@ -61,11 +60,10 @@ func (d *dockerImageDestination) PutManifest(manifest []byte) error {
 	defer res.Body.Close()
 	if res.StatusCode != http.StatusCreated {
 		body, err := ioutil.ReadAll(res.Body)
-		if err == nil {
-			logrus.Debugf("Error body %s", string(body))
+		if err != nil {
+			return err
 		}
-		logrus.Debugf("Error uploading manifest, status %d, %#v", res.StatusCode, res)
-		return fmt.Errorf("Error uploading manifest to %s, status %d", url, res.StatusCode)
+		return fmt.Errorf("Error uploading manifest to %s, status %d, %#v, %s", url, res.StatusCode, res, body)
 	}
 	return nil
 }
@@ -73,21 +71,18 @@ func (d *dockerImageDestination) PutManifest(manifest []byte) error {
 func (d *dockerImageDestination) PutLayer(digest string, stream io.Reader) error {
 	checkURL := fmt.Sprintf(blobsURL, d.ref.RemoteName(), digest)
 
-	logrus.Debugf("Checking %s", checkURL)
 	res, err := d.c.makeRequest("HEAD", checkURL, nil, nil)
 	if err != nil {
 		return err
 	}
 	defer res.Body.Close()
 	if res.StatusCode == http.StatusOK && res.Header.Get("Docker-Content-Digest") == digest {
-		logrus.Debugf("... already exists, not uploading")
+		// already exists, not uploading
 		return nil
 	}
-	logrus.Debugf("... failed, status %d", res.StatusCode)
 
 	// FIXME? Chunked upload, progress reporting, etc.
 	uploadURL := fmt.Sprintf(blobUploadURL, d.ref.RemoteName(), digest)
-	logrus.Debugf("Uploading %s", uploadURL)
 	// FIXME: Set Content-Length?
 	res, err = d.c.makeRequest("POST", uploadURL, map[string][]string{"Content-Type": {"application/octet-stream"}}, stream)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


docker/docker_image_dest.go
```diff
@@ -61,41 +60,36 @@ func (d *dockerImageDestination) PutManifest(manifest []byte) error {
 	defer res.Body.Close()
 	if res.StatusCode != http.StatusCreated {
 		body, err := ioutil.ReadAll(res.Body)
-		if err == nil {
-			logrus.Debugf("Error body %s", string(body))
+		if err != nil {
+			return err
```
mtrmac: I think we should be reporting the HTTP status code, instead of the error reading the rest of the body, here.

runcom: make sense

docker/docker_image_src.go
```diff
@@ -55,6 +55,15 @@ func (s *dockerImageSource) IntendedDockerReference() string {
 	return fmt.Sprintf("%s:%s", s.ref.Name(), s.tag)
 }
 
+// ManifestMIMETypes returns a slice of supported MIME types
+func manifestMIMETypes() []string {
```
mtrmac: What does “supported” mean here, exactly, considering that all callers in #73 only use V2S1?

Maybe we just don’t need this function?

runcom: that's probably why I un-exported it in the first place - will remove it

mtrmac:
Logging of the HTTP requests and replies is really useful, though, and can’t be replaced by logging by callers of the outer API.

It would be perfectly fine not log by default, but I would very much like (`skopeo --debug copy`) to keep logging. Perhaps add an (optional?) logger parameter, or a `SetLogger` method to the client, source/destination, and image?

mtrmac:
… at the very least comment the log lines out to make them easy to restore when debugging, instead of dropping them entirely.

runcom:
I think there's something wrong in logging inside libraries. If the libraries are well done, eventually what you need just lay outside the api

mtrmac:
No; if a library makes 10 HTTP requests to 5 different servers to obtain a single result, the API caller ordinarily does not care which one failed in which specific way (unless it wants to specifically react to that failure); but for debugging or diagnosing a system failure, the programmer (= me) or the sysadmin needs to know which one failed and how, even if the program does not know how to react to the failure—because the human does know.

mtrmac:
Passing a logger instance into the API would make the API side-effect-free, and would allow both not logging anything by default, and getting the information when necessary.

runcom:
alright, I'm going ahead and use `context.Context` (https://blog.golang.org/context). I've found it very useful in upstream `docker/distribution` pkg where top layer commands configure a context with a logger and the desired logging level. Then every interface/library user can configure the context as she wishes. This also means requiring each methods of our interfaces should now receive a context arg as its first arg - and I really like it (we refactored all the code in docker to use this pattern, even then a context is not specifically needed)

Are you fine with this?

mtrmac:
What would that look like? AFAICS `context` does nothing for logging explicitly, and using it naively seems to end up pretty ugly (`img.ctx.Value(loggerKey).(Logger).Debugf(…)`?)

I kind of liked how https://golang.org/doc/effective_go.html#embedding does this — or even have a private `log *log.Logger` field.

mtrmac:
(If you don’t want to deal with this, feel free to comment the logging out for now — just please do not throw it away completely.)

runcom:
I want to find the best approach otherwise I'll leave those there don't worry! However, that quite works by embedding a logger instance in the context (can't explain it all here but the implementation in docker/distribution/context is pretty simple

mtrmac:
True, with the helper `context.GetLogger(img.ctx).Errorf(…)` is somewhat better (but still pretty verbose) - OTOH note how rarely `GetLogger` is used in Docker.

runcom:
Do you think the fact that it's rare they call the logger from the context is for performance reasons? I don't understand if you like it or not because I do (generally having a context object  in the API is also good)

mtrmac:
No, I just mean that it is not an universal pattern in there. (Though, Docker seems to be directly calling `logrus.`$something all over the place, and you’re right that this is not a pattern to follow in a library.)

---

As for context objects in general, I am perfectly fine with _specific_ context objects carrying _specific_ state appropriate for the current use. An untyped, compile-time-unverifiable, difficult-to-grep `WithValue`/`Value` information transfer system feels a bit like giving up on the benefits of static typing; especially future API changes would be more difficult when users of the old API would keep compiling. Certainly that kind of abstraction may be necessary to transfer data across libraries from time to time, but as an external-facing API it is not thrilling.

The only thing `x/net/context.Context` actually _does_ is timeouts and cancellation, which we don’t currently implement at all. Now, certainly, that could be very useful for e.g. `GetManifest`/`GetLayer`, but it is pretty useless for parsing signatures. It is not obvious to me that the two should share much context or state.

Certainly we can have a `SomethingContext` class appropriate for HTTP and signature parsing and for any other operations within the `skopeo` project; that class that would essentially end up being a “service locator”, and implicitly bundle all external dependencies into an indivisible whole.

(OTOH the policy evaluation engine in mtrmac/policy-eval implicitly calls `Image.Manifest` which chains to `ImageSource.GetManifest`, so perhaps they do share a lot, often enough?)

---

I don’t know. I guess tl;dr right now it feels weird to use `x/net/context.Context` for things it is not particularly good at, and to not use it at all for those things it was designed to do well, but I might just lack the foresight.

runcom:
closing for now, but this is something worth looking into (I'll add an RFE to my todo list)

PR #73, opened 2016-05-24
add the possibility to choose image's MIME type

runcom:
This is tough, but as a user of skopeo libraries/pkgs I wish to be able to provide a list of mime types which I wish to get back (in order). Then I can check the returned mime type to understand it.

@mtrmac PTAL

### DIFF
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -55,11 +55,13 @@ func (s *dockerImageSource) IntendedDockerReference() string {
 	return fmt.Sprintf("%s:%s", s.ref.Name(), s.tag)
 }
 
-func (s *dockerImageSource) GetManifest() ([]byte, string, error) {
+func (s *dockerImageSource) GetManifest(mimetypes []string) ([]byte, string, error) {
 	url := fmt.Sprintf(manifestURL, s.ref.RemoteName(), s.tag)
 	// TODO(runcom) set manifest version header! schema1 for now - then schema2 etc etc and v1
 	// TODO(runcom) NO, switch on the resulter manifest like Docker is doing
-	res, err := s.c.makeRequest("GET", url, nil, nil)
+	headers := make(map[string][]string)
+	headers["Accept"] = mimetypes
+	res, err := s.c.makeRequest("GET", url, headers, nil)
 	if err != nil {
 		return nil, "", err
 	}
--- a/types/types.go
+++ b/types/types.go
@@ -27,8 +27,9 @@ type ImageSource interface {
 	// (not as the image itself, or its underlying storage, claims).  This can be used e.g. to determine which public keys are trusted for this image.
 	// May be "" if unknown.
 	IntendedDockerReference() string
-	// GetManifest returns the image's manifest along with its MIME type. The empty string is returned if the MIME type is unknown.  It may use a remote (= slow) service.
-	GetManifest() ([]byte, string, error)
+	// GetManifest returns the image's manifest along with its MIME type. The empty string is returned if the MIME type is unknown. The slice parameter indicates the supported mime types the manifest should be when getting it.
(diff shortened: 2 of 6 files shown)


types/types.go
```diff
@@ -28,7 +28,7 @@ type ImageSource interface {
 	// May be "" if unknown.
 	IntendedDockerReference() string
 	// GetManifest returns the image's manifest along with its MIME type. The empty string is returned if the MIME type is unknown.  It may use a remote (= slow) service.
-	GetManifest() ([]byte, string, error)
+	GetManifest([]string) ([]byte, string, error)
```
mtrmac: Document the parameter, at least in one of the parameter name and the comment.

docker/docker_image_src.go
```diff
@@ -55,11 +55,13 @@ func (s *dockerImageSource) IntendedDockerReference() string {
...
 
-func (s *dockerImageSource) GetManifest() ([]byte, string, error) {
+func (s *dockerImageSource) GetManifest(mimetypes []string) ([]byte, string, error) {
 	url := fmt.Sprintf(manifestURL, s.ref.RemoteName(), s.tag)
 	// TODO(runcom) set manifest version header! schema1 for now - then schema2 etc etc and v1
 	// TODO(runcom) NO, switch on the resulter manifest like Docker is doing
-	res, err := s.c.makeRequest("GET", url, nil, nil)
+	headers := make(map[string][]string)
+	headers["Accept"] = mimetypes
+	res, err := s.c.makeRequest("GET", url, headers, nil)
```
mtrmac: Sorry, I should have researched this back at #65 , but this is invalid HTTP.

Per https://tools.ietf.org/html/rfc7230 3.2.2
> A sender MUST NOT generate multiple header fields with the same field name in a message unless either the entire field value for that header field is defined as a comma-separated list…
and
https://tools.ietf.org/html/rfc7231 5.3.2 defines `Accept` as semicolon-separated, not comma-separated.

mtrmac: AFAICT #65 would be needed only for the `Set-Cookie` header.

runcom: there's probably something wrong in every go client around (docker/distribution in primis)
They actually `set` and `add` the `Accept` header

runcom:
```
Accept: application/vnd.docker.distribution.manifest.list.v2+json;application/vnd.docker.distribution.manifest.v1+json;application/vnd.docker.distribution.manifest.v2+json
```
A call with this accept header isn't even working for a registry (at least the docker hub) (you don't get the manifest list here....)

runcom: nor with comma, or space+comma or space+semicolon..

mtrmac: Yeah, `github.com/docker/distribution/registry/handlers.imageManifestHandler.GetImageManifest` is clearly non-conforming. \*sigh\*.

This looks good, then.

types/types.go
runcom: updated PTAL

mtrmac:
Per IRC discussion: setting the MIME type list to `nil` will be treated as “all supported non-list MIME types”, intended to allow getting a blob matching the required digest look nice.

mtrmac:
Also, only `ImageSource.GetManifest` allows supplying MIME types; the intent is for clients who want to parse the manifests to use an `ImageSource.

Clients who want to sue `skopeo`’s parsing should use `types.Image`, and then they don’t need to care about MIME types.  In fact, `types.Image` needs to decide among the various manifest alternatives which one to parse (and which one to match against the provided or signed manifest digest). So, `Image.Manifest` will not be all that useful for parsing the contents, it is basically useful only for verifying against a digest; therefore it will not have a MIME type parameter.

PR #74, opened 2016-05-24
Add SigningMechanism.ImportKeysFromBytes

mtrmac:
This will be needed for verification against specified public keys.

Also rerun `hack/vendor.sh` to pick up import support from `github.com/mtrmac/gpgme` .


signature/mechanism.go
```diff
@@ -13,6 +13,10 @@ import (
 // FIXME: Eventually expand on keyIdentity (namespace them between mechanisms to
 // eliminate ambiguities, support CA signatures and perhaps other key properties)
 type SigningMechanism interface {
+	// ImportKeysFromBlob imports public keys from the supplied blob and returns their identities.
+	// The blob is assumed to have an appropriate format (the caller is expected to know which one).
+	// NOTE: This may modify long-term state (e.g. key storage in a directory underlying the mechanism).
+	ImportKeysFromBlob(blob []byte) ([]string, error)
```
runcom: Can we call this `FromBytes` (I've seen this suffix pretty often when the arg is a byte slice) OTOH I think just `ImportKeys` is fine if we don't plan to import keys from other  stuff

runcom:
Just a nit, LGTM otherwise, let me know if you're fine with my comment, otherwise again I'll merge

signature/mechanism.go
runcom: much as you did here https://github.com/projectatomic/skopeo/pull/55/files#diff-230170c6339d5a6b9de551a0aeb6e74eR44

mtrmac:
Good point, renamed.

PR #75, opened 2016-05-24
Add docker/utils.ManifestMatchesDigest

mtrmac:
As opposed to callers just calling `utils.ManifestDigest(image.Manifest())`, this is a forward-compatible interface, allowing other digest algorithms to be added in the future.

Right now, we only support SHA-256, so the underlying implementation does not change anything.

### DIFF
--- a/docker/utils/manifest.go
+++ b/docker/utils/manifest.go
@@ -64,3 +64,16 @@ func ManifestDigest(manifest []byte) (string, error) {
 	hash := sha256.Sum256(manifest)
 	return "sha256:" + hex.EncodeToString(hash[:]), nil
 }
+
+// ManifestMatchesDigest returns true iff the manifest matches expectedDigest.
+// Error may be set if this returns false.
+// Note that this is not doing ConstantTimeCompare; by the time we get here, the cryptographic signature must already have been verified,
+// or we are not using a cryptographic channel and the attacker can modify the digest along with the manifest blob.
+func ManifestMatchesDigest(manifest []byte, expectedDigest string) (bool, error) {
+	// This should eventually support various digest types.
+	actualDigest, err := ManifestDigest(manifest)
+	if err != nil {
+		return false, err
+	}
+	return expectedDigest == actualDigest, nil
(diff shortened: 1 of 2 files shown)


mtrmac:
(Right now I need this for an `types.Image`, and with the various MIME types giving the full information to a `types.Image` seems reasonable. OTOH eventually the core should probably go into `docker/utils`, and it should also be used by `VerifyDockerManifestSignature`.

Feel free to defer merging this after I get to cleaning up `VerifyDockerManifestSignature`.)

docker/docker_image.go
```diff
@@ -54,6 +54,23 @@ func (i *dockerImage) Manifest() ([]byte, error) {
...
+	manifest, err := i.Manifest()
+	if err != nil {
+		return false, err
+	}
+	actualDigest, err := utils.ManifestDigest(manifest)
+	if err != nil {
+		return false, err
+	}
+	// Note that this is not doing ConstantTimeCompare; by the time we get here, the cryptographic signature must already have been verified,
+	// or we are not using a cryptographic channel and the attacker can modify the digest along with the manifest blob.
```
runcom: can you put this comment in the func comment above? so that godoc is aware of this and users also

runcom:
> Feel free to defer merging this after I get to cleaning up VerifyDockerManifestSignature.)

I'd rather wait for this than merging :)

docker/docker_image.go
mtrmac: Good point, will update.

mtrmac:
> > Feel free to defer merging this after I get to cleaning up VerifyDockerManifestSignature.)
> I'd rather wait for this than merging :)
Sure, marking as WIP for now.

mtrmac:
(Currently depends on, and includes, #79. I will rebase as needed.)

mtrmac:
Moved the functionality from `types.Image` directly to `docker/utils`; after that, having a helper in `types.Image` did not seem all that useful.

Now with a test, and ready to be merged.

PR #79, opened 2016-05-28
Make types.Image Docker-independent, add docker.GenericImageFromSource

mtrmac:
Following up on the discussions in #52 and #58 , this allows using `types.Image` for non-Docker `types.ImageSource` backends, instead moving the Docker-specific knowledge into `docker.Image` and the `skopeo inspect` implementation.

This will be useful for policy evaluation unit tests, by allowing testing against a `dir:…` `types.Image` without having to construct testing-only mocks and reimplement `Image` functionality like `MatchesManifestDigest` (#75) in the mocks.

As is, this is not a complete API cleanup (e.g. having the generic Image constructor as `docker.GenericImageFromSource` is clearly not the right thing to do) but it does move us closer to the desired end state, at least in the implementation if not in the external API.

See the individual commits for detailed comments.


runcom:
:+1: fine to me for now, feel free to merge it!

Issue #89, opened 2016-06-01
go1.4.2 compatibility


runcom:
Building skopeo in RHEL7 with go1.4.2 gives:
```
$ /usr/local/go1.4.2/go/bin/go build -o skopeo ./cmd/skopeo
# github.com/projectatomic/skopeo/signature
src/github.com/projectatomic/skopeo/signature/json.go:68: dec.Token undefined (type *json.Decoder has no field or method Token)
src/github.com/projectatomic/skopeo/signature/json.go:72: undefined: json.Delim
src/github.com/projectatomic/skopeo/signature/json.go:76: dec.Token undefined (type *json.Decoder has no field or method Token)
src/github.com/projectatomic/skopeo/signature/json.go:80: undefined: json.Delim
src/github.com/projectatomic/skopeo/signature/json.go:103: dec.Token undefined (type *json.Decoder has no field or method Token)
```
The fastest thing we can do to deal with this is to create a file `signature/json_go1.4.2.go` (or the like) and put there a build tag to be compiled only with go1.4.2 and lower - and another file to build with go > 1.4.2 (which has the undefined methods above in the errors). We can just move there `paranoidUnmarshalJSONObject` so we have two version of this function for the different go versions
.
@mtrmac sounds good to you? is `paranoidUnmarshalJSONObject` fine to be reimplemented w/o using `Token` and `Delim` which aren't available in go1.4.2?

mtrmac:
https://github.com/mtrmac/skopeo/tree/go142 , but please do not merge this. Let those who need to build against the old version apply this manually.

runcom:
Correct, let's keep this open for a while. I'll close it as soon as we switch to go 1.5+ in at least RHEL
Thanks

runcom:
@cgwalters then I think this patch https://github.com/mtrmac/skopeo/tree/go142 needs to be applied - same as we do for RHEL

cgwalters:
The patch strategy doesn't really work with the idea of continuous delivery, which is what CAHC is trying to do.  At least nontrivial patches to code which is likely to change.

I can freeze for now, but it seems clear that we need to pull in a newer golang.  This would be a good on-list discussion.

runcom:
> I can freeze for now, but it seems clear that we need to pull in a newer golang. This would be a good on-list discussion.

great

cgwalters:
[...]

PR #91, opened 2016-06-02
DO NOT MERGE: Go142 test

mtrmac:
Only to test #46 with #89 fixed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,4 +7,4 @@ notifications:
   email: false
 
 script:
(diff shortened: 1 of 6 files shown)


mtrmac:
As of https://rhn.redhat.com/errata/RHSA-2016-1538.html / https://lists.centos.org/pipermail/centos-announce/2016-August/022005.html , CentOS has golang 1.6.3. so this is no longer relevant.

PR #93, opened 2016-06-07
OpenShift native signatures

mtrmac:
This adds support for storing signatures in Atomic Registry, assuming (currently unmerged) server-side support per https://github.com/openshift/origin/pull/8371 and https://github.com/openshift/origin/pull/9181 .

This should eventually replace #44.

(Marked as WIP because the server side is missing, and because this currently requires a cluster-wide “update signatures” privilege, which seems impractical, so future changes of the API cannot be ruled out.  Also, this is currently done as 2 commits to preserve the 2 possible ways to upload signatures, after the decisions are final they should be squashed.)


mtrmac:
Updated for the final version of https://github.com/containers/image/pull/14 , I will re-vendor after that is merged.

mtrmac:
(Conceptually, this depends on https://github.com/projectatomic/skopeo/pull/146 to update the server to use the same API.   But we haven’t merged integration tests for signatures yet, so this inconsistency does not show up as a test failure yet. So, filing this as a separate PR for simpler review.)

runcom:
feel free to merge containers/image#14 and revendor here - I'll merge those in skopeo

mtrmac:
Revendored, please ACK.

runcom:
ack

PR #95, opened 2016-06-09
Card container_security_113 - Delete image support

jwhonce:
Add support to mark images for deletion from repository

Requires:
  * V2 API and schema
  * registry configured to allow deletes
  * run registry garbage collection to free up disk space

Signed-off-by: Jhon Honce <jhonce@redhat.com>

### DIFF
--- a/cmd/skopeo/delete.go
+++ b/cmd/skopeo/delete.go
@@ -1,0 +1,27 @@
+package main
+
+import (
+	"github.com/Sirupsen/logrus"
+	"github.com/codegangsta/cli"
+)
+
+func deleteHandler(context *cli.Context) {
+	if len(context.Args()) != 1 {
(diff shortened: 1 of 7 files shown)


jwhonce:
@runcom @mtrmac Please review and merge.  Thanks

rhatdan:
Does skopeo have any other interactive routines?  If not, I think we should just delete the image with out prompting the user for confirmation.  Then we could put the confirmation into the atomic command.

jwhonce:
@rhatdan I can move that functionality. I tend to error on the safe side of delete.

runcom:
Agree, I think atomic should handle the interactive confirmation

jwhonce:
@rhatdan @runcom Code updated

runcom:
LGTM @mtrmac

rhatdan:
LGTM

PR #97, opened 2016-06-11
Use the "default" policy section for unidentified images

mtrmac:
@runcom @rhatdan This is a minimal fix to make `--policy` useable, but probably not the right thing to do overall; please review the plan below before I start implementing it.

The context is that the image verification policy (per `signature.NewPolicyFromFile` and future-PR `--policy`) is scoped by `image.IntendedDockerReference` (“what the user wanted to pull”), allowing different policies for different registries/namespaces/repositories/tags:

```js
{
    "default": […],
    "specific": {
        "docker.io/openshift": […],
        "docker.io/openshift/origin": […],
        "localhost:8443": […],
        "localhost:8443/myns/official:specificrelease": […]
    }
}
```

The `dir:/path/to/directory/with/manifest/and/layer/tarballs` image specification does not fit into this paradigm because that location cannot be converted into an `IntendedDockerReference`. So, this PR has the minimal fix of letting such images use the `"default"` case, instead of failing to determine the appropriate policy section and rejecting them outright.

This kind of works, but more or less forces `"default"` to be either `"reject"` or `"insecureAcceptAnything"` (because a `"signedBy"` would again want a plausible `IntendedDockerReference` value), and if a user wanted to consider all locally fetched images as trusted (i.e. to make the trust decision at pull time, and then allow everything which has already been pulled), this would require `"default":` …`"insecureAcceptAnything"`, and that would also apply to all Docker and Atomic repositories without a more specific policy section. Especially the latter seems very undesirable.

A few options how to deal with this:

- Hard-code that images with no `IntendedDockerReference` are always allowed, i.e. hard-code the “verification only/primarily happens on pull and local users are trusted” model. This is what yum/dnf do, but it seems very likely that an organization would only want to pull/run/otherwise use only images coming from the site’s maintained repository and to prevent cluster nodes from running locally-generated images, however or by whomever they were generated.

- Special-case `"local"` in the policy, as a peer to `"default"` and `"specific"`. All `dir:` images would be treated the same, but it would be possible to have `"local"` set to `"insecureAcceptAnything"` and `"default"` to `"reject"`. Or would `"local"` be called `"unidentified"`? More importantly, how do exploded containers on shared storage fit into this story?

- Rework the policy from a set of hierarchical `IntendedDockerReference` scopes to a per-`ImageSource`-type set of hierarchies, each one with their own rules for hierarchies/rules. E.g.
  ```js
{
    "default": […],
// empty string after the "type:" designation means the default policy for that format.
    "atomic:": […],
    "atomic:localhost:8443": […],
    "atomic:localhost:8443/myns/official:specificrelease": […]
    "docker:": […],
    "docker:docker.io/openshift": […],
    "docker:docker.io/openshift/origin": […],
    "dir:": […],
// Now we can implement directory hierarchies, even if the format is not a dockerReference.
    "dir:/home": […],
    "dir:/home/trusteduser/knownrepo": […],
    "dir:/mnt/companyimages": […],
// And eventually:
    "extractedshareddir:": […],
    "extractedshareddir:/mnt/companyimages": […],
}
```
(where `type:` with an empty string would be the per-type default).

  Or perhaps with a two-level format (image type as a first level of hierarchy, various scopes as the second one), would that be cleaner? That would be one less string format to parse = one less case of in-band signaling, OTOH the defaults would need to be with an `""` empty string key, kind of ugly.

The last option, divorcing the policy from Docker reference formats and instead extending it to arbitrary image sources, seems to have least ugly cases, and is most forward-looking, at the cost of a little extra complexity. Does that make sense? Is there a more elegant solution I am missing?


rhatdan:
I like the final option,
@vbatts @nalind @simo5 PTAL

simo5:
A default that ends up being "allow all" is not very useful, making this more general and avoiding that pitfall sounds like a much more desirable option, so I am with @rhatdan

runcom:
@mtrmac 

> - Replace default/specific[dockerReference] with an
  imageType:[somenamespace] (as a single string or object hierarchy?),
  which would natively support dir: (and allow different semantics for
  dir:/file/system/path and docker://full-format/reference/with:hostname )


I think this will be the best thing we could do (and I like differentiating by image type, eventually we'll add support for `appc` images as well and IntendedDockerRef would be empty there as well - not sure how _our_ signatures fits in appc though)

runcom:
if you're all fine with this - @mtrmac can you rebase and I'll merge

mtrmac:
> if you're all fine with this - @mtrmac can you rebase and I'll merge

I am just about to start working on the per-reference-type namespacing discussed above, so by the time we sort out the repo move this will probably become moot.

mtrmac:
See https://github.com/containers/image/pull/15 , will be handled by namespacing the policy per transport type instead.

PR #98, opened 2016-06-11
Move docker.genericImage to a separate skopeo/image subpackage

mtrmac:
Signature verification in `skopeo --policy … copy` requires a `types.Image`, making `docker.GenericImageFromSource` increasingly awkward. Clean this up by moving the `genericImage` code into a separate `skopeo/image` subpackage.

See individual commits for more detailed comments.

PR #99, opened 2016-06-11
Add minimal comments to fixManifestLayers

mtrmac:
I have stared at that code often enough now that at least some minimal documentation seems warranted.

This does not really go into why duplicate layers can happen or why it is worth supporting that; the code originates from https://github.com/docker/docker/commit 504e67b867865a2835e8002c01087a2cfd7bfd0e , which does not explain either.

(This will conflict with other PRs, notably #98 . I will rebase as necessary.)

### DIFF
--- a/image/image.go
+++ b/image/image.go
@@ -212,11 +212,18 @@ func (i *genericImage) getLayer(dest types.ImageDestination, digest string) erro
 	return dest.PutLayer(digest, stream)
 }
 
+// fixManifestLayers, after validating the supplied manifest
+// (to use correctly-formatted IDs, and to not have non-consecutive ID collisions in manifest.History),
+// modifies manifest to only have one entry for each layer ID in manifest.History (deleting the older duplicates,
+// both from manifest.History and manifest.FSLayers).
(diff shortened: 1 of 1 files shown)


mtrmac:
Other things worth considering, if we wanted to diverge from the original `docker/docker/distribution` code:

- fail if `len(manifest.FSLayers) != len(manifest.History)`
- fail if `len(manifest.History) == 0`
- fail if, when removing entries with a duplicated ID, we would be removing non-duplicated `FSLayer` entries. (Can this ever happen? If it can, would it happen in this natural way, or would the `FSLayer` duplication perhaps use indexes shifted by one in either direction?)

runcom:
> fail if, when removing entries with a duplicated ID, we would be removing non-duplicated FSLayer entries. (Can this ever happen? If it can, would it happen in this natural way, or would the FSLayer duplication perhaps use indexes shifted by one in either direction?)

Pretty confident this cannot happen :confused:

runcom:
#98 merged, can you rebase?

mtrmac:
Rebased.

PR #101, opened 2016-06-13
Generalize [Get|Put]Layer

runcom:
preparation for https://trello.com/c/hhK1klNW/143-skopeo-pull-from-docker-and-convert-to-oci

This is actually needed because to convert to OCI image's manifest we also need to fetch the *config* from a v2s2 manifest which is referenced just a normal blob

### DIFF
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -65,12 +65,12 @@ func copyHandler(context *cli.Context) {
 		logrus.Fatalf("Error parsing manifest: %s", err.Error())
 	}
 	for _, layer := range layers {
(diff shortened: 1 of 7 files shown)


runcom:
@mtrmac PTAL

mtrmac:
Sure.  (I do wonder how long we can sustain this source/destination abstraction, same for all possible image formats… but it works for now.)

Could you, though, fix the removing and re-adding of `dockerImageSource.GetSignatures` so that we don’t have a commit which does not build = false positives in `git bisect`, please? Either drop the hunks with an (add --patch), or just squash the two commits.

runcom:
@mtrmac done, I'll merge after tests green

PR #102, opened 2016-06-14
DO NOT MERGE: OCI ImageDestination

runcom:
This doesn't do much except https://github.com/projectatomic/skopeo/commit/90b7bdd9cc546183f098b0901426478e1bfbcf1b which moves manifest utils to a top level pkg.

The rest implements OCI ImageDestination which basically outputs to a directory following https://github.com/opencontainers/image-spec/blob/master/image-layout.md

Note that the OCI image spec it's still WIP itself.

This is still WIP but I greatly appreciate any comment @mtrmac (there are some ugliness just to make this work so don't bother about them I will fix it of course, I'd like to have an impression on the direction first)

It's already kind of working (expect reference naming which has an issue upstream as well):
```
$ ./skopeo copy docker://docker.io/fedora oci:docker.io/fedora
INFO[0001] Downloading library/fedora/blobs/sha256:8b53fe19c255ee721e5b2441b8469fc01a6324e8564de89bf8485387b34f95e6 
INFO[0045] Downloading library/fedora/blobs/sha256:a20665eb1fe2912accb3d5dadaed360430df0d1aa46874875886947d61d3d4ee 

$ tree docker.io/fedora 
docker.io/fedora
├── blobs
│   ├── sha256-8b53fe19c255ee721e5b2441b8469fc01a6324e8564de89bf8485387b34f95e6
│   ├── sha256-a20665eb1fe2912accb3d5dadaed360430df0d1aa46874875886947d61d3d4ee
│   └── sha256-fdf695e9003ee97fcec524f3f6c5aa70a32b256eab051a6f1cb621b7cf77352a
├── oci-layout
└── refs
    └── TODO

2 directories, 5 files
```


runcom:
This PR has 2 commits from https://github.com/projectatomic/skopeo/pull/101 (I'll remove them once that is merged)

manifest/manifest.go
```diff
@@ -10,19 +10,31 @@ import (
 
 // FIXME: Should we just use docker/distribution and docker/docker implementations directly?
 
+// FIXME(runcom, mitr): should we havea mediatype pkg??
```
mtrmac: I don’t quite feel we need to have a separate package for everything. Sure, use packages to simplify dependencies and avoid loops and have clear separation between transports, conversions and signatures, and from time to time to simplify the callers’ notation to avoid too many words in a symbol reference, but at some point adding a complex difficult-to-navigate package structure could make understanding more difficult instead of simpler.

manifest/manifest.go
```diff
@@ -34,9 +46,10 @@ func GuessManifestMIMEType(manifest []byte) string {
 	}
 
 	switch meta.MediaType {
-	case DockerV2Schema2MIMEType, DockerV2ListMIMEType: // A recognized type.
+	case DockerV2Schema2MIMEType, DockerV2ListMIMEType, OCIV1DescriptorMIMEType, OCIV1ImageManifestMIMEType, OCIV1ImageManifestListMIMEType, OCIV1ImageSerializationRootfsTarGzipMIMEType, OCIV1ImageSerializationConfigMIMEType: // A recognized type.
 		return meta.MediaType
 	}
+	// TODO(mitr, runcom): flacky? just error out?
```
mtrmac: No, this is AFAICS the _only_ way the function can return `DockerV2Schema1MIMEType`, and recognizing that is essential for stripping the JWS signatures = computing the correct manifest digest.

mtrmac:
- In mtrmac/skopeo:verify-on-pull I have already moved the layer parsing out of `copy.go`; it would be good to integrate this here.  Shall I separate that part of the verify-on-pull branch into a separate, immediately mergeable, PR?

- Moving `skopeo/docker/utils` to `skopeo/manifest`: sure, why not; but when dropping the `Manifest` part of function names, it would read cleaner to keep the module reference as `manifest.Digest` instead of `mutils.Digest`. etc.

oci/oci.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	if err := blob.Sync(); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (d *ociImageDestination) PutSignatures(signatures [][]byte) error {
+	// TODO
+	return nil
```
mtrmac: If this TODO comment is not a short-term thing, the method should return an error instead of silently throwing away data.

manifest/manifest.go
runcom: :+1:

oci/oci.go
```diff
@@ -0,0 +1,110 @@
...
+
+func (d *ociImageDestination) PutBlob(digest string, stream io.Reader) error {
+	blobsPath := filepath.Join(d.dir, "blobs")
+	if _, err := os.Stat(blobsPath); err != nil && os.IsNotExist(err) {
+		if err := os.MkdirAll(blobsPath, 0777); err != nil {
+			return err
+		}
+	} else if err != nil {
+		return err
+	}
```
mtrmac: It seems this could be clearer with an `ensureParentDirectoryExists` helper function, instead of duplicating the computation of `blobsPath` here and `blobPath` below, and duplicating the `Stat`/`IsNotExist`/`MkdirAll` logic.

manifest/manifest.go
runcom: :+1:

runcom:
> In mtrmac/skopeo:verify-on-pull I have already moved the layer parsing out of copy.go; it would be good to integrate this here. Shall I separate that part of the verify-on-pull branch into a separate, immediately mergeable, PR?

would be cool - pls do if you can

> 
> Moving skopeo/docker/utils to skopeo/manifest: sure, why not; but when dropping the Manifest part of function names, it would read cleaner to keep the module reference as manifest.Digest instead of mutils.Digest. etc.

right the issue here is that we have so many vars called `manifest` - I'll see if I can replace them all

oci/oci.go
runcom: :+1:

oci/oci.go
runcom: yeah, as I wrote in the first comment above this is really WIP and right I'm fixing all these ugly stuff so don't bother to comment code review now - I would love to have an higher evaluation of the direction of the PR instead

oci/oci.go
runcom: but wait, then skopeo copy can't work well because it always error out on failed to upload signatures. any idea?

oci/oci.go
mtrmac: Sure; that is basically the `copy.go` vs. `types.Image` discussion, the rest is implementation details. I’ll prepare the Docker case.

(Note that having `copy.go` use `types.Image` would essentially force adding proper OCI support to `types.Image` — but then having `skopeo inspect` work on OCI is useful anyway, this is “only” a matter of timing :wink: )

oci/oci.go
mtrmac: `if len(signatures) != 0 { return $someError }`.

_At the moment_ images with signatures are nonexistent, so the above would work well enough. Later we will need a `skopeo copy --drop-signatures`.

runcom: alright, that makes sense, wasn't sure about that

mtrmac:
> > In mtrmac/skopeo:verify-on-pull I have already moved the layer parsing out of copy.go; it would be good to integrate this here. Shall I separate that part of the verify-on-pull branch into a separate, immediately mergeable, PR?

> would be cool - pls do if you can

#103 .

runcom:
@mtrmac PTAL while I import #103 here after merged

oci/oci.go
```diff
@@ -0,0 +1,147 @@
...
+		tag = x.Tag()
+	}
+	return ref, tag, nil
+}
+
+// NewOCIImageDestination returns an ImageDestination for writing to an existing directory.
+func NewOCIImageDestination(dir string) (types.ImageDestination, error) {
+	// oci dest follow docker's one - parse those here
+	// there's no clear direction on naming though but since image-spec is WIP just assume this for now..
+	ref, tag, err := parseDockerImageName(dir)
```
runcom: @mtrmac this is important but I guess it's fine as it is right now :(

mtrmac: In general paths are not valid Docker Image Names; Is even `/home/mitr/some-dir` valid?  `/home/mitr/I_LIKE_CAPS/some_dir` definitely isn’t valid.

runcom: right, that's why we error out if the destination name in `oci:docker.io/fedora` isn't a valid reference.

mtrmac: AFAICS `d.ref` is only used for `FullName` (i.e. to turn it back into a string), this seems not worth all the effort; just find the last `:` and split into directory + tag?

(Have I ever mentioned that I hate mini-languages and strings and in-band signalling in general, a tiny bit? :) Yeah yeah, we can’t avoid strings for command line.)

runcom: > (Have I ever mentioned that I hate mini-languages and strings and in-band signalling in general, a tiny bit? :) Yeah yeah, we can’t avoid strings for command line.)

haha now I know it

> AFAICS d.ref is only used for FullName (i.e. to turn it back into a string), this seems not worth all the effort; just find the last : and split into directory + tag?

was smelling to me as well - I'll do it

mtrmac:
Copying to preserve:

> I guess PutManifest would really benefit from a MIME type parameter.

> but should that parameter match the actual mime type in the manifest bytes? otherwise, it's just asking to covert bytes to the correct manifest right?

> Yes; conversion, if any would happen in the caller. And the caller may not know the manifest type either (see GetManifest). But if the caller does know, it seems a bit pointless, and potentially dangerous, to throw that away and guess.

> Right, we'll figure that out once I actually come here and hack on this TODO, can you move those comments in the main PR page instead of per commit? (sorry I don't want to lose them once I push force :()

oci/oci.go
runcom: > just find the last : and split into directory + tag?

how do we make sure directory is valid? valid as in `docker.io/fedora` for instance, I think we should require `oci:` prefix to enforce a valid destination, or not?

runcom:
@mtrmac fixing latest stuff before trying to merge with #103  :)

oci/oci_dest.go
```diff
@@ -0,0 +1,135 @@
...
+
+func (d *ociImageDestination) CanonicalDockerReference() (string, error) {
+	return "", fmt.Errorf("Can not determine canonical Docker reference for an OCI image")
+}
+
+func (d *ociImageDestination) PutManifest(m []byte) error {
+	if err := d.ensureParentDirectoryExists("refs"); err != nil {
+		return err
+	}
+	// TODO(runcom):
```
runcom: this is the only big todo left

oci/oci_dest.go
```diff
@@ -0,0 +1,135 @@
...
+	// convert the manifest to OCI v1 - beaware we can also receive a docker v2s1 manifest so we need to
+	// handle this! docker.io/library/busybox it's still on v2s1 for instance
+	digest, err := manifest.Digest(m)
+	if err != nil {
+		return err
+	}
+	desc := descriptor{}
+	desc.Digest = digest
+	mt := manifest.GuessMIMEType(m)
+	// TODO(runcom): beaware and add support for OCI manifest list
```
runcom: and this one also

runcom: @mtrmac but I think manifest list is an entirely new topic

oci/oci_dest.go
runcom: @mtrmac ptal at the last commit

cmd/skopeo/copy.go
```diff
@@ -55,12 +82,19 @@ func copyHandler(context *cli.Context) {
 	}
 	signBy := context.String("sign-by")
 
-	manifest, _, err := src.GetManifest([]string{utils.DockerV2Schema1MIMEType})
+	mimeTypes := []string{manifest.DockerV2Schema2MIMEType}
+	// if OCI image destination, then ask for v2s2 manifest for config
+	// this is ugly I know, we'll refactor
```
runcom: I have to take care of this one still..

oci/oci_dest.go
```diff
@@ -0,0 +1,168 @@
...
+func (d *ociImageDestination) CanonicalDockerReference() (string, error) {
+	return "", fmt.Errorf("Can not determine canonical Docker reference for an OCI image")
+}
+
+func createManifest(m []byte) ([]byte, string, error) {
+	om := ociManifest{}
+	mt := manifest.GuessMIMEType(m)
+	switch mt {
+	case manifest.DockerV2Schema1MIMEType:
+		return nil, "", fmt.Errorf("not yet implemented")
```
runcom: I think this should be just `can't do this` - we shouldn't bother doing this - schema1 and schema2 in this case differs even in the layer digests so I think we should not do this.

oci/oci_dest.go
mtrmac: I haven’t thought about manifests that much. I am leaning towards treating them as a repository of independent images rather than as a single image with multiple sets of data (what does it mean to “read the $name annotation” if there is a manifest list and multiple annotated manifests?), but I’m not sure that is intuitive to users either.

mtrmac: But, yeah, definitely, let’s leave manifests lists as a separate work item and PR.

oci/oci.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func (d *ociImageDestination) CanonicalDockerReference() (string, error) {
+	return "", fmt.Errorf("Can not determine canonical Docker reference for an OCI directory")
+}
+
+// TODO(runcom): manifest here it's just a image-spec/image.Descriptor and not bytes!!!
+func (d *ociImageDestination) PutManifest(manifest []byte) error {
+	digest := sha256.New()
+	_, err := digest.Write(manifest)
```
mtrmac: This can be shortened to `sha256.Sum256` BTW.

oci/oci_dest.go
mtrmac: “not yet implemented” is not (especially without the “yet”) a commitment to ever do that. If it is in principle possible to do, even if difficult, "can’t do this” is a bit misleading.  But, really, anything works, we can always reword.

oci/oci.go
runcom: :+1:

oci/oci_dest.go
runcom: I think I'll leave _yet_ because in theory it should be possible to convert a v2s1 to v2s2 as it's what Docker I believe does when pulling busybox for instance (busybox still serve v2s1 form the Docker Hub). Makes sense?

oci/oci.go
runcom: wait, where are you commenting?! there's no such code anymore? :confused:

oci/oci_dest.go
```diff
@@ -51,30 +59,55 @@ func (d *ociImageDestination) CanonicalDockerReference() (string, error) {
...
 func (d *ociImageDestination) PutManifest(m []byte) error {
 	if err := d.ensureParentDirectoryExists("refs"); err != nil {
 		return err
 	}
-	// TODO(runcom):
-	// make a function to create an OCI manifest from a distribution one which comes here!!!
-	// convert the manifest to OCI v1 - beaware we can also receive a docker v2s1 manifest so we need to
-	// handle this! docker.io/library/busybox it's still on v2s1 for instance
-	digest, err := manifest.Digest(m)
+	ociMan, mt, err := createManifest(m)
```
mtrmac: Creating the real manifest so late in `ImageDestination` won’t work if we want to sign manifests; `copy --sign-by` will not have the right manifest to sign.

For now, especially when the conversion is only a matter of changing a single field, I guess this could live in `copy.go`. Or, perhaps, just make this a TODO for now (we can’t sign anyway), I can live with that.

Eventually we will need to make that a library, and I can’t immediately see a clean way. `image.ByConvertingImageToMIMEType` perhaps? That also seems fairly ugly and it is not immediately clear that it can work.  (For that matter, `signature.SignDockerManifest` will need to be replaced/generalized as well.)

oci/oci_dest.go
mtrmac: Either one works on me. I am not an authority on English by any means.

oci/oci_dest.go
runcom: makes sense, adding a todo for this since signatures in OCI will be another topic not yet defined upstream as well

oci/oci_dest.go
runcom: neither I am :) let's leave `not yet implemented`

runcom:
@mtrmac removed the WIP label on this PR - except #103 this should be good to go as it doesn't break anything on its own - I'd defer #103 after this so we have a more general approach to #103 now that we have OCI. Can you do another pass on this and eventually merge?

oci/oci_dest.go
```diff
@@ -0,0 +1,188 @@
...
+func (d *ociImageDestination) PutBlob(digest string, stream io.Reader) error {
+	if err := d.ensureParentDirectoryExists("blobs"); err != nil {
+		return err
+	}
+	var err error
+	defer func() {
+		if err != nil {
+			d.cleanup()
+		}
+	}()
```
mtrmac: (This seems like another argument for a set properties+commit model for `ImageDestination`, where `ImageDestination` would contain an `Abort` method called by the copy library on any failure.)

Either way, the `os.RemoveAll` implementation of cleanup feels a bit scary: if creating files in these directories is failing for a reason we haven’t anticipated, deleting everything (even files we haven’t created and have no idea are there) seems overzealous.

oci/oci_dest.go
```diff
@@ -0,0 +1,188 @@
...
+func createManifest(m []byte) ([]byte, string, error) {
+	om := ociManifest{}
+	mt := manifest.GuessMIMEType(m)
+	switch mt {
+	case manifest.DockerV2Schema1MIMEType:
+		// There a simple reason about not yet implementing this.
+		// OCI image-spec assure about backward compatibility with docker v2s2 but not v2s1
+		// generating a v2s2 is a migration docker does when upgrading to 1.10.3
+		// and I don't think we should bother about this now (I don't want to have migration code here in skopeo)
+		return nil, "", fmt.Errorf("not yet implemented")
```
mtrmac: This will eventually be shown to the user, so it should say what is not implemented. (“pushing Docker Registry v2 Schema 1 to OCI” or something)

cmd/skopeo/copy.go
```diff
@@ -55,12 +82,20 @@ func copyHandler(context *cli.Context) {
...
 	signBy := context.String("sign-by")
 
-	manifest, _, err := src.GetManifest([]string{utils.DockerV2Schema1MIMEType})
+	mimeTypes := []string{manifest.DockerV2Schema2MIMEType}
+	// TODO(runcom, mitr): will refactor in followup where copy becomes a library
+	// if not OCI image destination, then ask for v2s1 manifest as well for config
+	// this is ugly I know, we'll refactor
+	if !strings.HasPrefix(context.Args()[1], ociPrefix) {
+		mimeTypes = append(mimeTypes, manifest.DockerV2Schema1MIMEType)
+	}
```
mtrmac: This still needs solving.

From the POV of `copy.go`, it seems it would be best to:
- Always ask for OCI?, v2s2, v2s1, in that order (we might want to parametrize this depending on destination later, but now it seems unnecessary)
- Implement `manifestBlobs`/`Image.LayerDigests` for OCI and v2s2 as well. Then `copy.go` won’t care what the source uses.
- Keep the version check in `OCIDestination` to refuse v2s1.

Or do we need to keep preferring v2s1 because of OpenShift or something? AFAICS `skopeo copy` is a new functionality anyway, so does any of the other projects depend on skopeo preferring v2s1 right now?

mtrmac:
> Can you do another pass on this and eventually merge?

Quickly sent through the code, not yet a detailed line-by-line review.  We should progress on merge this, #103 and #112.

#112 is probably the easiest to start with, currently just blocked on a failing test.

Then #103 and this; this one needs rebasing at the very least, then I wouldn’t mind (too much :) ) starting with this one and me integrating it into #103.

mtrmac:
Or, if you like, #103 is ready for merging right now.

oci/oci_dest.go
runcom: yeah, thought about that, already on it

runcom:
I rebased #112 and merged it

runcom:
I'll rebase now that #103 is merged.

runcom:
@mtrmac PTAL - I've added 2 more commits after rebasing

image/image.go
```diff
@@ -75,28 +75,49 @@ func (i *genericImage) Signatures() ([][]byte, error) {
...
+		config, err := ioutil.ReadAll(rawConfig)
+		if err != nil {
+			return nil, err
+		}
+		v1 := &v1Image{}
+		if err := json.Unmarshal(config, v1); err != nil {
+			return nil, err
+		}
+		return &types.ImageInspectInfo{
+			DockerVersion: v1.DockerVersion,
```
runcom: not sure how do I retrieve tha Tag here

oci/oci_dest.go
runcom: if you're ok with this I would defer this change to another PR where we fix this in every place (like in copy.go as well)

image/image.go
mtrmac: The tag is not a property of an image at all, so it really doesn’t belong in there.  Just document in `types.ImageInspectInfo` that the `Tag` field is legacy fo v2s1 and will not be supported for other formats?

cmd/skopeo/copy.go
```diff
@@ -74,7 +74,7 @@ func copyHandler(context *cli.Context) error {
 	}
 
 	// FIXME: We need to call PutManifest after PutBlob and PutSignatures. This seems ugly; move to a "set properties" + "commit" model?
-	if err := dest.PutManifest(manifest); err != nil {
+	if err := dest.PutManifest(m); err != nil {
```
mtrmac: The 3 `cmd/skopeo/copy.go` hunks can be dropped: after #103 `copy.go` does not import `skopeo/manifest`, so we can use the more descriptive `manifest` variable name.

image/image.go
```diff
@@ -48,7 +48,12 @@ func (i *genericImage) IntendedDockerReference() string {
 // NOTE: It is essential for signature verification that Manifest returns the manifest from which LayerDigests is computed.
 func (i *genericImage) Manifest() ([]byte, error) {
 	if i.cachedManifest == nil {
-		m, _, err := i.src.GetManifest([]string{manifest.DockerV2Schema1MIMEType})
+		mts := []string{
```
mtrmac: We have enough electrons to afford a readable name like `manifestMIMETypes` ☺ or perhaps just have a multi-line argument to `GetManifest`.

image/image.go
```diff
@@ -70,28 +75,49 @@ func (i *genericImage) Signatures() ([][]byte, error) {
...
 	}
-	ms1, ok := m.(*manifestSchema1)
-	if !ok {
-		return nil, fmt.Errorf("error retrivieng manifest schema1")
-	}
-	v1 := &v1Image{}
-	if err := json.Unmarshal([]byte(ms1.History[0].V1Compatibility), v1); err != nil {
-		return nil, err
+	switch t := m.(type) {
+	case *manifestSchema1:
```
mtrmac: This would be much cleaner as methods on the `genericManifest` interface.

image/image.go
mtrmac: Also note that `inspect.go` will need to reasonably deal with the missing `Tag`, at least add a `json:omitempty` annotation, and perhaps make sure that the documentation is still correct.

image/image.go
mtrmac: … or, perhaps, a `getv1Image` method on the `genericManifest` interface, and keeping the `ImageInspectInfo` construction common in the two cases? Either way, after this type switch is gone, the rest does not matter that much.

image/image.go
```diff
@@ -70,28 +75,49 @@ func (i *genericImage) Signatures() ([][]byte, error) {
...
-	return &types.ImageInspectInfo{
-		Tag:           ms1.Tag,
-		DockerVersion: v1.DockerVersion,
-		Created:       v1.Created,
-		Labels:        v1.Config.Labels,
-		Architecture:  v1.Architecture,
-		Os:            v1.OS,
-		Layers:        ms1.GetLayers(),
-	}, nil
+	return nil, errors.New("error inspecting image")
```
mtrmac: This should say a bit more about what kind of error or what the user is supposed to do. (Unsupported manifest type? Internal error unexpected type inconsistency?) Or, with the move to a `genericManifest` method, this line could perhaps entirely disappear.

image/image.go
```diff
@@ -122,6 +148,31 @@ type genericManifest interface {
...
+	Digest    string `json:"digest"`
+}
+
+type manifestSchema2 struct {
+	Config descriptor   `json:"config"`
+	Layers []descriptor `json:"layers"`
+}
+
+func (m *manifestSchema2) String() string {
+	// TODO: does it make sense to have string here?
```
mtrmac: If nothing is using the method, remove it from the interface, and all implementations.

… actually, there is one use, in `LayersCommand`. This needs to be rethought somehow.  (FWIW I have no idea how `skopeo layers` is supposed to be used when the caller does not know what the name of the created directory is and has no way to figure that out.)

image/image.go
```diff
@@ -157,29 +208,40 @@ func sanitize(s string) string {
...
 	if err != nil {
 		return nil, err
 	}
-	mschema1 := &manifestSchema1{}
-	if err := json.Unmarshal(manblob, mschema1); err != nil {
-		return nil, err
-	}
-	if err := fixManifestLayers(mschema1); err != nil {
-		return nil, err
+	mt := manifest.GuessMIMEType(manblob)
```
mtrmac: `i.Manifest` may have known the MIME type; we should use that, if known, and only guess if unknown.  (Perhaps an `i.manifestAndMIMEType` doing the caching, and `i.Manifest` a public method which calls `i.manifestAndMIMEType` and discards the MIME type? Or do we want to make the MIME type public for conversions anyway?)

image/image.go
```diff
@@ -157,29 +208,40 @@ func sanitize(s string) string {
...
+		//}
+		return &mschema1, nil
+	case manifest.DockerV2Schema2MIMEType:
+		v2s2 := manifestSchema2{}
+		if err := json.Unmarshal(manblob, &v2s2); err != nil {
+			return nil, err
+		}
+		return &v2s2, nil
+	default:
+		return nil, fmt.Errorf("Unable to unmarshal into %s", mt)
```
mtrmac: “into”? IIUC, this is “unsupported manifest MIME type %s”.

image/image.go
runcom: Right will fix

manifest/fixtures/v2s2.manifest.json
```diff
@@ -1,26 +0,0 @@
-{
```
mtrmac: Why is this being removed? The tests depend on it.

On the contrary, we should have an OCI fixture and test for `GuessMIMEType`.

manifest/manifest.go
```diff
@@ -48,7 +45,7 @@ func GuessMIMEType(manifest []byte) string {
 	}
 
 	switch meta.MediaType {
-	case DockerV2Schema2MIMEType, DockerV2ListMIMEType, OCIV1DescriptorMIMEType, OCIV1ImageManifestMIMEType, OCIV1ImageManifestListMIMEType, OCIV1ImageSerializationRootfsTarGzipMIMEType, OCIV1ImageSerializationConfigMIMEType, OCIV1ImageSerializationCombinedMIMEType: // A recognized type.
+	case DockerV2Schema2MIMEType, DockerV2ListMIMEType, OCIV1ImageManifestMIMEType, OCIV1ImageManifestListMIMEType: // A recognized type.
```
mtrmac: Actually test fixtures for all of these would be nice.

manifest/fixtures/v2s2.manifest.json
runcom: Mistake, will fix

oci/oci_dest.go
```diff
@@ -0,0 +1,188 @@
...
+
+// NewOCIImageDestination returns an ImageDestination for writing to an existing directory.
+func NewOCIImageDestination(dir string) (types.ImageDestination, error) {
+	dest := dir
+	sep := strings.LastIndex(dir, ":")
+	tag := "latest"
+	if sep != -1 {
+		dest = dir[:sep]
+		tag = dir[sep+1:]
+		if !refRegexp.MatchString(tag) {
```
mtrmac: It seems to me that all of the above would be more readable with the `dir` parameter name and `dest` variable name exchanged.

Also, use `if`/`else`? Longer but perhaps more readable than the variable overwrites.

image/image.go
runcom: Sticked with golang idioms of naming variables the shortest I could

oci/oci_dest.go
mtrmac: I’m completely fine with not doing the commit/abort model here; but I would also be happier not having the dangerous `RemoveAll` (i.e. `d.cleanup`) around. Just fail and let the user deal with the messy unfinished directory structure manually.

oci/oci_dest.go
```diff
@@ -0,0 +1,188 @@
...
+		return err
+	}
+	var err error
+	defer func() {
+		if err != nil {
+			d.cleanup()
+		}
+	}()
+	// TODO(mitr, runcom): this breaks signatures entirely since at this point we're creating a new manifest
+	// and signatures don't apply anymore. Will fix.
```
mtrmac: Perhaps also add a comment to `PutSignatures` to point out that just adding signature support in that method without handling the manifest conversion first would be useless.

oci/oci_dest.go
runcom: Alright

oci/oci_dest.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func (d *ociImageDestination) cleanup() {
+	os.RemoveAll(d.dir)
+}
+
+func (d *ociImageDestination) ensureParentDirectoryExists(parent string) error {
+	path := filepath.Join(d.dir, parent)
+	if _, err := os.Stat(path); err != nil && os.IsNotExist(err) {
+		if err := os.MkdirAll(path, 0700); err != nil {
```
mtrmac: Use 0755 perhaps? This is using 0700, and files within these directories are created using 0644. They don’t _have_ to be consistent (the 0700 on the directory prevails) but it would be nice for them to be.

mtrmac:
> @mtrmac PTAL - I've added 2 more commits after rebasing

Read this line by line, that should be all.

image/image.go
mtrmac: > The tag is not a property of an image at all, so it really doesn’t belong in there

What I mean by that is that a single manifest (as identified by its digest) can AFAIK carry multiple tags at once, and the tags can be added/removed later. In that model, storing the tag within the manifest, or asking an image what tag it “is”, does not make sense.  Or I may be mistaken about the Docker tagging model.

image/image.go
mtrmac: OK, leave it

runcom:
On it, thanks a lot

image/image.go
runcom: You're right to me

image/image.go
runcom: there's a notable comment in there
```
LayersCommand(layers ...string) error // configure download directory?
```
why don't we extend layercommand to receive a directory name for destination?

image/image.go
runcom: I've added a comment in types/types.go about Tag + json omitempty

image/image.go
mtrmac: Sure, why not. I don’t know the history behind `skopeo layers` and I didn’t take the time to understand it, so all my refactoring was aiming at isolating the parts I don’t understand without changing their behavior; anything you decide to do with that command is fine with me.

image/image.go
mtrmac: That omitempty needs to go to `inspect.go:inspectOutput` I think.

image/image.go
runcom: I think also we should try not to break what @giuseppe did for system containers.

Now the maybe-issues are::

- the manfiest can be v2s2 (till now it was only v2s1)
- in the layers you also get the _config_ layer (till now the config was embedded in v2s1)

@giuseppe, could you tell me if system containers can be adapted to this? or it's time to use `skopeo copy` in atomic

runcom: the situation is:
```
$ mkdir testbusybox                      
$ ./skopeo copy docker://docker.io/busybox dir:testbusybox
$ ll testbusybox 
total 660K
-rw-rw-r--. 1 amurdaca amurdaca 1.5K Jun 25 11:08 2b8fd9751c4c0f5dd266fcae00707e67a2545ef34f9a29354585f93dac906749.tar
-rw-rw-r--. 1 amurdaca amurdaca 652K Jun 25 11:07 8ddc19f16526912237dd8af81971d5e4dd0587907234be2b83e249518d5b673f.tar
-rw-r--r--. 1 amurdaca amurdaca  505 Jun 25 11:08 manifest.json
$ cat testbusybox/manifest.json 
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
   "config": {
      "mediaType": "application/octet-stream",
      "size": 1459,
      "digest": "sha256:2b8fd9751c4c0f5dd266fcae00707e67a2545ef34f9a29354585f93dac906749"
   },
   "layers": [
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 667590,
         "digest": "sha256:8ddc19f16526912237dd8af81971d5e4dd0587907234be2b83e249518d5b673f"
      }
   ]
}%                                                                                                    
```

runcom: @giuseppe you can see from the above that now you get a manifest v2s2 and you also get the layer for the config which I think it's needed for you
```
$ cat testbusybox/2b8fd9751c4c0f5dd266fcae00707e67a2545ef34f9a29354585f93dac906749.tar
{"architecture":"amd64","config":{"Hostname":"55cd1f8f6e5b","Domainname":"","User":"","AttachStdin":false,"AttachStdout":false,"AttachStderr":false,"Tty":false,"OpenStdin":false,"StdinOnce":false,"Env":["PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"],"Cmd":["sh"],"Image":"sha256:e732471cb81a564575aad46b9510161c5945deaf18e9be3db344333d72f0b4b2","Volumes":null,"WorkingDir":"","Entrypoint":null,"OnBuild":null,"Labels":{}},"container":"764ef4448baa9a1ce19e4ae95f8cdd4eda7a1186c512773e56dc634dff208a59","container_config":{"Hostname":"55cd1f8f6e5b","Domainname":"","User":"","AttachStdin":false,"AttachStdout":false,"AttachStderr":false,"Tty":false,"OpenStdin":false,"StdinOnce":false,"Env":["PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"],"Cmd":["/bin/sh","-c","#(nop) CMD [\"sh\"]"],"Image":"sha256:e732471cb81a564575aad46b9510161c5945deaf18e9be3db344333d72f0b4b2","Volumes":null,"WorkingDir":"","Entrypoint":null,"OnBuild":null,"Labels":{}},"created":"2016-06-23T23:23:37.198943461Z","docker_version":"1.10.3","history":[{"created":"2016-06-23T23:23:36.73131105Z","created_by":"/bin/sh -c #(nop) ADD file:9ca60502d646bdd815bb51e612c458e2d447b597b95cf435f9673f0966d41c1a in /"},{"created":"2016-06-23T23:23:37.198943461Z","created_by":"/bin/sh -c #(nop) CMD [\"sh\"]","empty_layer":true}],"os":"linux","rootfs":{"type":"layers","diff_ids":["sha256:8ac8bfaff55af948c796026ee867448c5b5b5d9dd3549f4006d9759b25d4a893"]}}%

**Comment by runcom** [comment]
@mtrmac test failures seems related?

File: manifest/manifest.go (hunk 24 above)
Comment (runcom) [inline manifest/manifest.go]: added

File: image/image.go (hunk 15 above)
Comment (runcom) [inline image/image.go]: fixed

**Comment by mtrmac** [comment]
> @mtrmac test failures seems related?

Yeah; without actually placing any breakpoints in, it seems that this changes the manifest we get from the Docker Hub from v1s1 to v2s2, and the OpenShift registry does not support v2s2 and rejects it. Oops.

A short-term fix would be to either find s1-only images hosted on the Docker Hub (do they exist, or have they converted everything to s2?), or perhaps to modify the tests to pull some local images (fixtures?) from any one of the 3? in-container Docker registries instead of relying on the Docker Hub (which would be nice anyway).

Longer-term, we absolutely need to figure this out. I suppose it is easy enough to add `SupportedDestinationMIMETypes` to an `ImageDestination`, and to backpropagate that information through `Image` (which would need a new parameter) to `ImageSource` (which already supports it in `GetManifest`).

(It is also conceptually really awkward; a tagged Docker Reference, even if not retagged, does not specify a unique image, you can get something completely different depending on what destination you want to store the image in. *sigh*, no helping this now, I suppose.)

File: image/image.go [hunk 28]
```diff
@@ -122,6 +148,31 @@ type genericManifest interface {
...
+	// TODO: does it make sense to have string here?
+	return ""
+}
+
+func (m *manifestSchema2) GetLayers() []string {
+	blobs := []string{}
+	for _, layer := range m.Layers {
+		blobs = append(blobs, layer.Digest)
+	}
+	blobs = append(blobs, m.Config.Digest)
```
runcom: I'd like to sort this out before moving to containers/image - so when a v2s2 manifest is retrieved and we look for layers I'm appending the config digest here because it's needed (as it was in v2s1, but v2s1 had the config embedded into the manifest itself) so now we _have_ _to_ retrieve the config somehow. Would it be better by adding a `Config()` to the `generalManifest` interface? or is this ok? /cc @mtrmac

image/image.go
runcom: +1 for having it public - see https://github.com/containers/image/pull/19/files

image/image.go
runcom: nvm tried a fix into https://github.com/containers/image/pull/19

runcom:
I'll keep this open until we merge:

https://github.com/containers/image/pull/19
https://github.com/containers/image/pull/17

after those are merged, I'll split the OCI bits to containers/image and refactor code here

image/image.go
mtrmac: Yeah, see the conversation around containers/image#19.

image/image.go
mtrmac: ACK.

mtrmac:
#17 / #19 as it is now does not, AFAICS, fix the integration test failures (pulling s2 from Docker Hub and trying to save it into OpenShift, which only supports s1).

runcom:
they don't I'm just tracking progress in containers/image to go forward with this one

runcom:
> eah; without actually placing any breakpoints in, it seems that this changes the manifest we get from the Docker Hub from v1s1 to v2s2, and the OpenShift registry does not support v2s2 and rejects it. Oops.
> 
> A short-term fix would be to either find s1-only images hosted on the Docker Hub (do they exist, or have they converted everything to s2?), or perhaps to modify the tests to pull some local images (fixtures?) from any one of the 3? in-container Docker registries instead of relying on the Docker Hub (which would be nice anyway).

right, using images with v2s1 only manifests should fix the test - I just pushed to test my fix out

runcom:
right, test passed but we should look forward to fix this as you described in https://github.com/projectatomic/skopeo/pull/102#issuecomment-228747350

mtrmac:
> right, test passed but we should look forward to fix this as you described in #102 (comment)

Good to confirm that this is the culprit; if this is fixed, it would be nice to revert to using the `library/` images again so that we can have confidence in this working for the common users and uses.

runcom:
Right, we still have to fix the underlying issue though :( taking  care of this tomorrow probably

cmd/skopeo/copy.go
```diff
@@ -18,13 +18,13 @@ func copyHandler(context *cli.Context) error {
...
 		return fmt.Errorf("Error initializing %s: %v", context.Args()[0], err)
 	}
-	src := image.FromSource(rawSource)
 
 	dest, err := parseImageDestination(context, context.Args()[1])
 	if err != nil {
 		return fmt.Errorf("Error initializing %s: %v", context.Args()[1], err)
 	}
 	signBy := context.String("sign-by")
+	src := image.FromSource(rawSource, dest.SupportedImageDestinationMIMEType())
```
mtrmac: Yah, it works!  Stylistically it would be a bit more readable to do all of `dest` first and `source` second, then, so that we have a clear, simple `dest` paragraph and `src` paragraph.

runcom: +1 will do it after we merge containers/image#26 and do the full rebase here

PR #103, opened 2016-06-14
Move parsing layer digests from copy.go to types.Image

mtrmac:
See individual commits for detailed comments.

### DIFF
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -1,54 +1,24 @@
 package main
 
 import (
-	"encoding/json"
 	"errors"
 	"fmt"
 
-	"github.com/projectatomic/skopeo/manifest"
+	"github.com/projectatomic/skopeo/image"
 	"github.com/projectatomic/skopeo/signature"
 	"github.com/urfave/cli"
 )
-
-// FIXME: Also handle schema2, and put this elsewhere:
-// docker.go contains similar code, more sophisticated
-// (at the very least the deduplication should be reused from there).
-func manifestLayers(manifest []byte) ([]string, error) {
-	man := manifestSchema1{}
-	if err := json.Unmarshal(manifest, &man); err != nil {
-		return nil, err
-	}
-
-	layers := []string{}
-	for _, layer := range man.FSLayers {
-		layers = append(layers, layer.BlobSum)
-	}
-	return layers, nil
-}
-
-// FIXME: this is a copy from docker_image.go and does not belong here.
-type manifestSchema1 struct {
-	Name     string
-	Tag      string
-	FSLayers []struct {
-		BlobSum string `json:"blobSum"`
-	} `json:"fsLayers"`
-	History []struct {
-		V1Compatibility string `json:"v1Compatibility"`
-	} `json:"history"`
-	// TODO(runcom) verify the downloaded manifest
-	//Signature []byte `json:"signature"`
-}
 
 func copyHandler(context *cli.Context) error {
 	if len(context.Args()) != 2 {
 		return errors.New("Usage: copy source destination")
 	}
 
-	src, err := parseImageSource(context, context.Args()[0])
+	rawSource, err := parseImageSource(context, context.Args()[0])
 	if err != nil {
 		return fmt.Errorf("Error initializing %s: %v", context.Args()[0], err)
 	}
+	src := image.FromSource(rawSource)
 
 	dest, err := parseImageDestination(context, context.Args()[1])
 	if err != nil {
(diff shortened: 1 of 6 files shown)


cmd/skopeo/copy.go
```diff
@@ -1,71 +1,40 @@
...
+	src := image.FromSource(rawSource)
 
 	dest, err := parseImageDestination(context, context.Args()[1])
 	if err != nil {
 		logrus.Fatalf("Error initializing %s: %s", context.Args()[1], err.Error())
 	}
 	signBy := context.String("sign-by")
 
-	manifest, _, err := src.GetManifest([]string{utils.DockerV2Schema1MIMEType})
+	manifest, err := src.Manifest()
```
runcom: if we use image here, how do we deal with mime type selection? eg. I need a v2s2 for OCI generally (can hack up something from v2s1 but it's ugly)

mtrmac: We don’t, here.

AFAICS there aren’t any cases when the images has both v2s1 and v2s2 and we would want to use the v2s1 one; the “right” one (and also the one that matches a digest reference) is v2s2. Is that incorrect?

For signature verification to make sense, we really need a `types.Image` to be a _single_ image; not a set of several possibly completely different images, with different methods and different parameters to these methods choosing any of them.

So I expect that `types.Image` would, for example:
- Prefer OCI to v2s2, and v2s2 to v2s1
- Given a single `ImageSource` with the same contents, always choose the same version.

If we really needed to refer to the v2s1 manifest of a combined v2s1+v2s2 `ImageSource`, that would be a different `types.Image` object. (`image.FromSourceWithSpecifiedMIMEType`?) and very rarely used.

runcom: alright got it. But Image(Source|Destination) need to handle different manifests because a registry does not assure you'll get what you asked but only what's supported, is that correct?

mtrmac: That’s not essential in my thinking (we could instead first detect what the registry contains, and then create an appropriate `ImageSource`), but it is the easiest way to do it at the moment, yes.

I do wonder about v2s2→OCI conversions and the like; perhaps that would find it useful to have an `ImageDestination` with a known intended MIME type? At the moment the types are mutually exclusive, so “ociImageDestination supports OCI and none of v2s1 v2s2“ is enough, which is a property of the type, not of the instance.

image/image.go
```diff
@@ -172,6 +179,32 @@ func (i *genericImage) getSchema1Manifest() (manifest, error) {
...
+		res = append(res, digest)
+	}
+	return res
+}
+
+// LayerDigests returns a list of layer digests referenced by this image.
+// The list will not contain duplicates; it is not intended to correspond to the "history" or "parent chain" of a Docker image.
+// NOTE: It is essential for signature verification that LayerDigests is computed from the same manifest which is returned by Manifest().
+func (i *genericImage) LayerDigests() ([]string, error) {
+	m, err := i.getSchema1Manifest()
```
runcom: another question, I'm lost thanks for bearing with me - how do we handle this type of switch https://github.com/projectatomic/skopeo/pull/102/files#diff-0afbed05940006aa6ff131fc915c90b5R16?

mtrmac: I think that would happen somewhere around `genericImage.Manifest` (note that `getSchema1Manifest already calls `.Manifest`)—before we ever return anything manifest-dependent to any user of this object, we need to decide which MIME version to use; and then we commit to that decision.

You have already built the `manifest` interface as an abstraction, so I guess `getSchema1Manifest` would end up a `getParsedManifest` which returns an object implementing `manifest` for whatever MIME version was selected as described above.  Then both `LayerDigests` and `Inspect` would use this `manifest` implementation.

All this requires deciding what to do for the non-v2s1 data returned by `Inspect` and `LayersCommand`; we will need to do that eventually, there’s no avoiding that.

mtrmac: Alternatively, it might be worth considering to make the MIME version decision in `image.FromSource` already, and then have a `v2s1Image`, `v2s2Image`, `OCIImage` as separate implementations of `types.Image`, and eliminate the separate `manifest` interface.

But that’s not at all necessary to do now; with only `types.Image` and `image.FromSource` as public API, we can make this change anytime, if it ever seemed like a good idea.

runcom: > Alternatively, it might be worth considering to make the MIME version decision in image.FromSource already, and then have a v2s1Image, v2s2Image, OCIImage as separate implementations of types.Image, and eliminate the separate manifest interface.

+1



> But that’s not at all necessary to do now; with only types.Image and image.FromSource as public API, we can make this change anytime, if it ever seemed like a good idea.

right

runcom:
can you rebase this?

mtrmac:
Rebased.

PR #104, opened 2016-06-15
expose blob size

runcom:

### DIFF
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -65,7 +65,8 @@ func copyHandler(context *cli.Context) {
 		logrus.Fatalf("Error parsing manifest: %s", err.Error())
 	}
 	for _, layer := range layers {
-		stream, err := src.GetBlob(layer)
+		// TODO(mitr): do not ignore the size param returned here
+		stream, _, err := src.GetBlob(layer)
 		if err != nil {
 			logrus.Fatalf("Error reading layer %s: %s", layer, err.Error())
 		}
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"strconv"
 
 	"github.com/Sirupsen/logrus"
 	"github.com/projectatomic/skopeo/docker/utils"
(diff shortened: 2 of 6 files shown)


cmd/skopeo/copy.go
```diff
@@ -65,7 +65,7 @@ func copyHandler(context *cli.Context) {
 		logrus.Fatalf("Error parsing manifest: %s", err.Error())
 	}
 	for _, layer := range layers {
-		stream, err := src.GetBlob(layer)
+		stream, _, err := src.GetBlob(layer)
```
mtrmac: It would be nice to use the size here. At least add a comment, please, I need to add digest verification here anyway and that will make sure I don’t forget.

docker/docker_image_src.go
```diff
@@ -78,18 +79,19 @@ func (s *dockerImageSource) GetManifest(mimetypes []string) ([]byte, string, err
...
-		return nil, err
+		return nil, 0, err
 	}
 	if res.StatusCode != http.StatusOK {
 		// print url also
-		return nil, fmt.Errorf("Invalid status code returned when fetching blob %d", res.StatusCode)
+		return nil, 0, fmt.Errorf("Invalid status code returned when fetching blob %d", res.StatusCode)
 	}
-	return res.Body, nil
+	size, _ := strconv.ParseInt(res.Header.Get("Content-Lenght"), 10, 64)
```
mtrmac: - Typo in header name! (the urgency of having working tests for all of this is increasing… I do need to get that done.)
- `strconv.ParseInt` does not _always_ return 0 on error. (I don’t really care whether we would fail or just return ”unknown” on failure, but silently returning a different value on overflow is not great.)

cmd/skopeo/copy.go
runcom: :+1:

types/types.go
```diff
@@ -30,7 +30,7 @@ type ImageSource interface {
 	// It may use a remote (= slow) service.
 	GetManifest([]string) ([]byte, string, error)
 	// Note: Calling GetBlob() may have ordering dependencies WRT other methods of this type. FIXME: How does this work with (docker save) on stdin?
-	GetBlob(digest string) (io.ReadCloser, error)
+	GetBlob(digest string) (io.ReadCloser, int64, error)
```
mtrmac: Should document the return value.  Also, _if_ the value is defined to be optional, return 0 if unknown, this should be documented here.  (Alternatively, it might be a mandatory return value, and then the Docker client would fail if the header is missing or invalid.)

docker/docker_image_src.go
runcom: Fixed both

types/types.go
```diff
@@ -30,7 +30,8 @@ type ImageSource interface {
 	// It may use a remote (= slow) service.
 	GetManifest([]string) ([]byte, string, error)
 	// Note: Calling GetBlob() may have ordering dependencies WRT other methods of this type. FIXME: How does this work with (docker save) on stdin?
-	GetBlob(digest string) (io.ReadCloser, error)
+	// the second argument returned is the size of the blob. If not known 0 is returned
```
mtrmac: s/the second argument/The second return value/.

mtrmac:
Apart from that one comment nitpick, LGTM; with that fixed, feel free to merge after the tests pass.

PR #113, opened 2016-06-20
update github.com/urfave/cli to v1.17.0

runcom:
Took @GrantSeltzer PR #111 and just updated the library to the latest version

Thx @GrantSeltzer, I left you as author of course in commit, will merge after tests are green

Signed-off-by: Antonio Murdaca <runcom@redhat.com>

PR #114, opened 2016-06-20
Codegangsta to urfave cli

GrantSeltzer:
- Updated the name of the cli library from codegangsta to urfave (Moved)
- Upgraded the library
- Changed definitions of functions that are used as app actions to reflect the library changes. Now these functions must return errors. We're still using logrus but return nils at end of the functions.


runcom:
Ca you put your second commit here on top of the commit when have in #113? This pr is missing the actual vendoring process (./hack/vendor.sh call) while #113 is ok, then you can squash the two commits so git bisect is fine

cmd/skopeo/layers.go
```diff
@@ -2,20 +2,23 @@ package main
...
 // TODO(runcom): document args and usage
 var layersCmd = cli.Command{
 	Name:  "layers",
 	Usage: "get images layers",
-	Action: func(c *cli.Context) {
+	Action: func(c *cli.Context) error {
 		img, err := parseImage(c)
 		if err != nil {
 			logrus.Fatal(err)
+			return err
```
runcom: is this required? isn't Fatal (like Fatalf) just stop execution? (also below)

cmd/skopeo/layers.go
```diff
@@ -2,20 +2,23 @@ package main
...
 		}
 		if err := img.Layers(c.Args().Tail()...); err != nil {
 			logrus.Fatal(err)
+			return err
```
runcom: ditto

cmd/skopeo/inspect.go
```diff
@@ -45,7 +45,6 @@ var inspectCmd = cli.Command{
 		}
 		if c.Bool("raw") {
 			fmt.Println(string(rawManifest))
-			return
```
runcom: mmm this is wrong, here you need to `return nil`

GrantSeltzer:
LGTM!

runcom:
@mtrmac want to have a look at this?

mtrmac:
Sure, tests pass, and hopefully somebody has checked that the commands still seem to actually work…

And I very much like that this version of `cli` actually supports unit testing without adding ugly wrappers in #16:

 - There is now an `App.Writer` and `App.ErrWriter`
 - `Action` handlers can return an error (turned into `logrus.Fatal` at the top-level `main`) instead of calling `logrus.Fatalf` immediately.

Grant, if you still have the patch which turns `logrus.Fatal` into `return` in the various places, I would love to have that.  (If not, no harm done, I can redo that when rebasing #16.)

[Oh, and now working on the integration tests, I am much more confident in thinking that #16 is desirable. The edit-compile-test cycle time for integration tests is pretty horrible, so the basic error handling and trivial local cases like `dir:` should run as local unit tests.]

runcom:
@mtrmac great!!

runcom:
@GrantSeltzer can you do the `logrus.Fatal[f]` to `return err` conversion as you did before (but we dropped that just not to change that much)?

Feel free to add another commit on top of this (instead of amending to the first one)

GrantSeltzer:
@runcom I'm on it

GrantSeltzer:
@runcom @mtrmac What are your thoughts on instances that we use logrus.Fatalf? for example:

`			logrus.Fatalf("Error computing manifest digest: %s", err.Error())
`
Keep those? Just return the errror? or Concatenate the errors and return that?

mtrmac:
@GrantSeltzer 

Yes, or
```golang
return fmt.Errorf("Error computing manifest digest: %s", err.Error())
```
which would be a smaller change.  Either works fine, yours is a bit more efficient.

Just take care to preserve the `": "` to keep the output readable, please.

runcom:
> return fmt.Errorf("Error computing manifest digest: %s", err.Error())

right just remove `.Error()` here and use either %s or %v for errors (I like %v for errors, we do that in docker as well)

runcom:
Sorry guys, can we do this with `fmt.Errorf` and using either `%s` or `%v` w/o `.Error()` calls?

runcom:
FWIW `return errors.New(("Error initializing " + string(context.Args()[0])) + err.Error())` I don't like this
just
```
return fmt.Errorf("Error initializing %s: %v", context.Args()[0], err)
```
is far better to me..

GrantSeltzer:
Sure, will change as soon as I get home

cmd/skopeo/copy.go
```diff
@@ -41,70 +41,70 @@ type manifestSchema1 struct {
 
 func copyHandler(context *cli.Context) error {
 	if len(context.Args()) != 2 {
-		logrus.Fatal("Usage: copy source destination")
+		return fmt.Errorf("Usage: copy source destination")
```
runcom: w/o formatting, this would be just `errors.New("")`

GrantSeltzer: So would this be better as logrus.Fatal?

runcom: nope,it would be:
```
return errors.New("Usage: copy source destination")
```

image/image.go
```diff
@@ -238,7 +237,7 @@ func fixManifestLayers(manifest *manifestSchema1) error {
 		}
 	}
 	if imgs[len(imgs)-1].Parent != "" {
-		return errors.New("Invalid parent ID in the base layer of the image.")
+		return fmt.Errorf("Invalid parent ID in the base layer of the image.")
```
runcom: nitpicking - this was good with errors.New :p

GrantSeltzer: @runcom :+1:

PR #117, opened 2016-06-22
Add Atomic registry integration tests

mtrmac:
See the individual commits for detailed comments.

### DIFF
--- a/integration/copy_test.go
+++ b/integration/copy_test.go
@@ -1,0 +1,102 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/go-check/check"
+	"github.com/projectatomic/skopeo/docker/utils"
+)
+
+func init() {
+	check.Suite(&CopySuite{})
+}
+
+type CopySuite struct {
+	cluster *openshiftCluster
+}
+
+func (s *CopySuite) SetUpSuite(c *check.C) {
+	if os.Getenv("SKOPEO_CONTAINER_TESTS") != "1" {
+		c.Skip("Not running in a container, refusing to affect user state")
+	}
+
+	s.cluster = startOpenshiftCluster(c)
+
+	for _, stream := range []string{"unsigned"} {
+		isJSON := fmt.Sprintf(`{
+			"kind": "ImageStream",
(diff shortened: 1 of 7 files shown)


integration/copy_test.go
```diff
@@ -0,0 +1,102 @@
...
+// Streaming (skopeo copy)
+func (s *CopySuite) TestCopyStreaming(c *check.C) {
+	dir1, err := ioutil.TempDir("", "streaming-1")
+	c.Assert(err, check.IsNil)
+	defer os.RemoveAll(dir1)
+	dir2, err := ioutil.TempDir("", "streaming-2")
+	c.Assert(err, check.IsNil)
+	defer os.RemoveAll(dir2)
+
+	// FIXME: It would be nice to use one of the local Docker registries instead of neeeding an Internet connection.
```
runcom: I will fix this as we move forward with this, trust me ha

integration/utils.go
```diff
@@ -14,11 +15,11 @@ func consumeAndLogOutput(c *check.C, id string, f io.ReadCloser, err error) {
 			f.Close()
 			c.Logf("Output %s: Closed", id)
 		}()
-		buf := make([]byte, 0, 1024)
+		buf := make([]byte, 1024)
 		for {
 			c.Logf("Output %s: waiting", id)
 			n, err := f.Read(buf)
-			c.Logf("Output %s: got %d,%#v: %#v", id, n, err, buf[:n])
+			c.Logf("Output %s: got %d,%#v: %s", id, n, err, strings.TrimSuffix(string(buf[:n]), "\n"))
```
runcom: can we use "strings.TrimSpace`?

runcom:
LGTM, will merge after questions answers (no blocking)

integration/utils.go
mtrmac: I’d rather preserve the output as much as possible (e.g. including leading whitespace for indentation, but mostly for the principle of the thing). The content of `c.Logf` is only shown to the user when the test fails, so this does not add any clutter in the common case.

The `TrimSuffix` is in there only so that _if_ the test fails and the log is printed, the log lines are not double-spaced (one newline per log item, one within the logged line in `buf`).

runcom: :+1: fine

PR #123, opened 2016-06-24
Build binary in a docker container

duglin:
so that people don't need to install all dependencies just to build.

Make it so that "make binary" does nothing if nothing changed.

Also use ${DEST} in a few more spots for consistency

Signed-off-by: Doug Davis <dug@us.ibm.com>

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -1,0 +1,5 @@
+FROM ubuntu:16.04
+RUN apt-get update
+RUN apt-get install -y golang git-core libgpgme11-dev
+ENV GOPATH=/
+WORKDIR /src/github.com/projectatomic/skopeo
--- a/Makefile
+++ b/Makefile
@@ -25,8 +25,18 @@ GIT_COMMIT := $(shell git rev-parse HEAD 2> /dev/null || true)
 
 all: binary
 
-binary:
-	go build -ldflags "-X main.gitCommit=${GIT_COMMIT}" -o ${DEST}skopeo ./cmd/skopeo
+binary: skopeo
+
+# Build a docker image (skopeobuild) that has everything we need to build.
+# Then do the build and the output (skopeo) should appear in current dir
+skopeo: cmd/skopeo
+	docker build ${DOCKER_BUILD_ARGS} -f Dockerfile.build -t skopeobuildimage .
+	docker run --rm -v ${PWD}:/src/github.com/projectatomic/skopeo \
+		skopeobuildimage make binary-local
+
+# Build w/o using Docker containers
+binary-local:
+	go build -ldflags "-X main.gitCommit=${GIT_COMMIT}" -o skopeo ./cmd/skopeo
 
 build-container:
 	docker build ${DOCKER_BUILD_ARGS} -t "$(DOCKER_IMAGE)" .
--- a/hack/make/test-integration
+++ b/hack/make/test-integration
@@ -8,7 +8,7 @@ bundle_test_integration() {
 
 # subshell so that we can export PATH without breaking other things
 (
-	make binary
+	make binary-local
 	make install-binary
 	export GO15VENDOREXPERIMENT=1
 	bundle_test_integration


Makefile
```diff
@@ -25,14 +25,27 @@ GIT_COMMIT := $(shell git rev-parse HEAD 2> /dev/null || true)
...
 all: binary
 
-binary:
+binary: ${DEST}skopeo
+
+# Build a docker image (skopeobuild) that has everything we need to build.
+# Then do the build and the output (skopeo) shold appear in ${DEST} dir (or .).
+${DEST}skopeo: cmd/skopeo
+	docker build ${DOCKER_BUILD_ARGS} -f Dockerfile.build -t skopeobuildimage .
+	-docker rm -f skopeobuild
```
runcom: is this correct?

duglin: Yes, if for some reason the container in the next step is still around we want to delete it. But if not there then we should ignore the error, hence the "-"

runcom:
not sure why tests failing - docker isn't running ...

Makefile
runcom: then I suck at Makefiles..

duglin:
I think its because we're now calling docker from a container - trying to fix...

duglin:
Fixed but I'm not sure its the right solution.

Dockerfile.build
```diff
@@ -0,0 +1,5 @@
+FROM ubuntu:16.04
+RUN apt-get update
+RUN apt-get install -y golang git-core libgpgme11-dev
+ENV GOPATH=/
+WORKDIR /src/github.com/projectatomic/skopeo
```
mtrmac: ATM I think the development happens on Fedora.  OTOH having the build Dockerfile use Ubuntu and the test one use Fedora/CentOS might be good for interoperability :)

Makefile
```diff
@@ -25,14 +25,27 @@ GIT_COMMIT := $(shell git rev-parse HEAD 2> /dev/null || true)
 
 all: binary
 
-binary:
+binary: ${DEST}skopeo
+
+# Build a docker image (skopeobuild) that has everything we need to build.
+# Then do the build and the output (skopeo) shold appear in ${DEST} dir (or .).
+${DEST}skopeo: cmd/skopeo
+	docker build ${DOCKER_BUILD_ARGS} -f Dockerfile.build -t skopeobuildimage .
```
mtrmac: Am I correct in thinking that the `skopeobuildimage` container does not change on every build? If so, it would be nice to be able to reuse it across builds.

duglin: @mtrmac correct, once built this step in the Makefile should almost be a no-op because docker will use the cache for each step in `Dockefile.build`.

Makefile
```diff
@@ -25,14 +25,27 @@ GIT_COMMIT := $(shell git rev-parse HEAD 2> /dev/null || true)
...
+	-docker rm -f skopeobuild
+	docker run -v ${PWD}:/src/github.com/projectatomic/skopeo \
+		--name skopeobuild --env DEST=/ skopeobuildimage make binary-local
+	docker cp skopeobuild:/skopeo ${DEST}skopeo
```
mtrmac: This `DEST` business feels awkward, I’d rather remove it than spread all over the place.  @runcom ?

- The rules above don’t depend on `DEST` being supported: the `docker cp` could just as well do 
```sh
docker cp skopeobuild:/src/github.com/projectatomic/skopeo/skopeo ….
```
- The caller needs to add a trailing `/` to `DEST`, which is unintuitive.
- `DEST` could be easily confused with the more relevant `DESTDIR`.  (I am interested in this because the policy stuff will make usage of `make install` more important; installing the binary without the policy could, depending on what we decide for the compiled-in defaults, result in a broken application.)

Makefile
mtrmac: Thanks, I forgot about the cache. Yes, the cache makes this cheap, and with the cache removing the container instead of permanently polluting the namespace is nice. Never mind.

Makefile
duglin: I don't have a strong feeling about `DEST` but removing it would clean some stuff up, yes.

runcom: I'm fine removing it, I confused it with destdir indeed

duglin: ok removed

Makefile
```diff
@@ -25,14 +25,25 @@ GIT_COMMIT := $(shell git rev-parse HEAD 2> /dev/null || true)
...
+# Build w/o using Docker containers
+binary-local:
+	go build -ldflags "-X main.gitCommit=${GIT_COMMIT}" -o skopeo ./cmd/skopeo
 
 build-container:
 	docker build ${DOCKER_BUILD_ARGS} -t "$(DOCKER_IMAGE)" .
 
 clean:
 	rm -f skopeo
+	# maybe we should also erase the Docker build image too - not sure yet
```
runcom: I wouldn't, @mtrmac?

duglin: I think it kind of depends on whether you consider "clean" to mean just " erase the binary" or "reset everything".

runcom: I honestly don't mind dropping this clean target entirely if it was up to me since I'm sure I've never used it and I just recompile overwriting everything. 
If you guys think otherwise then I won't mind either. I'll drop the ball on @mtrmac

Dockerfile.build
runcom: It's fine by now, ideally a test grid would take care of interoperability?

Makefile
duglin: well, with this PR a 'make binary' will not rebuild unless something changed  :-)  like a good Makefile should

mtrmac: The canonical reference for `make` targets, to the extent anything like that exists in Linux, is https://www.gnu.org/prep/standards/html_node/Makefile-Conventions.html#Makefile-Conventions , e.g.

> `clean`: Delete all files in the current directory that are normally created by building the program. Also delete files in other directories if they are created by this makefile. However, don’t delete the files that record the configuration. …

> `distclean`: Delete all files in the current directory (or created by this makefile) that are created by configuring or building the program. If you have unpacked the source and built the program without creating any other files, ‘make distclean’ should leave only the files that were in the distribution. …

> `maintainer-clean`: Delete almost everything that can be reconstructed with this Makefile. This typically includes everything deleted by distclean, plus more: C source files produced by Bison, tags tables, Info files, and so on.  …

i.e. `make clean` is “I don’t need the build state” or even “delete the build state to guarantee a fresh rebuild”.

This model really doesn't account for build processes which create non-filesystem artifacts (and even require root privileges).

And given the Docker image cache pointed out by @duglin, removing the image name would not save any space or remove state anyway.

So I think the `clean` target is fine as it is (though without that “not sure yet” comment, please: comments in code should document design and intent of the code, or perhaps to do items (though even that is a tiny bit controversial), but if we actually decided not to do something, having a comment doubting that decision would just cause us to repeatedly and unnecessarily revisit.):

- It is clearly useful to delete the built binary and guarantee a fresh rebuild.
- The build container is fairly unrelated to the built binary and we shouldn’t need to rebuild it at all often.
- Deleting the name of the build container without affecting the cache does not affect the build process in a meaningful way.  (It _might_ be useful for `make clean` to flush the container cache so that a next rebuild truly rebuidls the build image from scratch — but honestly this is not worth automating, worrying about, and spending time on, at this point, IMHO).

Dockerfile.build
mtrmac: AFAICT the question here is:

- Will we have a noticeable amount of users building from scratch using this Makefile (as opposed to using a distribution which builds however it wants)
- Do we care enough what build toolchain they use when/if they come asking questions or asking for help?

and my answers for both are, at least at the moment, “not enough to worry about”; so if @duglin wants to use Ubuntu for this, let’s.

Makefile
duglin: I've removed the commentary for now

Dockerfile.build
runcom: that's my opinion as well :)

runcom:
I'd say this is good to go, @duglin care to rebase? I'll merge after

duglin:
rebased

PR #125, opened 2016-06-25
signature: docker: fix govet in sprintf

runcom:
```
go vet ./...
signature/docker.go:51: wrong number of args for format in Sprintf call: 1 needed but 2 args
```
I don't know how this went past the go vet check

### DIFF
--- a/signature/docker.go
+++ b/signature/docker.go
@@ -48,7 +48,7 @@ func VerifyDockerManifestSignature(unverifiedSignature, unverifiedManifest []byt
 				return err
 			}
 			if !matches {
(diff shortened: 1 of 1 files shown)


runcom:
@mtrmac ^^

mtrmac:
ACK. Should I still be merging this into skopeo? Can we somehow move the PR to the other repo? If not, could you clone it manually?

> I don't know how this went past the go vet check

This is not the first time something like this has happened. I am fairly sure that the `hack/*` logic for only validating updated files is somewhat broken, but I haven’t looked into this in detail.

runcom:
I'll move this right there in the new repo

runcom:
@mtrmac moved there - I'll merge it in the other repo

PR #127, opened 2016-06-28
cmd/skopeo: refactor layers command

runcom:


runcom:
@mtrmac PTAL

runcom:
this is basically the same as before but this is now using primitives from containers/image instead of having `LayersCommand` in the containers/image API (which was ugly)

cmd/skopeo/layers.go
```diff
@@ -9,13 +14,45 @@ var layersCmd = cli.Command{
...
+		src := image.FromSource(rawSource)
+		layers := c.Args().Tail()
+		if len(layers) == 0 {
+			ls, err := src.LayerDigests()
+			if err != nil {
+				return err
+			}
+			layers = ls
+		}
+		tmpDir, err := ioutil.TempDir(".", "layers-")
```
runcom: the only notable change is here - I don't believe we should output somewhere significant but it would be nice to at least log where this is going to write layers and manifest? wdyt?

runcom: I guess we can use c.Args()[0] eventually

runcom: so, looking at atomic I don't see any problem here since the layers are downloaded in a temp dir here https://github.com/projectatomic/atomic/blob/master/Atomic/util.py#L224 and then the layers are retrieved with https://github.com/projectatomic/atomic/blob/master/Atomic/atomic.py#L1117

I believe an anonymous temp dir in the cwd is fine for atomic and would not bother myself changing something now - if ppl needs to add a destination to `skopeo layers` then will do this otherwise this command will eventually go away as @giuseppe is already looking forward to use `skopeo copy`

runcom:
so this patch should not break atomic based on https://github.com/projectatomic/skopeo/pull/127#discussion_r68758886

cmd/skopeo/layers.go
mtrmac: (looking at Atomic:) Heh, that works.

Yeah, no need to expand this if the current user of the command can use the current version fine.

cmd/skopeo/layers.go
```diff
@@ -9,13 +14,45 @@ var layersCmd = cli.Command{
...
+		for _, l := range layers {
+			if !strings.HasPrefix(l, "sha256:") {
+				l = "sha256:" + l
+			}
+			r, _, err := rawSource.GetBlob(l)
+			if err != nil {
+				return err
+			}
+			if err := dest.PutBlob(l, r); err != nil {
+				r.Close()
```
mtrmac: Perhaps use `defer` for the `Close`? OTOH doing that in a loop is IIRC tricky.

runcom: right I did that because defer+for is tricky

mtrmac:
(the other notable change is that the manifest does not go through `fixManifestDigests`.)

Looks fine (and it’s great to get rid of `LayersCommand`), only blocked on `containers/image` update and replacing the vendoring commit.

runcom:
done the vendoring, I'll merge after tests finish

mtrmac:
LGTM pending tests.

PR #128, opened 2016-06-28
vendor containers/image 77c407390866f120f903fd1d70d34536e530d852

runcom:
Signed-off-by: Antonio Murdaca <runcom@redhat.com>

PR #129, opened 2016-06-28
Update for changed images.Type API

mtrmac:
Depends on https://github.com/containers/image/pull/24 .

### DIFF
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -26,24 +26,24 @@ func copyHandler(context *cli.Context) error {
 	}
 	signBy := context.String("sign-by")
 
-	manifest, err := src.Manifest()
+	manifest, _, err := src.Manifest()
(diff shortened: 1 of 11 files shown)


mtrmac:
This is a duplicate of #128, with a bit more variable renaming.

… and separating the vendor update commit and the code changes; I suppose I should squash them in the final version so that we don’t have an intermediate broken commit, shouldn’t I?

runcom:
closed #128 - also yes, pls squash so git bisect works fine

runcom:
mmm why isn't Travis using some Docker cache to speed up container build - I should look into it

runcom:
seems like there isn't any good way https://github.com/travis-ci/travis-ci/issues/5358 (unless building and publishing our test docker image, which I don't like). I believe caching can be easily achieved with Jenkins but I don't have clue where to host such a CI infra

PR #132, opened 2016-06-29
WIP: integration: do not pull from Docker Hub

runcom:
Include _frozen_ images in the repo integration directory to setup integration tests faster w/o relying on pulling from Docker Hub (plus, excrcise the copy operation with this)

Missing bit is uploading by tag a docker image but I left a TODO for now - it's probably ugly but we can live with this till we have upload tag (if we ever)

/@mtrmac PTAL

Signed-off-by: Antonio Murdaca runcom@redhat.com


runcom:
We can embed different docker images with different manifests also so we can test various things going on with manifests

runcom:
also, instead of embedding those images in the source tree - we can do the `skopeo copy docker://busybox:latest dir:directory` directly in the Dockerfile so we don't clutter the source here in the repo. The tests suite will just push the image to the test registry and that's it. WDYT?

integration/copy_test.go
```diff
@@ -75,11 +85,10 @@ func (s *CopySuite) TestCopyStreaming(c *check.C) {
...
 	defer os.RemoveAll(dir2)
 
-	// FIXME: It would be nice to use one of the local Docker registries instead of neeeding an Internet connection.
 	// streaming: docker: → atomic:
-	assertSkopeoSucceeds(c, "", "--debug", "copy", "docker://busybox:1-glibc", "atomic:myns/unsigned:streaming")
+	assertSkopeoSucceeds(c, "", "--debug", "copy", localDockerBusybox, "atomic:myns/unsigned:streaming")
 	// Compare (copies of) the original and the copy:
-	assertSkopeoSucceeds(c, "", "copy", "docker://busybox:1-glibc", "dir:"+dir1)
+	assertSkopeoSucceeds(c, "", "copy", localDockerBusybox, "dir:"+dir1)
 	assertSkopeoSucceeds(c, "", "copy", "atomic:myns/unsigned:streaming", "dir:"+dir2)
```
mtrmac: Does this actually work? The `copy` tests rely on the fact that `busybox:latest` and `busybox:1-glibc` are different images (because OpenShift signatures are attached ~to the maniefst digest, and we test what happens when a required signature is available on one image but not the other).

runcom: tests passes so not sure what's happening if you think they shouldn't 

Btw, if this looks good to you we can add the `1-glibc` image as part of our image fixtures

mtrmac:
> also, instead of embedding those images in the source tree - we can do the skopeo copy docker://busybox:latest dir:directory directly in the Dockerfile so we don't clutter the source here in the repo. The tests suite will just push the image to the test registry and that's it. WDYT?

Interesting; that would amortize the pull cost for development, _and_ keep the Travis tests (which don’t cache) checking that we continue to work with the Docker Hub.

(It would also force us to keep the `dir:` format more fixed (or to rebuild the container on format changes)—which is nice, keeping the format fixed is desirable.)

Why `dnf install skopeo`, can’t I
```
COPY ./skopeo /tmp/skopeo
RUN /tmp/skopeo … && rm /tmp/skopeo
```
or would that break the caching?  If it broke the caching, there would be no point.

Another option would be to keep the image in fixtures, but use some much smaller images. We don’t need them to be useful after all.

---

Not being able to upload by tag and test pulling by tag is awkward, though. For v2s1 `ImageDestination`, the best we can do in `PutManifest` is to parse it and to check that the `tag` matches, can’t we? So perhaps let’s do that.

integration/copy_test.go
```diff
@@ -11,10 +11,13 @@ import (
 )
 
 func init() {
-	check.Suite(&CopySuite{})
+	check.Suite(&CopySuite{
+		ss: &SkopeoSuite{},
```
mtrmac: This doesn’t seem awfully shorter than setting up the one registry we need, and doing the copy (and it sets up 4! of them).

Perhaps add a `setupRegistryV2WithPreparedImagesAt` helper, and use it in both suites?

runcom:
> Not being able to upload by tag and test pulling by tag is awkward, though. For v2s1 ImageDestination, the best we can do in PutManifest is to parse it and to check that the tag matches, can’t we? So perhaps let’s do that.

but V2S2 is coming shortly :/



> or would that break the caching? If it broke the caching, there would be no point.

we can build skopeo itself in the Dockerfile - do that and rm the binary - the cache will be invalidated but it's ok to rebuild as last step because the source changes in development

integration/copy_test.go
mtrmac: Ah, sorry; the signature tests are not yet merged upstream.  Anyway, yes, please add the other image.

integration/copy_test.go
runcom: right, wasn't sure

mtrmac:
> > Not being able to upload by tag and test pulling by tag is awkward, though. For v2s1 ImageDestination, the best we can do in PutManifest is to parse it and to check that the tag matches, can’t we? So perhaps let’s do that.

> but V2S2 is coming shortly :/

The tests for `docker:`→`atomic:` need a v2s1 manifest ATM. (Can we simulate the Hub’s behavior of serving both with the same tag, or is that a Hub-specific unavailable functionality?)

runcom:
> Can we simulate the Hub’s behavior of serving both with the same tag, or is that a Hub-specific unavailable functionality?

not sure what you're asking. I don't think you can pull v2s2 if you pushed v2s1 if this is what you're asking

mtrmac:
> > or would that break the caching? If it broke the caching, there would be no point.

> we can build skopeo itself in the Dockerfile - do that and rm the binary - the cache will be invalidated but it's ok to rebuild as last step because the source changes in development

To me, copying the images from the Hub in a `Dockerfile` is useful primarily for the caching, so that we don't wait for the Hub on every test run. If we were doing this in a `Dockerfile` in a non-cached step which is done on every test run, then there’s really no point, we get neither speed improvement nor Hub independence.

mtrmac:
> >  Can we simulate the Hub’s behavior of serving both with the same tag, or is that a Hub-specific unavailable functionality?

> not sure what you're asking. I don't think you can pull v2s2 if you pushed v2s1 if this is what you're asking

Either that or pull v2s1 if I pushed v2s2: Basically will we continue to test that our `Accept:` header contents are set as necessary for the Hub?

runcom:
> pull v2s1 if I pushed v2s2

this can be done by _not_ asking v2s2 in the first place because the registry translates on the fly the manifest from v2s2 to v2s1

PR #133, opened 2016-06-29
add possibility to download to OCI image-layout

runcom:
replacement for #102 - currently blocked on containers/image#23 bits (this branch contains code from containers/image#23 from my personal fork) 

/cc @mtrmac 

### DIFF
--- a/cmd/skopeo/layers.go
+++ b/cmd/skopeo/layers.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/containers/image/directory"
 	"github.com/containers/image/image"
+	"github.com/containers/image/manifest"
 	"github.com/urfave/cli"
 )
 
@@ -18,7 +19,12 @@ var layersCmd = cli.Command{
 		if err != nil {
 			return err
 		}
-		src := image.FromSource(rawSource)
+		src := image.FromSource(rawSource, []string{
+			// TODO: skopeo layers only support these now
+			// eventually we'll remove this command altogether...
+			manifest.DockerV2Schema1SignedMIMEType,
+			manifest.DockerV2Schema1MIMEType,
+		})
 		blobDigests := c.Args().Tail()
 		if len(blobDigests) == 0 {
 			b, err := src.BlobDigests()
--- a/integration/copy_test.go
+++ b/integration/copy_test.go
@@ -55,13 +55,22 @@ func (s *CopySuite) TestCopySimple(c *check.C) {
 
 	// FIXME: It would be nice to use one of the local Docker registries instead of neeeding an Internet connection.
 	// "pull": docker: → dir:
-	assertSkopeoSucceeds(c, "", "copy", "docker://busybox:latest", "dir:"+dir1)
+	assertSkopeoSucceeds(c, "", "copy", "docker://estesp/busybox:latest", "dir:"+dir1)
 	// "push": dir: → atomic:
 	assertSkopeoSucceeds(c, "", "--debug", "copy", "dir:"+dir1, "atomic:myns/unsigned:unsigned")
 	// The result of pushing and pulling is an unmodified image.
 	assertSkopeoSucceeds(c, "", "copy", "atomic:myns/unsigned:unsigned", "dir:"+dir2)
 	out := combinedOutputOfCommand(c, "diff", "-urN", dir1, dir2)
 	c.Assert(out, check.Equals, "")
+
+	// docker v2s2 -> OCI image layout
+	// ociDest will be created by oci: if it doesn't exist
+	// so don't create it here to exercise auto-creation
+	ociDest := "busybox-latest"
+	defer os.RemoveAll(ociDest)
+	assertSkopeoSucceeds(c, "", "copy", "docker://busybox:latest", "oci:"+ociDest)
+	_, err = os.Stat(ociDest)
+	c.Assert(err, check.IsNil)
 
 	// FIXME: Also check pushing to docker://
 }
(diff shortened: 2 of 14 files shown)


runcom:
there something wrong with this I cannot spot

runcom:
so here is the funny thing - sending v2s1 in accept headers isn't returning anymore a v2s1 but always a v2s2...

runcom:
alright, having a default mime types list in `FromSource` is breaking this PR as opposed to #102 which was working because the case 1) docker -> dir pulls a v2s2 manifest (because dir don't care) and 2) dir -> atomic breaks because it wants v2s1

runcom:
this was the relevant conversation https://github.com/containers/image/pull/26#discussion_r68958299

mtrmac:
Ouch, that is painful. OTOH the ability to get v2s2, or anything else, into a `dir:` is valuable, and “copying to a `dir:` pulls v2s1 because otherwise our test would break” is not a good rationale for our users :)

For now, I guess let’s run the streaming test with a known-v2s1 image?

Longer-term we might want to allow the user to choose what to use in `copy` (this applies primarily to manifest lists and the arch dependencies, but that infrastructure or something very similar could be used to force using specific formats)—but that is clearly not something we need or want to build right now.

cmd/skopeo/layers.go
```diff
@@ -18,7 +19,12 @@ var layersCmd = cli.Command{
 		if err != nil {
 			return err
 		}
-		src := image.FromSource(rawSource)
+		src := image.FromSource(rawSource, []string{
+			// TODO: skopeo layers only support these now
+			// eventually we'll remove this command altogether...
+			manifest.DockerV2Schema1SignedMIMEType,
+			manifest.DockerV2Schema1MIMEType,
+		})
```
mtrmac: Really? AFAICS the code is generic and works for everything.

mtrmac:
Also it Would Be Nice™ to have an integration test for OCI, but that is absolutely not a blocker for me.

cmd/skopeo/layers.go
runcom: I think the goal is to deprecate this command, or blindly call out copy

mtrmac: Yeah, sure, but that is not a rationale for this change to `layers.go` nor does it explain how “skopeo layers only support these now”.

Or are you saying that the code would work fine but you want to intentionally restrict it to v2s1 to force users to use `skopeo copy`?

runcom: The former :) atomic uses layers command with v2s1 right now, so I won't change anything let's make sure to fix copy instead also

runcom:
I will add it as soon as we Marge the changes in containers/image

runcom:
The issue however is real because official docker images are getting rebuilt as we speak and they all will be v2s2 and docker-dir dir-atomic will always be broken (while I think Docker-atomic isn't broken)

mtrmac:
I don’t really care about the `docker`→`dir`→`atomic` workflow as an use case; as far as I am concerned, `dir` is primarily a debugging tool, or at most an internal implementation detail of `atomic pull`.  (Is there consensus on this?)

If `skopeo copy` does the right thing for `docker`→`atomic`, that’s more important.

(Also note that the test doesn’t test `docker`→`dir`→`atomic`, it just assumes that `docker`→ will always return the same image, which is no longer the case.  (Technically it was never the case, that test breaks if the content on the Docker Hub changes between the two commands, but I digress…) The purpose of the test is to show that `skopeo copy` does not modify the image; it is perfectly fine to demonstrate that with v2s1, though we could eventually in some vague future have such a test for every MIME type.)

Also, over time OpenShift will, hopefully, add v2s2 support, and most of the issue will disappear or at least become less noticeable.

runcom:
Awesome I'm less worried now thanks

mtrmac:
I think you found some v2s1 images (or perhaps there are some intentionally v2s1 images?) for the other test experiments; if that image is (and will remain?) v2s1, the streaming test with that image should succeed IIUC.

runcom:
It will, I'm about to fix this in a moment

runcom:
@mtrmac PTAL - I've added a small smoke test for Docker->OCI - I'll re-iterate on tests later in followup PRs since I'm taking care of others as well

This should be good to merge after tests

runcom:
wait, this needs containers/image#28

integration/copy_test.go
```diff
@@ -55,14 +55,23 @@ func (s *CopySuite) TestCopySimple(c *check.C) {
...
 	assertSkopeoSucceeds(c, "", "--debug", "copy", "dir:"+dir1, "atomic:myns/unsigned:unsigned")
 	// The result of pushing and pulling is an unmodified image.
 	assertSkopeoSucceeds(c, "", "copy", "atomic:myns/unsigned:unsigned", "dir:"+dir2)
 	out := combinedOutputOfCommand(c, "diff", "-urN", dir1, dir2)
 	c.Assert(out, check.Equals, "")
 
+	// docker v2s2 -> OCI image layout
+	// ociDest will be created by oci: if it doesn't exist
+	// so don't create it here to exercise auto-creation
+	ociDest := "busybox-lateset"
```
mtrmac: s/lateset/latest/

mtrmac:
Great, thanks for making that work!; LGTM after re-vendoring from containers/image master.

runcom:
for some reason Travis failed integration tests on `TestCopy*` but running test locally everything is fine.

runcom:
@mtrmac 
```
time="2016-07-01T07:42:58Z" level=debug msg="Error body {\"errors\":[{\"code\":\"UNKNOWN\",\"message\":\"unknown error\",\"detail\":{\"ErrStatus\":{\"kind\":\"Status\",\"apiVersion\":\"v1\",\"metadata\":{},\"status\":\"Failure\",\"message\":\"ImageStreamMapping \\\"unsigned\\\" is invalid: tag: Required value\",\"reason\":\"Invalid\",\"details\":{\"name\":\"unsigned\",\"kind\":\"ImageStreamMapping\",\"causes\":[{\"reason\":\"FieldValueRequired\",\"message\":\"Required value\",\"field\":\"tag\"}]},\"code\":422}}}]}\n" 
```
do you know what can cause this in Travis while locally is fine?

runcom:
FWIW I'm not lying locally it works:
```
+ go test -test.timeout=10m -check.v github.com/projectatomic/skopeo/integration
SKIP: check_test.go:82: SkopeoSuite.TestCanAuthToPrivateRegistryV2WithoutDockerCfg (we need to restore --username --password flags!)
SKIP: check_test.go:89: SkopeoSuite.TestNeedAuthToPrivateRegistryV2WithoutDockerCfg (mock empty docker config)
PASS: check_test.go:98: SkopeoSuite.TestNoNeedAuthToPrivateRegistryV2ImageNotFound	0.011s
PASS: check_test.go:73: SkopeoSuite.TestVersion	0.005s
PASS: copy_test.go:48: CopySuite.TestCopySimple	11.175s
PASS: copy_test.go:79: CopySuite.TestCopyStreaming	10.946s
PASS: signing_test.go:64: SigningSuite.TestSignVerifySmoke	2.825s
OK: 5 passed, 2 skipped
PASS
ok  	github.com/projectatomic/skopeo/integration	36.931s
```

mtrmac:
Some work is happening on the OpenShift PR, changing the API; so probably the server has changed and broken the tests.

You very likely still have a locally built (or locally cached) container with the old OpenShift version, so tests work for you, but Travis rebuilds with no cache and uses the new version.

I was promised the current API would be archived in another branch, so we can just switch until we update the client, let me have a look…

mtrmac:
> I was promised the current API would be archived in another branch, so we can just switch until we update the client, let me have a look…

I can’t find that archive, so I rescued the old commit from one of my Docker images: #139 .

runcom:
all good! merging finally :) Thanks for the help @mtrmac

PR #134, opened 2016-06-30
inspect: provide description

TomasTomecek:
I'm not really sure about the content, this is what I was able to figure out from code.

Suggest edits, please.

### DIFF
--- a/cmd/skopeo/inspect.go
+++ b/cmd/skopeo/inspect.go
@@ -27,6 +27,12 @@ type inspectOutput struct {
 var inspectCmd = cli.Command{
 	Name:  "inspect",
 	Usage: "inspect images on a registry",
+	Description: "the image is specified in this form: \"{prefix}{image}\", " +
+		"supported prefixes are:\n" +
+		"     * docker:// — docker image reference\n" +
+		"     * atomic:   — atomic registry reference\n" +
+		"     * dir:      — local directories (for debugging)\n" +
+		"   e.g. docker://fedora",
 	Flags: []cli.Flag{
 		cli.BoolFlag{
 			Name:  "raw",


mtrmac:
The content is OK, to be a bit more precise, check what `man1/skopeo.1` says.

I am genuinely not sure about having that full documentation with examples in `--help`, though; it will be used all over the place and it kind of seems that we should present that as a common `skopeo` concept, perhaps by referring to the man page. OTOH having immediate examples in `--help` is clearly immediately useful, and we can share the text across subcommands easily. I think I’ll let @runcom make the call.

More generally, we have grown into having this transport`:`name concept all over the place, and we need to think about how to talk about that.

@runcom ?

- It needs a name we use everywhere. (Internally the API calls it an `image name`, `man1/skopeo.1` uses `image-name` mostly but calls it `-location` in the one place where we actually document the format)  I suppose the `location` was wrong, and we should just call it an `image name` everywhere. Or `skopeo name` or `namespaced container reference` or something, to avoid confusion with Docker References and URLs, at the cost of being slightly ridiculous?
- The canonical documentation should probably end up in the `skopeo.1` man page as a separate paragraph, to be referenced everywhere else.
- Separately, the output of `--help` for `skopeo` and all its subcommands needs attention (#137)

TomasTomecek:
so, what shall I do with this?

runcom:
@TomasTomecek sorry, been busy with other stuff, I'll try to have a look at this this week

mtrmac:
@TomasTomecek #149 improves the `--help` output a bit, and reworks the man page to include a new “Image Names” section documenting the formats.

Do you think this is useful enough? I suppose it might still be worth it to explicitly point to the man page to explain what `IMAGE-NAME` means in each of the `--help` outputs.

TomasTomecek:
#149 definitely solves my issue, will close this then; thanks!

PR #139, opened 2016-07-01
Keep using an old version of https://github.com/openshift/origin/pull/9181

mtrmac:
I don’t know how to checkout a specific untagged commit ( 9ff4bf43548c758b6767b639b335681285fece48 ) from the original repo, so I have forked the project and fetched that commit from a cached Docker image.

We should instead update the containers/image client for the new API ASAP, and then the github.com/mtrmac/origin repo should be removed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,7 +44,8 @@ RUN set -x \
 	&& yum install -y which git tar wget hostname util-linux bsdtar socat ethtool device-mapper iptables tree findutils nmap-ncat e2fsprogs xfsprogs lsof docker iproute \
 	&& export GOPATH=$(mktemp -d) \
 	# && git clone git://github.com/openshift/origin "$GOPATH/src/github.com/openshift/origin" \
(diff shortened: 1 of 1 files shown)


runcom:
seems fine to me - but can I run it on my laptop with my gopath? (I assume no)

mtrmac:
> seems fine to me - but can I run it on my laptop with my gopath? (I assume no)

Why not? All this does is change the repo which is pulled from github, when building the container. It has no effect on your non-docker environment.

runcom:
> Why not? All this does is change the repo which is pulled from github, when building the container. It has no effect on your non-docker environment.

alright, I missed that, thanks for clarifying

runcom:
feel free to merge then

mtrmac:
Actually this breakage couldn’t have been caused by the openshift/origin#9181 API change because mtrmac/image:openshift-native-signatures is not yet merged; this must have been a signature-independent change. Looking into it now…

PR #141, opened 2016-07-01
Fix integration tests on CentOS

mtrmac:
This fixes `--version` integration test on CentOS, as noticed by https://github.com/projectatomic/skopeo/pull/91 .  The underlying cause is:

- `Makefile` builds with `-ldflags "-X var=value"`, while go 1.4.2 only supports `"-X var value"`.  This  auses CentOS builds to be built without the specific commit information
- The `--version` integration test assumes that commit information will always follow the version number.

Changing either one of these would fix the build, changing the integration test has the advantage that we don't have to use the obsolete `-X` syntax and suffer warnings on newer Go versions.


runcom:
:+1:

PR #146, opened 2016-08-01
Update OpenShift

mtrmac:
With #139 , we have frozen OpenShift at an old version to keep our signature storage client working.

This updates OpenShift to the final version of https://github.com/openshift/origin/pull/9181 .

(Conceptually, this should be committed simultaneously with https://github.com/projectatomic/skopeo/pull/93 . But we haven’t merged integration tests for signatures yet, so this inconsistency does not show up as a test failure yet.  So, filing this as a separate PR for simpler review.)


mtrmac:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;yWz1COwb&#x2F;71-5-container-image-signing-and-verification-with-a-single-key"> (5) Container image signing and verification with a single key</a>

runcom:
lgtm

PR #148, opened 2016-08-04
Verify blobs

mtrmac:
This adds verification of blobs against the expected digests as they are being copied.

Depends on https://github.com/containers/image/pull/48 , I will revendor after it is merged.


mtrmac:
Revendored.

runcom:
lgtm

Issue #137, opened 2016-07-01
Fix --help output


mtrmac:
The outputs of `skopeo --help` and `skopeo $command --help` are not helpful enough; we need to review them and add the missing documentation.

PR #149, opened 2016-08-08
Improve man page and --help

mtrmac:
Should, at least partially, address #134 and #137.

### DIFF
--- a/man1/skopeo.1
+++ b/man1/skopeo.1
@@ -6,64 +6,86 @@
 ..
 .TH "skopeo" "1" "2016-04-21" "Linux" "Linux Programmer's Manual"
 .SH NAME
-skopeo \(em Inspect Docker images and repositories on registries
+skopeo \(em Various operations with container images images and container image registries
 .SH SYNOPSIS
-\fBskopeo copy\fR [\fB--sign-by=\fRkey-ID] source-location destination-location
-.PP
-\fBskopeo delete\fR source-location
-.PP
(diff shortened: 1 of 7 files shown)


rhatdan:
These LGTM.

But I would prefer if we rewrote these all in markdown, so they could easily be displayed on web sites and it would make it easier for people to contribute.

More examples is always helpful also.

man1/skopeo.1
```diff
@@ -6,121 +6,158 @@
...
 .TP
-.B standalone-sign
+\fBskopeo standalone-sign\fP \fImanifest\fP \fIdocker-reference\fP \fIkey-fingerprint\fP \%\fB--output\fP|\fB-o\fP \fIsignature\fP
 Create a signature using local files.
-This is primarily a debugging tool and should not be part of your normal operational workflow.
-.sp
-.B manifest
-path to file containing manifest of image
+This is primarily a debugging tool, or useful for special cases,
+and usually should not be a part of your normal operational workflow;
```
rhatdan: I thought we were going to remove this message.  Lets not say it is just a debugging tool.

mtrmac: That’s why there is the new “or useful for special cases” case.

mtrmac:
> But I would prefer if we rewrote these all in markdown, so they could easily be displayed on web sites and it would make it easier for people to contribute.

Sure, that could be useful; it is not on the critical path for signing, though, so I have filed #151 to track the markdown conversion.

Thanks for the review!

PR #150, opened 2016-08-09
Makefile: use :Z for volume mount (SELinux)

amitsaha:
As per http://www.projectatomic.io/blog/2015/06/using-volumes-with-docker-can-cause-problems-with-selinux/, use `:Z` to volume mount. Without this, make fails with permission issue on SELinux enabled host.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ binary: skopeo
 # Then do the build and the output (skopeo) should appear in current dir
 skopeo: cmd/skopeo
 	docker build ${DOCKER_BUILD_ARGS} -f Dockerfile.build -t skopeobuildimage .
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

runcom:
Ping @duglin does docker have this selinux label also?

mtrmac:
(I am, philosophically, a bit unhappy with a simple `make all` changing context of files in my home directory. But then `--security-opt label:disable` is not better, so I don’t have any better idea.)

rhatdan:
You could force a relabel back to the correct label in after the docker run.

restorecon -F-R ${PWD}

runcom:
Fixed in #171

PR #155, opened 2016-08-10
WIP: bash completion

mtrmac:
This only adds `--generate-bash-completion` for the subcommand names; needs integration with the bash completion infrastructure (available in `github.com/urfave/cli`).

The API may also support completion for non-subcommands (options, arguments), but AFAICS provides little to no infrastructure for it, so it would be noticeably more work.


mtrmac:
(Consider this just a trivial starting point, feel free to take this over.)

runcom:
closing, @mikebrow is taking care of this in https://github.com/projectatomic/skopeo/pull/173

PR #50, opened 2016-05-09
Correct man page formatting

jwhonce:
Signed-off-by: Jhon Honce <jhonce@redhat.com>

PR #157, opened 2016-08-10
Verify signatures on pull

mtrmac:
Adds `/etc/atomic/policy.json`, and a command-line option to override (`skopeo --policy some_path copy …`), and enforces this policy in `skopeo copy`.

`make install` now installs `default-policy.json` to `/etc/atomic/policy.json`; the policy is set to `insecureAcceptAnything` for all images by default. (This may not be desirable long-term, but it is a reasonable transition strategy for now I think. The policy always uses the most specific matching scope, so it is easy to e.g. pin a key and require signatures for a specific repository namespace.)

Note that with merging this, `./skopeo` will abort if the policy in `/etc/atomic` does not exist. This should not affect real deployments, but may be noticeable with local checkouts. A bit regrettable, but we do need to fail closed if a policy can’t be read.  (Use `./skopeo --policy default-policy.json …` in a checkout.)

This depends on, and vendors in, unmerged https://github.com/containers/image/pull/49 and https://github.com/containers/image/pull/50 , I will revendor as necessary.

@lsm5 , note that adding the config file will affect packaging.

@rhatdan , any comments on the path `/etc/atomic/policy.json`?

### DIFF
--- a/cmd/skopeo/main.go
+++ b/cmd/skopeo/main.go
@@ -5,6 +5,7 @@ import (
 	"os"
 
 	"github.com/Sirupsen/logrus"
+	"github.com/containers/image/signature"
 	"github.com/projectatomic/skopeo/version"
 	"github.com/urfave/cli"
 )
@@ -50,6 +51,11 @@ func createApp() *cli.App {
 			Name:  "tls-verify",
 			Usage: "verify certificates",
 		},
+		cli.StringFlag{
+			Name:  "policy",
+			Value: "",
+			Usage: "Path to a signature verification policy file",
+		},
 	}
 	app.Before = func(c *cli.Context) error {
 		if c.GlobalBool("debug") {
@@ -74,4 +80,20 @@ func main() {
 	if err := app.Run(os.Args); err != nil {
 		logrus.Fatal(err)
 	}
+}
+
+// getPolicyContext handles the global "policy" flag.
+func getPolicyContext(c *cli.Context) (*signature.PolicyContext, error) {
+	policyPath := c.GlobalString("policy")
+	var policy *signature.Policy // This could be cached across calls, if we had an application context.
+	var err error
+	if policyPath == "" {
+		policy, err = signature.DefaultPolicy(nil)
+	} else {
+		policy, err = signature.NewPolicyFromFile(policyPath)
+	}
+	if err != nil {
+		return nil, err
+	}
+	return signature.NewPolicyContext(policy)
 }
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ export GO15VENDOREXPERIMENT=1
 PREFIX ?= ${DESTDIR}/usr
 INSTALLDIR=${PREFIX}/bin
 MANINSTALLDIR=${PREFIX}/share/man
+CONTAINERSSYSCONFIGDIR=${DESTDIR}/etc/containers
 BASHINSTALLDIR=${PREFIX}/share/bash-completion/completions
 GO_MD2MAN ?= /usr/bin/go-md2man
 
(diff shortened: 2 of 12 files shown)


mtrmac:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;yWz1COwb&#x2F;71-5-container-image-signing-and-verification-with-a-single-key"> (5) Container image signing and verification with a single key</a>

lsm5:
On Wed, Aug 10, 2016 at 04:03:54PM -0700, Miloslav Trmač wrote:
> @lsm5 , note that adding the config file will affect packaging.

sure thing, thanks for the headsup
-- 
Lokesh
Freenode: lsm5
GPG: 0xC7C3A0DD

runcom:
I like `/etc/atomic/policy.json`

Unsure about failing hard if not `/etc/atomic` exists though - I guess I have no strong opinion, perhaps just tell something useful to the user suffices

LGTM though, reviewed #50 as well

mtrmac:
> Unsure about failing hard if not `/etc/atomic` exists though - I guess I have no strong opinion, perhaps just tell something useful to the user suffices

```
$ ./skopeo copy dir:a dir:b
FATA[0000] Error loading verification policy: open /etc/atomic/policy.json: no such file or directory 
```

runcom:
> Unsure about failing hard if not /etc/atomic exists though - I guess I have no strong opinion, perhaps just tell something useful to the user suffices
> $ ./skopeo copy dir:a dir:b
> FATA[0000] Error loading verification policy: open /etc/atomic/policy.json: no such file or directory

yup, saying I'm fine with that :)

rhatdan:
Should the path be vendor neutral?  Or is this a path handed to container/images?

runcom:
after meeting: `/etc/container` seems ok

mtrmac:
The path comes from `containers/image`, so that various tools all use the same file.

To be precise, the path would be `/etc/container/policy.json`? (I can’t find any package owning `/etc/container` in F24, so I’m not sure where to look for other users of that directory and possible problems.)

mtrmac:
Revendored to use `/etc/containers/policy.json` (pending re-review of https://github.com/containers/image/pull/49 ), and simplified `getPolicyContext` a little.

cmd/skopeo/main.go
```diff
@@ -75,3 +81,19 @@ func main() {
 		logrus.Fatal(err)
 	}
 }
+
+// getPolicyContext handles the global "policy" flag.
+func getPolicyContext(c *cli.Context) (*signature.PolicyContext, error) {
+	policyPath := c.GlobalString("policy")
+	var policy *signature.Policy // This could be cached across calls, if we had an application context.
```
runcom: Can be contacted in a single var block

runcom: *compacted

runcom:
Just a nit, LGTM otherwise

Feel free to merge containers/image bit, re vendor here and merge here

cmd/skopeo/main.go
mtrmac: Yeah, OTOH it is easier to attach the “This could be cached” comment to the `policy` variable when it is on a separate line.

mtrmac:
Thanks for the review!

I have decided to drop the dependency on https://github.com/containers/image/pull/50 ; the cost is that the link in `skopeo.1` will be pointing at a non-existent URL for a time, the benefit is that this makes https://github.com/containers/image/pull/50 a branch with no relationships and greatly simplifies the structure of outstanding branches.

cmd/skopeo/main.go
runcom: I believe it's a somehow golang idiom, you can still add comments on top of each variable in a var block

PR #160, opened 2016-08-11
fix dependencies

mikebrow:
Clean up dependencies.

Signed-off-by: Mike Brown <brownwm@us.ibm.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: all binary build clean install install-binary shell test-integration
+.PHONY: all binary build-container build-local clean install install-binary shell test-integration
 
 export GO15VENDOREXPERIMENT=1
 
@@ -28,11 +28,9 @@ MANPAGES_MD = $(wildcard docs/*.md)
 
 all: binary docs
 
-binary: skopeo
-
 # Build a docker image (skopeobuild) that has everything we need to build.
 # Then do the build and the output (skopeo) should appear in current dir
-skopeo: cmd/skopeo
+binary: cmd/skopeo
 	docker build ${DOCKER_BUILD_ARGS} -f Dockerfile.build -t skopeobuildimage .
 	docker run --rm -v ${PWD}:/src/github.com/projectatomic/skopeo \
 		skopeobuildimage make binary-local
@@ -57,7 +55,7 @@ install: install-binary install-docs
 	# TODO(runcom)
 	#install -m 644 completion/bash/skopeo ${BASHINSTALLDIR}/
 
-install-binary: skopeo
+install-binary: ./skopeo
 	install -d -m 0755 ${INSTALLDIR}
 	install -m 755 skopeo ${INSTALLDIR}
 


Makefile
```diff
@@ -28,11 +28,11 @@ MANPAGES_MD = $(wildcard docs/*.md)
...
 all: binary docs
 
-binary: skopeo
+binary: skopeo-build
 
 # Build a docker image (skopeobuild) that has everything we need to build.
 # Then do the build and the output (skopeo) should appear in current dir
-skopeo: cmd/skopeo
+skopeo-build: cmd/skopeo
 	docker build ${DOCKER_BUILD_ARGS} -f Dockerfile.build -t skopeobuildimage .
```
rhatdan: Probably should be phony

mikebrow: agreed

mtrmac: What is this actually fixing? “fix dependencies” does not make it obvious what is the failure case.

mikebrow: The failure case is make the skopeo binary.. touch a source file.. make the skopeo binary. Make fails to build the skopeo binary because it does not check the skopeo binary against the source tree. Required make clean in between builds.

mtrmac: Ah, true.  In that case, please just fold this into the `binary` target and we don't need the `skopeo-build` one (two phony targets chaining) at all.

mtrmac: /me is starting to be tempted to use Autoconf or something equally horrible just to outsource all of this.  (Not that I know of anything which integrates build-in-container with accurate golang dependency tracking.)

mikebrow:
Comments addressed. @mtrmac Cheers.

mikebrow:
rebased

mikebrow:
@mtrmac FYI I went ahead and got rid of the phony binary target as well. Is much cleaner just using the real binary `skopeo` as the target for all.

mtrmac:
> @mtrmac FYI I went ahead and got rid of the phony binary target as well. Is much cleaner just using the real binary skopeo as the target for all.

Doesn’t this break the purpose of your PR?

> The failure case is make the skopeo binary.. touch a source file.. make the skopeo binary. Make fails to build the skopeo binary because it does not check the skopeo binary against the source tree. Required make clean in between builds.

Because `skopeo` is a non-phony target, `make skopeo` does not rebuild after source files are changed. And with #163 it is useful for `skopeo` to be non-phony, otherwise `make all && make install` would build it twice.

So, a non-phony `skopeo` dependency (for `make install` to fail without a build), and a phony `build` target (to force rebuilds) seems to work for most cases. It does not allow `make install` to automatically succeed in a new checkout without a `make {all,build}` first, but that is not truly essential and at least we have a reasonable error message now.

mikebrow:
Sigh.  Suppose this is why folks fall to using .sh builds & installs. PITA but yes better to build every time than not, and it's ok to force the make all before the sudo make install, so.. here you go :-)

Makefile
```diff
@@ -57,7 +55,7 @@ install: install-binary install-docs
 	# TODO(runcom)
 	#install -m 644 completion/bash/skopeo ${BASHINSTALLDIR}/
 
-install-binary: skopeo
+install-binary: ./skopeo
```
mtrmac: What does this change do?

mtrmac:
Thanks!

Makefile
mtrmac: *shrug* it works either way, let’s not drag this PR out any longer. I would still be interested to know though :)

mikebrow: It does nothing wrt typical read/write/does the file exist type operations. The `./` only holds meaning on execution, to resolve which path which executable we want to execute type issues.  It's just a habitual thing in this context :-)

PR #161, opened 2016-08-11
add a source debug build

mikebrow:
Adds `make debug`.. allows a developer to build a source debuggable version of skopeo. For example, using delve.

Signed-off-by: Mike Brown <brownwm@us.ibm.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,13 @@ MANINSTALLDIR=${PREFIX}/share/man
 #BASHINSTALLDIR=${PREFIX}/share/bash-completion/completions
 GO_MD2MAN ?= /usr/bin/go-md2man
 
+ifeq ($(DEBUG), 1)
+  GOGCFLAGS += "-N -l"
+else
+  GOGCFLAGS ?= ""
+  DEBUG ?= 0
+endif
+
 GIT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD 2>/dev/null)
 DOCKER_IMAGE := skopeo-dev$(if $(GIT_BRANCH),:$(GIT_BRANCH))
 # set env like gobuildtag?
@@ -26,6 +33,10 @@ GIT_COMMIT := $(shell git rev-parse HEAD 2> /dev/null || true)
 
 MANPAGES_MD = $(wildcard docs/*.md)
 
+#   make all DEBUG=1
+#     Note: Uses the -N -l go compiler options to disable compiler optimizations
+#           and inlining. Using these build options allows you to subsequently
+#           use source debugging tools like delve.
 all: binary docs
 
 # Build a docker image (skopeobuild) that has everything we need to build.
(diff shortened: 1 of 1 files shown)


runcom:
@duglin ptal

Makefile
```diff
@@ -41,6 +41,20 @@ skopeo: cmd/skopeo
 binary-local:
 	go build -ldflags "-X main.gitCommit=${GIT_COMMIT}" -o skopeo ./cmd/skopeo
 
+#   make debug
+#     Note: Uses the -N -l go compiler options to disable compiler optimizations
+#           and inlining. Using these build options allows you to subsequently
+#           use source debugging tools like delve.
+debug: debug-skopeo docs
```
mtrmac: Is this useful? I would expect that when debugging, you only need the binary.

Makefile
```diff
@@ -41,6 +41,20 @@ skopeo: cmd/skopeo
...
+
+debug-skopeo: cmd/skopeo
+	docker build ${DOCKER_BUILD_ARGS} -f Dockerfile.build -t skopeobuildimage .
+	docker run --rm -v ${PWD}:/src/github.com/projectatomic/skopeo \
+		skopeobuildimage make debug-binary-local
+
+debug-binary-local:
+	go build -ldflags "-X main.gitCommit=${GIT_COMMIT}" -gcflags $(GOGCFLAGS) -o skopeo ./cmd/skopeo
+
```
mtrmac: All of this duplication, and top-level targets, feels a bit unclean; would it be reasonably easy to have `make {binary,binary-local} DEBUG=1` or something like that, and use the DEBUG flag to decide whether to set the `-gcflags`?

Makefile
mikebrow: make install currently requires the doc be built first .. otherwise yeah sure we don't have to build docs

mtrmac: The just added #163 could be another way to deal with that.

Makefile
mikebrow: I suppose..

mikebrow:
@mtrmac Switched to a make ... DEBUG=1 pattern

Makefile
```diff
@@ -35,11 +47,12 @@ binary: skopeo
 skopeo: cmd/skopeo
 	docker build ${DOCKER_BUILD_ARGS} -f Dockerfile.build -t skopeobuildimage .
 	docker run --rm -v ${PWD}:/src/github.com/projectatomic/skopeo \
-		skopeobuildimage make binary-local
+		skopeobuildimage make binary-local ${DEBUGENV}
```
mtrmac: Can just do `DEBUG=$(DEBUG)` here without the extra `DEBUGENV` variable.

mikebrow: Works, adds the DEBUG=0 to the output of the make all which I suppose is ok.

mikebrow:
rebased...

mtrmac:
Merged, thanks!

mtrmac:
Bah, I can’t merge this manually it seems.

@mikebrow do you want to try rebasing again?  Sorry about that… Or we can wait a few days for @runcom or somebody else with admin status on this project to allow force-pushing.

mikebrow:
Saw the merge and stopped watching for additional messages, sry bout that.  Rebased.

PR #162, opened 2016-08-11
bump to v0.1.14

rhatdan:


mtrmac:
@rhatdan PTAL at #163 first, so that we don't have users confused why `make install` fails.  Lokesh has dealt with this for Fedora, but having a cleaner upstream solution for the release would be nice.

rhatdan:
I did lets merge that and then I will repush this fix and switch the tag

runcom:
mmm what happened here? my F24 already has `v0.1.14`

PR #163, opened 2016-08-11
Make the install-* targets depend on things they are installing

mtrmac:
This ensures that we are not installing e.g. an obsolete version of the man page after the Markdown version is updated.

Note that this greatly benefits from the `skopeo` target being non-phony, otherwise `(make install)` would rebuild the binary.


mikebrow:
That works!

rhatdan:
LGTM

PR #165, opened 2016-08-16
Improve manifest digest handling

mtrmac:
- Modify `skopeo inspect --raw` to output an _unmodified_ manifest.
 - Provide `skopeo manifest-digest $manifest_file` to compute a manifest digest, including the v2s1 signature stripping.

Both will be called by `atomic`.


runcom:
lgtm

rebase and merge

PR #169, opened 2016-08-17
Makefile cleanup

mtrmac:
This is a follow-up to #161 , cleaning up a few rough edges.  I will rebase this after #161 is merged.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,10 +10,7 @@ MANINSTALLDIR=${PREFIX}/share/man
 GO_MD2MAN ?= /usr/bin/go-md2man
 
 ifeq ($(DEBUG), 1)
-  GOGCFLAGS += "-N -l"
-else
-  GOGCFLAGS ?= ""
-  DEBUG ?= 0
+  override GOGCFLAGS += -N -l
 endif
 
 GIT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD 2>/dev/null)
@@ -44,11 +41,11 @@ all: binary docs
 binary: cmd/skopeo
 	docker build ${DOCKER_BUILD_ARGS} -f Dockerfile.build -t skopeobuildimage .
 	docker run --rm -v ${PWD}:/src/github.com/projectatomic/skopeo \
-		skopeobuildimage make binary-local DEBUG=$(DEBUG)
+		skopeobuildimage make binary-local $(if $(DEBUG),DEBUG=$(DEBUG))
 
 # Build w/o using Docker containers
 binary-local:
-	go build -ldflags "-X main.gitCommit=${GIT_COMMIT}" -gcflags $(GOGCFLAGS) -o skopeo ./cmd/skopeo
+	go build -ldflags "-X main.gitCommit=${GIT_COMMIT}" -gcflags "$(GOGCFLAGS)" -o skopeo ./cmd/skopeo
 
 
 build-container:


rhatdan:
LGTM

Makefile
```diff
@@ -9,6 +9,10 @@ MANINSTALLDIR=${PREFIX}/share/man
 #BASHINSTALLDIR=${PREFIX}/share/bash-completion/completions
 GO_MD2MAN ?= /usr/bin/go-md2man
 
+ifeq ($(DEBUG), 1)
+  override GOGCFLAGS += -N -l
```
mikebrow: Why these changes? 
Without the quotes around -N -l, make all DEBUG=1 fails.
Without the else case to set GOGCFLAGS to "" make all fails.

Makefile
```diff
@@ -10,10 +10,7 @@ MANINSTALLDIR=${PREFIX}/share/man
 GO_MD2MAN ?= /usr/bin/go-md2man
 
 ifeq ($(DEBUG), 1)
-  GOGCFLAGS += "-N -l"
-else
-  GOGCFLAGS ?= ""
-  DEBUG ?= 0
+  override GOGCFLAGS += -N -l
```
mikebrow: Why these changes? 
Without the quotes around -N -l, make all DEBUG=1 fails.
Without the else case to set GOGCFLAGS to "" make all fails.
Without the DEBUG ?= 0 `DEBUG=` is passed to the make command.

mikebrow: why override +=?

mtrmac: See the comments in https://github.com/projectatomic/skopeo/pull/169/commits/694af0507d9a2471e7a2637d18df53dc1ad35d56 , sorry for not pointing to them.

Makefile
```diff
@@ -10,10 +10,7 @@ MANINSTALLDIR=${PREFIX}/share/man
 GO_MD2MAN ?= /usr/bin/go-md2man
 
 ifeq ($(DEBUG), 1)
-  GOGCFLAGS += "-N -l"
-else
-  GOGCFLAGS ?= ""
-  DEBUG ?= 0
```
mikebrow: Doesn't break anything, but you still get `DEBUG=` below on the call to make in the container. I thought it was a bit cleaner to say DEBUG=0 in the false case.

Makefile
mikebrow: Moving the "" down and forcing override means GOGCFLAGS can't be set outside the make file. I didn't see a case where "GOGCFLAGS+=" was ignoring the debug=1 flag?

mtrmac: If you set `GOGCFLAGS=` on the command line, it completely overrides what `Makefile` does and the `GOGCFLAGS += …` line does nothing, making `DEBUG=1` ineffective.

Makefile
mtrmac: It would be even clearer to not say anything ;)

The `binary:` target could contain something like `$(if DEBUG=1,DEBUG=1,)` (untested) to do that, but that’s getting too complicated for just passing a single boolean down (for my, entirely subjective and perhaps weird and wrong, taste).

Makefile
mikebrow: Cool.  Learned a new trick today. So override doesn't really override but not overriding does. That "makes" sense ;-)

Makefile
mikebrow: Cool.. here you go this works:
skopeobuildimage make binary-local $(if $(DEBUG),DEBUG=$(DEBUG))

mikebrow: and 	
`go build -ldflags "-X main.gitCommit=${GIT_COMMIT}" $(if $(GOGCFLAGS),-gcflags "$(GOGCFLAGS)") -o skopeo ./cmd/skopeo` to also remove the empty -gcflags case.. if you want

mtrmac: > `skopeobuildimage make binary-local $(if $(DEBUG),DEBUG=$(DEBUG))`

Thanks, that’s more readable than I thought it would be, PR updated to use this.

runcom:
lgtm

needs a rebase then merge

PR #171, opened 2016-08-18
Fix selinux perms in Makefile binary build

aweiteka:
As a new contributor I required this change to my environment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ all: binary docs
 # Then do the build and the output (skopeo) should appear in current dir
 binary: cmd/skopeo
 	docker build ${DOCKER_BUILD_ARGS} -f Dockerfile.build -t skopeobuildimage .
-	docker run --rm -v ${PWD}:/src/github.com/projectatomic/skopeo \
+	docker run --rm --security-opt label:disable -v $$(pwd):/src/github.com/projectatomic/skopeo \
 		skopeobuildimage make binary-local $(if $(DEBUG),DEBUG=$(DEBUG))
 
 # Build w/o using Docker containers


mtrmac:
Yes, I seem to remember hitting that… but please just use `-v $$(pwd):`… , we don’t need an extra variable.

aweiteka:
> we don’t need an extra variable.

ACK. Updated.

Makefile
```diff
@@ -43,8 +43,8 @@ all: binary docs
 # Then do the build and the output (skopeo) should appear in current dir
 binary: cmd/skopeo
 	docker build ${DOCKER_BUILD_ARGS} -f Dockerfile.build -t skopeobuildimage .
-	docker run --rm -v ${PWD}:/src/github.com/projectatomic/skopeo \
-		skopeobuildimage make binary-local DEBUG=$(DEBUG)
```
mtrmac: Why was the `DEBUG=` dropped?

runcom:
LGTM

@aweiteka can you rebase?

Makefile
```diff
@@ -43,7 +43,7 @@ all: binary docs
 # Then do the build and the output (skopeo) should appear in current dir
 binary: cmd/skopeo
 	docker build ${DOCKER_BUILD_ARGS} -f Dockerfile.build -t skopeobuildimage .
-	docker run --rm -v ${PWD}:/src/github.com/projectatomic/skopeo \
+	docker run --rm -v $$(pwd):/src/github.com/projectatomic/skopeo:Z \
```
mtrmac: FWIW this `:Z` subsumes #150.  (And I have no objections to that, to be explicit.)

runcom: I totally missed that Z, I don't have any objection either, if it's ok for @rhatdan we can close #150 and merge this

rhatdan: I think it would be better to disable selinux for this container then relabeling the skopeo directory.

--security-opt label=disable

cgwalters:
Personally I now use pet containers for development work which have all of the build dependencies installed, and notably here have SELinux disabled because I'm using overlayfs on my workstation.

rhatdan:
You can now run SELinux and overlayfs together if you use a kernel from this repo.

```
cat /etc/yum.repos.d/_copr_pcmoore-kernel-secnext.repo 
[pcmoore-kernel-secnext]
name=Copr repo for kernel-secnext owned by pcmoore
baseurl=https://copr-be.cloud.fedoraproject.org/results/pcmoore/kernel-secnext/fedora-rawhide-$basearch/
skip_if_unavailable=True
gpgcheck=1
gpgkey=https://copr-be.cloud.fedoraproject.org/results/pcmoore/kernel-secnext/pubkey.gpg
enabled=1
```

aweiteka:
@runcom this should be all set.

runcom:
@aweiteka waiting on tests to merge, thx

PR #177, opened 2016-08-30
Docker lookaside tests

mtrmac:
This includes https://github.com/projectatomic/skopeo/pull/170 , and adds integration tests for it.

Sadly, those tests depend on something like https://github.com/mtrmac/image/tree/docker-push-to-tag to support reasonable pushing to Docker registries, and that is not really ready. (Though perhaps it could be argued that it is better than status quo already?)

### DIFF
--- a/integration/copy_test.go
+++ b/integration/copy_test.go
@@ -1,11 +1,11 @@
 package main
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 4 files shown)


mtrmac:
After https://github.com/containers/image/pull/98 , these tests now work and can be merged.

runcom:
@mtrmac alright, feel free to rebase if you have to and merge this then

PR #179, opened 2016-08-31
Vendor containers/storage

nalind:
Vendor containers/storage, and its dependencies github.com/mistifyio/go-zfs and github.com/pborman/uuid.


mtrmac:
I guess it would be cleanest to have a single PR which vendors this as one commit, and then the appropriate version of `containers/image` as another commit.  In the meantime, is there a benefit? Am I right in guessing that it would allow the https://github.com/containers/image/pull/63 tests to pass?

nalind:
This should fix most of the problem that's keeping the test for containers/image#63 from passing, but I think it still needs to conditionally define the _btrfs_noversion_ and _libdm_no_deferred_remove_ build tags.

rhatdan:
Is this still being worked on?

nalind:
@rhatdan Yes, the PR is current.

rhatdan:
@mtrmac @runcom PTAL

runcom:
@rhatdan we need to first merge https://github.com/containers/image/pull/63 and that's pending review by Miloslav

rhatdan:
Ok we just need him to make sure nothing in Signatures gets broken.

nalind:
Updated to pull in the current version of containers/image as well.

PR #185, opened 2016-09-01
Add --remove-signatures to (skopeo copy)

mtrmac:
This exposes https://github.com/containers/image/pull/68, and is necessary to allow copying signed images into destinations which don't support signatures.

(We could instead silently drop signatures when copying into such destinations, I think that is worse. `atomic pull` can easily hard-code `skopeo copy --remove-signatures docker://$source docker-daemon:$destination` when we know that `docker-daemon:` doesn’t support signatures; after we add such support, we would drop the option from `atomic`.)


runcom:
lgtm

PR #192, opened 2016-09-06
Refer to the policy file as a trust policy file.

rhatdan:


rhatdan:
Fixes https://github.com/projectatomic/skopeo/issues/191

cmd/skopeo/copy.go
```diff
@@ -14,11 +14,11 @@ func copyHandler(context *cli.Context) error {
 		return errors.New("Usage: copy source destination")
 	}
 
-	policyContext, err := getPolicyContext(context)
+	trustContext, err := getTrustContext(context)
 	if err != nil {
-		return fmt.Errorf("Error loading verification policy: %v", err)
+		return fmt.Errorf("Error loading trust verification: %v", err)
```
mtrmac: I am not a native speaker but “trust verification” does not make much sense to me. How do you load a "verification”?

cmd/skopeo/main.go
```diff
@@ -82,18 +82,18 @@ func main() {
...
 }
 
-// getPolicyContext handles the global "policy" flag.
-func getPolicyContext(c *cli.Context) (*signature.PolicyContext, error) {
-	policyPath := c.GlobalString("policy")
-	var policy *signature.Policy // This could be cached across calls, if we had an application context.
+// getTrustContext handles the global "trust" flag.
+func getTrustContext(c *cli.Context) (*signature.TrustContext, error) {
+	trustPath := c.GlobalString("trust")
+	var trust *signature.Trust // This could be cached across calls, if we had an application context.
```
mtrmac: This just won’t work without corresponding changes in containers/image providing the API.

Makefile
```diff
@@ -61,7 +61,7 @@ clean:
 	rm -f skopeo docs/*.1
 
 install: install-binary install-docs install-completions
-	install -D -m 644 default-policy.json ${CONTAINERSSYSCONFIGDIR}/policy.json
+	install -D -m 644 default-trust.json ${CONTAINERSSYSCONFIGDIR}/trust.json
```
mtrmac: This policy path is hard-coded inside `containers/image`, needs updating there as well.

(As can be seen by this PR not changing the path of anything consuming the image.)

docs/skopeo.1.md
```diff
@@ -136,9 +135,9 @@ Verify a signature using local files, digest will be printed on success.
 show help for `skopeo`
 
 # FILES
-  **/etc/containers/policy.json**
-  Default signature verification policy file, if **--policy** is not specified.
-  The policy format is documented in https://github.com/containers/image/blob/master/docs/policy.json.md .
+  **/etc/containers/trust.json**
+  Default signature verification trust file, if **--trust** is not specified.
+  The trust format is documented in https://github.com/containers/image/blob/master/docs/trust.json.md .
```
mtrmac: Another external reference, https://github.com/containers/image/pull/50 needs to be updated.

cmd/skopeo/copy.go
rhatdan: I agree,  I was just throwing this together for discussion.  I liked this slightly better the 
Error loading verification trust.

"Error loading trust policy" is probably ok.

mtrmac:
I am perfectly fine with renaming `--policy` to `--trust`, but we do need a $name for the file or the concept which works in situations like "the image is not signed and $name requires a signature” or “the signature for this image is not accepted per $name”.  “signature verification policy” is a mouthful but works; “trust”, to me, doesn’t.

rhatdan:
I think we can drop this back and just leave policy in lots of places but begin talking about it as a "trust" policy.

@aweiteka WDYT?

mtrmac:
The two really user-visible pieces to which we kind of commit to in the public release are the `--policy` option to `skopeo`, and the `/etc/containers/policy.json` path. (We could of course change the default names in future releases and read the older names for compatibility, but it would be a bit awkward.)

I am perfectly fine with changing these two to `--trust` and `trust.json`; it is less clear what names to use for the internal `signature.Policy{,Context}` API naming and the error messages, but we can change them ~anytime.

rhatdan:
I simplified the original pull request to just change the way we refer to the file,  as a trust policy file.  Lets talk at the meeting.

rhatdan:
Fixes https://github.com/projectatomic/skopeo/issues/191

rhatdan:
@aweiteka and I have agreed on this.

@mtrmac @runcom PTAL and lets get this merged.

runcom:
LGTM as is (@mtrmac?)

mtrmac:
ACK. A similar patch to containers/image would touch 3 error messages.

PR #202, opened 2016-09-13
Change atomic URI

### LINKED ISSUE #189: Do not hard code registry port
As a registry admin I am able to configure the registry to serve on any port. When using a port other than 5000 I get this error:

        "Invalid format of docker reference: expecting port 5000"

See https://github.com/projectatomic/skopeo/blob/4421e7ea2face14ce4530ed6e6df66b6f94d544f/vendor/github.com/containers/image/openshift/openshift.go#L164

Could we discover the port using `~/.docker/config.json`?

### LINKED ISSUE #194: atomic type awkward
The atomic type is awkward when using with "copy" workflow. The primary object the user knows about is a docker image, locally or in the registry. What we do with the atomic type is infer the docker image reference. This means we don't have the registry URL or port (see #189).

What seems to make more sense is to drop atomic transport type and use a flag to enable writing to openshift API.

Copy:
```
skopeo copy --sign-by me@example.com --atomic docker-daemon:registry.example.com/my/image docker://registry.example.com/my/image
```

Standalone:
```
skopeo standalone-sign --atomic MANIFEST registry.example.com/my/image KEY-FINGERPRINT
```

runcom:
Vendor containers/image#81

Fix #189 
Fix #194 

### DIFF
--- a/integration/copy_test.go
+++ b/integration/copy_test.go
@@ -98,9 +98,9 @@ func (s *CopySuite) TestCopySimple(c *check.C) {
 	// "pull": docker: → dir:
 	assertSkopeoSucceeds(c, "", "copy", "docker://estesp/busybox:latest", "dir:"+dir1)
 	// "push": dir: → atomic:
-	assertSkopeoSucceeds(c, "", "--debug", "copy", "dir:"+dir1, "atomic:myns/unsigned:unsigned")
+	assertSkopeoSucceeds(c, "", "--debug", "copy", "dir:"+dir1, "atomic:localhost:5000/myns/unsigned:unsigned")
 	// The result of pushing and pulling is an unmodified image.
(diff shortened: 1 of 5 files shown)


mtrmac:
ACK.

mtrmac:
cc @baude : `atomic:$ns/$stream:$tag` → `atomic`:$dockerregistry[:$docker-port]/$ns/$stream:$tag`

runcom:
@mtrmac why is there a leading / in your example with the dockerhost? That's not supposed to be there right?

mtrmac:
@runcom I can’t see a leading `/`; the first `/` separates host name and namespace, e.g. `localhost/mynamespace/myimagestream:latest` or `localhost:555/mynamespace/myimagestream:latest`.

runcom:
@mtrmac sorry - I should pay more attention before commenting (I though $docker-port]/ was $dockerhost]/)

PR #203, opened 2016-09-13
Adds --policy-insecure-allow-anything flag

### LINKED ISSUE #181: avoid policy.json check?
On a fresh machine install I just needed skopeo to pull some oci images needed for testing and I realized the tool is failing with:

```
$ skopeo copy docker://busybox oci:busybox
FATA[0000] Error loading verification policy: open /etc/containers/policy.json: no such file or directory
```

@mtrmac do you think we can relax this somehow? I know I can use `default-policy.json` tbh but.. wdyt?

kushaldas:
This patch adds a new flag --policy-insecure-allow-anything
Closes #181, we can now directly use the tool with the
above mentioned flag wihout using a policy file.

Signed-off-by: Kushal Das mail@kushaldas.in

### DIFF
--- a/cmd/skopeo/main.go
+++ b/cmd/skopeo/main.go
@@ -32,6 +32,10 @@ func createApp() *cli.App {
 			Name:  "debug",
 			Usage: "enable debug output",
 		},
+		cli.BoolFlag{
+			Name:  "insecure-policy",
+			Usage: "run the tool without any policy check",
+		},
 		cli.StringFlag{
 			Name:  "username",
 			Value: "",
@@ -92,7 +96,9 @@ func getPolicyContext(c *cli.Context) (*signature.PolicyContext, error) {
 	policyPath := c.GlobalString("policy")
 	var policy *signature.Policy // This could be cached across calls, if we had an application context.
 	var err error
-	if policyPath == "" {
+	if c.GlobalBool("insecure-policy") {
+		policy = &signature.Policy{Default: []signature.PolicyRequirement{signature.NewPRInsecureAcceptAnything()}}
+	} else if policyPath == "" {
 		policy, err = signature.DefaultPolicy(nil)
 	} else {
 		policy, err = signature.NewPolicyFromFile(policyPath)


runcom:
I'm +1 on this (I don't want to have this in containers/image also)

@mtrmac wdyt?

rhatdan:
Lets just call this --insecure. You guys must be better typers then I am.

cmd/skopeo/main.go
```diff
@@ -92,7 +96,10 @@ func getPolicyContext(c *cli.Context) (*signature.PolicyContext, error) {
 	policyPath := c.GlobalString("policy")
 	var policy *signature.Policy // This could be cached across calls, if we had an application context.
 	var err error
-	if policyPath == "" {
+	if c.GlobalBool("policy-insecure-allow-anything") {
+		insecurepolicy := []byte(`{"default":[{"type":"insecureAcceptAnything"}]}`)
+		policy, err = signature.NewPolicyFromBytes(insecurepolicy)
```
mtrmac: This would be more type-safe, and visibly break if we ever were unfortunate enough to change the format, as (roughly, untested)

```go
&Policy{
  Default: PolicyRequirements{NewPRInsecureAcceptAnything()}
}
```

mtrmac:
I am a little uneasy with `--insecure` because it is so general; wouldn’t users expect this to also (only?) disable TLS certificate verification? (And then, would they use `--insecure` to bypass one check without noticing that they have disabled others?)

I would be happier with any of / any permutation of `--insecure-policy`, `--insecure-trust`, etc. Y’all are right that `accept-anything` is unnecessarily verbose.

(I am not good at UI, and I am perfectly willing to concede this to the consensus opinion; I just wanted to make sure that this issue has been considered.)

cmd/skopeo/main.go
```diff
@@ -32,6 +32,10 @@ func createApp() *cli.App {
 			Name:  "debug",
 			Usage: "enable debug output",
 		},
+		cli.BoolFlag{
+			Name:  "policy-insecure-allow-anything",
+			Usage: "run the tool without any policy check",
+		},
```
mtrmac: (One of these days we should decide on the order of these options, i.e. their order in `--help`. Alphabetical? By topic?  But not today, this is not a reason to change this PR.)

runcom: meh honestly

runcom:
> would be happier with any of / any permutation of --insecure-policy, --insecure-trust, etc. Y’all are right that accept-anything is unnecessarily verbose.

I'm happy as well with either of those - @rhatdan do you have any preference?

rhatdan:
--trust-all  @aweiteka wants us to talk about trust.

mtrmac:
@kushaldas Sorry, it seems nobody has noticed your update. The code LGTM, the only outstanding question is how much discussion needs to happen around naming (the code does `--insecure-policy`).

(`docs/skopeo.1.md` will also need updating—_after_ there is a consensus on the naming.)

rhatdan:
@aweiteka What do you think.  --insecure-policy or --trust-all

runcom:
I would call it insecure-policy to remark that it's still insecure to run this way, trust-all makes me think of a secure policy which happens to allow anything.

rhatdan:
Ok lets go with --insecure-policy  and update the man page and the bash completions.

kushaldas:
What all needs to be done for this PR to be merged?

runcom:
This PR needs a rebase and fixes from @rhatdan comment above

rhatdan:
@kushaldas Could you update your pull request?

mtrmac:
This was merged in #319.

yanivpaz:
I cant see this flag

skopeo version 0.1.34



AME:
   skopeo copy - Copy an IMAGE-NAME from one location to another

USAGE:
   skopeo copy [command options] SOURCE-IMAGE DESTINATION-IMAGE

DESCRIPTION:


  Container "IMAGE-NAME" uses a "transport":"details" format.

  Supported transports:
  containers-storage, dir, docker, docker-archive, docker-daemon, oci, oci-archive, ostree, tarball

  See skopeo(1) section "IMAGE NAMES" for the expected format


OPTIONS:
   --additional-tag value                                          additional tags (supports docker-archive)
   --remove-signatures                                             Do not copy signatures from SOURCE-IMAGE
   --sign-by FINGERPRINT                                           Sign the image using a GPG key with the specified FINGERPRINT
   --format MANIFEST TYPE, -f MANIFEST TYPE                        MANIFEST TYPE (oci, v2s1, or v2s2) to use when saving image to directory using the 'dir:' transport (default is manifest type of source)
   --authfile value                                                path of the authentication file. Default is ${XDG_RUNTIME_DIR}/containers/auth.json
   --src-creds USERNAME[:PASSWORD], --screds USERNAME[:PASSWORD]   Use USERNAME[:PASSWORD] for accessing the registry
   --src-cert-dir PATH                                             use certificates at PATH (*.crt, *.cert, *.key) to connect to the registry or daemon
   --src-tls-verify value                                          require HTTPS and verify certificates when talking to the container registry or daemon (defaults to true)
   --src-shared-blob-dir DIRECTORY                                 DIRECTORY to use to share blobs across OCI repositories
   --src-daemon-host HOST                                          use docker daemon host at HOST (docker-daemon: only)
   --dest-creds USERNAME[:PASSWORD], --dcreds USERNAME[:PASSWORD]  Use USERNAME[:PASSWORD] for accessing the registry
   --dest-cert-dir PATH                                            use certificates at PATH (*.crt, *.cert, *.key) to connect to the registry or daemon
   --dest-tls-verify value                                         require HTTPS and verify certificates when talking to the container registry or daemon (defaults to true)
   --dest-shared-blob-dir DIRECTORY                                DIRECTORY to use to share blobs across OCI repositories
   --dest-daemon-host HOST                                         use docker daemon host at HOST (docker-daemon: only)
   --dest-ostree-tmp-dir DIRECTORY                                 DIRECTORY to use for OSTree temporary files
   --dest-compress                                                 Compress tarball image layers when saving to directory using the 'dir' transport. (default is same compression type as source)

PR #205, opened 2016-09-13
TLS verification in docker registries

mtrmac:
Depends on https://github.com/containers/image/pull/87 .

- Flip `--tls-verify` default to true    
- Document better what `--tls-verify` does
- ... and sprinkle `--tls-verify=false` over integration tests.

cc @baude


runcom:
Ack

PR #222, opened 2016-10-02
main: give a pointer to install the policy

vbatts:
Signed-off-by: Vincent Batts vbatts@hashbangbash.com

### DIFF
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -35,6 +35,9 @@ func copyHandler(context *cli.Context) error {
 
 	policyContext, err := getPolicyContext(context)
 	if err != nil {
+		if os.IsNotExist(err) {
+			return fmt.Errorf("No system policy found. Install the default-policy.json from the source repo: %v", err)
+		}
 		return fmt.Errorf("Error loading trust policy: %v", err)
 	}
 	defer policyContext.Destroy()


cmd/skopeo/copy.go
```diff
@@ -17,6 +17,9 @@ func copyHandler(context *cli.Context) error {
 
 	policyContext, err := getPolicyContext(context)
 	if err != nil {
+		if os.IsNotExist(err) {
+			return fmt.Errorf("No system policy found. Install the default-policy.json from the source repo: %v", err)
```
runcom: We're in the process of having a generic skopeo-policy RPM which skopeo will require on installation.

I would drop the "installa the default-policy....."  and point instead to this new rpm. 

@mtrmac wdyt?

vbatts: fair. I just pulled the cloned source and hit the error, and thought that any kind of information would be helpful.

rhatdan:
@runcom Should this be closed or merged?

runcom:
I think this  is mergeable, @mtrmac? This also needs a rebase I guess

mtrmac:
https://github.com/projectatomic/skopeo/pull/222#discussion_r81578632 ; we should either be able to distinguish between a production / development environment and only point to `make install` in development, or we can just close this and rely on RPM packaging for production and keep the local development a bit painful. I don’t really have a preference between these two alternatives, but I would prefer not printing a `Install the default-policy.json from the source repo` message in RPM builds.

vbatts:
rebased and updated the message to not refer to the source repo, but just an expected path.

rhatdan:
@vbatts Looks like you need a rebase.

vbatts:
rebased

vbatts:
rebased. PTAL

mtrmac:
I still think “Install … from the source repo” is not the best thing to say for packaged deployments, but if @runcom likes this, I can live with that.

(Looking into the macOS test failures.)

rhatdan:
This PR has not been looked at in almost a year, can I close?

vbatts:
lol yes

PR #223, opened 2016-10-03
cmd/skopeo/inspect: fix tag not being populated

runcom:
@baude @mtrmac @rhatdan PTAL

Close https://github.com/containers/image/pull/118

### DIFF
--- a/cmd/skopeo/inspect.go
+++ b/cmd/skopeo/inspect.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/containers/image/docker"
 	"github.com/containers/image/manifest"
+	"github.com/docker/docker/reference"
 	"github.com/urfave/cli"
 )
 
@@ -56,9 +57,13 @@ var inspectCmd = cli.Command{
 		if err != nil {
 			return err
 		}
+		var tag string
+		if tagged, ok := img.Reference().DockerReference().(reference.NamedTagged); ok {
+			tag = tagged.Tag()
+		}
 		outputData := inspectOutput{
 			Name: "", // Possibly overridden for a docker.Image.
-			Tag:  imgInspect.Tag,
+			Tag:  tag,
 			// Digest is set below.
 			RepoTags:      []string{}, // Possibly overriden for a docker.Image.
 			Created:       imgInspect.Created,


cmd/skopeo/inspect.go
```diff
@@ -56,9 +57,13 @@ var inspectCmd = cli.Command{
...
 			return err
 		}
+		var tag string
+		if tagged, ok := img.Reference().DockerReference().(reference.NamedTagged); ok {
+			tag = tagged.Tag()
+		}
 		outputData := inspectOutput{
 			Name: "", // Possibly overridden for a docker.Image.
-			Tag:  imgInspect.Tag,
+			Tag:  tag,
```
rhatdan: Is tag being "" ok?

runcom: It can't unless it's a digest

rhatdan: Should the code throw an  error if tag is "" then?

runcom: If it's tagged then tag cannot not be empty, if it's a digest then this code isn't hit and you get an empty tag below in the response which is wanted because an image can't be both tagged and with a digest

rhatdan:
```
+		var tag string
+		if tagged, ok := img.Reference().DockerReference().(reference.NamedTagged); ok {
+			tag = tagged.Tag()
+		}
```
So it can never be not OK?

runcom: It can be if it's a reference with a digest, at that point populating Tag doesn't make sense and it's empty in the inspect response

baude: @runcom applied your patch and LGTM!

runcom: cool

runcom:
@mtrmac wanna merge?

cmd/skopeo/inspect.go
rhatdan: Then LGTM

mtrmac:
Is this a bug at all? What needs the value, for what purpose? Quoting `types.go`:
```go
// The Tag field is a legacy field which is here just for the Docker v2s1 manifest. It won't be supported
// for other manifest types.
```

In the old code, `skopeo inspect` fetches the `tag` value from the remote manifest; it is kind of weird piece of information to ask for, but it is a reasonable and logical that `skopeo inspect` would be a command to fetch data from the registry and return a value available in that data.

With this change, `skopeo inspect` just passes the provided transport`:`transport-specific reference through our parsing mechanisms, and then returns a part of the result of this parsing, completely independent of the remote repo.

I’m _perfectly fine_ with `skopeo` providing a “parse Docker reference to components and reformat as needed” functionality, but it seems weird to bundle this into the `skopeo inspect` “fetch a remote manifest and return information from that” subcommand. I was always thinking of the usage of “`skopeo inspect` and read `name` to expand a Docker reference” as a stopgap taking advantage of an existing weird case, until skopeo exposes that in a more structured way, not as a template to further expansion to `skopeo inspect`.

https://github.com/projectatomic/atomic/issues/578 suggested a new skopeo subcommand for expanding the reference, this seems broadly in the same category. Perhaps we need a `skopeo DRcalc` (by analogy with `ipcalc), with `--tag`, `--fully-expanded`, `--canonical`, and so on.

cmd/skopeo/inspect.go
```diff
@@ -56,9 +57,13 @@ var inspectCmd = cli.Command{
 		if err != nil {
 			return err
 		}
+		var tag string
+		if tagged, ok := img.Reference().DockerReference().(reference.NamedTagged); ok {
```
mtrmac: (note that `DockerReference` can be `nil`; toes this handle it correctly?)

runcom: Need to handle that yes

runcom:
In short, @baude asked for this because it seemed a bug to have tag exposed for V1 and for V2. While you're right V2 isn't carrying such it might be still OK to report this as part of inspect (in V2 case just as a parsed transport reference).

baude:
I agree.

mtrmac:
No, let’s not muddy `inspect` any more, and make it any more essential.

`skopeo docker-reference --tag $docker-reference`, or something like that, please.

runcom:
I'm not sure about that, then let's remove this altogether and just use "docker/distribution" references.. That library should then be translated to python in some way. I don't like a command "docker-reference".

mtrmac:
I did ask “What needs the value, for what purpose?”.  Are we considering changing it only for the (incorrectly perceived) symmetry, or does `atomic` or something else really need to compute the value?

(I would, very mildly, prefer a `skopeo docker-reference` or `skopeo DRcalc` or whatever allowing `atomic` or anyone else to use the canonical `docker/docker/reference` implementation to `atomic` reimplementing `docker/docker/reference` in Python, to avoid the risk of behavior drift. OTOH I can see that the overhead of extra subprocesses may be unattractive, and if, as I understand, `atomic` is already reimplementing that package in Python, then tag extraction is pretty much the most trivial aspect of that reimplementation.)

baude:
I have been working on a python replacement.  I'll begin to merge it into Atomic.

PR #224, opened 2016-10-04
Add insecureAcceptAnything to default docker-daemon transport

### LINKED ISSUE #212: Do not reject when destination is remote
With default policy set to "type: reject" I am unable to "push" an image from docker-daemon to docker. This is unexpected. I want to protect my host to reject yet that should not prevent me from pushing from docker-daemon. Is there a way to configure policy to accept all from source type docker-daemon with default reject?

aweiteka:
Resolves #212

### DIFF
--- a/default-policy.json
+++ b/default-policy.json
@@ -3,5 +3,12 @@
         {
             "type": "insecureAcceptAnything"
         }
-    ]
+    ],
+    "transports":
+        {
+            "docker-daemon":
+                {
+                    "": [{"type":"insecureAcceptAnything"}]
+                }
+        }
 }


mtrmac:
Yeah, as #212 says, this is blocked on, or should be a part of, https://github.com/containers/image/pull/79

rhatdan:
@aweiteka what is the state of this?

mtrmac:
AFAICS now that `docker-daemon:` is in master, this should work, and be mergeable, after a rebase. An integration test would also be nice but I have no idea how to conveniently integration-test `docker-daemon` ATM anyway, so let’s not block on that.

aweiteka:
signed off commit; rebase pushed

mtrmac:
Thanks!

runcom:
This isn't impacting the Docker integration right? Afaict Docker does not need this field from the policy when verifying on pull

mtrmac:
No, this is “push”.  And, really, in the default configuration, redundant with the global default which is the same—but this means that _if_ the user restricts the global default, push is still allowed unless they explicitly restrict it as well.

PR #88, opened 2016-05-31
Add a policy evaluation library

mtrmac:
This is now mostly ready for review; in particular the public API structure of `PolicyContext`, and the internal abstraction of `PolicyRequirement` and `PolicyReferenceMatch`, are done.

The one thing I still want to get done before merging is to eliminate duplication between `signature.go` and `policy_eval_signedby.go`.

See individual commit messages for a little more detailed descriptions.

PR #226, opened 2016-10-06
vendor containers/image

runcom:
specifically https://github.com/containers/image/pull/115 and also add an integration test

### DIFF
--- a/integration/copy_test.go
+++ b/integration/copy_test.go
@@ -96,8 +96,11 @@ func fileFromFixture(c *check.C, inputPath string, edits map[string]string) stri
 	return path
 }
 
-// The most basic (skopeo copy) use:
-func (s *CopySuite) TestCopySimple(c *check.C) {
+func (s *CopySuite) TestCopyFailsWithManifestList(c *check.C) {
+	assertSkopeoFails(c, ".*can not copy docker://estesp/busybox:latest: manifest contains multiple images.*", "copy", "docker://estesp/busybox:latest", "dir:somedir")
+}
+
+func (s *CopySuite) TestCopySimpleAtomicRegistry(c *check.C) {
 	dir1, err := ioutil.TempDir("", "copy-1")
 	c.Assert(err, check.IsNil)
 	defer os.RemoveAll(dir1)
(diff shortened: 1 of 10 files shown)


runcom:
@mtrmac PTAL

runcom:
I think it's time to setup the schema 1 registry in tests :) I'll do it

mtrmac:
If it’s all the same to you, I’d prefer running the tests against a v2 registry with s2; s1 images’ signatures eventually _should_ break with fixed s1 upload (#88), so converting most of the tests from Docker Hub estesp (s1+s2+list) against OpenShift to s2 against v2 registry would be nicer.

runcom:
> If it’s all the same to you, I’d prefer running the tests against a v2 registry with s2; s1 images’ signatures eventually should break with fixed s1 upload (#88), so converting most of the tests from Docker Hub estesp (s1+s2+list) against OpenShift to s2 against v2 registry would be nicer.

wait, v2s2 broke something in openshift here, I'll revert some changes and you can see

runcom:
@mtrmac do you know why the current test with `docker://busybox:latest` which is v2s2 will make atomic registry fails with 500?

mtrmac:
Atomic Registry by default refuses to accept s2.  https://github.com/mtrmac/skopeo/tree/docker-push-to-tag enables that, but then we are running tests in a non-default configuration, not a great idea.

That’s why I was suggesting a docker/distribution registry + s2, instead of Atomic. (And why this is so much work.)

runcom:
To summarize, you want that test to push the image to an in-test docker/distribution V2 registry, instead of the atomic registry, right?

mtrmac:
At least `TestCopySignatures`, yes; perhaps others may be needed. There should still be _something_ testing against the Atomic Registry.

runcom:
Right, we can find an s1 image to still test _something_ for the atomic registry, fine. I'll finish this tomorrow morning. The v2s2 registry is already setup in the test infra so it shouldn't be _that_ hard I hope.

mtrmac:
Yes, at least `TestCopyDockerSigstore` is already using it.

runcom:
@mtrmac tests fixed, PTAL

integration/copy_test.go
```diff
@@ -15,17 +15,24 @@ import (
...
 
 type CopySuite struct {
+	ss       *SkopeoSuite
 	cluster  *openshiftCluster
 	registry *testRegistryV2
 	gpgHome  string
 }
 
+func (s *CopySuite) SetUpTest(c *check.C) {
+	s.ss.SetUpTest(c)
```
mtrmac: `SkopeoSuite` is setting up a fresh set of _five_ registries for _every single test_.  That seems a fair bit of unnecessary overhead, especially when `CopySuite.SetUpSuite` is already setting up a v2 registry.

integration/copy_test.go
```diff
@@ -107,13 +118,33 @@ func (s *CopySuite) TestCopySimple(c *check.C) {
 
 	// FIXME: It would be nice to use one of the local Docker registries instead of neeeding an Internet connection.
 	// "pull": docker: → dir:
-	assertSkopeoSucceeds(c, "", "copy", "docker://estesp/busybox:latest", "dir:"+dir1)
+	assertSkopeoSucceeds(c, "", "copy", "docker://estesp/busybox:amd64", "dir:"+dir1)
```
mtrmac: Heh :) Nice and small.

integration/copy_test.go
runcom: fair.

runcom:
Should be good to go if tests agree @mtrmac

PR #233, opened 2016-10-12
add sigstore default configuration

runcom:
@mtrmac @baude PTAL

We'll soon ship a skopeo sub-pkg which will just contain `/etc/containers/policy.json` and `/etc/containers/registries.d/default.yaml` so that docker (after projectatomic/docker#200) and atomic can require it in their spec w/o having N specs shipping those files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ INSTALLDIR=${PREFIX}/bin
 MANINSTALLDIR=${PREFIX}/share/man
 CONTAINERSSYSCONFIGDIR=${DESTDIR}/etc/containers
 REGISTRIESDDIR=${CONTAINERSSYSCONFIGDIR}/registries.d
+SIGSTOREDIR=${DESTDIR}/var/lib/atomic/sigstore
 BASHINSTALLDIR=${PREFIX}/share/bash-completion/completions
 GO_MD2MAN ?= /usr/bin/go-md2man
 
@@ -62,8 +63,9 @@ clean:
 	rm -f skopeo docs/*.1
 
 install: install-binary install-docs install-completions
+	install -d -m 755 ${SIGSTOREDIR}
 	install -D -m 644 default-policy.json ${CONTAINERSSYSCONFIGDIR}/policy.json
-	install -d -m 755 ${REGISTRIESDDIR}
+	install -D -m 644 default.yaml ${REGISTRIESDDIR}/default.yaml
 
 install-binary: ./skopeo
 	install -D -m 755 skopeo ${INSTALLDIR}/skopeo
--- a/default.yaml
+++ b/default.yaml
@@ -1,0 +1,26 @@
+# This is a default registries.d configuration file.  You may
+# add to this file or create additional files in registries.d/.
+#
+# sigstore: indicates a location that is read and write
+# sigstore-staging: indicates a location that is only for write
+#
+# sigstore and sigstore-staging take a value of the following:
+#   sigstore:  {schema}://location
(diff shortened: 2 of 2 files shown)


default.yaml
```diff
@@ -0,0 +1,26 @@
...
+# sigstore: indicates a location that is read and write
+# sigstore-staging: indicates a location that is only for write
+#
+# sigstore and sigstore-staging take a value of the following:
+#   sigstore:  {schema}://location
+#
+# For reading signatures, schema may be http, https, or file.
+# For writing signatures, schema may only be file.
+
+# This is the default signature write location for registries.
```
mtrmac: for docker registries, to be pedantic.

runcom:
> Should this have some kind of Makefile integration? As it is, it does nothing and is, at best, an example to be stowed away in docs.

right, let's install it with make also

mtrmac:
BTW, now thinking about what we do with `policy.json`, the way we unconditionally overwrite users’ configuration seems pretty risky. Both of these files should be installed only if nothing already exists in the destination.

runcom:
@mtrmac PTAL, fixed your comments

runcom:
> BTW, now thinking about what we do with policy.json, the way we unconditionally overwrite users’ configuration seems pretty risky. Both of these files should be installed only if nothing already exists in the destination.

we have config(noreplace) in the RPM, I believe who does make from source knows what he's doing

Makefile
```diff
@@ -62,8 +62,10 @@ clean:
 	rm -f skopeo docs/*.1
 
 install: install-binary install-docs install-completions
+	install -d -m 755 ${CONTAINERSSYSCONFIGDIR}
```
mtrmac: Doesn’t the `-D` immediately below make this unnecessary?

Makefile
```diff
@@ -62,8 +62,10 @@ clean:
...
 	install -D -m 644 default-policy.json ${CONTAINERSSYSCONFIGDIR}/policy.json
 	install -d -m 755 ${REGISTRIESDDIR}
+	install -D -m 644 default.yaml ${REGISTRIESDDIR}/default.yaml
```
mtrmac: Similarly, the line above can now be dropped perhaps.

default.yaml
```diff
@@ -0,0 +1,26 @@
...
+# sigstore and sigstore-staging take a value of the following:
+#   sigstore:  {schema}://location
+#
+# For reading signatures, schema may be http, https, or file.
+# For writing signatures, schema may only be file.
+
+# This is the default signature write location for docker registries.
+default-docker:
+#  sigstore: file:///var/lib/atomic/sigstore
+  sigstore-staging: file:///var/lib/atomic/sigstore
```
mtrmac: Should we be creating this directory as well?

Makefile
runcom: fixed

default.yaml
runcom: fixed

runcom:
@mtrmac can you rebase your integrate-all-the-things branch so that it contains this commit? could you ping me once done, thanks!

default.yaml
rhatdan: container image registries.

mtrmac: No, this does _not_ apply to OpenShift.

mtrmac:
> @mtrmac can you rebase your integrate-all-the-things branch so that it contains this commit? could you ping me once done, thanks!

Done, sorry for the delay.

PR #236, opened 2016-10-20
Rename type insecureAcceptAnything to accept

aweiteka:
Required for pending change in upstream library https://github.com/containers/image/pull/132

Signed-off-by: Aaron Weitekamp aweiteka@redhat.com

### DIFF
--- a/default-policy.json
+++ b/default-policy.json
@@ -1,7 +1,7 @@
 {
     "default": [
         {
-            "type": "insecureAcceptAnything"
+            "type": "accept"
         }
     ]
 }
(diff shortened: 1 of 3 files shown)


rhatdan:
Don't we need to get this into skopeo before we ship rhel7.3.1?

aweiteka:
We need this in https://github.com/containers/image/pull/132 first, then this can go in. We can stage support in (atomic CLI)[https://github.com/projectatomic/atomic/blob/master/Atomic/trust.py] so it will be backwards compatible.

rhatdan:

aweiteka:
At this point I consider skopeo an "expert" tool so I'm okay just mapping this to 'accept' in atomic CLI or any other tools that call/wrap skopeo.

rhatdan:
Then can you close this pull request.

PR #241, opened 2016-11-01
Add add-signature and verify-signature

nalind:
Add add-signature and verify-signature commands, for adding a signature to an image in its current location, and for verifying an image in place.

### DIFF
--- a/cmd/skopeo/signing.go
+++ b/cmd/skopeo/signing.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 
 	"github.com/containers/image/signature"
+	"github.com/containers/image/transports"
 	"github.com/urfave/cli"
 )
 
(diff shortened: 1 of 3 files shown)


runcom:
Ping @mtrmac

cmd/skopeo/signing.go
```diff
@@ -88,3 +89,137 @@ var standaloneVerifyCmd = cli.Command{
...
+		return errors.New("Usage: skopeo add-signature reference key-fingerprint")
+	}
+	ref := context.Args()[0]
+	fingerprint := context.Args()[1]
+
+	imageRef, err := transports.ParseImageName(ref)
+	if err != nil {
+		return fmt.Errorf("Error parsing %s: %v", ref, err)
+	}
+	name := transports.ImageName(imageRef)
```
mtrmac: Why not use `ref` (the original user input) directly? Using exactly the users' input is simpler, and could make it less confusing.

cmd/skopeo/signing.go
```diff
@@ -88,3 +89,137 @@ var standaloneVerifyCmd = cli.Command{
...
+
+	// TODO: the signed named reference is always relative to the transport.  Is this the right value?
+	signReference := imageRef.StringWithinTransport()
```
mtrmac: No; it would be `imageRef.DockerReference` _if it is not `nil`_.  The `docker-reference` field in a signature is using the `containers/image/docker/reference` semantics, it is not transport-specific.

If `imageRef.DockerReference` is `nil`, the user would have to specify the signed identity explicitly (probably as an option, not as a mandatory argument). Or it can be left unimplemented for now.

cmd/skopeo/signing.go
```diff
@@ -88,3 +89,137 @@ var standaloneVerifyCmd = cli.Command{
...
+
+	signatures, err := src.GetSignatures()
+	if err != nil {
+		return fmt.Errorf("Error reading signatures from %s: %v", name, err)
+	}
+
+	signatures = append(signatures, signature)
+
+	err = dest.PutSignatures(signatures)
+	if err != nil {
```
mtrmac: (Nit: the two lines above, and similarly for `Commit` below, could be combined.)

cmd/skopeo/signing.go
```diff
@@ -88,3 +89,137 @@ var standaloneVerifyCmd = cli.Command{
...
+		return fmt.Errorf("Error writing to %s: %v", name, err)
+	}
+	defer dest.Close()
+
+	signatures, err := src.GetSignatures()
+	if err != nil {
+		return fmt.Errorf("Error reading signatures from %s: %v", name, err)
+	}
+
+	signatures = append(signatures, signature)
```
mtrmac: This will keep adding signatures even if a signature already exists for the (fingerprint, `signReference`) combination. That of course doesn’t hurt, and we don’t easily provide a way to avoid that; I guess we could only explicitly document that we might do nothing in the future, if the required signature is already present.

cmd/skopeo/signing.go
```diff
@@ -88,3 +89,137 @@ var standaloneVerifyCmd = cli.Command{
...
+		return errors.New("Usage: skopeo verify-signature reference key-fingerprint")
+	}
+	ref := context.Args()[0]
+	expectedFingerprint := context.Args()[1]
+
+	imageRef, err := transports.ParseImageName(ref)
+	if err != nil {
+		return fmt.Errorf("Error parsing %s: %v", ref, err)
+	}
+	name := transports.ImageName(imageRef)
```
mtrmac: Should just use `ref` directly, see above.

cmd/skopeo/signing.go
```diff
@@ -88,3 +89,137 @@ var standaloneVerifyCmd = cli.Command{
...
+	if err != nil {
+		return fmt.Errorf("Error reading signatures from %s: %v", name, err)
+	}
+
+	if len(signatures) == 0 {
+		return fmt.Errorf("No signatures found: %v", err)
+	}
+
+	// TODO: the signed named reference is always relative to the transport.  Is this the right value?
+	expectedReference := imageRef.StringWithinTransport()
```
mtrmac: No, see above.

cmd/skopeo/signing.go
nalind: The value we get back can be different from the passed-in value (IIRC the docker reference will include the default `latest` tag if the incoming value doesn't contain one, and having a signature check fail because either the signed or expected value included that tag and the other one didn't got annoying pretty fast.

cmd/skopeo/signing.go
nalind: Fixing.

nalind:
The use case, at the time, was checking that the storage backend implementation was fully functional in terms of signing and verifying images, preferably without implicitly depending on the feature being supported in other backends.

I could see add-signature being useful in a future registry setup, but yeah, I was hedging on the wording there.  I could also see verify-signature being used to defer the signature check until, or repeat the signature check when, we actually try to start up a container using the image, if we want to allow for things like key revocation, but that wasn't on my mind at the time.

cmd/skopeo/signing.go
nalind: Okay, sounds simple enough - default to DockerReference().String() if possible, overridden with a CLI flag, and if neither is possible, return an error.  Fixing.

cmd/skopeo/signing.go
nalind: Leaving this, pending conversation above.

nalind:
Okay, other than whether we produce error messages using the passed-in image reference or a value retrived from `transports.ImageName()`, I think all feedback has been incorporated.  If I've missed something, though, please remind me.

rhatdan:
@mtrmac Is this ok to merge?

cmd/skopeo/signing.go
mtrmac: Yeah, but `name` is not used (neither now, nor IIRC in the original variant) for building the `signReference` value.

cmd/skopeo/signing.go
```diff
@@ -88,3 +89,179 @@ var standaloneVerifyCmd = cli.Command{
...
+		if signReference == nil {
+			return fmt.Errorf("No reference associated with image %s: %v", ref, err)
+		}
+		reference = signReference.String()
+	} else {
+		signRef, err := imageRef.Transport().ParseReference(reference)
+		if err != nil {
+			return fmt.Errorf("Error parsing %s: %v", reference, err)
+		}
+		signReference := signRef.DockerReference()
```
mtrmac: This won’t work; if `imageRef.DockerReference()` (in the `reference == ""` case) is `nil`, i.e. the user _needs_ to supply a reference, `signRef.DockerReference()` will most likely also be `nil` in this case (it will be calling the same implementation).  Either just use the input `reference` string as is, or, to validate it, parse it using`containers/image/docker/reference.ParseNamed` possibly with `.WithDefaultTag`.

mtrmac:
> The use case, at the time, was checking that the storage backend implementation was fully functional in terms of signing and verifying images, preferably without implicitly depending on the feature being supported in other backends.

That’s a fairly weak reason to expose this (or, to be precise, _exactly_ this) to users… using the `dir:` backend as a counterpart, or writing one-off throwaway versions of the tools, or testing in unit tests, could work for those purposes without adding a CLI.

(The case for `standalone-sign` and `standalone-verify` is similarly weak, better only because these building blocks are the bare minimum and once they are present, everything else can be built in shell scripts.)

> I could see add-signature being useful in a future registry setup, but yeah, I was hedging on the wording there.

Yeah, I’m sure there are workflows where it will be extremely useful, but I am worried about someone just calling `skopeo add-signature docker://mycompany/importantproduct $supertrustedkey` without realizing that they aren’t in control of what they are signing. How do we nudge users to protect themselves?

Some possibilities:

 * Do a signature check on the original image before signing it. (That can get naively self-referential - `add-signature` would refuse to sign by $key because it wants a signature by $key first.)
 * Require the user to explicitly specify the image digest, with a very explicit name like `--manifest-digest-I-know-is-expected` (I suck at naming).

Perhaps both? Then the user can either sign images accepted by the policy (`skopeo add-signature --reference example.com/production-repo/product:$version atomic://qa-repo/product:$version $production-key`), or sign images with an explicit reference (`skopeo add-signature --manifest-digest-I-know-is-expected $digest atomic://production-repo/product:$version $production_key`).

> I could also see verify-signature being used to defer the signature check until, or repeat the signature check when, we actually try to start up a container using the image, if we want to allow for things like key revocation, but that wasn't on my mind at the time.

I’d kind of expect these checks to be done by Go code instead of calling an executable, but exposing re-checking to shell scripts could work. But I do strongly think that such re-verification should happen against the policy, not against a single specified digest+reference. (That’s actually simpler to implement, see `IsRunningImageAllowed` in `containers/image/copy/copy.go`.)

And again, how do we nudge users not to do `skopeo verify docker://mycompany/importantproduct` and then assume that further references to that are OK? One possibility would be to refuse to verify remote repositories but to still allow local ones (`dir:`, `docker-daemon:` etc.)—the local ones are still racy but presumably under the same administrative control; though that feels pretty draconian, there should be a better way.

rhatdan:

mtrmac:
Something a bit like `add-signature` was suggested as useful for OpenShift, but OpenShift would more likely want a Go API than a CLI; and in the mean time the `ImageDestination` API has been interpreted more restrictively in that just calling a `PutSignatures` without anything else is not necessarily legal; so the bulk of the work would be in changing this API rather than the CLI.

As for `verify-signature`, I still have concerns about TOCTOU vulnerabilities; and the suggested use case of verifying signatures on container start etc. will be handled differently (again by calling a Go API—and one specialized for verifying signatures of already-extracted images, at that).

rhatdan:
@nalind @mtrmac This needs to be updated or closed

rhatdan:
I am closing since no one has touched in another 6 months.

PR #242, opened 2016-11-02
Fix skopeo layers and deprecate it

runcom:
As title says. @mtrmac PTAL

@giuseppe I know you're using `skopeo layers` in your sys containers, would be nice to migrate that to `mkdir dir && skopeo copy docker://image_name dir:here_are_the_layers`

### DIFF
--- a/cmd/skopeo/layers.go
+++ b/cmd/skopeo/layers.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"io/ioutil"
 	"strings"
 
@@ -15,10 +16,13 @@ import (
 var layersCmd = cli.Command{
 	Name:      "layers",
 	Usage:     "Get layers of IMAGE-NAME",
-	ArgsUsage: "IMAGE-NAME",
+	ArgsUsage: "IMAGE-NAME [LAYER...]",
 	Action: func(c *cli.Context) error {
+		if c.NArg() == 0 {
+			return errors.New("please specify an image")
+		}
 		rawSource, err := parseImageSource(c, c.Args()[0], []string{
-			// TODO: skopeo layers only support these now
+			// TODO: skopeo layers only supports these now
 			// eventually we'll remove this command altogether...
 			manifest.DockerV2Schema1SignedMediaType,
 			manifest.DockerV2Schema1MediaType,
--- a/cmd/skopeo/layers.go
+++ b/cmd/skopeo/layers.go
@@ -2,7 +2,9 @@ package main
 
 import (
 	"errors"
+	"fmt"
 	"io/ioutil"
+	"os"
 	"strings"
 
 	"github.com/containers/image/directory"
@@ -12,14 +14,14 @@ import (
 	"github.com/urfave/cli"
 )
 
-// TODO(runcom): document args and usage
 var layersCmd = cli.Command{
 	Name:      "layers",
 	Usage:     "Get layers of IMAGE-NAME",
 	ArgsUsage: "IMAGE-NAME [LAYER...]",
 	Action: func(c *cli.Context) error {
+		fmt.Fprintln(os.Stderr, `DEPRECATED: skopeo layers is deprecated in favor of skopeo copy`)
 		if c.NArg() == 0 {
-			return errors.New("please specify an image")
+			return errors.New("Usage: layers imageReference [layer...]")
 		}
 		rawSource, err := parseImageSource(c, c.Args()[0], []string{
 			// TODO: skopeo layers only supports these now


cmd/skopeo/layers.go
```diff
@@ -15,10 +17,14 @@ import (
 var layersCmd = cli.Command{
 	Name:      "layers",
 	Usage:     "Get layers of IMAGE-NAME",
-	ArgsUsage: "IMAGE-NAME",
+	ArgsUsage: "IMAGE-NAME [LAYER...]",
 	Action: func(c *cli.Context) error {
+		fmt.Println(`DEPRECATED: skopeo layers is deprecated in favor of skopeo copy `)
```
mtrmac: Shouldn’t this go to stderr? I guess I can see a case for either… just something to consider.

cmd/skopeo/layers.go
```diff
@@ -15,10 +17,14 @@ import (
...
+		if c.NArg() == 0 {
+			return errors.New("please specify an image")
+		}
```
mtrmac: ACK on having the check; for consistency, this should say `usage: correct usage`, like e.g. https://github.com/projectatomic/skopeo/blob/master/cmd/skopeo/delete.go#L12

mtrmac:
> would be nice to migrate that to `mkdir dir && skopeo copy docker://image_name dir:here_are_the_layers`

(Do we document and commit to the `dir:` format as an ABI? I guess we need to… Not that having `skopeo layers` undocumented ever stopped anyone 😈 )

runcom:
> (Do we document and commit to the dir: format as an ABI? I guess we need to… Not that having skopeo layers undocumented ever stopped anyone  )

I think we should commit to `oci:` instead here

mtrmac:
> I think we should commit to `oci:` instead here

That might involve a schema or metadata conversion… which might actually be desirable, depending on the situation.

runcom:
alright, fixed your comments - we'll talk about atomic moving to copy in another issue maybe (no urgency right now)

mtrmac:
Thanks! LGTM pending tests.

rhatdan:
Lets remove the layers information in the man page.  Perhaps remove layers from command completion.  I would prefer we make it dissapear if we don't want users using it any more.

giuseppe:
one advantage of `skopeo layers` is that I could specify which layers to download, so that `atomic pull` will only fetch the missing ones (a layer could be already stored locally, from another image, or didn't change in a newer version of the image).  How can I do it with `copy`?

mtrmac:
`copy` can’t do that for `dir:`; it would be possible to write an `ostree:` transport for containers/image and then `skopeo copy` would be able to skip redownloading, but that’s obviously not all that attractive given the existing Python implementation.

giuseppe:
could we add a filter to `skopeo copy` that allows to specify which layers to consider as `skopeo layers` currently does?  We can drive it from atomic in the same way as we currently do.

mtrmac:
In principle, sure, it’s code. Though it is a bit unclear what this does to the semantics of `skopeo copy` (will the manifest still contain all layers?), and some cases (schema conversion) really need the layers at least read if not written. So, it would be pretty ugly in an already pretty complex copy pipeline.

Separately, what semantics exactly do you need? Is it just a blind “download one layer, the caller is reading and interpreting the manifest and dealing with any semantics”, or it the caller actually expecting intelligence and conversions etc. from skopeo?

**In particular, should signatures be verified when copying the manifest and data** for your use case?

giuseppe:
what we do in atomic is to first fetch the manifest, parse it, and then retrieve all the layers that are not already stored in the OSTree repository.  What is needed from atomic is the possibility to fetch a subset of the layers in an image.
The only intelligence needed is for the signatures.  Are signatures verified on the manifest?

mtrmac:
Right now, `skopeo inspect` does not enforce the signature policy, only `skopeo copy` does. It could be added to `inspect`, at least as an option.  After that, whatever is downloading thes layer *must* verify the layer blobs against their digests, which current `skopeo layers` isn’t doing either.

@runcom Opinions on the path forward?

It seems to me that we either ignore signatures, in which case it seems most natural to revive `skopeo layers`, Or we plan to support them for pulling to `ostree:`, in which case a native Golang `ostree:` transport would be conceptually simpler than carving up `skopeo copy` into pieces and then having `atomic` reassemble it back together. OTOH https://github.com/projectatomic/atomic/blob/master/Atomic/syscontainers.py is pretty big.

… after all, a `skopeo copy` restricted to only a smaller set of layers starts looking as a pretty attractive option…

giuseppe:
would it be possible to check signatures having only a subset of the layers?  Having something like `--skip-layers=....` seems to me like a quite generic solution and won't require to change what we already have in place.

rhatdan:
In the future we would also want to support container/storage as a back end for skopy copy also.  Is there a simple way to get that to happen also?

mtrmac:
containers/storage will be integrated natively ( https://github.com/containers/image/pull/63 ), so these disassemble/reassemble-at-the-CLI-level concerns do not apply.  (Also, it _always_ reads all inputs in https://github.com/containers/image/pull/63/files#diff-d77114c36cb773f7266e73a470816298R218 so it is not as interested in skipping existing layers.)

mtrmac:
> … after all, a `skopeo copy` restricted to only a smaller set of layers starts looking as a pretty attractive option…

> would it be possible to check signatures having only a subset of the layers? Having something like --skip-layers=.... seems to me like a quite generic solution and won't require to change what we already have in place.

Well, “generic” only to the `atomic`-like “dig into the internal implementation details” users. But yes, it does seem the simplest way to integrate the two; https://github.com/containers/image/pull/159 is not _pretty_ but it seems fairly acceptable, at least if we add some tests… @runcom ?

rhatdan:
@mtrmac Are you saying that there will be a 
skopeo copy docker:image storage
So that atomic command could pull an image from a registry and install it directly into a local container/storage repository?

mtrmac:
> Are you saying that there will be a skopeo copy docker:image storage
> So that atomic command could pull an image from a registry and install it directly into a local container/storage repository? 

I’d expect `skopeo copy docker:busybox storage:busybox` to work with https://github.com/containers/image/pull/63 , already today, yes.

rhatdan:
Awesome, if this was merged...

PR #243, opened 2016-11-03
Add static compile target in Makefile

hustcat:
I want to run `skopeo` on CentOS with low glibc version, so static compile is better for me.

Signed-off-by: Ye Yin <eyniy@qq.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ MANPAGES_MD = $(wildcard docs/*.md)
 #     Note: Uses the -N -l go compiler options to disable compiler optimizations
 #           and inlining. Using these build options allows you to subsequently
 #           use source debugging tools like delve.
-all: binary docs
+all: binary binary-static docs
 
 # Build a docker image (skopeobuild) that has everything we need to build.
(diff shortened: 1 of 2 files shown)


runcom:
Seems ok

Makefile
```diff
@@ -36,7 +36,7 @@ MANPAGES_MD = $(wildcard docs/*.md)
 #     Note: Uses the -N -l go compiler options to disable compiler optimizations
 #           and inlining. Using these build options allows you to subsequently
 #           use source debugging tools like delve.
-all: binary docs
+all: binary binary-static docs
```
mtrmac: With this, `make all` _both_ runs `make binary`→`make binary-local`, and runs `make binary-static`→`make binary-static-local`, i.e. a `skopeo` binary is created twice, with the static one overwriting the dynamic one. That doesn’t really make sense.

@hustcat What was this _intended_ to do?

hustcat: Oh, sorry. How could delete `binary-static` from `all` target?

PR #248, opened 2016-11-15
Use docker/distribution/digest

Crazykev:
use docker/distribution/digest instead of string, 
mainly test patch containers/image#163, should not merge before that.

cc @runcom

### DIFF
--- a/cmd/skopeo/layers.go
+++ b/cmd/skopeo/layers.go
@@ -11,6 +11,7 @@ import (
 	"github.com/containers/image/image"
 	"github.com/containers/image/manifest"
 	"github.com/containers/image/types"
+	"github.com/docker/distribution/digest"
 	"github.com/urfave/cli"
 )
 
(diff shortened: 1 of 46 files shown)


cmd/skopeo/layers.go
```diff
@@ -70,9 +78,6 @@ var layersCmd = cli.Command{
 		defer dest.Close()
 
 		for _, digest := range blobDigests {
-			if !strings.HasPrefix(digest, "sha256:") {
-				digest = "sha256:" + digest
-			}
```
mtrmac: This would change behavior. Unless @runcom or someone else knows that this is no longer necessary, we need to preserve this, probably above `ParseDigest` above.

Crazykev: @mtrmac sorry for that, it do change the behavior. Will fix.

Crazykev:
@mtrmac fixed and updated image vendor as well.

mtrmac:
LGTM, please drop the `hack/vendor.sh` part, and rerun `hack/vendor.sh`.

Crazykev:
@mtrmac updated.

Crazykev:
@mtrmac OK, checked, it's already the latest image master branch.

mtrmac:
LGTM pending tests.

runcom:
tests green, merging.

PR #252, opened 2016-11-23
Only build one skopeo binary by (make all)

mtrmac:
Both `make binary` and `make binary-static` compile the code and create a skopeo binary, so `make all` should only depend on one of them.


rhatdan:
LGTM but an alternative would have been to create skopeo-static as the binary.  But I don't think most people want to build skopeo-static.

PR #254, opened 2016-11-25
use user/pass flags

### LINKED ISSUE #253: docker:// transport requires "docker login" and doesn't obey --username and --password
If you try to do a `skopeo copy` to a Docker registry repository that doesn't exist you get an error:

```
% skopeo copy docker://opensuse docker://cyphar/nonexistant
Copying blob sha256:467db25190688bc9dc1d5fd6dbced1ac56f55d93a942987f448e62ad2614e46e
 0 B / 46.97 MB [--------------------------------------------------------------]
FATA[0004] Error writing blob: not authorized to read from destination repository cyphar/nonexistent
```

The above example does work if you first do a `docker push cyphar/nonexistent` but that's quite annoying. I'm not sure if this should be transparently handled by containers/image or if we should be implementing this policy.

runcom:
We already use global flags for docker specific stuff. This patch enables `--username` and `--password` to be passed down to `containers/image` to setup docker's registries auth.

Fixes #253 

@mtrmac @cyphar PTAL

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,11 +49,12 @@ Copying images
 -
 `skopeo` can copy container images between various storage mechanisms,
 e.g. Docker registries (including the Docker Hub), the Atomic Registry,
-and local directories:
+local directories, and local OCI-layout directories:
 
 ```sh
 $ skopeo copy docker://busybox:1-glibc atomic:myns/unsigned:streaming
 $ skopeo copy docker://busybox:latest dir:existingemptydirectory
+$ skopeo copy docker://busybox:latest oci:busybox_ocilayout
 ```
 
 Deleting images
@@ -65,14 +66,10 @@ $ skopeo delete docker://localhost:5000/imagename:latest
 
 Private registries with authentication
 -
-When interacting with private registries, `skopeo` first looks for the Docker's cli config file (usually located at `$HOME/.docker/config.json`) to get the credentials needed to authenticate. When the file isn't available it falls back looking for `--username` and `--password` flags. The ultimate fallback, as Docker does, is to provide an empty authentication when interacting with those registries.
+When interacting with private registries, `skopeo` first looks for `--creds` (for `skopeo inspect|delete`) or `--src-creds|--dest-creds` (for `skopeo copy`) flags. If those aren't provided, it looks for the Docker's cli config file (usually located at `$HOME/.docker/config.json`) to get the credentials needed to authenticate. The ultimate fallback, as Docker does, is to provide an empty authentication when interacting with those registries.
 
 Examples:
 ```sh
-# on my system
-$ skopeo --help | grep docker-cfg
-   --docker-cfg "/home/runcom/.docker"	Docker's cli config for auth
-
 $ cat /home/runcom/.docker/config.json
 {
 	"auths": {
@@ -88,16 +85,22 @@ $ skopeo inspect docker://myregistrydomain.com:5000/busybox
 {"Tag":"latest","Digest":"sha256:473bb2189d7b913ed7187a33d11e743fdc2f88931122a44d91a301b64419f092","RepoTags":["latest"],"Comment":"","Created":"2016-01-15T18:06:41.282540103Z","ContainerConfig":{"Hostname":"aded96b43f48","Domainname":"","User":"","AttachStdin":false,"AttachStdout":false,"AttachStderr":false,"Tty":false,"OpenStdin":false,"StdinOnce":false,"Env":null,"Cmd":["/bin/sh","-c","#(nop) CMD [\"sh\"]"],"Image":"9e77fef7a1c9f989988c06620dabc4020c607885b959a2cbd7c2283c91da3e33","Volumes":null,"WorkingDir":"","Entrypoint":null,"OnBuild":null,"Labels":null},"DockerVersion":"1.8.3","Author":"","Config":{"Hostname":"aded96b43f48","Domainname":"","User":"","AttachStdin":false,"AttachStdout":false,"AttachStderr":false,"Tty":false,"OpenStdin":false,"StdinOnce":false,"Env":null,"Cmd":["sh"],"Image":"9e77fef7a1c9f989988c06620dabc4020c607885b959a2cbd7c2283c91da3e33","Volumes":null,"WorkingDir":"","Entrypoint":null,"OnBuild":null,"Labels":null},"Architecture":"amd64","Os":"linux"}
 
 # let's try now to fake a non existent Docker's config file
-$ skopeo --docker-cfg="" inspect docker://myregistrydomain.com:5000/busybox
-FATA[0000] Get https://myregistrydomain.com:5000/v2/busybox/manifests/latest: no basic auth credentials
+$ cat /home/runcom/.docker/config.json
+{}
 
-# passing --username and --password - we can see that everything goes fine
-$ skopeo --docker-cfg="" --username=testuser --password=testpassword inspect docker://myregistrydomain.com:5000/busybox
+$ skopeo inspect docker://myregistrydomain.com:5000/busybox
+FATA[0000] unauthorized: authentication required
+
+# passing --creds - we can see that everything goes fine
+$ skopeo inspect --creds=testuser:testpassword docker://myregistrydomain.com:5000/busybox
 {"Tag":"latest","Digest":"sha256:473bb2189d7b913ed7187a33d11e743fdc2f88931122a44d91a301b64419f092","RepoTags":["latest"],"Comment":"","Created":"2016-01-15T18:06:41.282540103Z","ContainerConfig":{"Hostname":"aded96b43f48","Domainname":"","User":"","AttachStdin":false,"AttachStdout":false,"AttachStderr":false,"Tty":false,"OpenStdin":false,"StdinOnce":false,"Env":null,"Cmd":["/bin/sh","-c","#(nop) CMD [\"sh\"]"],"Image":"9e77fef7a1c9f989988c06620dabc4020c607885b959a2cbd7c2283c91da3e33","Volumes":null,"WorkingDir":"","Entrypoint":null,"OnBuild":null,"Labels":null},"DockerVersion":"1.8.3","Author":"","Config":{"Hostname":"aded96b43f48","Domainname":"","User":"","AttachStdin":false,"AttachStdout":false,"AttachStderr":false,"Tty":false,"OpenStdin":false,"StdinOnce":false,"Env":null,"Cmd":["sh"],"Image":"9e77fef7a1c9f989988c06620dabc4020c607885b959a2cbd7c2283c91da3e33","Volumes":null,"WorkingDir":"","Entrypoint":null,"OnBuild":null,"Labels":null},"Architecture":"amd64","Os":"linux"}
+
+# skopeo copy example:
+$ skopeo copy --src-creds=testuser:testpassword docker://myregistrydomain.com:5000/private oci:local_oci_image
 ```
 If your cli config is found but it doesn't contain the necessary credentials for the queried registry
-you'll get an error. You can fix this by either logging in (via `docker login`) or providing `--username`
-and `--password`.
+you'll get an error. You can fix this by either logging in (via `docker login`) or providing `--creds` or `--src-creds|--dest-creds`.
+
 Building
 -
 To build the manual you will need go-md2man.
@@ -116,7 +119,7 @@ $ brew install gpgme
 $ make binary-local
 ```
 
-You may need to install additional development packages: gpgme-devel and libassuan-devel
+You may need to install additional development packages: `gpgme-devel` and `libassuan-devel`
 ```sh
 $ dnf install gpgme-devel libassuan-devel
 ```
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -7,8 +7,24 @@ import (
 
 	"github.com/containers/image/copy"
 	"github.com/containers/image/transports"
+	"github.com/containers/image/types"
 	"github.com/urfave/cli"
 )
+
+// contextsFromGlobalOptions returns source and destionation types.SystemContext depending on c.
+func contextsFromGlobalOptions(c *cli.Context) (*types.SystemContext, *types.SystemContext, error) {
+	sourceCtx, err := contextFromGlobalOptions(c, "src-creds")
+	if err != nil {
+		return nil, nil, err
+	}
+
+	destinationCtx, err := contextFromGlobalOptions(c, "dest-creds")
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return sourceCtx, destinationCtx, nil
+}
 
 func copyHandler(context *cli.Context) error {
 	if len(context.Args()) != 2 {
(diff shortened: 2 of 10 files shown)


mtrmac:
I am not really thrilled with the generic option names: if `--username` and `--password` never worked, nothing stops us from naming them `--docker-username` / `--docker-password`. 

I am worried that w really shouldn’t send the same password to several different servers. OTOH the distinction really isn’t between transport types, but between individual hostnames, and that would be a horrible UI, so a generic `--username` and fervently hoping that the users know what they are doing might be the best I can do.

This should also restore `TestCanAuthToPrivateRegistryV2WithoutDockerCfg` and perhaps some of the neighboring tests, though.

After that, I guess, LGTM.

mtrmac:
(As for “we already do use generic option names for Docker-specific options”, that's partly legacy. And partly using `--cert-path` for an unwanted service is not a vulnerability, but sending a password to an unwanted service is. But, really, I am bad at designing UIs,)

runcom:
> This should also restore TestCanAuthToPrivateRegistryV2WithoutDockerCfg and perhaps some of the neighboring tests, though

Sort of.. I have to mock an empty Docker cfg first. I'll add that though, thx. 

> I am not really thrilled with the generic option names: if --username and --password never worked, nothing stops us from naming them --docker-username / --docker-password.

We used to have user/pass flag in the very first skopeo versions :/

mtrmac:
Yeah; I am not _thrilled_ with `--password` but I can live with that.

runcom:
@mtrmac what if we add `--src-username` and `--src-password` and the same for `dest`. That's not horrible to me and IIUC should solve your concern around wrongly sending a password to a wrong server.

mtrmac:
That’s great idea. We can very confidently expect that the sets of servers underlying a transport(`atomic:` native API + Docker registry API, `docker:` native API + sigstore) must be under a shared control.

Yes to `--{src,dest}-{username,password}`, please.

runcom:
Alright, do note that I'll make these new flags available only to `copy`. `inspect` and the now deprecated `layers` command will continue to behave as usual (w/o any user/pass flag set).

runcom:
Or wait, I can add username and password flag locally for inspect/layers commands. I'll do this way.

runcom:
@mtrmac PTAL, before going ahead and add docs/tests/`containers/image` bits, I would like to have your ack on the direction with `copy`. I'll also fix `layers|inspect` once we agree on `copy`.

rhatdan:
I don't like options with dashes ("-") in them.  Could we look at other tools like ssh? Would
```
--src dwalsh/password@registry --dest mitr/foobar@remote
```
be better?

runcom:
@rhatdan fine, but `--src` isn't clear enough because it's not saying we're talking about credentials

runcom:
We don't need to specify the registry also, that's implied by src and dest in the command and we don't need that (something like that is how the Docker auth config file works but that's another story, here in the cli we directly honor user/pass combination without checking the registry they'll be used to authenticate against)

runcom:
@rhatdan @mtrmac BTW, I'm fine compacting "username:password" in the cli as it  doesn't make sense to have 2 flags for user/pass when you must always use both. I'll rework this PR once we agree on flags naming though.

rhatdan:
@runcom It would help me if you gave a couple of examples of what you are proposing for a CLI.

runcom:
Like `--src-creds runcom:password` - modulo no dash in the flag

rhatdan:
I like that better

mtrmac:
> BTW, I'm fine compacting "username:password" in the cli as it doesn't make sense to have 2 flags for user/pass when you must always use both. I'll rework this PR once we agree on flags naming though.

That feels a bit off, as a general security habit I like to avoid formatting/parsing and in-band signalling. More importantly, it _may_ make sense to provide an username but not a password (e.g. when authenticating using a Kerberos ticket, or a smartcard)

runcom:
> BTW, I'm fine compacting "username:password" in the cli as it doesn't make sense to have 2 flags for user/pass when you must always use both. I'll rework this PR once we agree on flags naming though.
> That feels a bit off, as a general security habit I like to avoid formatting/parsing and in-band signalling. More importantly, it may make sense to provide an username but not a password (e.g. when authenticating using a Kerberos ticket, or a smartcard)

alright, no parsing then, any idea regarding Dan's issue on naming?

rhatdan:
Why wouldn't --cred dwalsh
Cause the passwd checks to happen to use kerberos like sshd does?

I guess if you had a case where there was a password without a user then we have a problem.

mtrmac:
> Why wouldn't --cred dwalsh
Cause the passwd checks to happen to use kerberos like sshd does?

Good point, that works fine.

runcom:
> Why wouldn't --cred dwalsh
> Cause the passwd checks to happen to use kerberos like sshd does?
> Good point, that works fine.

I probably missed something, are we ok with `--creds username:password`? I didn't understand Dan's example :confused:

mtrmac:
> I don't like options with dashes ("-") in them. 

This could also be handled by providing aliases:
```go
cli.StringFlag {
    Name: "src-username, suser"
    // …
}
```

and then
```
skopeo copy --suser mitr --spass sourcepass $source --duser Miloslav --dpass destpass $dest
```
(or even `--su`/`--sp`/`--du`/`--dp` but that’s probably too unreadable).

OTOH it _is_ more typing than
```
skopeo copy --screds mitr:sourcepass $source --dcreds Miloslav:destpass $dest
```

rhatdan:
runcom I am just giving a username not a password
--creds dwalsh 
Means Username = dwalsh No password
--creds dwalsh:foobar
Means Username=dwalsh Password=foobar.

cmd/skopeo/copy.go
```diff
@@ -32,10 +32,14 @@ func copyHandler(context *cli.Context) error {
...
 
-	return copy.Image(contextFromGlobalOptions(context), policyContext, destRef, srcRef, &copy.Options{
+	sourceCtx, destinationCtx := contextsFromGlobalOptions(context)
+
+	return copy.Image(policyContext, destRef, srcRef, &copy.Options{
 		RemoveSignatures: removeSignatures,
 		SignBy:           signBy,
 		ReportWriter:     os.Stdout,
+		SourceCtx:        sourceCtx,
+		DestinationCtx:   destinationCtx,
```
mtrmac: I guess this is unavoidable :/

mtrmac:
Overall, I like the separation of username/password but it does mean quite a bit of extra typing.

WRT the too-long-option names, we can provide shortcuts but it would still be nice to at least allow using long readable ones (`--dcreds` is also pretty cryptic).

= overall I don’t have a strong opinion on the single/two-argument decision.

mtrmac:
> We can very confidently expect that the sets of servers underlying a transport(`atomic:` native API + Docker registry API, `docker:` native API + sigstore) must be under a shared control.

Sadly, we can’t, when uploading to `docker.io` for docker/distribution vs. company-specific sigstore.

Ouch.

(This does not _immediately_ affect this PR because that is only setting a docker/distribution username:password. But any more such cases, and telling the user to run `docker login` so that we have a hostname-associated configuration stored and not a bazillion of command-line options will start to look pretty good.)

runcom:
> runcom I am just giving a username not a password
> --creds dwalsh
> Means Username = dwalsh No password
> --creds dwalsh:foobar
> Means Username=dwalsh Password=foobar.

Yup, thanks @rhatdan - I was asking if @mtrmac is fine with that also, Miloslav are you?

mtrmac:
I don’t have a strong opinion on `creds` vs. `username`+`password`; separating the source and destination still seems worthwhile to me.

runcom:
@mtrmac I was talking just about the "user:pass" format which should be parsed

mtrmac:
> @mtrmac I was talking just about the "user:pass" format which should be parsed

Yeah, I don’t have a strong opinion. If you like that, go for it.

runcom:
@mtrmac @rhatdan PTAL again, before going ahead and add docs/tests/`containers/image` bits, I would like to have your ack about `copy`. I'll also fix `layers|inspect` once we agree on `copy`.
The patch right now allows:
```bash
$ skopeo copy --src-creds dwalsh:password --dest-creds runcom docker://busybox docker://runcom/credentials
```

cmd/skopeo/utils.go
```diff
@@ -1,22 +1,55 @@
...
+	}
+	return up[0], up[1], nil
+}
+
+func getDockerAuth(creds string) (*types.DockerAuthConfig, error) {
+	username, password, err := parseCreds(creds)
+	if err != nil {
+		return nil, err
+	}
+	fmt.Println(username)
```
rhatdan: Looks like debug outpuy

runcom: Yup I'll drop this

rhatdan:
Missing changes to man pages and bash completions.

runcom:
Yup, if @mtrmac ack on the changes done so far, I'll update man, docs and everything else (adding tests also)

runcom:
This is ready to be reviewed fully. @mtrmac @rhatdan added man changes, README changes, tests, completions (it's auto-generated so...). What's missing is to create a PR for `containers/image` changes, I'll do that shortly after this has been reviewed.

README.md
```diff
@@ -65,14 +66,10 @@ $ skopeo delete docker://localhost:5000/imagename:latest
 
 Private registries with authentication
 -
-When interacting with private registries, `skopeo` first looks for the Docker's cli config file (usually located at `$HOME/.docker/config.json`) to get the credentials needed to authenticate. When the file isn't available it falls back looking for `--username` and `--password` flags. The ultimate fallback, as Docker does, is to provide an empty authentication when interacting with those registries.
+When interacting with private registries, `skopeo` first looks for the Docker's cli config file (usually located at `$HOME/.docker/config.json`) to get the credentials needed to authenticate. When the file isn't available it falls back looking for `--creds` (for `skopeo inspect`) or `--src-creds|--dest-creds` (for `skopeo copy`) flags. The ultimate fallback, as Docker does, is to provide an empty authentication when interacting with those registries.
```
rhatdan: I think this is wrong.  If the user specified --creds, --src-creds, --dest-creds, it should override content in config.json rather then fall back.

docs/skopeo.1.md
```diff
@@ -70,6 +66,10 @@ Uses the system's trust policy to validate images, rejects images not trusted by
 
   **--sign-by=**_key-id_ add a signature using that key ID for an image name corresponding to _destination-image_
 
+  **--src-creds** _username:password_ for accessing the source registry
```
rhatdan: Should this be 
```
_username[:password]_
```

rhatdan:
Still missing bash completion fixes.

runcom:
> Still missing bash completion fixes.

where is bash completion? `hack/make/bash_autocomplete` auto generates the bash completion on `make install` - not sure what I should fix :confused:

README.md
runcom: > I think this is wrong. If the user specified --creds, --src-creds, --dest-creds, it should override content in config.json rather then fall back.

Yeah, it's already this way, this piece of the README is wrong, I'll fix it.

docs/skopeo.1.md
runcom: very likely, I'll fix it

rhatdan:
Looks like it only generates one level of bash completions.  Lets ignore this for now and we can open an issue to have a much more complete bach completions.

README.md
```diff
@@ -88,16 +85,22 @@ $ skopeo inspect docker://myregistrydomain.com:5000/busybox
...
+$ cat /home/runcom/.docker/config.json
+{}
+
+$ skopeo inspect docker://myregistrydomain.com:5000/busybox
+FATA[0000] unauthorized: authentication required
 
-# passing --username and --password - we can see that everything goes fine
-$ skopeo --docker-cfg="" --username=testuser --password=testpassword inspect docker://myregistrydomain.com:5000/busybox
+# passing --creds - we can see that everything goes fine
+$ skopeo --docker-cfg="" --creds=testuser:testpassword inspect docker://myregistrydomain.com:5000/busybox
```
mtrmac: `--docker-cfg` seems not to exist at all

runcom:
> Looks like it only generates one level of bash completions. Lets ignore this for now and we can open an issue to have a much more complete bach completions.

ack, there's already an issue upstream which tracks this https://github.com/urfave/cli/issues/188

cmd/skopeo/delete.go
```diff
@@ -18,7 +18,14 @@ func deleteHandler(context *cli.Context) error {
 		return fmt.Errorf("Invalid source name %s: %v", context.Args()[0], err)
 	}
 
-	if err := ref.DeleteImage(contextFromGlobalOptions(context)); err != nil {
+	ctx := contextFromGlobalOptions(context)
+	if context.IsSet("creds") {
+		ctx.DockerAuthConfig, err = getDockerAuth(context.String("creds"))
```
mtrmac: If everyone is going to use it like this, shouldn't it be folded into `contextFromGlobalOptions(context, creds-option-name)`?

cmd/skopeo/copy.go
```diff
@@ -54,5 +84,15 @@ var copyCmd = cli.Command{
 			Name:  "sign-by",
 			Usage: "Sign the image using a GPG key with the specified `FINGERPRINT`",
 		},
+		cli.StringFlag{
+			Name:  "src-creds, screds",
+			Value: "",
+			Usage: "Use `USERNAME:PASSWORD` for accessing the source registry",
```
mtrmac: Could this use `USERNAME[:PASSWORD]`, i.e. does that work with the template-variable markers? (If so, please fix everywhere)

cmd/skopeo/utils.go
```diff
@@ -1,30 +1,67 @@
...
 		DockerCertPath:              c.GlobalString("cert-path"),
-		DockerInsecureSkipTLSVerify: !tlsVerify,
+		DockerInsecureSkipTLSVerify: !c.GlobalBoolT("tls-verify"),
 	}
 }
 
-// ParseImage converts image URL-like string to an initialized handler for that image.
+// parseImage converts image URL-like string to an initialized handler for that image.
 // The caller must call .Close() on the returned Image.
 func parseImage(c *cli.Context) (types.Image, error) {
```
mtrmac: Should `parseImageSource` be also updated to support `--creds`? (This would make it work for `skopeo layers`, though I am more concerned about the API symmetry than `layers` as such)

runcom:
@mtrmac @rhatdan updated

cmd/skopeo/utils.go
```diff
@@ -1,30 +1,72 @@
...
 	"github.com/containers/image/transports"
 	"github.com/containers/image/types"
 	"github.com/urfave/cli"
 )
 
-// contextFromGlobalOptions returns a types.SystemContext depending on c.
-func contextFromGlobalOptions(c *cli.Context) *types.SystemContext {
-	tlsVerify := c.GlobalBoolT("tls-verify")
+func contextFromGlobalOptions(c *cli.Context, credsFlag string) (*types.SystemContext, error) {
+	ctx := newBaseSystemContext(c)
```
mtrmac: I don’t think we need the separate `newBaseSystemContext` function now that `credsFlag` is handled inline.  (Eventually we may support `credsFlag == ""` to turn that off, but we don’t need that now.)

This does work just fine, feel free to merge as is.

rhatdan:
Bash Completions https://github.com/projectatomic/skopeo/pull/256

cmd/skopeo/utils.go
runcom: done, thx

runcom:
> Bash Completions #256

thanks Dan, we'll merge this one and re-iterate on #256

runcom:
my bad, I have to fix containers/image and revendor here. I'll do it asap, sorry

PR #255, opened 2016-11-25
Dockerfile: remove inline comments

runcom:
Newer docker version (> 1.12.x) now perform a pre-build syntax check. Inline comments are no longer allowed.


mtrmac:
LGTM.  It would be nice to preserve at least the comment about the version conflict, but do feel free to merge as is.

PR #259, opened 2016-12-01
[wip] integration: do not pull from Docker Hub

runcom:
- Fix https://github.com/projectatomic/skopeo/pull/132
- Requires https://github.com/containers/image/pull/148
- Cleanup the `Dockerfile` by removing registry V1 since we don't support it at all

When https://github.com/containers/image/pull/148 is merged - we'll load images on the test registries when the test suites come up.

@mtrmac PTAL

### DIFF
--- a/hack/download-frozen-image-v2.sh
+++ b/hack/download-frozen-image-v2.sh
@@ -1,0 +1,121 @@
+#!/bin/bash
+set -e
+
+# hello-world                      latest              ef872312fe1b        3 months ago        910 B
+# hello-world                      latest              ef872312fe1bbc5e05aae626791a47ee9b032efa8f3bda39cc0be7b56bfe59b9   3 months ago        910 B
+
+# debian                           latest              f6fab3b798be        10 weeks ago        85.1 MB
+# debian                           latest              f6fab3b798be3174f45aa1eb731f8182705555f89c9026d8c1ef230cbf8301dd   10 weeks ago        85.1 MB
+
+if ! command -v curl &> /dev/null; then
+	echo >&2 'error: "curl" not found!'
+	exit 1
+fi
+
+usage() {
+	echo "usage: $0 dir image[:tag][@digest] ..."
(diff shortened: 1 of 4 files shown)


hack/download-frozen-image-v2.sh
```diff
@@ -0,0 +1,121 @@
+#!/bin/bash
```
runcom: we'll likely remove this as well once we have `docker-archive` transport and directly use skopeo probably

mtrmac:
This doesn’t actually remove any Hub references from the tests yet, does it?

runcom:
> This doesn’t actually remove any Hub references from the tests yet, does it?

not yet, waiting on the `docker-archive` functionality at least

mtrmac:
If I understand correctly, the idea is to use that shell script → `docker-archive:` → `docker load` → `docker push` to recreated the needed images in our registry; is that right?

If so, why not just do `docker pull`&&`docker push`, perhaps both in the Dockerfile?

runcom:
@mtrmac we don't have `docker` in the docker container - and I wouldn't have it also.
The idea is to download the images in the `Dockerfile`, and at test suite setup `skopeo copy ...` from tarballs directly.

mtrmac:
Oops, that was stupid of me.

In that case, we could also use `dir:` instead of `docker-archive:`, that involves fewer manifest conversions. OTOH the conversions, both in the new script and in `docker-archive:`, are already implemented, so reworking the script to use the `dir:` format would be quite unnecessary work.

cyphar:
:+1: However we have to make sure that that `docker://` source is properly tested with the live Docker Hub. Maybe make it a separate suite or something -- but we still need to verify that the upstream Docker Hub still works with us.

runcom:
> However we have to make sure that that docker:// source is properly tested with the live Docker Hub. Maybe make it a separate suite or something -- but we still need to verify that the upstream Docker Hub still works with us.

of course, yes

rhatdan:

rhatdan:
@runcom Either update this or close it.

hack/download-frozen-image-v2.sh
cyphar: We've now had `docker-archive` for a while, so we can drop this. :wink:

PR #263, opened 2016-12-03
update c/image and docker/go-connections

runcom:
Vendor https://github.com/containers/image/pull/180 - This won't change any flags or anything anywhere for now. It's needed to not block development in `containers/image`.

@mtrmac PTAL and merge this one - afterward I'll fix `skopeo copy` to have src and dest certs directories. (already on my TODO :))

Signed-off-by: Antonio Murdaca <runcom@redhat.com>

PR #264, opened 2016-12-04
cmd: per command tls flags

runcom:
Based on #263 

@mtrmac @rhatdan PTAL

### DIFF
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -13,12 +13,12 @@ import (
 
 // contextsFromGlobalOptions returns source and destionation types.SystemContext depending on c.
 func contextsFromGlobalOptions(c *cli.Context) (*types.SystemContext, *types.SystemContext, error) {
-	sourceCtx, err := contextFromGlobalOptions(c, "src-creds")
+	sourceCtx, err := contextFromGlobalOptions(c, "src-")
 	if err != nil {
 		return nil, nil, err
 	}
 
-	destinationCtx, err := contextFromGlobalOptions(c, "dest-creds")
+	destinationCtx, err := contextFromGlobalOptions(c, "dest-")
 	if err != nil {
 		return nil, nil, err
 	}
@@ -86,6 +86,24 @@ var copyCmd = cli.Command{
 			Name:  "dest-creds, dcreds",
 			Value: "",
 			Usage: "Use `USERNAME[:PASSWORD]` for accessing the destination registry",
+		},
+		cli.StringFlag{
+			Name:  "src-cert-dir",
+			Value: "",
+			Usage: "use certificates at `PATH` (*.crt, *.cert, *.key) to connect to the source registry",
+		},
+		cli.BoolTFlag{
+			Name:  "src-tls-verify",
+			Usage: "require HTTPS and verify certificates when talking to the docker source registry (defaults to true)",
+		},
+		cli.StringFlag{
+			Name:  "dest-cert-dir",
+			Value: "",
+			Usage: "use certificates at `PATH` (*.crt, *.cert, *.key) to connect to the destination registry",
+		},
+		cli.BoolTFlag{
+			Name:  "dest-tls-verify",
+			Usage: "require HTTPS and verify certificates when talking to the docker destination registry (defaults to true)",
 		},
 	},
 }
--- a/cmd/skopeo/delete.go
+++ b/cmd/skopeo/delete.go
@@ -18,7 +18,7 @@ func deleteHandler(context *cli.Context) error {
 		return fmt.Errorf("Invalid source name %s: %v", context.Args()[0], err)
 	}
 
-	ctx, err := contextFromGlobalOptions(context, "creds")
+	ctx, err := contextFromGlobalOptions(context, "")
 	if err != nil {
 		return err
 	}
@@ -38,6 +38,15 @@ var deleteCmd = cli.Command{
 			Name:  "creds",
 			Value: "",
 			Usage: "Use `USERNAME[:PASSWORD]` for accessing the registry",
+		},
+		cli.StringFlag{
+			Name:  "cert-dir",
+			Value: "",
+			Usage: "use certificates at `PATH` (*.crt, *.cert, *.key) to connect to the registry",
+		},
+		cli.BoolTFlag{
+			Name:  "tls-verify",
+			Usage: "require HTTPS and verify certificates when talking to docker registries (defaults to true)",
 		},
 	},
 }
--- a/cmd/skopeo/layers.go
+++ b/cmd/skopeo/layers.go
@@ -19,6 +19,7 @@ var layersCmd = cli.Command{
 	Name:      "layers",
 	Usage:     "Get layers of IMAGE-NAME",
 	ArgsUsage: "IMAGE-NAME [LAYER...]",
+	Hidden:    true,
 	Action: func(c *cli.Context) error {
 		fmt.Fprintln(os.Stderr, `DEPRECATED: skopeo layers is deprecated in favor of skopeo copy`)
 		if c.NArg() == 0 {
--- a/cmd/skopeo/main.go
+++ b/cmd/skopeo/main.go
@@ -30,14 +30,10 @@ func createApp() *cli.App {
 			Name:  "debug",
 			Usage: "enable debug output",
 		},
-		cli.StringFlag{
-			Name:  "cert-path",
-			Value: "",
-			Usage: "use certificates at `PATH` (cert.pem, key.pem) to connect to the registry",
-		},
 		cli.BoolTFlag{
-			Name:  "tls-verify",
-			Usage: "require HTTPS and verify certificates when talking to docker registries (defaults to true)",
+			Name:   "tls-verify",
+			Usage:  "require HTTPS and verify certificates when talking to docker registries (defaults to true)",
+			Hidden: true,
 		},
 		cli.StringFlag{
 			Name:  "policy",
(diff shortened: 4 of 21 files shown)


cmd/skopeo/utils.go
```diff
@@ -9,12 +9,20 @@ import (
 	"github.com/urfave/cli"
 )
 
-func contextFromGlobalOptions(c *cli.Context, credsFlag string) (*types.SystemContext, error) {
+func contextFromGlobalOptions(c *cli.Context, credsFlag, certPathFlag, tlsVerifyFlag string) (*types.SystemContext, error) {
 	ctx := &types.SystemContext{
-		RegistriesDirPath:           c.GlobalString("registries.d"),
+		RegistriesDirPath: c.GlobalString("registries.d"),
+		// DEPRECATED: keep this here for backward compatibility, but override
+		// them if per subcommand flags are provided (see below).
```
runcom: This could entirely go away since we changed the way we lookup certs in `--cert-path` (no more `cert.pem`, `cert.key`), @mtrmac what do you think about getting rid of the global flags altogether?

cmd/skopeo/copy.go
```diff
@@ -87,5 +87,23 @@ var copyCmd = cli.Command{
 			Value: "",
 			Usage: "Use `USERNAME[:PASSWORD]` for accessing the destination registry",
 		},
+		cli.StringFlag{
+			Name:  "src-cert-path",
```
rhatdan: Options are a little wordy,  -path adds nothing to this.

cmd/skopeo/copy.go
```diff
@@ -87,5 +87,23 @@ var copyCmd = cli.Command{
...
+			Value: "",
+			Usage: "use certificates at `PATH` (*.crt, *.cert, *.key) to connect to the source registry",
+		},
+		cli.BoolTFlag{
+			Name:  "src-tls-verify",
+			Usage: "require HTTPS and verify certificates when talking to the docker source registry (defaults to true)",
+		},
+		cli.StringFlag{
+			Name:  "dest-cert-path",
```
rhatdan: Remove -path

cmd/skopeo/copy.go
```diff
@@ -87,5 +87,23 @@ var copyCmd = cli.Command{
 			Value: "",
 			Usage: "Use `USERNAME[:PASSWORD]` for accessing the destination registry",
 		},
+		cli.StringFlag{
+			Name:  "src-cert-path",
+			Value: "",
+			Usage: "use certificates at `PATH` (*.crt, *.cert, *.key) to connect to the source registry",
+		},
+		cli.BoolTFlag{
+			Name:  "src-tls-verify",
```
rhatdan: Remove -verify

cmd/skopeo/copy.go
```diff
@@ -87,5 +87,23 @@ var copyCmd = cli.Command{
...
+			Usage: "require HTTPS and verify certificates when talking to the docker source registry (defaults to true)",
+		},
+		cli.StringFlag{
+			Name:  "dest-cert-path",
+			Value: "",
+			Usage: "use certificates at `PATH` (*.crt, *.cert, *.key) to connect to the destination registry",
+		},
+		cli.BoolTFlag{
+			Name:  "dest-tls-verify",
```
rhatdan: remove -verify

cmd/skopeo/layers.go
```diff
@@ -19,8 +19,25 @@ var layersCmd = cli.Command{
 	Name:      "layers",
 	Usage:     "Get layers of IMAGE-NAME",
 	ArgsUsage: "IMAGE-NAME [LAYER...]",
+	Hidden:    true,
```
mtrmac: This would be worth calling out… if not a separate PR, at least a separate commit?

cmd/skopeo/main.go
```diff
@@ -31,13 +31,15 @@ func createApp() *cli.App {
...
+			Name:   "cert-path",
+			Value:  "",
+			Usage:  "use certificates at `PATH` (cert.pem, key.pem) to connect to the registry",
+			Hidden: true,
 		},
 		cli.BoolTFlag{
-			Name:  "tls-verify",
-			Usage: "require HTTPS and verify certificates when talking to docker registries (defaults to true)",
+			Name:   "tls-verify",
+			Usage:  "require HTTPS and verify certificates when talking to docker registries (defaults to true)",
```
mtrmac: If these two are deprecated and don’t do anything, should this say so instead of describing what these options don’t do?  Or perhaps the `Usage` is never used for `Hidden` flags?

cmd/skopeo/main.go
```diff
@@ -54,6 +56,12 @@ func createApp() *cli.App {
 		if c.GlobalBool("debug") {
 			logrus.SetLevel(logrus.DebugLevel)
 		}
+		if c.GlobalIsSet("cert-path") {
+			logrus.Warn("'--cert-path' is deprecated, please set this on the specific subcommand")
+		}
+		if c.GlobalIsSet("tls-verify") {
+			logrus.Warn("'--tls-verify' is deprecated, please set this on the specific subcommand")
```
mtrmac: I am tempted to say that these cases should be a fatal error, same as any other unknown flag.

mtrmac: … assuming that we would silently ignore them; as long as we do use the values, we should only warn of course.

cmd/skopeo/utils.go
mtrmac: If `--cert-path` is not interoperable with the original flag, I don’t really see the point in introducing a new version which does something else but warns.

Looking at various pastebins I’ve seen so far, `--tls-verify=false` is used pretty frequently and keeping it working would be nice.

cmd/skopeo/utils.go
```diff
@@ -9,12 +9,20 @@ import (
 	"github.com/urfave/cli"
 )
 
-func contextFromGlobalOptions(c *cli.Context, credsFlag string) (*types.SystemContext, error) {
+func contextFromGlobalOptions(c *cli.Context, credsFlag, certPathFlag, tlsVerifyFlag string) (*types.SystemContext, error) {
```
mtrmac: This is unwieldy… can this instead be a `flagPrefix string` (`""`, "src-"`, "dest-"`)?

cmd/skopeo/delete.go
```diff
@@ -39,5 +39,14 @@ var deleteCmd = cli.Command{
...
 		},
+		cli.StringFlag{
+			Name:  "cert-path",
+			Value: "",
+			Usage: "use certificates at `PATH` (*.crt, *.cert, *.key) to connect to the registry",
+		},
+		cli.BoolTFlag{
+			Name:  "tls-verify",
+			Usage: "require HTTPS and verify certificates when talking to docker registries (defaults to true)",
+		},
```
mtrmac: Repeating this all over… could we have a `credsFlagList(prefix string, kind string) []cli.Flag)` in `utils.go` containing the canonical version of these flags, and the various subcommands just concatenating that with their own private flags?

(`kind` would be something like `""`/`"source"`/"destination"` for use in the `Usage:` fields.)

That way both the flag definitions and uses would be in a single place in `utils.go`.

cmd/skopeo/utils.go
runcom: > > Looking at various pastebins I’ve seen so far, --tls-verify=false is used pretty frequently and keeping it working would be nice.

that means that if you use `skopeo copy` with just `--tls-verify` it'll be applied to both src and dest, is that ok with you?

runcom:
> I assume you are certain enough that no existing user will be broken by this?

I assume so - and really, we're fixing something as copy should not send certs to both src and dest (I can keep compat with inspect and layers though)

cmd/skopeo/main.go
runcom: hidden doesn't show anything right

cmd/skopeo/utils.go
runcom: sure

cmd/skopeo/delete.go
runcom: fine

runcom:
@mtrmac right now the main skopeo user is `atomic` - and it's not using `--cert-path` at all from what I can see - sure `--tls-verify` must be kept for backward compatibility there (assuming you're ok that a global `--tls-verify=true` as is right now, will switch off tls verification for src *AND* dest, as is right now :))

cmd/skopeo/main.go
rhatdan: I would go with Fatal.  Allowing the user to specify them and expecting something to happen, is not a good solution.  Especially for a security one like tls-verify.

cmd/skopeo/utils.go
mtrmac: > that means that if you use `skopeo copy` with just `--tls-verify` it'll be applied to both src and dest,

That’s ugly, sure, but if this PR is already adding a warning about the flag being deprecated, it seems acceptable to me.

cmd/skopeo/main.go
runcom: I believe we can just Warn with `--tls-verify` because we still need to support that - I'll totally drop `--cert-path` though since hardly anyone is using that and there's no point in keeping backward compatibility.

cmd/skopeo/layers.go
runcom: moved to a separate commit.

runcom:
@mtrmac apart from https://github.com/projectatomic/skopeo/pull/264#discussion_r90896815 (which I'm not totally sure of) I fixed your comments, PTAL again

mtrmac:
> apart from https://github.com/projectatomic/skopeo/pull/264#discussion_r90896815 (which I'm not totally sure of)

Yeah, splicing strings is ugly...

LGTM, make sure to drop the non-`master` c/i branch in `vendor.sh`

PR #269, opened 2016-12-08
Makefile: run "go" with $(BUILDTAGS)

nalind:
Run the "go" command with the $(BUILDTAGS) makefile variable passed in as build tags.  We don't currently set it, but we'll need to eventually, and adding it now does no harm.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,10 +52,10 @@ binary-static: cmd/skopeo
 
 # Build w/o using Docker containers
 binary-local:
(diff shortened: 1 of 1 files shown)


runcom:
:+1: just waiting for tests and I'll merge it

nalind:
Hmm, maybe BUILDFLAGS instead of BUILDTAGS?

runcom:
> Hmm, maybe BUILDFLAGS instead of BUILDTAGS?

I don't mind, BUILDTAGS is fine since go uses `-tags` :)

nalind:
Works for me.  It's not like any of this is set in stone.  Thanks!

PR #273, opened 2016-12-09
Remove an obsolete FIXME

mtrmac:
With https://github.com/containers/image/pull/183 , this no longer applies.


runcom:
lgtm will merge after tests

PR #279, opened 2016-12-28
[donotmerge] integration: use ::compress option for compression tests

cyphar:
Due to pending changes in containers/image, it is necessary that we
explicitly make compression enabled in the tests which require it. The
old behaviour was based on broken semantics (that compression would
never be un-done if you're changing compression policies).

### DIFF
--- a/integration/copy_test.go
+++ b/integration/copy_test.go
@@ -111,11 +111,11 @@ func (s *CopySuite) TestCopySimpleAtomicRegistry(c *check.C) {
 
 	// FIXME: It would be nice to use one of the local Docker registries instead of neeeding an Internet connection.
 	// "pull": docker: → dir:
-	assertSkopeoSucceeds(c, "", "copy", "docker://estesp/busybox:amd64", "dir:"+dir1)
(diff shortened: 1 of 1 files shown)


cyphar:
DO NOT MERGE. This depends on containers/image#193.

integration/copy_test.go
```diff
@@ -297,7 +297,7 @@ func (s *CopySuite) TestCopyCompression(c *check.C) {
 		c.Assert(err, check.IsNil)
 
 		assertSkopeoSucceeds(c, "", "--tls-verify=false", "copy", "dir:fixtures/"+t.fixture, t.remote)
-		assertSkopeoSucceeds(c, "", "--tls-verify=false", "copy", t.remote, "dir:"+dir)
+		assertSkopeoSucceeds(c, "", "--tls-verify=false", "copy", t.remote, "dir:"+dir+"::compress")
```
mtrmac: See the discussion in https://github.com/containers/image/issues/192 and https://github.com/containers/image/pull/193 — this significantly weakens the test, now we don't really know whether the compression has happened on push or pull.

rhatdan:

cyphar:
I think we decided it wasn't a good idea, but that means I'll have to rethink #280.

PR #280, opened 2016-12-28
integration: add test for OCI <-> docker-archive translation

cyphar:
There have been a long string of bugs discovered while testing this
functionality, and it's quite important the we ensure that this doesn't
break.

### DIFF
--- a/integration/copy_test.go
+++ b/integration/copy_test.go
@@ -151,6 +151,43 @@ func (s *CopySuite) TestCopySimple(c *check.C) {
 	c.Assert(err, check.IsNil)
 }
 
+// Make sure that docker-archive is identical to docker.
+func (s *CopySuite) TestCopyArchive(c *check.C) {
+	ar1, err := ioutil.TempDir("", "ar-1")
+	c.Assert(err, check.IsNil)
+	defer os.RemoveAll(ar1)
+	ar2, err := ioutil.TempDir("", "ar-2")
+	c.Assert(err, check.IsNil)
+	defer os.RemoveAll(ar2)
+	dir1, err := ioutil.TempDir("", "copy-1")
+	c.Assert(err, check.IsNil)
+	defer os.RemoveAll(dir1)
+	dir2, err := ioutil.TempDir("", "copy-2")
+	c.Assert(err, check.IsNil)
+	defer os.RemoveAll(dir2)
+
+	// FIXME: It would be nice to use one of the local Docker registries instead of neeeding an Internet connection.
+	// "pull": docker: → docker-archive:
+	assertSkopeoSucceeds(c, "", "copy", "docker://busybox", "docker-archive:"+ar1+"/archive.tar:busybox")
+	// "copy": docker-archive: → dir:
+	assertSkopeoSucceeds(c, "", "copy", "docker-archive:"+ar1+"/archive.tar", "dir:"+dir1)
+	// "pull": docker: → dir:
+	assertSkopeoSucceeds(c, "", "copy", "docker://busybox", "dir:"+dir2)
+	out := combinedOutputOfCommand(c, "diff", "-urN", dir1, dir2)
+	c.Assert(out, check.Equals, "")
+
+	// docker-archive -> OCI image layout
+	ociDest := "busybox-latest"
+	defer os.RemoveAll(ociDest)
+	assertSkopeoSucceeds(c, "", "copy", "docker-archive:"+ar1+"/archive.tar", "oci:"+ociDest)
+	_, err = os.Stat(ociDest)
+	c.Assert(err, check.IsNil)
+
+	// OCI image layout -> docker-archive
+	assertSkopeoSucceeds(c, "", "copy", "oci:"+ociDest, "docker-archive:"+ar2+"/archive.tar")
+	c.Assert(err, check.IsNil)
+}
+
 // Streaming (skopeo copy)
 func (s *CopySuite) TestCopyStreaming(c *check.C) {
 	dir1, err := ioutil.TempDir("", "streaming-1")


cyphar:
DO NOT MERGE. This depends on containers/image#193 (thus also #279) and containers/image#143.

integration/copy_test.go
```diff
@@ -151,6 +151,43 @@ func (s *CopySuite) TestCopySimple(c *check.C) {
...
+	// "copy": docker-archive: → dir:
+	assertSkopeoSucceeds(c, "", "copy", "docker-archive:"+ar1+"/archive.tar", "dir:"+dir1)
+	// "pull": docker: → dir:
+	assertSkopeoSucceeds(c, "", "copy", "docker://busybox", "dir:"+dir2)
+	out := combinedOutputOfCommand(c, "diff", "-urN", dir1, dir2)
+	c.Assert(out, check.Equals, "")
+
+	// docker-archive -> OCI image layout
+	ociDest := "busybox-latest"
+	defer os.RemoveAll(ociDest)
```
mtrmac: Why is this one not a `TempDir` as above? (Perhaps check how `TestCopyDirSignature` does this, with so many directories.)

integration/copy_test.go
```diff
@@ -151,6 +151,43 @@ func (s *CopySuite) TestCopySimple(c *check.C) {
 	c.Assert(err, check.IsNil)
 }
 
+// Make sure that docker-archive is identical to docker.
```
mtrmac: This is not really describing the test… what would “a local file being identical to a remote repository” (or is it “to a daemon”) mean? And docker does nothing with OCI…

I guess, a test is a test is a test, and it _is_ useful even if not explained, but knowing what is being tested could make it easier in the future to figure out whether a failure is a bug in the test or in the software being tested.

integration/copy_test.go
```diff
@@ -151,6 +151,43 @@ func (s *CopySuite) TestCopySimple(c *check.C) {
...
+	defer os.RemoveAll(dir1)
+	dir2, err := ioutil.TempDir("", "copy-2")
+	c.Assert(err, check.IsNil)
+	defer os.RemoveAll(dir2)
+
+	// FIXME: It would be nice to use one of the local Docker registries instead of neeeding an Internet connection.
+	// "pull": docker: → docker-archive:
+	assertSkopeoSucceeds(c, "", "copy", "docker://busybox", "docker-archive:"+ar1+"/archive.tar:busybox")
+	// "copy": docker-archive: → dir:
+	assertSkopeoSucceeds(c, "", "copy", "docker-archive:"+ar1+"/archive.tar", "dir:"+dir1)
```
mtrmac: Is it essential that this uncompressess the layers (per #279 and https://github.com/containers/image/pull/193 ), or does it not matter here?

integration/copy_test.go
```diff
@@ -151,6 +151,43 @@ func (s *CopySuite) TestCopySimple(c *check.C) {
...
+	// "pull": docker: → dir:
+	assertSkopeoSucceeds(c, "", "copy", "docker://busybox", "dir:"+dir2)
+	out := combinedOutputOfCommand(c, "diff", "-urN", dir1, dir2)
```
mtrmac: Sure, layers should match (either uncompressed or unmodified); but why should we expect the manifest, which has been thrown away in the copy to `docker-archive:ar1`, and generated from scratch in the copy from it, to match (and keep matching) byte-for-byte the manifest stored in the remote `docker://busybox` repo?

integration/copy_test.go
```diff
@@ -151,6 +151,43 @@ func (s *CopySuite) TestCopySimple(c *check.C) {
...
+	// docker-archive -> OCI image layout
+	ociDest := "busybox-latest"
+	defer os.RemoveAll(ociDest)
+	assertSkopeoSucceeds(c, "", "copy", "docker-archive:"+ar1+"/archive.tar", "oci:"+ociDest)
+	_, err = os.Stat(ociDest)
+	c.Assert(err, check.IsNil)
+
+	// OCI image layout -> docker-archive
+	assertSkopeoSucceeds(c, "", "copy", "oci:"+ociDest, "docker-archive:"+ar2+"/archive.tar")
+	c.Assert(err, check.IsNil)
```
mtrmac: Are these two copies intended to be just a minimal smoke test, or is there a more specific kind of failure this is testing for?

cyphar:
#305 wouldn't have happened if we had these tests in place... *sigh*

integration/copy_test.go
cyphar: There were a variety of failures that would happen if you tried to test going from OCI to Docker, back to OCI then back to Docker. This is a smoke test to ensure that at the very least we don't have asymmetric issues with one side of the conversion.

cyphar: For example, the most recent series of issues we've seen would've been caught by this smoke test.

cyphar:
I'm going to make a separate PR for doing regular `OCI <-> Docker`.

mtrmac:
> #305 wouldn't have happened if we had these tests in place... sigh

FWIW, note that https://github.com/containers/image/pull/148 is still pending.

cyphar:
Yeah, which is why I've split out #306 so we get /some/ tests in place.

integration/copy_test.go
cyphar: This does matter for the test below. However, since the original idea behind containers/image#193 was NACK'd, the current system will have the blobs decompressed in the `docker-archive:` case but I don't think it'll happen in the `docker://` case.

But first lets get the other stuff merged and we can figure out how to make a nice integration test once the code works.

integration/copy_test.go
cyphar: I'll split this test to look like #306, where we have a test that checks something that _must_ be invariant but also an extra test that checks the things-that-should-be-the-same, with a comment explaining the situation.

integration/copy_test.go
cyphar: Yes, it should be a `TempDir`.

integration/copy_test.go
cyphar: However, I will add some extra round-trip style tests (like #306) once the two important PRs are merged. I'm not going to waste time writing lots of detailed tests until we all agree on the implementations (especially semantics like compression, which massively change what invariants we can test).

rhatdan:
Is this pull request still being worked on ?

cyphar:
I haven't touched this in a while, but we still should include these tests. Also, there are some other issues that I've found in `skopeo` (such as the tags being lost bug) that we need to make sure don't get regressed on.

rhatdan:
@cyphar This needs an update or it needs to be closed.

cyphar:
/me will rebase (and probably) rewrite this.

rhatdan:
@cyphar Nothing has happened to this PR in 6 months, should I close?

rhatdan:
@cyphar I am closing this since nothing has happened in over a year. Reopen if you are still interested in workin on it.

PR #135, opened 2016-06-30
DO NOT MERGE: No hack/vendor.sh

runcom:
Fix https://github.com/projectatomic/skopeo/issues/131

Based on https://github.com/projectatomic/skopeo/pull/133 and missing #136, I'll rebase as needed to finally merge this.

This is using [https://github.com/Masterminds/glide](glide) and [https://github.com/sgotti/glide-vc](glide-vc) to do the vendoring w/o bothering ourself. Please read more on those repos but it's actually awesome and flexible (like updating just 1 dep instead of waiting for all of them to be merged, etc)

PR #284, opened 2017-01-09
switch to vndr

runcom:
vndr is almost exactly the same as our old good hack/vendor.sh. Except
it's cleaner and it allows to re-vendor just one dependency if needed
(which we do a lot for containers/image).

close #135


runcom:
@mtrmac PTAL - from now on, there would be no need to run hack/vendor.sh to vendor _every_ dependency. Instead `vndr github.com/containers/image` should just vendor c/image instantly (since it's tiny).

hack/vendor.sh
```diff
@@ -1,46 +1,15 @@
...
-rm -rf vendor/
-source 'hack/.vendor-helpers.sh'
+# This file is just wrapper around vndr (github.com/LK4D4/vndr) tool.
+# For updating dependencies you should change `vendor.conf` file in root of the
+# project. Please refer to https://github.com/LK4D4/vndr/blob/master/README.md for
+# vndr usage.
 
-clone git github.com/urfave/cli v1.17.0
-clone git github.com/containers/image master
-clone git github.com/opencontainers/go-digest master
```
mtrmac: This is the only item which has not been replicated in the new `vendor.conf`, and it does seem necessary:

`rm -rf /vendor/*/; vndr` does not recreate `vendor/github.com/opencontainers/go-digest`.

runcom: great catch :+1:

runcom:
> Also, this supersedes #135, doesn’t it?

it does, yes

hack/vendor.sh
runcom: fixed

mtrmac:
:+1:

PR #285, opened 2017-01-12
Makefile: use go-md2man in the skopeobuildimage

AkihiroSuda:
Previously, the doc cannot be built if go-md2man is not installed on the host.

So I suggest using go-md2man installed in the skopeobuildimage (https://github.com/projectatomic/skopeo/pull/243/files#diff-d92c0cc6cc3b84dda3c92547bcf56981R3)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,6 @@ CONTAINERSSYSCONFIGDIR=${DESTDIR}/etc/containers
 REGISTRIESDDIR=${CONTAINERSSYSCONFIGDIR}/registries.d
 SIGSTOREDIR=${DESTDIR}/var/lib/atomic/sigstore
 BASHINSTALLDIR=${PREFIX}/share/bash-completion/completions
-GO_MD2MAN ?= /usr/bin/go-md2man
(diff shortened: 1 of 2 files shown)


runcom:
mm I don't like this :smile: it's like providing a container to execute `make` because host doesn't have it. I think it's fine way it is right now, sorry.

AkihiroSuda:
OK, thank you

mtrmac:
> it's like providing a container to execute make because host doesn't have it. 

That’s not _obviously_ nonsensical to me, if we wanted the build to be reproducible that way; e.g. we could have `make all-local` and a `make all` which runs `make all-local` in a container (and then we would drop all of the other `-local` targets, and have all targets but the top level in a single non-container variant; and we would only start one container per `make`, not one per individual created file).

But I suppose the real issue with this is that, we (definitely me, and IIRC you as well) almost always use the non-container build targets, so the in-container ones are minimally maintained. Given that, adding more minimally-maintained build paths would mostly make things more fragile and less consistent.

PR #286, opened 2017-01-12
Makefile: /usr/bin/go-md2man -> go-md2man

AkihiroSuda:
When user installed go-md2man via `go get`, it is unlikely to be installed as `/usr/bin/go-md2man`.

So I suggest changing `/usr/bin/go-md2man` to just `go-md2man`.


mtrmac:
Thanks! Makes sense to me, but I’ll defer to @runcom .

runcom:
Yes, this makes sense to me as well.

PR #289, opened 2017-01-18
Fix Skopeo tests with latest containers/image and kubernetes

erikh:
@runc0m here you go, this is based off erikh/image@kube-fix.


runcom:
Let's see if tests pass here :) @mtrmac

vendor.conf
```diff
@@ -27,6 +27,7 @@ github.com/mtrmac/gpgme master
 # openshift/origin' k8s dependencies as of OpenShift v1.1.5
 github.com/golang/glog 44145f04b68cf362d9c4df2182967c2275eaefed
 k8s.io/kubernetes 4a3f9c5b19c7ff804cbc1bf37a15c044ca5d2353 https://github.com/openshift/kubernetes
+k8s.io/apimachinery master
```
runcom: @mtrmac do we need to use this from openshift?

erikh:
success!

vendor.conf
mtrmac: Openshift doesn't have a forked repo of `k8s.io/apimachinery` right now, so we don’t really have a choice unless we wanted to force all developers of `containers/image` to maintain a `$GOPATH` with `k8s.io/kubernetes` checked out from the `openshift/kubernetes` fork.

Anyway, the …`pkg/util/errors` packages are identical between `k8s.io/apimachinery` and `master` of `openshift/kubernetes` (though we had an older fork in here). The …`pkg/util/net` packages differ a bit, but that seems trivial enough (HTTP 2 enabling/disabling mechanism).

mtrmac:
LGTM.

After https://github.com/containers/image/pull/214 is merged, please restore the `github.com/containers/image master` line, rerun `vndr` and update this PR.

runcom:
@erikh please go ahead and restore `github.com/containers/image master` line, re-run `vndr` an update this PR (as @mtrmac said). Thanks!

mtrmac:
Merged in #291. Thanks for the fix!

PR #298, opened 2017-01-30
Vendor in non-gpgme support, and update tests and build machinery

mtrmac:
This vendors in https://github.com/containers/image/pull/207 , and:

 - Fixes the build scripts to propagate `BUILDTAGS` throughout the various subroutines and helpers
 - Modifies the signing tests to be mark them as skipped when building with the `openpgp` backend (which does not support signing). It would be nice to split the tests into a signing/verification part, and continuing to test verification (of signatures in fixtures), but this works for now. The core verification code hopefully has sufficient coverage in `containers/image/signature`.


runcom:
:+1:

PR #306, opened 2017-02-15
integration: add OCI <-> Docker roundtrip tests

cyphar:

### DIFF
--- a/integration/copy_test.go
+++ b/integration/copy_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
+	"log"
 	"net/http"
 	"net/http/httptest"
 	"os"
@@ -13,6 +14,7 @@ import (
 	"github.com/containers/image/manifest"
 	"github.com/go-check/check"
 	"github.com/opencontainers/go-digest"
+	"github.com/opencontainers/image-tools/image"
 )
 
 func init() {
@@ -216,7 +218,14 @@ func (s *CopySuite) TestCopyOCIRoundTrip(c *check.C) {
 	out := combinedOutputOfCommand(c, "diff", "-urN", oci1+"/blobs", oci2+"/blobs")
 	c.Assert(out, check.Equals, "")
 
+	// For some silly reason we pass a logger to the OCI library here...
+	logger := log.New(os.Stderr, "", 0)
+
 	// TODO: Verify using the upstream OCI image validator.
+	err = image.ValidateLayout(oci1, nil, logger)
+	c.Assert(err, check.IsNil)
+	err = image.ValidateLayout(oci2, nil, logger)
+	c.Assert(err, check.IsNil)
 
 	// Now verify that everything is identical. Currently this is true, but
 	// because we recompute the manifests on-the-fly this doesn't necessarily
--- a/integration/copy_test.go
+++ b/integration/copy_test.go
@@ -186,6 +186,46 @@ func (s *CopySuite) TestCopyStreaming(c *check.C) {
 	// FIXME: Also check pushing to docker://
 }
 
+// OCI round-trip testing. It's very important to make sure that OCI <-> Docker
+// conversion works (while skopeo handles many things, one of the most obvious
+// benefits of a tool like skopeo is that you can use OCI tooling to create an
+// image and then as the final step convert the image to a non-standard format
+// like Docker). But this only works if we _test_ it.
+func (s *CopySuite) TestCopyOCIRoundTrip(c *check.C) {
+	const ourRegistry = "docker://" + v2DockerRegistryURL + "/"
+
+	oci1, err := ioutil.TempDir("", "oci-1")
+	c.Assert(err, check.IsNil)
+	defer os.RemoveAll(oci1)
+	oci2, err := ioutil.TempDir("", "oci-2")
+	c.Assert(err, check.IsNil)
+	defer os.RemoveAll(oci2)
+
+	// Docker -> OCI
+	assertSkopeoSucceeds(c, "", "--tls-verify=false", "--debug", "copy", "docker://busybox", "oci:"+oci1+":latest")
+	// OCI -> Docker
+	assertSkopeoSucceeds(c, "", "--tls-verify=false", "--debug", "copy", "oci:"+oci1+":latest", ourRegistry+"original/busybox:oci_copy")
+	// Docker -> OCI
+	assertSkopeoSucceeds(c, "", "--tls-verify=false", "--debug", "copy", ourRegistry+"original/busybox:oci_copy", "oci:"+oci2+":latest")
+	// OCI -> Docker
+	assertSkopeoSucceeds(c, "", "--tls-verify=false", "--debug", "copy", "oci:"+oci2+":latest", ourRegistry+"original/busybox:oci_copy2")
+
+	// TODO: Add some more tags to output to and check those work properly.
+
+	// First, make sure the OCI blobs are the same. This should _always_ be true.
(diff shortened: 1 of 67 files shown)


runcom:
:+1: 

awesome, thanks @cyphar

cyphar:
Wait, wait. I pushed this so I can test it because Docker takes forever here. :wink:

runcom:
> Wait, wait. I pushed this so I can test it because Docker takes forever here.

yeah, it was an OK from the on the intention :smile:

cyphar:
Grrr. Integration tests are failing because we don't have `docker-archive`. Maybe if I implement `docker://` -> `oci:` conversion this will become easier?

runcom:
> Maybe if I implement docker:// -> oci: conversion this will become easier?

don't we have that already? :confused:

cyphar:
Actually, we do... I'm confused why I'm getting this error:

```
time="2017-02-15T15:26:41Z" level=fatal msg="Error creating an updated image manifest: Conversion of image manifest from application/vnd.docker.distribution.manifest.v1+prettyjws to application/vnd.oci.image.manifest.v1+json is not implemented" 
```

What is `ourRegistry` running and why the hell is it broken?

integration/copy_test.go
```diff
@@ -186,6 +186,29 @@ func (s *CopySuite) TestCopyStreaming(c *check.C) {
...
+
+	oci1, err := ioutil.TempDir("", "oci-1")
+	c.Assert(err, check.IsNil)
+	defer os.RemoveAll(oci1)
+	oci2, err := ioutil.TempDir("", "oci-2")
+	c.Assert(err, check.IsNil)
+	defer os.RemoveAll(oci2)
+
+	// Docker -> OCI
+	assertSkopeoSucceeds(c, "", "--tls-verify=false", "--debug", "--remove-signatures", "copy", "docker://estesp/busybox:amd64", "oci:"+oci1+":latest")
```
runcom: just use `busybox:latest` - the `amd64` tag is v2s1 only...

runcom:
@cyphar you have some flags-related issues in the tests and also, please take into account my comment https://github.com/projectatomic/skopeo/pull/306#discussion_r101304344 about not using that busybox tag

cyphar:
@runcom 

```
time="2017-02-15T16:40:37Z" level=fatal msg="can not copy docker://estesp/busybox:latest: manifest contains multiple images" 
```

:open_mouth:

runcom:
@cyphar I'm sorry I meant the official `busybox` image not `library/busybox`

integration/copy_test.go
```diff
@@ -186,6 +186,31 @@ func (s *CopySuite) TestCopyStreaming(c *check.C) {
...
+	defer os.RemoveAll(oci2)
+
+	// Docker -> OCI
+	assertSkopeoSucceeds(c, "", "--tls-verify=false", "--debug", "copy", "docker://busybox", "oci:"+oci1+":latest")
+	// OCI -> Docker
+	assertSkopeoSucceeds(c, "", "--tls-verify=false", "--debug", "copy", "oci:"+oci1+":latest", ourRegistry+"busybox:oci_copy")
+	// Docker -> OCI
+	assertSkopeoSucceeds(c, "", "--tls-verify=false", "--debug", "copy", ourRegistry+"busybox:oci_copy", "oci:"+oci2+":another-tag")
+	// OCI -> Docker
+	assertSkopeoSucceeds(c, "", "--tls-verify=false", "--debug", "copy", "oci:"+oci2+":another-tag", ourRegistry+"busybox:oci_copy2")
```
mtrmac: Is there an obvious kind of comparison that can be done in here to make sure that we do not only don’t report any error, but that we actually copy and convert the images succesfully?

Perhaps check that `oci1`’s and `busybix_oci_copy2` list of layer digests is exactly the same? That the config is identical between `oci` and `oci2`, between the two `ourRegistry/busybox` copies?

cyphar: Yeah, I was thinking of doing that. First I'm just trying to get this to pass (I'm getting weird registry errors). Still working on it. And I can't run it locally because Docker appears to be broken on my machine.

mtrmac:
:+1: assuming the tests pass, it is not immediately obvious to me why they are / have been failing.

runcom:
There seems to be another issue

cyphar:
After running this locally, it looks like this is actually another issue with manifest conversion:

```
ERRO[0129] response completed with error                 err.code=manifest unknown err.detail=errors verifying manifest: unrecognized manifest content type  err.message=manifest unknown go.version=go1.7.5 http.request.host=localhost:5555 http.request.id=3c0d5fa4-9df2-4d45-8
d85-a2e81d3a69fe http.request.method=GET http.request.remoteaddr=127.0.0.1:58722 http.request.uri=/v2/busybox/manifests/latest http.request.useragent=Go-http-client/1.1 http.response.contenttype=application/json; charset=utf-8 http.response.duration=1.544431ms http.response
.status=404 http.response.written=84 instance.id=2334cfb3-48f9-4a5d-8692-0cbbc26c1cf8 vars.name=busybox vars.reference=latest version=v2.1.0+unknown
127.0.0.1 - - [16/Feb/2017:09:42:16 +0000] "GET /v2/busybox/manifests/latest HTTP/1.1" 404 84 "" "Go-http-client/1.1"

```

Which basically ends up boiling down to "we didn't change the manifest type when converting to Docker distribution from OCI":

```
%  cat /tmp/registry/docker/registry/v2/blobs/sha256/56/560799532c865dc38f12595b8a7eb8b602c9fcb608df90abe89499e6a0135483/data
{"schemaVersion":2,"config":{"mediaType":"application/vnd.oci.image.config.v1+json","size":575,"digest":"sha256:34166908e15c5badddab5191bca8bd0e4f429aeeecfba5ac37b387d9bb57b65b"},"layers":[{"mediaType":"application/vnd.oci.image.layer.v1.tar+gzip","size":677628,"digest":"sha256:4b0bc1c4050b03c95ef2a8e36e25feac42fd31283e8c30b3ee5df6b043155d3c"}]}
```

cyphar:
So the plus side is that this test is already finding bugs. The downside is that this is a bug in the first place.

cyphar:
Alright, this is actually caused by the same issue as containers/image#237. So it has been fixed in `containers/image`.

runcom:
> Alright, this is actually caused by the same issue as containers/image#237. So it has been fixed in containers/image.

can you revendor as part of this PR?

cyphar:
@runcom Doing it as we speak. It required re-vendoring other things. Oh, and I'm also adding `image-tools` validation so we can be even more sure about the validity of the output.

runcom:
@cyphar you may need to revendor the image-spec as well https://travis-ci.org/projectatomic/skopeo/builds/202194334#L1698

cyphar:
@runcom Trust me, it's much more than that. Wait for the commit message, I've had to revendor at least 7 projects.

runcom:
> @runcom Trust me, it's much more than that. Wait for the commit message, I've had to revendor at least 7 projects.

I see, ping me when done pls, I want this merged asap

cyphar:
@runcom There, done. PTAL.

cyphar:
The "warnings" in the log are because opencontainers/image-spec#548 (or opencontainers/image-spec#560) hasn't been merged yet. But the actual validation _is_ working as expected, and this is similar validation to the kind I'm doing in `umoci` (though in `umoci` it's much more extreme).

cyphar:
I will be adding **many** more tests once `docker-archive` is merged. This is just a stop-gap to guarantee things don't break in the meantime.

cyphar:
:heart:

Issue #304, opened 2017-02-07
Stop vendoring master!


cyphar:
Inside `vendor.conf` you have several projects that you are vendoring master from. While that might work in principle, I'm currently trying to backport `containers/image` patches and it's quite hard to figure out what version of `containers/image` is being used.

Please just stick to commit hashes!

runcom:
let's do that, yeah

cyphar:
:smile_cat:

mtrmac:
I don’t care strongly either way, I can see that it would be useful. My only concern is that the `vendor.conf` should, as much as possible, clearly document whether we are _required_ to use precisely these commits (e.g. for the set of dependencies which we take from `docker/docker` and where we want to use exactly the same consistent versions) or whether we generally want to keep up with `master` and these commits are just a point-in-time snapshot which is perfectly fine to update.

cyphar:
How about we add comments to `vendor.conf` what was checked out? The current system of adding everything in a commit message /sounds/ fine but I've run into several cases where the committer didn't write what commit was checked out.

mtrmac:
> How about we add comments to `vendor.conf` what was checked out?

Are you suggesting
```
github.com/containers/image master # somecommitid
```
?

AFAICS the `vndr` output does not say which version was used, so I don’t know where the user would get that information in a race-free way. (Also, it’s not quite obvious that people who forget to write commit messages would not forget to update comments.) Manually checking each remote repo for tag mappings before running `vndr` would be way too much work, and eliminate about 50% of the value of an automated vendoring tool.

Other vendoring systems do understand the difference between the desired tag (which can move over time by itself) and the commit ID that was actually checked out (which is constant, and can be updated only explicitly); hopefully the golang ecosystem will mature.

cyphar:
@mtrmac 

Yes, that's what I was suggesting. But as you say, it doesn't really work -- which is why we should not allow `master` to be used. I was just trying to work around your wish the "master semantically means any version". I disagree with the entire concept and I think it's blatantly dangerous (if it doesn't matter what version, then why are you vendoring it -- if you have to vendor something then you should be specifying _what_ you versioned).

> Manually checking each remote repo for tag mappings before running vndr would be way too much work, and eliminate about 50% of the value of an automated vendoring tool.

You could also use a tag if you want. The point is so that someone who wants to actually fucking use the information in `vendor.conf` doesn't have to spelunk the `git` history.

cyphar:
[...]

Issue #240, opened 2016-10-28
Error writing manifest: can't create an OCI manifest from Docker V2 schema 1 manifest


dohnto:
Hello, I am trying to examine a docker image which was created with older docker (version 1.9.1). I pushed that image to `dohnto/busybox:test` using skopeo. In order to do that, i convert the image to OCI.

```
skopeo version 0.1.17-dev commit: 507d09876d0620d46de7c8c20eb9b58d2dffa482
```

So when I run

```
skopeo copy docker://docker.io/dohnto/busybox:test oci:tmp
```

I got:

```
Getting image source manifest
Getting image source signatures
Getting image source configuration
Uploading blob sha256:56bec22e355981d8ba0878c6c2f23b21f422f30ab0aba188b54f1ffeff59c190
 559.41 KB / ? [---------------------------=----------------------------------] 
Uploading blob sha256:a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4
 0 B / ? [--------------------------------------------------------------------=]
Uploading manifest to image destination
FATA[0007] Error writing manifest: can't create an OCI manifest from Docker V2 schema 1 manifest
```

How to solve this issue? Thank you.

runcom:
Sorry but schema1 isn't compatible with OCI images by definition. Only schema2. There are no plans to support this conversion atm also.

runcom:
,

mtrmac:
The schema2→OCI conversion in `oci:` should really be eventually replaced by an OCI manifest support in `containers/image/image`. Then it should be plausible enough to do a s1→OCI conversion by using a temporary s2 `memoryImage`.

mtrmac:
> The schema2→OCI conversion in `oci:` should really be eventually replaced by an OCI manifest support in `containers/image/image`.

This has happened now.

> Then it should be plausible enough to do a s1→OCI conversion by using a temporary s2 `memoryImage`.

Yes; we even have a `manifestSchema1.OCIConfig`, so we are quite close.

jonboulle:
between this and containers/image, I always look at the wrong repo when looking for existing issues...
in any case, this is what I'm working on in containers/image#380

rhatdan:
@dohnto @mtrmac @jonboulle @vrothberg Does this issue still exists or can we close this issue?

mtrmac:
https://github.com/containers/image/pull/380 has implemented this, e.g.

```
[...]

PR #312, opened 2017-02-21
vendor: update dependencies

erikh:
This updates all the dependencies, but in particular fixes a dep for go-digest so it and containers/image can coexist.


runcom:
Looks good to me @erikh

mtrmac:
I don’t think vendoring _any_ version in skopeo can fix the `vendor`-caused type mismatch breaking #240.

We do need to figure out how to keep containers/image and skopeo’s `vendor.conf` in sync (and how to make sure they don’t stagnate on years-old versions), but that’s not quite the same thing I guess.

erikh:
part of the big problem is that we're not using SHAs and instead tracking master. This causes trouble because we can't look back and see what master was (easily, at least) based on a revision history because it's been wiped out by `vndr`.

Another option here would be to eliminate the `vendor` directory before running skopeo tests; this might  eliminate some problems, but not all.

I think the better solution is -- like you said -- to keep vendor.conf in sync. I wonder if we could source it from a gist or something?

erikh:
Seems pretty ghetto but it'd work! :D

mtrmac:
> part of the big problem is that we're not using SHAs and instead tracking master. This causes trouble because we can't look back and see what master was (easily, at least) based on a revision history because it's been wiped out by `vndr`.

We already have #304 for the digests vs. master discussion, let’s not duplicate it here please.

> Another option here would be to eliminate the `vendor` directory before running skopeo tests; this might  eliminate some problems, but not all.

Like, test skopeo against whatever happens to be in $GOPATH? I am probably misunderstanding.

> I think the better solution is -- like you said -- to keep vendor.conf in sync. I wonder if we could source it from a gist or something?

Repeating myself from above, I don’t think that using the exact same version of `go-digest` fixes the mismatched type names between containers/image and skopeo. That can only be fixed by containers/image, when used in skopeo, not having its own vendored copy of go-digest, of _any_ version of go-digest.

erikh:
I can confirm this patch fixes it. Can you provide an alternative if you
are unhappy with this patch? Both skopeo and image are broken right now.

mtrmac:
> I can confirm this patch fixes it. 

To be clear, fixes what? I am referring to the recent failure in https://github.com/containers/image/pull/240 :

```
cmd/skopeo/inspect.go:86: cannot assign "github.com/projectatomic/skopeo/vendor/github.com/containers/image/vendor/github.com/opencontainers/go-digest".Digest to outputData.Digest (type "github.com/projectatomic/skopeo/vendor/github.com/opencontainers/go-digest".Digest) in multiple assignment
```
This is a simple `type Digest string` in both packages / versions, this is not an interface mismatch which can be solved by bumping either of the dependencies; the issue is that the `type` command has created a named type incompatible with any other named type, and both projects having independent vendoring causes them to create two different named types.

And I have posted the suggestion twice in https://github.com/containers/image/pull/240:
> When test-skopeo copies containers/image into skopeo’s `vendor` subdirectory, it must not copy containers/image/vendor in there as well.

erikh:
I think this is resolved; do we need to keep this one around?

mtrmac:
If this was intended to fix the `digest.Digest` type mismatches in #240, I think #240 has fixed it the right way and this is not necessary. Or was this intended to fix something else?

erikh:
Will close for now, we can always re-open if necessary.

PR #315, opened 2017-03-03
fix copy panic on http response with tls-verify true

runcom:
@mtrmac added a test also for https://github.com/containers/image/pull/246


integration/copy_test.go
```diff
@@ -463,3 +463,10 @@ func (s *SkopeoSuite) TestCopySrcAndDestWithAuth(c *check.C) {
 	assertSkopeoSucceeds(c, "", "--tls-verify=false", "copy", "--dest-creds=testuser:testpassword", "docker://busybox", fmt.Sprintf("docker://%s/busybox:latest", s.regV2WithAuth.url))
 	assertSkopeoSucceeds(c, "", "--tls-verify=false", "copy", "--src-creds=testuser:testpassword", "--dest-creds=testuser:testpassword", fmt.Sprintf("docker://%s/busybox:latest", s.regV2WithAuth.url), fmt.Sprintf("docker://%s/test:auth", s.regV2WithAuth.url))
 }
+
+func (s *CopySuite) TestCopyNoPanicOnHTTPResponseWOTLSVerifyFalse(c *check.C) {
+	const ourRegistry = "docker://" + v2DockerRegistryURL + "/"
+
+	assertSkopeoFails(c, ".*server gave HTTP response to HTTPS client.*",
+		"copy", ourRegistry+"foobar", "dir:test")
```
mtrmac: This is a tiny bit brittle because this can just as well on “directory `test` does not exist”. This does work fine enough that actually creating a temporary directory is unnecessary work, but an one-liner comment that this is known to assume that the source will be used before the destination would be worth considering. _NOT_ a blocker.

mtrmac:
Oh, and thanks for taking the care to write a test!

PR #316, opened 2017-03-04
Vendor mtrmac/image:change-sigstore-layout

mtrmac:
This demonstrates https://github.com/containers/image/pull/247 .  There is a `TestCopyDockerSigstore` which tests that the write staging / read HTTP paths for a sigstore are interoperable, but it does not prescribe a specific layout, so it does not need updating.


mtrmac:
@runcom Re-vendored from master, no other changes in this repo. PTAL.

runcom:
:+1:

PR #318, opened 2017-03-19
Fix wrong naming of cert-dir argument in inspect command

enoodle:
A wrong name of this parameter prevents it from being used for the context (and at all). This is also different than the docs.

### DIFF
--- a/integration/check_test.go
+++ b/integration/check_test.go
@@ -85,6 +85,13 @@ func (s *SkopeoSuite) TestNeedAuthToPrivateRegistryV2WithoutDockerCfg(c *check.C
 	assertSkopeoFails(c, wanted, "--tls-verify=false", "inspect", fmt.Sprintf("docker://%s/busybox:latest", s.regV2WithAuth.url))
 }
 
+func (s *SkopeoSuite) TestCertDirInsteadOfCertPath(c *check.C) {
+	wanted := ".*flag provided but not defined: -cert-path.*"
+	assertSkopeoFails(c, wanted, "--tls-verify=false", "inspect", fmt.Sprintf("docker://%s/busybox:latest", s.regV2WithAuth.url), "--cert-path=/")
+	wanted = ".*unauthorized: authentication required.*"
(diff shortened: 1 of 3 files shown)


runcom:
Great catch!!! Could you add a regression test (and rebase)?

enoodle:
@runcom I wasn't sure where to add the tests, PTAL

integration/check_test.go
```diff
@@ -85,6 +85,13 @@ func (s *SkopeoSuite) TestNeedAuthToPrivateRegistryV2WithoutDockerCfg(c *check.C
 	assertSkopeoFails(c, wanted, "--tls-verify=false", "inspect", fmt.Sprintf("docker://%s/busybox:latest", s.regV2WithAuth.url))
 }
 
+func (s *SkopeoSuite) TestCertDirInsteadOfCertPath(c *check.C) {
+	wanted := ".*flag provided but not defined: -cert-path.*"
+	assertSkopeoFails(c, wanted, "--tls-verify=false", "inspect", fmt.Sprintf("docker://%s/busybox:latest", s.regV2WithAuth.url), "--cert-path=/")
+	wanted = ".*unauthorized: authentication required.*"
+	assertSkopeoFails(c, wanted, "--tls-verify=false", "inspect", fmt.Sprintf("docker://%s/busybox:latest", s.regV2WithAuth.url), "--cert-dir=/etc/docker/certs.d/")
```
mtrmac: @runcom 

Hum, this really only tests that the option name is acceptable; with `--tls-verify=false` and a HTTP-only registry the certificates in `--cert-dir` do not really matter. So the test is almost misleading, I’m not sure that it is really better than no test to be honest.

Preferably we should fix the
> `FIXME: Set up TLS for the docker registry port instead of using "--tls-verify=false" all over the place.`
todos, and then any smoke test of `skopeo inspect` would naturaly be a test for `--cert-dir`.

OTOH I don’t think it is at all reasonable to block this PR on doing that work.

mtrmac:
Thanks! LGTM assuming tests pass, and whatever @runcom decides about testing this.

integration/check_test.go
runcom: > OTOH I don’t think it is at all reasonable to block this PR on doing that work.

yup, I'll try to address that in a follow up

runcom:
let's keep the todo and test that in a follow up - this was a clearly silly bug worth fixing anyway

PR #332, opened 2017-04-29
Run-time manifest type support (e.g. schema1/schema2) autodetection re-vendor + integration tests

mtrmac:
This vendors in https://github.com/containers/image/pull/264 and adds integration tests for it.

See the individual commit messages for more details.

Note that the added integration test currently tests OpenShift with `acceptschema2:{true,false}`; a test against upstream docker/distribution registries (with the s2 and s1 versions already built by `Dockerfile`) exists, but is disabled because it fails—not on the s2→s1 conversion, but on s1→s1 copies: the s1-only registry requires updating the tag value, which we currently don’t do.

Updating the tag in s1 uploads is actually ready in  https://github.com/containers/image/pull/88 , but it breaks other integration tests (which incorrectly assume that s1 uploads don’t change the manifest) fairly badly, so we will need to update the other tests to support https://github.com/containers/image/pull/88 , and then we can enable the disabled tests in this PR.

### DIFF
--- a/integration/copy_test.go
+++ b/integration/copy_test.go
@@ -21,12 +21,16 @@ func init() {
 	check.Suite(&CopySuite{})
 }
 
-const v2DockerRegistryURL = "localhost:5555" // Update also policy.json
+const (
+	v2DockerRegistryURL   = "localhost:5555" // Update also policy.json
+	v2s1DockerRegistryURL = "localhost:5556"
+)
 
 type CopySuite struct {
-	cluster  *openshiftCluster
-	registry *testRegistryV2
-	gpgHome  string
+	cluster    *openshiftCluster
+	registry   *testRegistryV2
+	s1Registry *testRegistryV2
+	gpgHome    string
 }
 
 func (s *CopySuite) SetUpSuite(c *check.C) {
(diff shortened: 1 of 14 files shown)


mtrmac:
@runcom Final version ready for review.

integration/copy_test.go
```diff
@@ -535,3 +544,58 @@ func (s *CopySuite) TestCopyNoPanicOnHTTPResponseWOTLSVerifyFalse(c *check.C) {
 	assertSkopeoFails(c, ".*server gave HTTP response to HTTPS client.*",
 		"copy", ourRegistry+"foobar", "dir:test")
 }
+
+func (s *CopySuite) TestCopySchemaConversion(c *check.C) {
+	// Test conversion / schema autodetection both for the OpenShift embedded registry…
+	s.testCopySchemaConversionRegistries(c, "docker://localhost:5005/myns/schema1", "docker://localhost:5006/myns/schema2")
+	// … and for various docker/distribution registry versions.
+	if false {
```
runcom: is there an issue to track this? otherwise I would just get rid of this...an issue is more visible

runcom:
just a question, LGTM otherwise

integration/copy_test.go
mtrmac: That's already fixed in the follow-up #337 (https://github.com/projectatomic/skopeo/pull/337/commits/e45df01d9377fa3477d524cdf151923efbbfc7ce )

PR #336, opened 2017-05-06
DO NOT MERGE: Re-vendor to allow using schema2 when pushing to OpenShift (blocked on #332)

mtrmac:
This vendors in https://github.com/containers/image/pull/266 . Similarly to that, it depends on, and includes, #332.

No new tests are added here, this PR exists only to demonstrate that the re-vendored containers/image passes with the tests added in #332.


mtrmac:
> No new tests are added here, this PR exists only to demonstrate that the re-vendored containers/image passes with the tests added in #332.

Dropping this to avoid an unnecessary review step. The re-vendor will be included in #337.

PR #337, opened 2017-05-06
Update name/tag embedded into schema1 manifests

### LINKED ISSUE #290: `skopeo cp` does not properly support Docker Registry v2-1
In the Docker Registry protocol v2-1, there is namespace and tag metadata in the manifest. These values need to be rewritten when `skopeo cp` is pushing the manifest to another repository.

I noticed this bug when someone pointed out that `skopeo cp` does not work with Quay.io

mtrmac:
This vendors in https://github.com/containers/image/pull/88 , updates tests to pass now that schema1 manifests are being modified by the upload, and enables several tests which were previously broken by our incorrect schema1 support.  Fixes #290 .

Similarly to https://github.com/containers/image/pull/88, depends on, and includes, #336. I will rebase as necessary.

See the individual commit messages for more details.

### DIFF
--- a/integration/copy_test.go
+++ b/integration/copy_test.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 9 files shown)


mtrmac:
@runcom PTAL as well, merging https://github.com/containers/image/pull/88 without this would break CI.

runcom:
seems ok to me, we'll see what tests say on CI

PR #340, opened 2017-05-10
Simplify infrastructure setup, and make registry timeouts less strict

mtrmac:
This moves various setup steps from `SetUpTest` to `SetUpSuite` so that we don’t start up processes over and over; that is _slightly_ less reliable, but more efficient.

Also increases the registry startup timeout from 0.5 to 5 seconds, hopefully avoiding some CI flakes.

Finally, completely stops starting some of the registries which we don’t use, with an `if false` to keep the existing code structure. (Alternatively we could drop that, and stop building the docker/registry code altogether.

### DIFF
--- a/integration/check_test.go
+++ b/integration/check_test.go
@@ -12,9 +12,6 @@ import (
 const (
 	privateRegistryURL0 = "127.0.0.1:5000"
 	privateRegistryURL1 = "127.0.0.1:5001"
-	privateRegistryURL2 = "127.0.0.1:5002"
-	privateRegistryURL3 = "127.0.0.1:5003"
-	privateRegistryURL4 = "127.0.0.1:5004"
 )
 
 func Test(t *testing.T) {
@@ -26,10 +23,7 @@ func init() {
 }
 
 type SkopeoSuite struct {
-	regV1         *testRegistryV1
 	regV2         *testRegistryV2
-	regV2Shema1   *testRegistryV2
-	regV1WithAuth *testRegistryV1 // does v1 support auth?
 	regV2WithAuth *testRegistryV2
 }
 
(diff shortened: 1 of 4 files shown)


mtrmac:
(This is not intended to go into today's release.)

runcom:
> This moves various setup steps from SetUpTest to SetUpSuite so that we don’t start up processes over and over; that is slightly less reliable, but more efficient.

:cry: can we compare how efficient is this? I don't really like leaving registries states behind when jumping to a new test (the docker test suite, far bigger than this, still setup registries in `SetUpTest` afaict)

> 
> Also increases the registry startup timeout from 0.5 to 5 seconds, hopefully avoiding some CI flakes.

:+1: 

> 
> Finally, completely stops starting some of the registries which we don’t use, with an if false to keep the existing code structure. (Alternatively we could drop that, and stop building the docker/registry code altogether.

let's drop for sure the V1 stuff as skopeo isn't intended to support that at all

integration/check_test.go
```diff
@@ -34,25 +34,21 @@ type SkopeoSuite struct {
...
-
-}
-
-func (s *SkopeoSuite) SetUpTest(c *check.C) {
 	_, err := exec.LookPath(skopeoBinary)
 	c.Assert(err, check.IsNil)
 
-	s.regV1 = setupRegistryV1At(c, privateRegistryURL0, false) // TODO:(runcom)
+	if false { // not used
+		s.regV1 = setupRegistryV1At(c, privateRegistryURL0, false) // TODO:(runcom)
```
runcom: drop we won't support V1

integration/check_test.go
```diff
@@ -34,25 +34,21 @@ type SkopeoSuite struct {
...
+	}
 	s.regV2 = setupRegistryV2At(c, privateRegistryURL1, false, false)
-	s.regV2Shema1 = setupRegistryV2At(c, privateRegistryURL2, false, true)
-	s.regV1WithAuth = setupRegistryV1At(c, privateRegistryURL3, true) // not used
+	if false { // not used
+		s.regV2Shema1 = setupRegistryV2At(c, privateRegistryURL2, false, true)
+		s.regV1WithAuth = setupRegistryV1At(c, privateRegistryURL3, true)
```
runcom: drop we won't support V1

integration/check_test.go
```diff
@@ -34,25 +34,21 @@ type SkopeoSuite struct {
...
 
-	s.regV1 = setupRegistryV1At(c, privateRegistryURL0, false) // TODO:(runcom)
+	if false { // not used
+		s.regV1 = setupRegistryV1At(c, privateRegistryURL0, false) // TODO:(runcom)
+	}
 	s.regV2 = setupRegistryV2At(c, privateRegistryURL1, false, false)
-	s.regV2Shema1 = setupRegistryV2At(c, privateRegistryURL2, false, true)
-	s.regV1WithAuth = setupRegistryV1At(c, privateRegistryURL3, true) // not used
+	if false { // not used
+		s.regV2Shema1 = setupRegistryV2At(c, privateRegistryURL2, false, true)
```
runcom: don't we test V2S1? do we test it with openshift?

mtrmac:
> > This moves various setup steps from SetUpTest to SetUpSuite so that we don’t start up processes over and over; that is slightly less reliable, but more efficient.

> :cry: can we compare how efficient is this? I don't really like leaving registries states behind when jumping to a new test (the docker test suite, far bigger than this, still setup registries in `SetUpTest` afaict)

Honestly it was less than the variation between test runs. I’ll drop that part. (FWIW `copy_tests.go`, where most of the tests actually are, does use `SetUpSuite` already. But then starting an OpenShift cluster takes much longer.)

integration/check_test.go
mtrmac: Yes, `CopySuite.TestCopySchemaConversion` tests both OpenShift and non-OpenShift registries, and relies on registries set up in `CopySuite.SetUpSuite`, not on this one.

runcom: Cool, let's drop it then altogether

mtrmac:
Updated:
* `SkopeoSuite` keeps setting up new registry servers for each test. (`CopySuite` keeps usingper-suite registries.)
* The unused registry servers are completely removed instead of `if false`.
* The `docker/registry` server is no longer being built in `Dockerfile`.

runcom:
Awesome, LGTM pending tests

PR #343, opened 2017-05-12
README.md updates

mtrmac:
This primarily restructures the “building” section, including some fixes. Also drops a few completed items from `README.md`:

- Add Fedora instructions for installing go-md2man
- Move documentation build instructions to the end, with a separate header
- Create “Building {in a container,without a container}” subsections
- Restructure the “Building without a container” version
- Start all command examples with `$`
- Drop finished work from TODO
- Add a warning about in-container builds

---

Note the last commit: e.g. on Fedora, `make build` creates an Ubuntu binary which does not even start (`./skopeo: error while loading shared libraries: libdevmapper.so.1.02.1: cannot open shared object file: No such file or directory`). I expect it’s even worse on macOS.

Perhaps we might want to move the “Building without a container” subsection in front of the “Building in a container” subsection?

Alternatively we might also document `make binary-static` (which works around the Fedora problem, but not the macOS problem—and does not actually work now because the container does not install static libraries), making the instructions even more complex?

For now, this is leaving the ordering and instructions in this respect as is.

---

@hferentschik @duglin FYI

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,10 +103,14 @@ you'll get an error. You can fix this by either logging in (via `docker login`) 
 
 Building
 -
-To build the `skopeo` binary you need at least Go 1.5 because it uses the latest `GO15VENDOREXPERIMENT` flag. Also, make sure to clone the repository in your `GOPATH` - otherwise compilation fails.
+To build the `skopeo` binary you need at least Go 1.5 because it uses the latest `GO15VENDOREXPERIMENT` flag.
+
+### Building without a container
+Make sure to clone the repository in your `GOPATH` - otherwise compilation fails.
+
 ```sh
 $ git clone https://github.com/projectatomic/skopeo $GOPATH/src/github.com/projectatomic/skopeo
-$ cd $GOPATH/src/github.com/projectatomic/skopeo && make all
+$ cd $GOPATH/src/github.com/projectatomic/skopeo && make binary-local
 ```
 
 To build localy on OSX:
@@ -118,6 +122,12 @@ $ make binary-local
 You may need to install additional development packages: `gpgme-devel` and `libassuan-devel`
 ```sh
 $ sudo dnf install gpgme-devel libassuan-devel btrfs-progs-devel device-mapper-devel
+```
+
+### Building in a container
+If your system has a `docker` command and can run Linux containers,
+```sh
+$ make binary # Or (make all) to also build documentation, see below.
 ```
 
 ### Building documentation
--- a/README.md
+++ b/README.md
@@ -103,11 +103,6 @@ you'll get an error. You can fix this by either logging in (via `docker login`) 
 
 Building
 -
-To build the manual you will need go-md2man.
-```sh
-Debian$ sudo apt-get install go-md2man
-Fedora$ sudo dnf install go-md2man
-```
 To build the `skopeo` binary you need at least Go 1.5 because it uses the latest `GO15VENDOREXPERIMENT` flag. Also, make sure to clone the repository in your `GOPATH` - otherwise compilation fails.
 ```sh
 $ git clone https://github.com/projectatomic/skopeo $GOPATH/src/github.com/projectatomic/skopeo
@@ -124,6 +119,18 @@ You may need to install additional development packages: `gpgme-devel` and `liba
 ```sh
 $ sudo dnf install gpgme-devel libassuan-devel btrfs-progs-devel device-mapper-devel
 ```
+
+### Building documentation
+To build the manual you will need go-md2man.
+```sh
+Debian$ sudo apt-get install go-md2man
+Fedora$ sudo dnf install go-md2man
+```
+Then
+```sh
+$ make docs
+```
+
 Installing
 -
 If you built from source:
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ $ sudo make install
 ```
 `skopeo` is also available from Fedora 23 (and later):
 ```sh
-sudo dnf install skopeo
+$ sudo dnf install skopeo
 ```
 TODO
 -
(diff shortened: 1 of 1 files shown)


runcom:
:+1: @hferentschik would you take a look at this?

README.md
```diff
@@ -103,26 +103,40 @@ you'll get an error. You can fix this by either logging in (via `docker login`)
...
 Building
 -
-To build the manual you will need go-md2man.
+To build the `skopeo` binary you need at least Go 1.5 because it uses the latest `GO15VENDOREXPERIMENT` flag.
+
+### Building in a container
+If your system has a `docker` command and can run Linux containers,
 ```sh
-$ sudo apt-get install go-md2man
+$ make binary # Or (make all) to also build documentation, see below.
```
hferentschik: The docs make more sense now, but when I actually do run `make binary` I get:

    ....
    Step 4 : WORKDIR /src/github.com/projectatomic/skopeo
     ---> Running in bf527540083c
     ---> f817788091bf
    Removing intermediate container bf527540083c
    Successfully built f817788091bf
    docker run --rm --security-opt label:disable -v $(pwd):/src/github.com/projectatomic/skopeo \
    		skopeobuildimage make binary-local  BUILDTAGS='btrfs_noversion libdm_no_deferred_remove     containers_image_openpgp'
    CGO_ENABLED=1 GOARCH=amd64 GOOS=darwin go build -ldflags "-X     main.gitCommit=07c798ff82f82a68571b03eb9be3469baf483d93" -gcflags "" -tags "btrfs_noversion     libdm_no_deferred_remove containers_image_openpgp" -o skopeo ./cmd/skopeo
    # net
    could not determine kind of name for C.AI_MASK
    Makefile:60: recipe for target 'binary-local' failed
    make: *** [binary-local] Error 2
    make: *** [binary] Error 2

README.md
```diff
@@ -103,26 +103,40 @@ you'll get an error. You can fix this by either logging in (via `docker login`)
...
 ```
-To build the `skopeo` binary you need at least Go 1.5 because it uses the latest `GO15VENDOREXPERIMENT` flag. Also, make sure to clone the repository in your `GOPATH` - otherwise compilation fails.  
+(Note that the created executable is a Linux executable, and depends on dynamic libraries which may only be available only in a container of a similar Linux distribution.)
```
hferentschik: So what is the long term vision for skopeo? Will it always be a Linux CLI tool? There is not intention to make it work on other OS types?

README.md
```diff
@@ -103,26 +103,40 @@ you'll get an error. You can fix this by either logging in (via `docker login`)
...
+
+### Building without a container
+Install the necessary dependencies:
 ```sh
-$ git clone https://github.com/projectatomic/skopeo $GOPATH/src/github.com/projectatomic/skopeo
-$ cd $GOPATH/src/github.com/projectatomic/skopeo && make all
+Fedora$ dnf install gpgme-devel libassuan-devel
+macOS$ brew install gpgme
```
hferentschik: Would it be worth to document how to compile without gpgme? containers/image has the _containers_image_openpgp_ build flag.

README.md
rhatdan: I think we would be open to pull requests that would allow it to run on other OS.  But most of the people working on it are primarily Linux

README.md
mtrmac: Yes, that flag should work as well. Do you think it is useful enough to list it in here as _another_ option, instead of just telling the users to install `gpgme`? I’m worried that the more options and variants are in the README, the more users are likely to just give up.

README.md
mtrmac: I’m afraid I have no idea; random googling suggests that this is a problem when cross-compiling from Linux to macOS.
> ```
> CGO_ENABLED=1 GOARCH=amd64 GOOS=darwin go build …
> ```

How did the `CGO_ENABLED=1 GOARCH=amd64 GOOS=darwin` get there? I can’t see anything like that in the `Makefile`. Is `go` doing some magic autodetection, or is that a local modification?

README.md
mtrmac: This paragraph only applies to the “building in a container” section; `make binary-local` should work just fine (though, as you can see, because we don’t have a working CI for macOS, I’m afraid it’s too easy to break it without noticing).

I’ve updated the “building in/without a container” sections to make the trade-offs clearer, and on balance it _is_ probably better to start with the “without a container” alternative (the build instructions are longer, but much more broadly useful), so I have moved it before the in-container one.

mtrmac:
Added a commit which significantly expands the section on handling `containers/image` updates. @giuseppe does that make sense?

README.md
```diff
@@ -184,10 +184,20 @@ In order to update an existing dependency:
 - update the relevant dependency line in `vendor.conf`
 - run `vndr github.com/pkg/errors`
 
-In order to test out new PRs from [containers/image](https://github.com/containers/image) to not break `skopeo`:
+When new PRs for [containers/image](https://github.com/containers/image) break `skopeo` (i.e. `contaners/image` tests fail in `make test-skopeo`):
```
giuseppe: `s|contaners|containers|`

README.md
```diff
@@ -184,10 +184,20 @@ In order to update an existing dependency:
...
 
+- create out a new branch in your `skopeo` checkout and switch to it
 - update `vendor.conf`. Find out the `containers/image` dependency; update it to vendor from your own branch and your own repository fork (e.g. `github.com/containers/image my-branch https://github.com/runcom/image`)
 - run `vndr github.com/containers/image`
+- make any other necessary changes in the skopeo repo (e.g. add other dependencies now requried by `containers/image`, or update skopeo for changed `containers/image` API)
```
giuseppe: s|requried|required|

mtrmac:
Thanks, typos updated.

@runcom Could you take another look, please?

runcom:
@mtrmac looks great, thanks a lot

PR #347, opened 2017-05-26
Support /etc/docker/certs.d

### LINKED ISSUE #250: Support cert-based client-server authentication
Per https://docs.docker.com/engine/security/certificates/

This is being used by Red Hat to provide images only to customers with a valid entitlement subscription.

mtrmac:
This vendors in https://github.com/containers/image/pull/277 .

Fixes #250 , by supporting `certs.d` automatically instead of relying on the user of `skopeo` to manually supply a `--`*`cert-dir`.


mtrmac:
Tested per https://bugzilla.redhat.com/show_bug.cgi?id=1455575 .

PR #351, opened 2017-06-01
vendor.conf: add ostree-go

giuseppe:
it is used by containers/image for pulling images to the OSTree storage.

This is required in preparation of: https://github.com/containers/image/pull/279


mtrmac:
We typically create a single skopeo PR, initially (for test run purposes) vendoring from the unmerged containers/image branch and then updated to vendor from containers/image `master`; but this works as well, if @runcom is OK with it.

mtrmac:
Scratch that, sorry, NAK as is; we do vendoring differently in `containers/image` and `skopeo`.

In `containers/image`, only `vendor.conf` is checked into `git`, and the dependencies are pulled by the test process.

In skopeo, both `vendor.conf` and the resulting `vendor` directory is checked into `git`; so, merely adding a line to `vendor.conf` would not fix the test failures in https://github.com/containers/image/pull/279 .

I.e., please run `vndr` on top of this PR, and add the resulting changes to the same commit.

giuseppe:
@mtrmac I've pushed a new version with the ostree-go files vendor.

Is there any wish for Skopeo to use `trash` as well for maintaining `vendor`?

mtrmac:
ACK, thanks. (Still up to @runcom whether he wants this stand-alone PR.)

> Is there any wish for Skopeo to use `trash` as well for maintaining `vendor`?

I was personally hoping to avoid having to care until Golang finally standardizes on something :) At least for me personally, `vndr` works well enough that I don’t see an urgent need to change; of course there might be a great features I don’t know  I am missing.

giuseppe:
@mtrmac as for another PR, I won't probably open one to re-vendor `containers/image`.  The change for using ostree go bindings is not really so urgent and it can probably wait until next sync.

From my limited experience: the only advantage with `trash` is that it maintains a cache, most libs will be shared with `containers/image`.  It also opens up to the possibility of dropping `vendor` completely (as regenerating it is quite cheap from the cache) as `containers/image` does.

runcom:
can't see why we need to merge this now when https://github.com/containers/image/pull/279 is still in-flight, so let's just wait and do a commit when that's merged and bumped here.

mtrmac:
> can't see why we need to merge this now when containers/image#279 is still in-flight,
@runcom Because containers/image#279 tests fail without a vendor update, we need _something_ in skopeo.

So, I guess let’s do the traditional “DO NOT MERGE: Vendor containers/image $branch” PR?

runcom:
> So, I guess let’s do the traditional “DO NOT MERGE: Vendor containers/image $branch” PR?

Yep

giuseppe:
renamed to [donotmerge]..  I'll give a run for the containers/image PR with this branch to be sure tests will still pass once we will merge this.

mtrmac:
@giuseppe I’m sorry we keep being unclear (I’ll try to clarify the README in https://github.com/projectatomic/skopeo/pull/343 , and please do speak up if it is insufficient): the point is not special naming conventions of the PR, but having a skopeo PR which both integrates the new proposed containers/image version and contains any other updates necessary so that skopeo keeps building and working.

(When we initially split the skopeo and containers/image repos, we have had huge problems with the two getting out of sync, mostly by API changes, where updates of containers/image would break skopeo and nobody would notice that skopeo needs updating. Hence `make test-skopeo` to detect this in containers/image, and _if_ `make test-skopeo` fails, creating these “do not merge” “pre-PRs” to show what _exactly_ will be updated in skopeo; once that “pre-PR” has tests passing, the containers/image PR is merged and the “pre-PR” is converted into a regular PR vendoring from `containers/image:master` and immediately merged.)

mtrmac:
> the point is not special naming conventions of the PR, but having a skopeo PR which both integrates the new proposed containers/image version and contains any other updates necessary so that skopeo keeps building and working.

i.e. something like https://github.com/projectatomic/skopeo/pull/352 , (which seems to fail because skopeo’s CI also needs updating to install dependencies)

giuseppe:
@mtrmac thanks for the explanation.  (I think) I got it now :)  About the containers/image PRs I've opened https://github.com/containers/image/pull/279 is currently blocked on https://github.com/containers/image/pull/281.  I'll rebase it and do the changes once 281 is merged.  At that point we can amend the changes in the donotmerge PR for Skopeo and ensure the tests are still passing.

Does that make sense?

mtrmac:
Yes, perfect.

mtrmac:
Thanks!

PR #352, opened 2017-06-02
DO NOT MERGE: Vendor in giuseppe/image:use-ostree-golang-bindings

mtrmac:
This requires also adding github.com/ostreedev/ostree-go.

This is supposed to demonstrate how https://github.com/containers/image/pull/279 is incorporated into skopeo, right now the test suite fails.


mtrmac:
This is now correctly done in #351.

PR #356, opened 2017-06-09
Added go-md2man to fix build process from source.

gianarb:
I tried to build skopeo from source:

```bash
go get github.com/projectatomic/skopeo
cd $GOPATH/github.com/projectatomic/skopeo
make
```

This was the output.

```
Step 4/4 : WORKDIR /src/github.com/projectatomic/skopeo
 ---> b5eab880bf01
 Removing intermediate container 6695eac353b8
 Successfully built b5eab880bf01
 docker run --rm --security-opt label:disable -v
 $(pwd):/src/github.com/projectatomic/skopeo \
 skopeobuildimage make binary-local  BUILDTAGS='
 libdm_no_deferred_remove'
 go build -ldflags "-X
 main.gitCommit=f0d830a8ca959076ee3b78c71d861972a734c036"
 -gcflags "" -tags "libdm_no_deferred_remove" -o skopeo
 ./cmd/skopeo
 go-md2man -in docs/skopeo.1.md -out docs/skopeo.1.tmp && touch
 docs/skopeo.1.tmp && mv docs/skopeo.1.tmp docs/skopeo.1
 /bin/sh: 1: go-md2man: not found
 Makefile:69: recipe for target 'docs/skopeo.1' failed
 make: *** [docs/skopeo.1] Error 127
```

I added `go-md2man`.
Thanks a lot for your work!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,6 +37,7 @@ RUN set -x \
 ENV GOPATH /usr/share/gocode:/go
 ENV PATH $GOPATH/bin:/usr/share/gocode/bin:$PATH
 RUN go get github.com/golang/lint/golint
+RUN go get github.com/cpuguy83/go-md2man
 WORKDIR /go/src/github.com/projectatomic/skopeo
 COPY . /go/src/github.com/projectatomic/skopeo
 


runcom:
go-md2man should be installed here https://github.com/gianarb/skopeo/blob/b8ecd7d0c29916b81de2ca8226b928681945b6b9/Dockerfile#L3

This is likely an issue with where it's installed in PATH

gianarb:
Is it working in your env @runcom ? I am using Ubuntu 16.10. Can we move it from dnf to `go get` or something like that?

mtrmac:
Actually, `make docs` (as called by `make`) does not run in a container. The recently updated `README.md` does, kind of, say that go-md2man needs to be installed locally, but `make` using a container and possibly failing anyway _is_ surprising.

Some options:
- Drop `docs` from the default target.
- Duplicate `docs` into an in-container/without-container pair as well, and run it in a container with `make all`
- Document the dependency better
- Ensure `skopeo` is packaged in distributions, and highlight distribution packages over building from source.
- Automatically install `go-md2man`—probably not, silently modifying user’s environment is pretty hostile.
- (Possibly rework `Makefile` with a variable, so that we don’t have every target twice… i.e. have `make binary INCONTAINER={true,false}` instead of `make binary{,-local}`)

PR #365, opened 2017-06-16
Remove docker references whereever possible

rhatdan:

### DIFF
--- a/docs/skopeo.1.md
+++ b/docs/skopeo.1.md
@@ -6,7 +6,7 @@ skopeo -- Various operations with container images and container image registrie
 # SYNOPSIS
 **skopeo** [_global options_] _command_ [_command options_]
 # DESCRIPTION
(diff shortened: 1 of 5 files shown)


docs/skopeo.1.md
```diff
@@ -41,7 +41,7 @@ Most commands refer to container images, using a _transport_`:`_details_ format.
 
   **--insecure-policy** Adopt an insecure, permissive policy that allows anything. This obviates the need for a policy file.
 
-  **--registries.d** _dir_ use registry configuration files in _dir_ (e.g. for docker signature storage), overriding the default path.
+  **--registries.d** _dir_ use registry configuration files in _dir_ (e.g. for container signature storage), overriding the default path.
```
runcom: this is actually spec'ed in the signature format (the docker transport) /cc @mtrmac

rhatdan: Maybe drop docker then, and just say signature storage.
Where there was a clear "docker specification" I left the word.

mtrmac: Actually, `/etc/containers/registries.d` was intended to be transport-independent (https://github.com/projectatomic/skopeo/issues/183 ) and the files are internally transport-namespaced (the `docker:` and `default-docker:` top-level keys in https://github.com/containers/image/blob/master/docs/registries.d.md ).

So, saying container is fine, though specifically pointing at these files for the docker-specific lookaside system might still be useful. But then users who need pointing at these files would also probably want a knowledge base article or something like that to guide them through the publishing process, it shouldn’t all hinge on this single obscure reference in a man page.

runcom: Fine then

runcom:
LGTM

PR #354, opened 2017-06-05
[DO NOT MERGE] vendor c/image for concurrent layers pull

runcom:
Vendor of https://github.com/containers/image/pull/283

**PLEASE USE LATEST VNDR TOOL**
vndr got some cool updates so update it with `go get -u github.com/LK4D4/vndr`

Signed-off-by: Antonio Murdaca <runcom@redhat.com>

PR #371, opened 2017-06-27
Bump and pin containers/storage and containers/image

nalind:
Bump containers/storage, and work around an API change in our vendored copy of containers/image.


mtrmac:
This bumps containers/storage without updating skopeo’s `vendor.conf`  ??!

(WRT the second commit, it would instead be preferable to temporarily vendor in the https://github.com/containers/image/pull/295 instead of patching generated files, e.g. like #354, and to revert to vendoring from `master` after the c/image PR is merged. This one-liner is trivial enough that the current approach is OK for demonstrating that tests pass.)

After https://github.com/containers/image/pull/295 is merged, please replace this by updating top-level `vendor.conf`, running `vndr` and committing the results.

nalind:
Okay, updated.  We were pulling both storage and image directly from master, so I've set their versions to the current master commit IDs.

mtrmac:
> We were pulling both storage and image directly from master, 

Oops, that was stupid of me. Sorry. In that case, up to you whether you want to pin containers/storage, but leave containers/image at `master`.

nalind:
Okay, reverting the pinning for both.

mtrmac:
:+1: pending tests.

runcom:
will merged after tests :+1:

PR #376, opened 2017-07-18
vendor c/image: fix auth handlers

### LINKED ISSUE #375: " no handler for negotiate authentication"
Attempting to interact with a private V2 registry, with creds provided in either a config.json or via command line, I receive:

```
sudo skopeo inspect --creds=username:password docker://private-registry/imagename:tag 
FATA[0000] no handler for negotiate authentication  
```

I'd run into the same error with Docker 1.12 using Fedora 25, and found this:  https://github.com/moby/moby/issues/30276#issuecomment-279368196

That worked around the issue for Docker itself.  Is this the same issue, and is there a workaround for Skopeo?  The thread suggests it's something in the RHEL version of the Docker rpm, so maybe this is something else.

Just taking a guess, I tried `skopeo --insecure-policy ... `,  as well as setting up a policy of type `insecureAcceptAnything` for the docker transport, but both resulted in the same error.  I suspect I may be barking up the wrong tree.

runcom:
fix #375 


mtrmac:
LGTM, thanks!

PR #377, opened 2017-07-19
Update to image-spec v1.0.0 and revendor

mtrmac:
Update from image-spec 1.0.0rc6 to the final one.

(Yes, this does not change a single bit of the code from image-spec that we actually use, so we are updating only to look good; OTOH it does add quite a few updates from other packages. Potentially notable is that `opencontainers/go-digest` now refuses uppercase digits in digest values.)


mtrmac:
@rhatdan FYI

TomSweeneyRedHat:
LGTM

rhatdan:
LGTM

runcom:
LGTM

rhatdan:
Can we get this merged?

PR #378, opened 2017-07-20
Update to image-spec v1.0.0 and revendor

mtrmac:
This is similar to #377, bumping the tag but the only code changes are in unrelated packages…

@rhatdan FYI


rhatdan:
Awesome
@runcom PTAL

runcom:
LGTM (we'll cut a new release later today, stay tuned!!!)

mtrmac:
(For the record, this updates runtime-spec, not image-spec.)

PR #387, opened 2017-07-26
Fix `make install` on OSX

### LINKED ISSUE #383: Make install on OSX
The `make install*` tasks don't work on OSX since the `install` is a BSD rather than a GNU one and some of the flags are a bit different.

It won't be too hard to fix... but would it reasonable to do some test to try and work out the platform in the Makefile?

Something like `if [ "$(uname)" == "Darwin" ]`

errm:
Fixes #383

Signed-off-by: Ed Robinson <ed.robinson@reevoo.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,13 @@
 
 export GO15VENDOREXPERIMENT=1
 
+ifeq ($(shell uname),Darwin)
+PREFIX ?= ${DESTDIR}/usr/local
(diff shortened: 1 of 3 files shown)


mtrmac:
Thanks! LGTM pending tests.

errm:
@mtrmac CI fixed, PTAL

PR #399, opened 2017-08-25
Modify skopeo tests

umohnani8:
The oci name changes in containers/image https://github.com/containers/image/pull/318 caused the skopeo test to fail

Signed-off-by: umohnani8 <umohnani@redhat.com>

### DIFF
--- a/integration/copy_test.go
+++ b/integration/copy_test.go
@@ -137,14 +137,20 @@ func (s *CopySuite) TestCopySimple(c *check.C) {
 	out := combinedOutputOfCommand(c, "diff", "-urN", dir1, dir2)
 	c.Assert(out, check.Equals, "")
 
-	// docker v2s2 -> OCI image layout
+	// docker v2s2 -> OCI image layout with image name
 	// ociDest will be created by oci: if it doesn't exist
 	// so don't create it here to exercise auto-creation
-	ociDest := "busybox-latest"
+	ociDest := "busybox-latest-image"
+	ociImgName := "busybox"
 	defer os.RemoveAll(ociDest)
-	assertSkopeoSucceeds(c, "", "copy", "docker://busybox:latest", "oci:"+ociDest)
+	assertSkopeoSucceeds(c, "", "copy", "docker://busybox:latest", "oci:"+ociDest+":"+ociImgName)
 	_, err = os.Stat(ociDest)
 	c.Assert(err, check.IsNil)
+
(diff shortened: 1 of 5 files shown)


umohnani8:
@mtrmac PTAL

integration/copy_test.go
```diff
@@ -137,14 +137,19 @@ func (s *CopySuite) TestCopySimple(c *check.C) {
...
+	ociImgName := "busybox"
 	defer os.RemoveAll(ociDest)
-	assertSkopeoSucceeds(c, "", "copy", "docker://busybox:latest", "oci:"+ociDest)
+	assertSkopeoSucceeds(c, "", "copy", "docker://busybox:latest", "oci:"+ociDest+":"+ociImgName)
 	_, err = os.Stat(ociDest)
 	c.Assert(err, check.IsNil)
+
+	// docker v2s2 -> OCI image layout without image name
+	ociDest = "busybox-latest-noimage"
+	assertSkopeoFails(c, ".*Error writing manifest: cannot save image with empty image.ref.name.*", "copy", "docker://busybox:latest", "oci:"+ociDest)
```
mtrmac: Testing this scenario might detect problems earlier if it were a unit test in the `oci` transport; testing that here doesn’t hurt, sure; but shouldn’t the `busybox-latest-noimage` directory also be cleaned up? The error is reported in `PutManifest`, which means `PutBlob` has already written files into the directory AFAICS.

(Even the old code should have used `ioutil.TempDir` instead of writing to the current directory, but that’s of course independent of this PR and not a blocker for it.)

umohnani8: Fixed.

vendor.conf
```diff
@@ -1,5 +1,6 @@
 github.com/urfave/cli v1.17.0
 github.com/containers/image master
+# github.com/containers/image master
```
mtrmac: Nit: please drop this comment.

umohnani8: whoops, my bad. Fixed.

mtrmac:
@umohnani8 Please update the expected error message:
```
... value string = "time=\"2017-08-31T15:43:50Z\" level=fatal msg=\"Error initializing destination oci:busybox-latest-noimage:: cannot save image with empty image.ref.name\" \n"
... regex string = "(?s).*Error writing manifest: cannot save image with empty image.ref.name.*"
```

mtrmac:
Thanks!

PR #411, opened 2017-09-07
Touch up a few tpyos

TomSweeneyRedHat:

Try two, hopefully without merge issues from github.  Touching up a few typos in the README.md as noted in a review earlier today.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 skopeo [![Build Status](https://travis-ci.org/projectatomic/skopeo.svg?branch=master)](https://travis-ci.org/projectatomic/skopeo)
 =
 
-_Please be aware `skopeo` is still work in progress and it currently supports only registry API V2_
+_Please be aware that `skopeo` is still a work in progress and it currently supports only registry API V2_
 
 `skopeo` is a command line utility for various operations on container images and image repositories.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,7 +1,7 @@
 skopeo [![Build Status](https://travis-ci.org/projectatomic/skopeo.svg?branch=master)](https://travis-ci.org/projectatomic/skopeo)
 =
 
-_Please be aware `skopeo` is still work in progress and it currently supports only registry API V2_
+_Please be aware that `skopeo` is still a work in progress and it currently supports only registry API V2_
```
rhatdan: I think calling skopeo a work in progress denigrates it usefulness.  I believe that we should just have a note saying that it currently support the registry API V2.  What api would we want it to support?  Definitely not V1?

rhatdan: This page should celebrate all that skopeo can do.  We need to add the icon and talk about moving container images between different types of container storages.  Lets pump up Skopeo!

mtrmac:
LGTM, thanks.

(Pumping up skopeo should definitely happen, but does not block this PR.)

README.md
TomSweeneyRedHat: I concur, but for the short term I'd like to get this one in as the typos need to be cleaned.  I'll work on a better update tomorrow (Friday) morn.

PR #420, opened 2017-09-16
Support copying a single image from manifest lists

mtrmac:
This vendors in https://github.com/containers/image/pull/346 , and adds global `--override-arch` and `--override-os` options.

See https://github.com/containers/image/pull/346 for more discussion about implications of transparently choosing an image, especially for macOS.


mtrmac:
@runcom PTAL

runcom:
LGTM

mtrmac:
> Do we need a variant flag too for arm since there can be 32/64 bit images under the same 'arm' arch?

I don’t think so; AFAICS Golang `GOARCH` has separate `arm`/`arm64` values, they are not “variants”.

PR #426, opened 2017-09-26
Update image-tools, and remove the duplicate Sirupsen/logrus vendor

mtrmac:
Now that https://github.com/opencontainers/image-tools/pull/176 has been merged, we can stop carrying two `logrus` checkouts.

Also re-vendors other packages, now that we can :)

@runcom PTAL.


runcom:
Looks ok but some check fails

mtrmac:
Yeah, on macOS (case-insensitive), because Travis does
```sh
git clone --depth=50 https://github.com/projectatomic/skopeo.git projectatomic/skopeo
cd projectatomic/skopeo
git fetch origin +refs/pull/426/merge:
git checkout -qf FETCH_HEAD
```

The initial clone is checked out as `Sirupsen`, with the duplicate `sirupsen` ignored. The checkout applies the diff, which removes `Sirupsen`, and there isn’t any `logrus` on the filesystem left.

A direct
```sh
 git clone -b single-logrus https://github.com/mtrmac/skopeo.git x
```

checks out the final state directly, with a `sirupsen` (lowercase only), and everything should work.

So, if it’s OK with you, let’s force merge this through (It does pass on Linux as is, at worst we would only break Mac again).

mtrmac:
@runcom ?

runcom:
So good to merge?!

mtrmac:
Yes, I think it is good enough. Up to you whether my reasoning is sound enough.

runcom:
It seems like I don't have power to merge?!

mtrmac:
Me neither, and manual push is also refusing this:
```
remote: error: GH006: Protected branch update failed for refs/heads/master.
remote: error: Required status check "continuous-integration/travis-ci/pr" is failing.
```

I guess this is somewhere in the project settings.

runcom:
@mtrmac ping

mtrmac:
@runcom Rebased. The tests on macOS are still going to fail, as explained above, and you’ll need your admin permissions (if any) to override the tests and merge.

mtrmac:
@runcom ping

runcom:
I don't have power to merge unfortunately, @rhatdan maybe?

rhatdan:
Nope not me

runcom:
@jwhonce maybe? :angel:

PR #427, opened 2017-09-29
Vendor in latest containers storage

rhatdan:
We want to get support into skopeo for handling
override_kernel_checks so that we can use overlay
backend on RHEL.


rhatdan:
@runcom @mtrmac @mrunalp @nalind PTAL

nalind:
Looks like a typical bump-a-dependency change, so LGTM.

mtrmac:
(Do we want to switch from vendoring `master` (i.e. random updates over time, but keeping fairly fresh) to a specific git commit (i.e. a stable version when a human decides that it’s a good time to update, but no updates if no-one pays attention)? I don’t really have a strong opinion on this, just noting that this is not _only_ a rebase.)

Not merging due to the test failure — updating containers/storage to support 1.7 is of course nice; looking at https://github.com/containers/image/pull/352 , we might update the Go version to 1.8 instead.

rhatdan:
I don't care about switching from master to a specific version.  I know that @mrunalp prefers the git commit in cri-o, since he believes if he gets a bug on some vendored piece of code, it would be easier to investigate what was happening in that version.

I believe we should move to go 1.8.  We need this version of Skopeo ASAP since OpenShift wants to use skopeo to move images between docker storage and container storage using overlay.

rhatdan:
How do I get the OSX Stuff to use a newer version of go?

rhatdan:
Switched back to master, we can talk about which is better later, and update to @nalind patch to allow containers/storage to build on golang 1.7

rhatdan:
@mtrmac @runcom PTAL, Can we get this merged?

runcom:
Looks ok, @mtrmac ?

PR #433, opened 2017-10-15
Add --no-creds flag to skopeo inspect

juanluisvaladas:
Implements https://github.com/projectatomic/skopeo/issues/421

Currently skopeo inspect allows to:
* Use the default credentials in $HOME/.docker.config
* Explicitly define credentials via de --creds flag

This implements a --no-creds flag which will query docker registries anonymously.

### DIFF
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -141,6 +141,16 @@ var copyCmd = cli.Command{
 			Value: "",
 			Usage: "Use `USERNAME[:PASSWORD]` for accessing the destination registry",
 		},
+		cli.BoolFlag{
+			Name:  "src-no-creds",
+			Value: "",
+			Usage: "access the source registry anonymously",
+		},
+		cli.BoolFlag{
+			Name:  "dest-no-creds",
+			Value: "",
+			Usage: "access the destination registry anonymously",
+		},
 		cli.StringFlag{
 			Name:  "src-cert-dir",
 			Value: "",
--- a/cmd/skopeo/utils.go
+++ b/cmd/skopeo/utils.go
@@ -27,6 +27,9 @@ func contextFromGlobalOptions(c *cli.Context, flagPrefix string) (*types.SystemC
 		DockerDaemonCertPath:              c.String(flagPrefix + "cert-dir"),
 		DockerDaemonInsecureSkipTLSVerify: !c.BoolT(flagPrefix + "tls-verify"),
 	}
+	if c.IsSet(flagPrefix+"creds") && c.IsSet(flagPrefix+"no-creds") {
+		return nil, errors.New("creds and no-creds cannot be specified at the same time")
+	}
 	if c.IsSet(flagPrefix + "tls-verify") {
 		ctx.DockerInsecureSkipTLSVerify = !c.BoolT(flagPrefix + "tls-verify")
 	}
@@ -36,6 +39,9 @@ func contextFromGlobalOptions(c *cli.Context, flagPrefix string) (*types.SystemC
 		if err != nil {
 			return nil, err
 		}
+	}
+	if c.Bool(flagPrefix + "no-creds") {
+		ctx.DockerAuthConfig = &types.DockerAuthConfig{}
 	}
 	return ctx, nil
 }
--- a/cmd/skopeo/delete.go
+++ b/cmd/skopeo/delete.go
@@ -51,6 +51,10 @@ var deleteCmd = cli.Command{
 			Value: "",
 			Usage: "Use `USERNAME[:PASSWORD]` for accessing the registry",
 		},
+		cli.BoolFlag{
+			Name:  "no-creds",
+			Usage: "access the registry anonymously",
+		},
 		cli.StringFlag{
 			Name:  "cert-dir",
 			Value: "",
--- a/cmd/skopeo/inspect.go
+++ b/cmd/skopeo/inspect.go
@@ -65,6 +65,10 @@ var inspectCmd = cli.Command{
 			Value: "",
 			Usage: "Use `USERNAME[:PASSWORD]` for accessing the registry",
 		},
+		cli.BoolFlag{
+			Name:  "no-creds",
+			Usage: "access the registry anonymously",
+		},
 	},
 	Action: func(c *cli.Context) (retErr error) {
 		ctx := context.Background()


juanluisvaladas:
Hi, the travis build failure looks unrelated to my changes, can it be relaunched?

TomSweeneyRedHat:
The changes LGTM.  Not sure what's going on with Travis.  I concur, doesn't look like an issue with your submit, but rather some Ruby versioning issues.  @runcom do you know what's up with that or someone we can talk to for it?

mtrmac:
That failure is likely unrelated, it was not entirely obvious so from a quick look whether this is something Travis can fix for us or whether we need to update our build scripts.

cmd/skopeo/utils.go
```diff
@@ -28,6 +28,9 @@ func contextFromGlobalOptions(c *cli.Context, flagPrefix string) (*types.SystemC
 			return nil, err
 		}
 	}
+	if c.BoolT(flagPrefix + "no-creds") {
```
mtrmac: (This affects almost all commands; so the documentation, and bash completions, should also be updated everywhere.)

mtrmac:
https://github.com/kubernetes-incubator/cri-o/pull/1015 eventually suggests there should be a way to override the default config paths, and use something else; that might indirectly help here (`--auth-file /dev/null` or so), but I’m still curious when would a user want to not use the credentials required to make things work.

juanluisvaladas:
The scenario I've encountered and why I have implemented it is the following:
1-I'm the docker registry administrator
2- User complains about not being able to reach it
3- test with his u

juanluisvaladas:
The scenario I've encountered and why I have implemented it is the following:

I'm an openshift / atomic-registry administrator. A user complains about not being able to pull an image that should be available for everyone. I want to test it as system:unauthenticated. As I'm authenticated as the cluster admin here I have to choices:
1- `docker logout && skopeo inspect docker://...`
2- `HOME=/non/existing skopeo inspect docker://...`

Besides I use atomic-registry (the version shipped in OCP). If you try to access atomic-registry with no credentials you'll be considered `system:anonymous` but if you try to access it with an expired token you'll get an access denied. In this scenario, for automated things it's very handy for me to be able to run it with `--no-creds` so that I don't have to worry about having an expired token.

My main use for skopeo is debugging registries and I think many people will end up using it the same way

juanluisvaladas:
I added the flag --no-creds for delete (which is working) and for copy --src-no-creds and --dest-no-creds. The copy flags aren't working but  apparently the --src-tls-verify and --dest-tls-verify aren't either. I'll investigate this as a separate issue as soon as this PR is accepted or denied (to avoid merging conflicts)

juanluisvaladas:
I added the flag --no-creds for delete (which is working) and for copy --src-no-creds and --dest-no-creds. The copy flags aren't working but  apparently the --src-tls-verify and --dest-tls-verify aren't either. I'll investigate this as a separate issue as soon as this PR is accepted or denied (to avoid merging conflicts)

TomSweeneyRedHat:
Ruby is making Travis unhappy.  I don't believe it's created by these changes.
"Homebrew must be run under Ruby 2.3! (RuntimeError)"

cmd/skopeo/copy.go
```diff
@@ -99,6 +99,16 @@ var copyCmd = cli.Command{
...
+			Name:  "src-no-creds",
+			Value: "",
+			Usage: "Access the source registry anonymously",
+		},
+		cli.StringFlag{
+			Name:  "dest-no-creds",
+			Value: "",
+			Usage: "Access the destination registry anonymously",
+		},
+		cli.StringFlag{
```
TomSweeneyRedHat: note to self.  We really should uppercase the first word in all Usage messages, or lower case them all.

cmd/skopeo/utils.go
TomSweeneyRedHat: 👍

juanluisvaladas:
@TomSweeneyRedHat I've tested manually and I'm not certain but I think it isn't working.

I'll test it in a serious manner but I'm doing this on my free time so it's gonna be ok hold until the weekend.

mtrmac:
> Hi, the travis build failure looks unrelated to my changes, can it be relaunched?

That was fixed in #435 ; if you rebase on top of current `master`, the tests should rerun automatically.

> The scenario I've encountered and why I have implemented it is the following:
>
> I'm an openshift / atomic-registry administrator. A user complains about not being able to pull an image that should be available for everyone. I want to test it as system:unauthenticated. As I'm authenticated as the cluster admin here I have to choices:

Thanks; that does make sense, though I’d like @runcom to also take a look at the UI.

> I added the flag --no-creds for delete (which is working) and for copy --src-no-creds and --dest-no-creds. The copy flags aren't working but apparently the --src-tls-verify and --dest-tls-verify aren't either. I'll investigate this as a separate issue as soon as this PR is accepted or denied (to avoid merging conflicts)

FWIW `skopeo copy --src-creds` does work for me.

cmd/skopeo/utils.go
mtrmac: One more thing, please make this fail if both `"creds"` and `"no-creds"` are specified.

rhatdan:
@juanluisvaladas Could you rebase your PR?

@mtrmac How does this look now?

cmd/skopeo/utils.go
mtrmac: `BoolT` means “default to true”; this should default to false.

rhatdan:
This PR Looks pretty good, but no one has worked on it in almost a year.  @juanluisvaladas Are you still interested in this?  @mtrmac Should we rebase this ourselves?

juanluisvaladas:
@rhatdan sorry I forgot about this, I will get this rebased and tested on Monday.

juanluisvaladas:
bump

cmd/skopeo/copy.go
```diff
@@ -141,6 +141,16 @@ var copyCmd = cli.Command{
...
+			Name:  "src-no-creds",
+			Value: "",
+			Usage: "access the source registry anonymously",
+		},
+		cli.StringFlag{
+			Name:  "dest-no-creds",
+			Value: "",
+			Usage: "access the destination registry anonymously",
+		},
```
mtrmac: The two `StringFlag`s should be `BoolFlag` I think.

rhatdan:
@mtrmac @giuseppe @vrothberg, @juanluisvaladas has not worked on this for a while, should we have someone else take it over and move it to completion, or close the PR?

vrothberg:
> @mtrmac @giuseppe @vrothberg, @juanluisvaladas has not worked on this for a while, should we have someone else take it over and move it to completion, or close the PR?

Let's keep it open. Someone else can certainly take over and push the changes over the finish line.

QiWang19:
:ok_hand: I'll take this. @QiWang19

mtrmac:
#646 was merged. Thanks!

PR #435, opened 2017-10-19
Work around broken brew in Travis

mtrmac:


mtrmac:
@runcom PTAL. This fixes / works around the recent test failures we’ve been encountering on macOS.

PR #437, opened 2017-10-21
fix inspect with nil image config

runcom:
Fix https://github.com/projectatomic/skopeo/issues/436

@mtrmac @strigazi @miabbott @rhatdan PTAL

I'll cut a new release when this is in


rhatdan:
LGTM

strigazi:
Thanks for the quick fix!

Tested and works fine,

but, I'm not sure where it needs to be done (or if it needs to be done at all), to show a warning or hint that labels are missing. We had this issue with kubernetes which was throwing the same error and I tried skopeo to help me debug the issue.

If this fix was in place in the first place the "error/problem" would be silently ignored.

What do you think?

runcom:
I don't think skopeo should say anything about this, it's not an image validator, best it can do is to not panic. If you have issues with labels in an image you would definitely do many things to figure it out and with this fix you aren't silently ignoring the error - you'll find the error again wherever you use labels :)

So this should be enough

TomSweeneyRedHat:
LGTM

runcom:
Ready to go now @mtrmac PTAL

mtrmac:
>> but, I'm not sure where it needs to be done (or if it needs to be done at all), to show a warning or hint that labels are missing.

> I don't think skopeo should say anything about this, it's not an image validator, best it can do is to not panic.

Well, it could also be reasonable to fail with an ”your config is invalid” error and just refuse to output potentially incorrect data.

More to the point, at least in OCI per https://github.com/opencontainers/image-spec/blob/master/config.md the whole `config` subobject in `config.json` is optional, so this is _not_ invalid.

It could, though, make good sense to change the output of `skopeo inspect` to completely omit the `Labels` key in the output JSON if the image does not have any list of labels (as opposed to having a list of labels which is empty). It would be a bit of a hassle to both output and parse. Or, I guess, we could always invent a `LabelsPresent` field.

mtrmac:
:+1: pending tests

runcom:
I'll tag a new skopeo later today.

PR #440, opened 2017-10-27
[DO NOT MERGE] - Aligning Docker version between containers/image and skopeo

hferentschik:
See also https://github.com/containers/image/pull/328


vendor.conf.tmp
```diff
@@ -0,0 +1,46 @@
+github.com/urfave/cli v1.17.0
```
mtrmac: Please drop this file.

vendor.conf
```diff
@@ -1,23 +1,25 @@
 github.com/urfave/cli v1.17.0
-github.com/containers/image master
+github.com/containers/image issue-327 https://github.com/hferentschik/image.git
```
runcom: can you vendor this from `master` again?

vendor.conf
```diff
@@ -1,23 +1,25 @@
...
 github.com/opencontainers/go-digest master
 gopkg.in/cheggaaa/pb.v1 ad4efe000aa550bb54918c06ebbadc0ff17687b9 https://github.com/cheggaaa/pb
 github.com/containers/storage master
 github.com/sirupsen/logrus v1.0.0
+github.com/Sirupsen/logrus v1.0.0
```
runcom: mmm which of these is correct? why do we have both?

vendor.conf
hferentschik: Sorry, that was my intention

vendor.conf
hferentschik: github.com/sirupsen/logrus v1.0.0 is the correct one now. There was a major issue around logrus when it got renamed from github.com/Sirupsen/logrus to github.com/sirupsen/logrus.

hferentschik:
@runcom pushed an update

vendor.conf
runcom: right, my question was more "should we keep both?!"

runcom:
> @runcom pushed an update

looks ok locally, will merge after Travis is done testing. Thanks!

hferentschik:
Thanks

vendor.conf
```diff
@@ -27,7 +28,7 @@ github.com/opencontainers/runc master
 github.com/opencontainers/image-spec v1.0.0
 # -- start OCI image validation requirements.
 github.com/opencontainers/runtime-spec v1.0.0
-github.com/opencontainers/image-tools 6d941547fa1df31900990b3fb47ec2468c9c6469
+github.com/opencontainers/image-tools da84dc9dddc823a32f543e60323f841d12429c51
```
mtrmac: This reverts image-tools to an older commit, notably using upper-case `Sirupsen`.  But this commit does not actually update the vendored package?!

vendor.conf
```diff
@@ -48,4 +49,3 @@ github.com/opencontainers/selinux master
 golang.org/x/sys master
 github.com/tchap/go-patricia v2.2.6
 github.com/BurntSushi/toml master
-github.com/pquerna/ffjson d49c2bc1aa135aad0c6f4fc2056623ec78f5d5ac
```
mtrmac: And this removed item also does not correspond to removing any files from `vendor`?!

runcom:
@mtrmac thanks for spotting that

PR #441, opened 2017-10-28
Create smaller testing containers

mtrmac:
Four small changes to reduce the size of the `skopeo-dev` containers.

Looking at `docker history`, the RPM installation step is now 605 MB instead of 611 MB, while installing many more packages; and the OpenShift build step is 456 MB instead of 1.82 GB.  Overall, the container size is decreased from 2.78 GB to 1.41 GB, almost by half.

See the individual commit messages for details.


runcom:
Amazing @mtrmac :+1:

PR #443, opened 2017-11-06
copy: add shared blob directory support for OCI sources/destinations

jonboulle:
Following on from containers/image#369, this adds command-line flags to `skopeo copy` which allow users to use a shared directory for blobs when working with OCI layouts.

@runcom this patch is nice and tiny, but is there a better way to scope flags to specific source/dest types?

### DIFF
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -121,5 +121,15 @@ var copyCmd = cli.Command{
 			Value: "",
 			Usage: "`DIRECTORY` to use for OSTree temporary files",
 		},
+		cli.StringFlag{
+			Name:  "src-shared-blob-dir",
+			Value: "",
+			Usage: "`DIRECTORY` to use to fetch retrieved blobs (OCI layout sources only)",
+		},
+		cli.StringFlag{
+			Name:  "dest-shared-blob-dir",
+			Value: "",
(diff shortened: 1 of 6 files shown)


mtrmac:
:+1: pending tests.

Currently there isn’t a better way.  In principle `parseImage*` + `contextFrom*` could work together to ensure that the user does not specify context options which would be ignored by the user-specified image reference.

runcom:
Well, we could enhance the OCI référence format to take options in the reference string itself but best to go with this now.

jonboulle:
That's.... odd 

```
From github.com:projectatomic/skopeo
   62c6899..0d0055d  master     -> origin/master
Updating 62c6899..0d0055d
Fast-forward
 CONTRIBUTING.md | 6 +++---
 1 file changed, 3 insertions(+), 3 deletions(-)
```

Anyway, updated now

mtrmac:
> Well, we could enhance the OCI référence format to take options in the reference string itself

That introduces undesirable string parsing and joining, and issues like ”are two images the same if they only differ in an option?”.

The `contextFromGlobalOptions` could just notice that it is setting options for a transport which does not match the transport of a reference. (Which would be even easier if `SystemContext` did not embed all options directly, but only carried a pointer to an optional separate structure for each transport/topic. We might eventually need to do that just to prevent `SystemContext` from being too big.)

Anyway, that’s for someone’s Copious Spare Time.

Issue #446, opened 2017-11-07
README.md shows an invalid example of skopeo copy an image on OCI format


marcosps:
In README.md we can find:

skopeo copy docker://busybox:latest oci:busybox_ocilayout

But, without a proper tag, this command returns an error:
FATA[0000] Error initializing destination oci:tmp:: cannot save image with empty image.ref.name 

A correct example of the command should be:
skopeo copy docker://busybox:latest oci:busybox_ocilayout:latest

Or am I missing something?

PR #447, opened 2017-11-07
README.md: Fix example of skopeo copy command

marcosps:
In README.md, there is an example of skopeo copy command to download an
image in OCI format, but the current code returns an error:

skopeo copy docker://busybox:latest oci:busybox_ocilayout
FATA[0000] Error initializing destination oci:tmp:: cannot save image with empty image.ref.name

If we add a tag after the oci directory, the problem is gone:
skopeo copy docker://busybox:latest oci:busybox_ocilayout:latest

Fixes: #446

Signed-off-by: Marcos Paulo de Souza <marcos.souza.org@gmail.com>


mtrmac:
Thanks, good catch!

PR #453, opened 2017-11-16
Use credentials from authfile for skopeo commands

umohnani8:
skopeo copy, delete, and inspect can now use credentials stored in the auth file
by the kpod login command
e.g kpod login docker.io -> skopeo copy dir:mydir docker://username/image

Signed-off-by: Urvashi Mohnani <umohnani@redhat.com>

### DIFF
--- a/docs/skopeo.1.md
+++ b/docs/skopeo.1.md
@@ -19,7 +19,7 @@ Most commands refer to container images, using a _transport_`:`_details_ format.
   An existing local directory _path_ storing the manifest, layer tarballs and signatures as individual files. This is a non-standardized format, primarily useful for debugging or noninvasive container inspection.
 
   **docker://**_docker-reference_
-  An image in a registry implementing the "Docker Registry HTTP API V2". By default, uses the authorization state in `$HOME/.docker/config.json`, which is set e.g. using `(docker login)`.
+  An image in a registry implementing the "Docker Registry HTTP API V2". By default, uses the authorization state in either `$XDG_RUNTIME_DIR/containers/auth.json`, which is set using `(kpod login)`. If the authorization state is not found there, `$HOME/.docker/config.json` is checked, which is set using `(docker login)`.
 
   **docker-archive:**_path_[**:**_docker-reference_]
   An image is stored in the `docker save` formated file.  _docker-reference_ is only used when creating such a file, and it must not contain a digest.
@@ -64,6 +64,11 @@ Uses the system's trust policy to validate images, rejects images not trusted by
 
   _destination-image_ use the "image name" format described above
 
+  **--authfile** _path_
+
+  Path of the authentication file. Default is ${XDG_RUNTIME\_DIR}/containers/auth.json, which is set using `kpod login`.
+  If the authorization state is not found there, $HOME/.docker/config.json is checked, which is set using `docker login`.
+
   **--format, -f** _manifest-type_ Manifest type (oci, v2s1, or v2s2) to use when saving image to directory using the 'dir:' transport (default is manifest type of source)
 
   **--remove-signatures** do not copy signatures, if any, from _source-image_. Necessary when copying a signed image to a destination which does not support signatures.
@@ -97,6 +102,11 @@ Mark _image-name_ for deletion.  To release the allocated disk space, you need t
 $ docker exec -it registry bin/registry garbage-collect /etc/docker/registry/config.yml
 ```
 
+  **--authfile** _path_
+
+  Path of the authentication file. Default is ${XDG_RUNTIME\_DIR}/containers/auth.json, which is set using `kpod login`.
+  If the authorization state is not found there, $HOME/.docker/config.json is checked, which is set using `docker login`.
+
   **--creds** _username[:password]_ for accessing the registry
 
   **--cert-dir** _path_ Use certificates at _path_ (*.crt, *.cert, *.key) to connect to the registry
@@ -113,6 +123,11 @@ Return low-level information about _image-name_ in a registry
   **--raw** output raw manifest, default is to format in JSON
 
   _image-name_ name of image to retrieve information about
+
+  **--authfile** _path_
+
+  Path of the authentication file. Default is ${XDG_RUNTIME\_DIR}/containers/auth.json, which is set using `kpod login`.
+  If the authorization state is not found there, $HOME/.docker/config.json is checked, which is set using `docker login`.
 
   **--creds** _username[:password]_ for accessing the registry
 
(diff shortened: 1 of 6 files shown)


umohnani8:
@rhatdan @TomSweeneyRedHat @mtrmac PTAL

umohnani8:
@mtrmac would you prefer the authfile flag be `flagPrefix+"authfile"` as well?

docs/skopeo.1.md
```diff
@@ -19,7 +19,7 @@ Most commands refer to container images, using a _transport_`:`_details_ format.
   An existing local directory _path_ storing the manifest, layer tarballs and signatures as individual files. This is a non-standardized format, primarily useful for debugging or noninvasive container inspection.
 
   **docker://**_docker-reference_
-  An image in a registry implementing the "Docker Registry HTTP API V2". By default, uses the authorization state in `$HOME/.docker/config.json`, which is set e.g. using `(docker login)`.
+  An image in a registry implementing the "Docker Registry HTTP API V2". By default, uses the authorization state in `$XDG_RUNTIME_DIR/containers/auth.json`, which is set e.g. using `(kpod login)`.
```
rhatdan: I think we should leave in that this works with either kpod login or docker login.  Since skopeo is used by the larger docker community.

docs/skopeo.1.md
```diff
@@ -245,6 +260,9 @@ $ skopeo standalone-verify busybox-manifest.json registry.example.com/example/bu
 Signature verified, digest sha256:20bf21ed457b390829cdbeec8795a7bea1626991fda603e0d01b4e7f60427e55
 ```
 
+# SEE ALSO
+kpod-login(1)
```
rhatdan: Also add docker-login(1)

docs/skopeo.1.md
```diff
@@ -64,6 +64,11 @@ Uses the system's trust policy to validate images, rejects images not trusted by
 
   _destination-image_ use the "image name" format described above
 
+  **--authfile** _path_
+
+  Path of the authentication file. Default is ${XDG_RUNTIME_DIR}/containers/auth.json
+  which is set using `kpod login`
+
```
TomSweeneyRedHat: Just double checking as I've not tried.  If you use the $HOME/.docker/config.json file as the auth file, is docker happy afterwards?  I.e. if you do 'kpod login --auth ~/.docker/config.json', can you then run a docker command that requires auth without it spitting up?  Really is not related to this review though.

docs/skopeo.1.md
TomSweeneyRedHat: Should we not this in the Buildah/kpod commands too?  I'm torn, we do use docker-login for those too, just don't want people to think we didn't think about that if we don't document it.  Then again I'd rather not add a docker dependency to them.  i.e. if docker changes auth at some point, then we'd have to adjust those for that rather than just dropping it.

docs/skopeo.1.md
umohnani8: fixed

docs/skopeo.1.md
umohnani8: I actually haven't tried that yet. It might not work because there is a slight difference in the way kpod and docker store the registry name in auth.json and config.json.

docs/skopeo.1.md
umohnani8: fixed.
@rhatdan should this be mentioned in buildah and kpod as well?

rhatdan: We can mention in buildah and kpod that docker login works, but don't need to specify it directly.  Since those tools are not as tied into the Docker infrastructure as skopeo is.

docs/skopeo.1.md
```diff
@@ -64,6 +64,11 @@ Uses the system's trust policy to validate images, rejects images not trusted by
 
   _destination-image_ use the "image name" format described above
 
+  **--authfile** _path_
+
+  Path of the authentication file. Default is ${XDG_RUNTIME_DIR}/containers/auth.json
+  which is set using `kpod login` or `docker login`
```
mtrmac: This seems to say `docker login` writes to `auth.json`, which is not true. (Also in two other instances.)

mtrmac:
> @mtrmac would you prefer the authfile flag be `flagPrefix+"authfile"` as well?

No, a single flag as you have implemented is better. The `authFile` can already carry configurations for multiple registries, so it doesn’t seem useful to support having a different one for sources and destinations.

docs/skopeo.1.md
TomSweeneyRedHat: If we have this in kpod and/or Buildah, can we touch it up there too please?

umohnani8: fixed.
@TomSweeneyRedHat there is no mention of docker login in buildah or kpod. Is it something we want to add?

umohnani8:
@mtrmac @runcom PTAL

mtrmac:
LGTM

runcom:
LGTM

rhatdan:
@umohnani8 Lets add something to kpod and buildah that mentions docker login.

docs/skopeo.1.md
```diff
@@ -19,7 +19,7 @@ Most commands refer to container images, using a _transport_`:`_details_ format.
   An existing local directory _path_ storing the manifest, layer tarballs and signatures as individual files. This is a non-standardized format, primarily useful for debugging or noninvasive container inspection.
 
   **docker://**_docker-reference_
-  An image in a registry implementing the "Docker Registry HTTP API V2". By default, uses the authorization state in `$HOME/.docker/config.json`, which is set e.g. using `(docker login)`.
+  An image in a registry implementing the "Docker Registry HTTP API V2". By default, uses the authorization state in either `$XDG_RUNTIME_DIR/containers/auth.json`, which is set using `(kpod login)` or in `$HOME/.docker/config.json`, which is set using `(docker login)`
```
TomSweeneyRedHat: 👍 
@umohnani8 I do think this should probably be added to kpod/buildah somewhere.  We're currently searching thorough containers/auth.json and then .docker/config.json in both of those now right?  If not, then this should be added when we are.

TomSweeneyRedHat: I think we probably ought to note precedence here too.  Perhaps:  "By default, uses the authorization state in $XDG_RUNTIME_DIR/containers/auth.json`, which is set using `(kpod login)`.  If the authorization state is not found there, then `$HOME/.docker/config.json` is checked which is set using `(docker login)`."

FWIW, missing a period at the end of your current line here.

docs/skopeo.1.md
```diff
@@ -64,6 +64,11 @@ Uses the system's trust policy to validate images, rejects images not trusted by
 
   _destination-image_ use the "image name" format described above
 
+  **--authfile** _path_
+
+  Path of the authentication file. Default is ${XDG_RUNTIME\_DIR}/containers/auth.json, which is set using `kpod login`
```
TomSweeneyRedHat: Missing period after kpod login.  A mention of precedence would probably be good here too.  Perhaps: "Default is ${XDG_RUNTIME\_DIR}/containers/auth.json, which is set using `kpod login`.  If the authorization state is not found there, then $HOME/.docker/config.json is checked, which is set using `docker login`."

docs/skopeo.1.md
umohnani8: fixed.

docs/skopeo.1.md
umohnani8: fixed

umohnani8:
Tests passed

PR #468, opened 2018-01-10
Re-vendor, notably opencontainers/image-spec to fix tests

mtrmac:
Anyone running `(vndr)` currently ends up with failing tests in OCI schema validation because `gojsonschema` has fixed its `"$ref"` interpretation, exposing inconsistent URI usage inside `image-spec/schema`.

So, this runs `(vndr)`, and uses `mtrmac/image-spec:id-based-loader` ( https://github.com/opencontainers/image-spec/pull/739 ) to make the tests pass again.  As soon as that PR is merged we should revert to using the upstream image-spec repo again.


mtrmac:
(Alternatively, we could leave the repo as is, and just keep telling people to run `vndr $your_repo` instead of `vndr`, giving up on rebasing the other libraries as well. I don’t have a strong preference.)

mtrmac:
@runcom PTAL.  The fix has been merged upstream, so this is primarily a rebase of `opencontainers/image-spec`.

runcom:
LGTM

PR #480, opened 2018-03-03
DO NOT MERGE: Include legacy support in docker-archive: and update (skopeo layers)

mtrmac:
This vendors in https://github.com/containers/image/pull/370 and updates `skopeo layers` to match. (Well, mostly, because we don’t know the right `isConfig` value if the user gives us blob digests on the command line—but the destination `dir:` transport ignores the value anyway.)

Also shows that the tests from https://github.com/containers/image/pull/370 succeed.


nlewo:
Do you know if a new release is planned? 
Would this patch be sufficient to justify a new release :/?

runcom:
Yup, we can create a new release after this.

runcom:
This can be revendored and I can merge it

nlewo:
I just submitted the PR #485 where I included a @mtrmac patch and vendor has been `revendorized`.
This PR could then be closed.

PR #491, opened 2018-04-07
Vendor in latest containers image

rhatdan:
We also want to add the containers/image/docs directory to the
vendor so that the skopeo rpm can pull the man pages for registries.conf
and policy.json directly from the repo.

Added new command to Makefile vendor-image, that will vendor in the latest
container/image, and maintain its docs directory.


rhatdan:
@runcom @mtrmac PTAL

Makefile
```diff
@@ -114,6 +114,9 @@ install-completions:
 shell: build-container
 	$(DOCKER_RUN_DOCKER) bash
 
+vendor-image:
+	vndr -whitelist '^github.com/containers/image/docs/.*' github.com/containers/image
```
mtrmac: Overall, `README.md` needs updating as well.

I’m worried having a special command for vendoring c/image (as opposed to anything else) will get overlooked, though.

Maybe
```make
vendor: vendor.conf
	vndr -whitelist '^github.com/containers/image/docs/.*'
```
and then document `make vendor`?  OTOH PRs frequently update only the c/image subpackage (people understandably don’t want to be on the hook for changes in other packages they don’t care for).

I don’t know what’s the usual practice.

rhatdan: How about we add a test that will fail if the docs are not there and report an error explaining that you should execute make vendor if you are changing containers/image.

rhatdan: I updated the README.md, also added test to check for docs.

rhatdan:
@runcom @mtrmac I added changes to the README and a test which will fail if some one vendors in containers/image without using the special Makefile command.
PTAL

mtrmac:
ACK.

I’m not _thrilled_ about treating c/image specially here, but it does work well enough.

Please drop `hack/vendor.sh` (it would be another place to update, and we don’t use it nor AFAICT need it for anything) and rebase.

rhatdan:
@mtrmac Fixed the way you wanted.
@runcom PTAL

rhatdan:
Shoot I just blew my patch away.

PR #497, opened 2018-04-29
Add skopeo rpm spec file to contrib

pixdrift:
Adding skopeo spec file to repo to provide public access to files required to build rpm. Location in repo follows buildah's existing convention/location.

Spec file is from the srpm from the following repo with the _commitid_ changed to REPLACEWITHCOMMITID to match buildah .spec convention:
https://cbs.centos.org/repos/virt7-container-common-candidate/source/SRPMS/

Please advise if there is an updated version available or if there is an alternate location to access the spec.

Signed-off-by: pixdrift <support@pixeldrift.net>


rhatdan:
@lsm5 @baude PTAL
Should we setup the autobuild like we do for buildah and podman?

pixdrift:
Is there a possibility of getting the same path/file convention for spec files used in all projectatomic git repos. Not necessarily `this` structure, but one that is the same for all.

I checked libpod which has it under `/contrib/spec/podman.spec.in`. I didn't want to push my luck and suggest a move in that repo too 😄

rhatdan:
I think it would be a good idea to get them all the same basic path.

lsm5:
@rhatdan RE: the daily autobuild for fedora rawhide, it happens on my daily-rebuild repo. See: https://travis-ci.org/lsm5/daily-rebuild/builds . It uses the fedora dist-git repos directly and doesn't depend on spec files included upstream. skopeo is included in the current daily rebuilds already. In case you need this spec file as a general pointer to `ohh how do i quickly build this?`, then LGTM.

I'll leave it to @baude to decide if the COPR build stuff he does for podman should be done here as well.

mtrmac:
I’m not _fundamentally_ against having a spec file in the repo (and if the buildah/podman/… universe wants to standardize on a convention (@baude?), fine); my primary concern is: will anyone maintain it, or at least notice that it is broken?

What do you actually need to _do_ with the spec file? It’s not quite clear to me from
>  to provide public access to files required to build rpm

who would benefit and how. The Fedora dist-git is just as public, for example.

Fedora is pretty clear in https://fedoraproject.org/wiki/Packaging:Guidelines#Spec_Maintenance_and_Canonicity that Fedora’s spec files are maintained in Fedora’s git; if they get out of sync, the Fedora git version is the canonical one.

This is not just academic: every time Fedora’s git bumps the `Release:` version, the spec in this repo should bump it as well, possibly to a higher number, so that the locally-built RPM is actually used. So, a Fedora-targeted spec file in here would be a hassle to maintain.

---

Another aspect is that the Fedora `skopeo` SRPM contains also the `skopeo-containers` subpackage, which does not really have much to do with skopeo itself — e.g. note the `Source1: storage.conf` source file.  If the SRPM from this repo is supposed to be drop-in-compatible, we would need to maintain copies of those fairly unrelated files here as well.

---

> @rhatdan RE: the daily autobuild for fedora rawhide, it happens on my daily-rebuild repo. See: https://travis-ci.org/lsm5/daily-rebuild/builds . It uses the fedora dist-git repos directly and doesn't depend on spec files included upstream.

Yeah, using the Fedora spec directly seems, at least for Fedora, a better approach.

_Maybe_ we should just mention in `README.md` that various distributions, like Fedora, maintain a `skopeo` package and building from source may be unnecessary? I don’t know, overall I am probably missing the point of this PR.

contrib/rpm/skopeo.spec
```diff
@@ -0,0 +1,475 @@
...
+%endif # centos
+Version:        0.1.27
+Release:        1.git%{shortcommit0}%{?dist}
+Summary:        Inspect Docker images and repositories on registries
+License:        ASL 2.0
+URL:            %{git0}
+Source0:        %{git0}/archive/%{commit0}/%{name}-%{shortcommit0}.tar.gz
+Source1:        storage.conf
+Source2:        containers-storage.conf.5.md
+Source3:        mounts.conf
```
mtrmac: Just to reemphasize, without having these config `Source*` files available, the build will fail.

lsm5:
On Mon, Apr 30, 2018 at 10:13:06AM -0700, Miloslav Trmač wrote:
> Another aspect is that the Fedora `skopeo` SRPM contains also the `skopeo-containers` subpackage, which does not really have much to do with skopeo itself — e.g. note the `Source1: storage.conf` source file.  If the SRPM from this repo is supposed to be drop-in-compatible, we would need to maintain copies of those fairly unrelated files here as well.

This file can be upstramed, I hope.

-- 
Lokesh
IRC: lsm5
GPG: 0xC7C3A0DD
https://keybase.io/lsm5

contrib/rpm/skopeo.spec
pixdrift: That's similar to how buildah is now. Building the tarball from the repo is fine, and as long as the conf files are available in the repo somewhere this can be easily overcome (although, I believe mounts.conf currently isn't in the repo).

If the file is in the repo here, CI jobs can be added to validate the spec is up to date / in sync with the version details etc. when the repo versions change.

As a consumer outside Red Hat, the biggest issue I have is finding a canonical source for the spec files that Red Hat uses for its build. I can download the SRPM after the fact, but I have no mechanism or opportunity to provide feedback (or pull requests) if there are issues in the spec files that Red Hat is producing (other than bugzilla).

I am not suggesting this has to be solved here, but my view is it's practical to store it next to the source that it packages.

mtrmac: > Building the tarball from the repo is fine, and as long as the conf files are available in the repo somewhere this can be easily overcome (although, I believe mounts.conf currently isn't in the repo)

None of the three are in _this_ repo.  If you are fine with getting these files from some repo different from this one, why not get the spec file from that same repo as well?

---

> As a consumer outside Red Hat, the biggest issue I have is finding a canonical source for the spec files that Red Hat uses for its build.

I’m not sure we can solve that here — **generally**, RHEL spec files are maintained internally and only published in SRPMs, or CentOS, after the relevant RHEL release is made public.  That is, AFAICT, not likely to change soon (the rationale I have quoted for making the Fedora repo, and not upstream repos, canonical for Fedora applies just as well for RHEL), so having _some_ spec file in this repo would not give you any more visibility to the internal RHEL repo before a release.

And because the RHEL spec files are not public before a release, a CI run in public GitHub cannot sync with them either.

>  I can download the SRPM after the fact, but I have no mechanism or opportunity to provide feedback (or pull requests) if there are issues in the spec files that Red Hat is producing (other than bugzilla).

Bugzilla is the _right_ place for RHEL feedback and contributions, in general.

---

Given the _general_ situation described above, there might perhaps be something specifically different about `skopeo`, due to Atomic Host and frequent RHEL Atomic releases and the like, which might give you easier access to WIP work, or something close to it, for `skopeo`.

But that would happen in the packaging efforts. Maybe @lsm5 can suggest something.

pixdrift: Thanks for the response @mtrmac.

These files are currently available in this repo under the contrib directory
`storage.conf`
`containers-storage.conf.5.md`

It's only mounts.conf that is currently missing.

I am not really fine from getting them elsewhere, it would be nice to be able to understand where they are coming from, and once again be able to have a source I can provide feedback/PRs to. If I do have to get these files from somewhere else, i'd ask what the process is for determining that 'somewhere else'?

This is a more general issue with Red Hat making spec files available for public comment/commit, and I should probably raise it back through Red Hat channels.

rhatdan:
mounts.conf and registries.conf are distro centric.  So we should probably provide empty templates.

rhatdan:
I have been thinking we should rename skopeo-containers to something like containers-config or something along those lines.  Since that is really what the package contains.  mounts.

rpm -q  skopeo-containers -l
/etc/containers
/etc/containers/policy.json
/etc/containers/registries.d
/etc/containers/registries.d/default.yaml
/etc/containers/storage.conf
/usr/share/containers
/usr/share/containers/mounts.conf
/usr/share/man/man5/containers-storage.conf.5.gz
/usr/share/rhel/secrets
/usr/share/rhel/secrets/etc-pki-entitlement
/usr/share/rhel/secrets/rhel7.repo
/usr/share/rhel/secrets/rhsm
/var/lib/atomic/sigstore

pixdrift:
@rhatdan, as an educational exercise for myself, what is the current source of the `mounts.conf` that is referenced in the skopeo .spec file? ie. where can I go to find the source of `mounts.conf` that is used in the RH shipped rpm?

pixdrift:
Is it worth mentioning here the `registries.conf` file from atomic-registries moving into skopeo-containers? which was discussed here:
https://github.com/projectatomic/buildah/issues/635

Would that involve moving the `registries.conf` into this repo? or would that file be included in the skopeo-container rpm spec file similar to mounts.conf?

pixdrift:
Added the `mounts.conf` file to contrib with the other files described in the .spec so this file is available if this PR is accepted. Will need to update to a more generic version that isn't distro specific as described by @rhatdan above.

rhatdan:
Currently mounts.conf and registries.conf is stored with the the spec file in the internal koji repo.  We could setup different contrib files under skopeo for each distribution.  Most of the Red Hat ones are just sorting the order of registries.conf and we have the RHEL secrets in fedora and Centos, since there is nothing stopping a user from using a RHEL subscription to develop RHEL containers while sitting on a Fedora or Centos Box.

Making these more public would probably be a good idea.

storage.conf is also customizable by the distros although I think there would be more concensus there.

What do you think about changing the subpackage name to containers-config.

contrib/mounts.conf
```diff
@@ -0,0 +1 @@
+/usr/share/rhel/secrets:/run/secrets
```
rhatdan: Lets create a subdir called RHEL and put this file there and then create an empty one for others to use.

pixdrift: I have moved this mounts.conf into rhel. If you can point me to other distro examples happy to include them in this PR.

pixdrift:
Using 'common' is a fairly standard convention in other RHEL packages, could `containers-common` work as a package name which also then doesn't limit it to configuration files/data? This may mask the intent if it is just for config though 😄

pixdrift:
@rhatdan, where to from here? is the plan to put `containers-common` in a new projectatomic repo with its own spec? If so, I can update this PR to remove the files that will be moved into `containers-common`.

pixdrift:
So the rename is fine and the spec builds cleanly, but the impact of the package name change from `skopeo-containers` to `containers-common` is far and wide.

So far I have identified the following packages that have `Requires` on `skopeo-containers`
```
buildah
podman
docker-common (which was a surprise)
```

How is this kind of effort usually co-ordinated?

rhatdan:
You should add a 
provides: skopeo-containers

Then we can slowly fix the other packages to require the new name.

pixdrift:
@rhatdan Cheers, thanks for the tip.

pixdrift:
Upgrade looks good/clean, running into this strange issue updating skopeo though:
```
Error: Package: 1:skopeo-0.1.27-1.git7159f56.el7.x86_64 (/skopeo-0.1.27-1.git7159f56.el7.x86_64)
           Requires: containers-common = 0.1.27-1.git7159f56.el7
           Installed: 1:containers-common-0.1.27-1.git7159f56.el7.x86_64 (installed)
               containers-common = 1:0.1.27-1.git7159f56.el7
```

pixdrift:
It's already in there:
https://github.com/pixdrift/skopeo/blob/28786e9c592f637c2740fa15b5a8bea68e09d16a/contrib/rpm/skopeo.spec#L184

pixdrift:
CI failure looks unrelated.

pixdrift:
Resulting packages built on CentOS 7.4 are here:
http://rpmbuild.pixeldrift.net/scratch/

Update was clean from the packages versions in CentOS Extras. If anyone can test the spec on Fedora would be appreciated. Any other feedback welcome.

rhatdan:
@lsm5 PTAL

contrib/rpm/skopeo.spec
lsm5: btw, any CentOS mentions in the spec file are RE: CentOS Virt SIG (rebuilt from rawhide) and not CentOS Base (rebuilt from RHEL).

@pixdrift so, the major change in the spec file here is renaming skopeo-containers to containers-common, yes?
While people are using the `spec_in_upstream` approach to do things like automatic COPR builds, this spec file would often end up being stale compared to the canonical dist-git spec. So, as long as you're ok with keeping this spec updated and upstream is ok with frequent PRs to it, fine by me. Issues on RHEL/CentOS packages are best reported via bugzilla like @mtrmac mentioned.

If you choose to send PRs to this repo for spec file changes, please make sure to rebase the spec file on what's in the srpm. Also, if you're a user of CentOS Virt SIG, it would be best to send PRs to the fedora rawhide dist-git for skopeo, since that's where CentOS Virt SIG is built from. HTH.

pixdrift: @lsm5, what's the method I should use to get access to the canonical source and the files required to build the rpm? That is what I was hoping to solve.

The package name change is the only spec change (original intention was just to include it and the missing source files). Not sure how to raise/co-ordinate the package name upstream, if bugzilla is the method I can use that.

pixdrift: @lsm5, it does help, thanks.

It feels as if the source for the rpms is 'locked away' in Red Hat, and the only way for an outsider to access them outside of Red Hat is to wait for Red Hat to produce a product.

It's a much bigger issue than this repo, i'll admit, and I appreciate that packaging is in the vendor space... but I don't think it's crazy for the project to provide / supply their own rpm spec file (or dpkg sources for that matter) as part of the project so an rpm can be built independent of any vendor.

Downloading the src rpm works but it's at the mercy of a vendor/supplier release schedule, the rpm spec from the vendor also doesn't describe where the components used to create the rpm originated from.

Rather than the rpm fall behind and need to be kept in sync with upstream, an alternative view is that it is kept in sync with the repo packaging through transparent (ie. visible) CI processes and can be included/involved in the validation process of the pipeline when version numbers / files change in the repo.

mtrmac: > While people are using the spec_in_upstream approach to do things like automatic COPR builds, this spec file would often end up being stale compared to the canonical dist-git spec. So, as long as you're ok with keeping this spec updated and upstream is ok with frequent PRs to it, fine by me.

I’m fine with keeping my hands off and letting you two, or whoever is relevant, maintain the spec file in here, with as little friction as possible (e.g. @lsm5 merging PRs without my involvement), _as long as there is consensus on the benefit_.

But, given
> Also, if you're a user of CentOS Virt SIG, it would be best to send PRs to the fedora rawhide dist-git for skopeo, since that's where CentOS Virt SIG is built from.

mtrmac: and
> for changes to rhel/centos base, file a bug and add your patch if any, and wait for package maintainers to ship builds with that fix.

vs.
> what's the method I should use to get access to the canonical source and the files required to build the rpm?

it is still not clear to me that this is actually solving anything: If I understand correctly, @pixdrift wants to be involved in the spec file changes during RHEL development, but this PR would not change anything about the status quo, which is that RHEL specs are only revealed after a public release of a supported package.

And the Fedora and CentOS packages are also maintained elsewhere, and @lsm5 wants to accept PRs there, not here.

So, is maintaining an _extra_ copy, which is (if I am reading the discussion correctly) _still not_ the primary place of RHEL development, making anyone’s life easier?

pixdrift:
All good. Thanks for the feedback/advice re: source of specs, I will close this out.

contrib/rpm/skopeo.spec
lsm5: @mtrmac yup, you're right. Having a spec here would mostly just be an added cost unless it's being used for something like COPR automation.

PR #499, opened 2018-05-01
Stop using docker.Image

mtrmac:
Instead, use `DockerReference()` to obtain the repository name (which also makes it work for other transports that support Docker references), and a check for `docker.Transport` + the newly introduced `docker.GetRepositoryTags`.

This will allow dropping `docker.Image` from `containers/image`, and maybe even all of `ImageReference.NewImage` (forcing callers to think about manifest lists, among other things).

@runcom PTAL (non-urgent), primarly WDYT about eventually dropping the `NewImage` interface and forcing callers to deal with `ImageSource` and `image.FromSource` / `FromUnparsedImage` themselves? Or maybe wrap that in something like `c/image/image.FromReference(types.ImageReference)` so that callers continue to need only one call instead of two?

### DIFF
--- a/cmd/skopeo/inspect.go
+++ b/cmd/skopeo/inspect.go
@@ -96,10 +96,10 @@ var inspectCmd = cli.Command{
 			return err
 		}
 		outputData := inspectOutput{
-			Name: "", // Possibly overridden for a docker.Image.
+			Name: "", // Set below if DockerReference() is known
 			Tag:  imgInspect.Tag,
 			// Digest is set below.
-			RepoTags:      []string{}, // Possibly overriden for a docker.Image.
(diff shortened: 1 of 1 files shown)


runcom:
> @runcom PTAL (non-urgent), primarly WDYT about eventually dropping the NewImage interface and forcing callers to deal with ImageSource and image.FromSource / FromUnparsedImage themselves? Or maybe wrap that in something like c/image/image.FromReference(types.ImageReference) so that callers continue to need only one call instead of two?

I'm all for removing `NewImage` and have people use the low-level apis (as that's really what I wanted at the beginning of c/image)

mtrmac:
Thanks; merging this at least, we can then drop `docker.Image` and consider the fate of `NewImage` in c/image PRs.

PR #501, opened 2018-05-08
skopeo-copy: docker-archive: multitag support

vrothberg:
Add multitag support when generating docker-archive tarballs via the
newly added '--aditional-tag' option, which can be specified multiple
times to add more than one tag.  All specified tags will be added to the
RepoTags field in the docker-archive's manifest.json file.

This change requires to vendor the latest containers/image with
commit a1a9391830fd08637edbe45133fd0a8a2682ae75.

Signed-off-by: Valentin Rothberg <vrothberg@suse.com>

### DIFF
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/containers/image/copy"
+	"github.com/containers/image/docker/reference"
 	"github.com/containers/image/manifest"
 	"github.com/containers/image/transports"
 	"github.com/containers/image/transports/alltransports"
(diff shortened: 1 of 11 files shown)


mtrmac:
Actually, you’re right, this was directly caused by the F28 update.  Should be fixed by https://github.com/projectatomic/skopeo/pull/503 (per tests in https://github.com/projectatomic/skopeo/pull/502 ).

Please rebase this PR on top of `master`, then the tests should pass.

vrothberg:
@mtrmac Awesome, thanks for fixing it! I just rebased the change.

mtrmac:
Thanks again!

vrothberg:
@mtrmac @runcom Could we prepare a new release of skopeo including this change? We need that for the Open Build Service, and I'd prefer to bump our package to a new version rather than backporting the change.

PR #503, opened 2018-05-10
Run (go vet) on all subpackages instead only of changed files

mtrmac:
Apparently, it was never documented to use `go vet $somefile.go` (but `go tool vet $somefile.go` was).

go 1.10 seems to do more checks within packages, and `$somefile.go` is interpreted as a package with only that file (even if other files from that package are in the same directory), leading to spurious `"undefined: $symbol"` errors.

So, just run `go vet` on `./...` (explicitly excluding `skopeo/vendor` for the benefit of Go 1.8). We only have three subpackages, so the savings, if any, from running `go vet` only on the modified subpackages would be small.

More importantly, on a toolchain update, `./...` allows us to see the newly detected issues all at once, instead of randomly waiting for a commit that changes one of the affected files for the failure to show up.

### DIFF
--- a/hack/make/validate-vet
+++ b/hack/make/validate-vet
@@ -1,28 +1,13 @@
 #!/bin/bash
 
-source "$(dirname "$BASH_SOURCE")/.validate"
+errors=$(go vet $(go list -e ./... | grep -v "$SKOPEO_PKG"/vendor))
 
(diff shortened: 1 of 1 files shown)


runcom:
:+1:

PR #504, opened 2018-05-12
Improve installation documentation a bit

mtrmac:
- _Start_ with installing distribution packages, instead of mentioning it after the user has already built everything from source.
- Note that both the binary and documentation needs to be built for `make install` to work.


rhatdan:
LGTM

rhatdan:
@runcom @TomSweeneyRedHat PTAL

PR #507, opened 2018-05-19
Include vendor/github.com/containers/image/docs

mtrmac:
@rhatdan RFC. This is an alternative to #491 which forces users who update dependencies to update a few more of them instead of their single package — OTOH that means that c/storage will be generally kept up-to-date without needing separate PRs.

(Does not include the discussed test to ensure that the documentation is never removed; that might still be needed.)


mtrmac:
(For the record the original test is preserved in https://github.com/projectatomic/libpod/pull/672#issuecomment-384363722 ).

rhatdan:
LGTM

PR #514, opened 2018-05-30
vendor: update containers/image

giuseppe:
Needed to pick up this change:

ostree: use the same thread for ostree operations

Since https://github.com/ostreedev/ostree/pull/1555, locking is
enabled by default in OSTree.  Unfortunately it uses thread-private
data and it breaks the Golang bindings.  Force the same thread for the
write operations to the OSTree repository.

### DIFF
--- a/vendor/github.com/containers/image/ostree/ostree_dest.go
+++ b/vendor/github.com/containers/image/ostree/ostree_dest.go
@@ -14,6 +14,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
 	"syscall"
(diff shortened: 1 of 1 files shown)


giuseppe:
Once this lands, can we get updated packages as well?

vendor/github.com/containers/image/docs/atomic-signature-embedded-json.json
```diff
@@ -1,66 +0,0 @@
-{
```
mtrmac: Please run `make vendor` instead of running `vndr` manually.

giuseppe: thanks, I didn't know about it.  Pushed an updated version

rhatdan:
@giuseppe Once this is merged do we need a new version of skopeo shipped?

giuseppe:
@rhatdan yes we need a new version of Skopeo, otherwise the ostree storage is completely broken with the latest ostree version

rhatdan:
@runcom could you create a new version and I will get it built into Fedora.

PR #519, opened 2018-06-29
cmd/skopeo: show full help on 'copy' with no args

vbatts:
Signed-off-by: Vincent Batts <vbatts@hashbangbash.com>

### DIFF
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"strings"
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

runcom:
LGTM

cmd/skopeo/copy.go
```diff
@@ -34,7 +33,7 @@ func contextsFromGlobalOptions(c *cli.Context) (*types.SystemContext, *types.Sys
 
 func copyHandler(c *cli.Context) error {
 	if len(c.Args()) != 2 {
-		return errors.New("Usage: copy source destination")
+		return cli.ShowCommandHelp(c, "copy")
```
mtrmac: This will 1) send the output to `stdout`, and nothing to `stderr`, and 2) exit with return code 0. Both seem undesirable (OTOH 1) is the general behavior of `urfave/cli` everywhere).

Arguably a plain `(skopeo copy)` won’t get into any script that ever worked, so that should not matter — but a `skopeo copy $src $dest`, with `$src` and `$dest` for some other reason empty, might.

vbatts: Newer master of urfave/cli has a further function (...AndExit) where you can provide an exit code. As for writing to stderr, there might be way but needs investigation

mtrmac: https://github.com/projectatomic/skopeo/pull/520 : print help on `stdout`, end with a fatal error to `stderr`, and return a non-zero exit status.

PR #522, opened 2018-07-06
Makefile: add binary-minimal and binary-static-minimal

AkihiroSuda:
These targets produce a pure-Go binary, without the following features:
* ostree
* devicemapper
* btrfs
* gpgme


mtrmac:
Thanks!

Rather than add 4! duplicate build targets which are likely to be rarely used, and therefore broken from time to time, could this instead be chosen by a `make` variable, so that maybe `make build-local DISABLE_CGO=1` sets the appropriate values of `BUILDTAGS`, and the existing rules continue to be used?

AkihiroSuda:
updated

mtrmac:
Works for me.  @runcom WDYT?

runcom:
works for me as well :)

PR #523, opened 2018-07-12
RFC: Reliable CLI parsing

mtrmac:
@runcom , others: RFC

While testing a related change I have noticed that `skopeo layers` sets `DockerDaemonInsecureSkipTLSVerify` always to `true`, because the `--tls-verify` option is not declared for `skopeo layers`; and the way `urfave/cli` is used, there was no connection between the option declaration and reading the value, so it’s way too easy to omit one or the other.  (In particular, `cli.BoolT` can return `false` if the option is not found, or on any other error, really, not only when the user has explicitly provided a `false` value.)

So, this motivated a comprehensive rework of the CLI data management; instead of copy&pasting strings and hoping to never make a typo (which is just not good enough for security options), create a `struct someCommandOptions` holding well-typed argument values, and set up the `cli.Flag` structures to write data in there; command handlers then use the values directly instead of using string lookup via `*cli.Context`.  See the ‘Create an "options" structure for each command’ commit message for a detailed writeup of the structure change.

Given the explicit data structures, we can then fully model an `optionalBool` and `optionalString` argument type, and actually completely distinguish between a missing and default-valued argument values (well, as far as we can get without a native Go “optional” type mechanism).

(WIP because I need to finish and clean up the tests.)

---

Finally, this allows explicitly modeling the shared image-related options `--src-*`/`--dest-*` for `copy`, or `--*` for other commands, and expressing the relationships as explicit data structure pointers. That, however, made it much more visible that `--authfile` (introduced in #453 , and I’m afraid this design is pretty much my fault) is an irregular case: it is shared across the sources/destinations, but it is not a global flag; that forces all users to manage the scope of the flag explicitly, see how `sharedOpts` needs to be added everywhere. @umohnani8 @rhatdan @TomSweeneyRedHat Would it be reasonable to make `--authfile` a global flag, _without_ backward compatibility (i.e. `skopeo --authfile=… copy` instead of `skopeo copy --authfile=…`)? I guess not, but I thought I’d at least ask.

### DIFF
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -194,7 +194,7 @@ func (opts *copyOptions) run(c *cli.Context) error {
 		destinationCtx.DockerArchiveAdditionalTags = append(destinationCtx.DockerArchiveAdditionalTags, namedTagged)
 	}
 
-	ctx, cancel := commandTimeoutContextFromGlobalOptions(c)
+	ctx, cancel := opts.global.commandTimeoutContext()
 	defer cancel()
 
 	_, err = copy.Image(ctx, policyContext, destRef, srcRef, &copy.Options{
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -32,6 +32,10 @@ func contextsFromGlobalOptions(c *cli.Context) (*types.SystemContext, *types.Sys
 }
 
 type copyOptions struct {
+	additionalTags    cli.StringSlice // For docker-archive: destinations, in addition to the name:tag specified as destination, also add these
+	removeSignatures  bool            // Do not copy signatures from the source image
+	signByFingerprint string          // Sign the image using a GPG key with the specified fingerprint
+	format            optionalString  // Force conversion of the image to a specified format
 }
 
 func copyCmd() cli.Command {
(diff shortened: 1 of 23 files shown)


rhatdan:
Already out of date.

@giuseppe PTAL

giuseppe:
LGTM

Also is there any user of `skopeo layers`?  IIRC it was already deprecated some time ago, and atomic that was using it at the time moved as well to `skopeo copy`.  Would it make sense to drop `skopeo layers` completely?

runcom:
I'd love to drop skopeo layers completely, there have been plenty of time to notice it's deprecated.

rhatdan:
Can we drop it from the help/man pages first and see if anyone notices?

mtrmac:
This is now finished, including comprehensive tests for `--tls-verify`, the original impetus.  Any more comments before I merge it? Opinions on moving `--authfile`?

rhatdan:
SGTM

rhatdan:
@mtrmac What do you want to do with this PR?

mtrmac:
> @mtrmac What do you want to do with this PR?

@rhatdan Get it merged, ideally.  The discussion of the `--authfile` change above was a bit unclear to me and I failed to follow up.

Though, I guess, on balance, I’m willing to live with the code ugliness in order not to break `--authfile`.

Let me rebase this some time…

rhatdan:
@runcom Lets get this merged.

rhatdan:
@mtrmac Rebase and I will merge.

mtrmac:
@rhatdan rebased.

rhatdan:
LGTM

PR #524, opened 2018-07-13
Introduce the sync command

flavio:
The skopeo sync command can sync images between a SOURCE and a destination.

The purpose of this command is to assist with the mirroring of container images from different docker registries to a single docker registry.

Right now the following transport matrix is implemented:

  * `docker://` -> `docker://`
  * `docker://` -> `dir:`
  * `dir:` -> `docker://`

The `dir:` transport is supported to handle the use case of air-gapped environments.
In this context users can perform an initial sync on a trusted machine connected to the internet; that would be a `docker://` -> `dir:` sync.
The target directory can be copied to a removable drive that can then be plugged into a node of the air-gapped environment. From there a `dir:` -> `docker://` sync will import all the images into the registry serving the air-gapped environment.

The image namespace is changed during the  `docker://` to `docker://` or `dir:` copy. The FQDN of the registry hosting the image will be added as new root namespace of the image. For example, the image `registry.example.com/busybox:latest` will be copied to `registry.local.lan/registry.example.com/busybox:latest`.

The image namespace is not changed when doing a
`dir:` -> `docker://` sync operation.

The alteration of the image namespace is used to nicely scope images coming from different registries (the Docker Hub, quay.io, gcr, other registries). That allows all of them to be hosted on the same registry without incurring in clashes and making their origin explicit.

## TODO

I hope you like this feature and the direction it's going. Once we agree on its final design we will update this PR to extend the current test suites.

## Future work

Currently sync will keep adding missing content from `SOURCE` to `DESTINATION`. It would be nice to add a `--delete` flag to remove from `DESTINATION` contents that are no longer available inside of `SOURCE. That would be a bit like rsync's `--delete` flag.

If wanted, that should be addressed with a separate PR.

Signed-off-by: Flavio Castelli <fcastelli@suse.com>
Co-authored-by: Marco Vedovati <mvedovati@suse.com>

### DIFF
--- a/cmd/skopeo/sync.go
+++ b/cmd/skopeo/sync.go
@@ -1,0 +1,549 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+
+	"github.com/containers/image/v5/copy"
+	"github.com/containers/image/v5/directory"
+	"github.com/containers/image/v5/docker"
+	"github.com/containers/image/v5/docker/reference"
+	"github.com/containers/image/v5/transports"
+	"github.com/containers/image/v5/types"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"github.com/urfave/cli"
+	"gopkg.in/yaml.v2"
+)
+
+// syncOptions contains information retrieved from the skopeo sync command line.
+type syncOptions struct {
+	global            *globalOptions    // Global (not command dependant) skopeo options
+	srcImage          *imageOptions     // Source image options
+	destImage         *imageDestOptions // Destination image options
+	removeSignatures  bool              // Do not copy signatures from the source image
+	signByFingerprint string            // Sign the image using a GPG key with the specified fingerprint
+	source            string            // Source repository name
+	destination       string            // Destination registry name
+	scoped            bool              // When true, namespace copied images at destination using the source repository name
+}
+
+// repoDescriptor contains information of a single repository used as a sync source.
+type repoDescriptor struct {
+	DirBasePath  string                 // base path when source is 'dir'
+	TaggedImages []types.ImageReference // List of tagged image found for the repository
+	Context      *types.SystemContext   // SystemContext for the sync command
+}
+
+// tlsVerify is an implementation of the Unmarshaler interface, used to
+// customize the unmarshaling behaviour of the tls-verify YAML key.
+type tlsVerifyConfig struct {
+	skip types.OptionalBool // skip TLS verification check (false by default)
+}
+
+// registrySyncConfig contains information about a single registry, read from
+// the source YAML file
+type registrySyncConfig struct {
+	Images      map[string][]string    // Images map images name to slices with the images' tags
+	Credentials types.DockerAuthConfig // Username and password used to authenticate with the registry
+	TLSVerify   tlsVerifyConfig        `yaml:"tls-verify"` // TLS verification mode (enabled by default)
+	CertDir     string                 `yaml:"cert-dir"`   // Path to the TLS certificates of the registry
+}
+
+// sourceConfig contains all registries information read from the source YAML file
+type sourceConfig map[string]registrySyncConfig
+
+func syncCmd(global *globalOptions) cli.Command {
+	sharedFlags, sharedOpts := sharedImageFlags()
+	srcFlags, srcOpts := dockerImageFlags(global, sharedOpts, "src-", "screds")
+	destFlags, destOpts := dockerImageFlags(global, sharedOpts, "dest-", "dcreds")
+
+	opts := syncOptions{
+		global:    global,
+		srcImage:  srcOpts,
+		destImage: &imageDestOptions{imageOptions: destOpts},
+	}
+
+	return cli.Command{
+		Name:  "sync",
+		Usage: "Synchronize one or more images from one location to another",
+		Description: fmt.Sprint(`
+
+	Copy all the images from a SOURCE to a DESTINATION.
+
+	Allowed SOURCE transports (specified with --src): docker, dir, yaml.
+	Allowed DESTINATION transports (specified with --dest): docker, dir.
+
+	See skopeo-sync(1) for details.
+	`),
+		ArgsUsage: "--src SOURCE-LOCATION --dest DESTINATION-LOCATION SOURCE DESTINATION",
+		Action:    commandAction(opts.run),
+		// FIXME: Do we need to namespace the GPG aspect?
+		Flags: append(append(append([]cli.Flag{
+			cli.BoolFlag{
+				Name:        "remove-signatures",
+				Usage:       "Do not copy signatures from SOURCE images",
+				Destination: &opts.removeSignatures,
+			},
+			cli.StringFlag{
+				Name:        "sign-by",
+				Usage:       "Sign the image using a GPG key with the specified `FINGERPRINT`",
+				Destination: &opts.signByFingerprint,
+			},
+			cli.StringFlag{
+				Name:        "src, s",
+				Usage:       "SOURCE transport type",
+				Destination: &opts.source,
+			},
+			cli.StringFlag{
+				Name:        "dest, d",
+				Usage:       "DESTINATION transport type",
+				Destination: &opts.destination,
+			},
+			cli.BoolFlag{
+				Name:        "scoped",
+				Usage:       "Images at DESTINATION are prefix using the full source image path as scope",
+				Destination: &opts.scoped,
+			},
+		}, sharedFlags...), srcFlags...), destFlags...),
+	}
+}
+
+// unmarshalYAML is the implementation of the Unmarshaler interface method
+// method for the tlsVerifyConfig type.
+// It unmarshals the 'tls-verify' YAML key so that, when they key is not
+// specified, tls verification is enforced.
+func (tls *tlsVerifyConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var verify bool
+	if err := unmarshal(&verify); err != nil {
+		return err
+	}
+
+	tls.skip = types.NewOptionalBool(!verify)
+	return nil
+}
+
+// newSourceConfig unmarshals the provided YAML file path to the sourceConfig type.
+// It returns a new unmarshaled sourceConfig object and any error encountered.
+func newSourceConfig(yamlFile string) (sourceConfig, error) {
+	var cfg sourceConfig
+	source, err := ioutil.ReadFile(yamlFile)
+	if err != nil {
+		return cfg, err
+	}
+	err = yaml.Unmarshal(source, &cfg)
+	if err != nil {
+		return cfg, errors.Wrapf(err, "Failed to unmarshal %q", yamlFile)
+	}
+	return cfg, nil
+}
+
+// destinationReference creates an image reference using the provided transport.
+// It returns a image reference to be used as destination of an image copy and
+// any error encountered.
+func destinationReference(destination string, transport string) (types.ImageReference, error) {
+	var imageTransport types.ImageTransport
+
+	switch transport {
+	case docker.Transport.Name():
+		destination = fmt.Sprintf("//%s", destination)
+		imageTransport = docker.Transport
+	case directory.Transport.Name():
+		_, err := os.Stat(destination)
+		if err == nil {
+			return nil, errors.Errorf(fmt.Sprintf("Refusing to overwrite destination directory %q", destination))
+		}
+		if !os.IsNotExist(err) {
+			return nil, errors.Wrap(err, "Destination directory could not be used")
+		}
+		// the directory holding the image must be created here
+		if err = os.MkdirAll(destination, 0755); err != nil {
+			return nil, errors.Wrapf(err, fmt.Sprintf("Error creating directory for image %s",
+				destination))
+		}
+		imageTransport = directory.Transport
+	default:
+		return nil, errors.Errorf("%q is not a valid destination transport", transport)
+	}
+	logrus.Debugf("Destination for transport %q: %s", transport, destination)
+
+	destRef, err := imageTransport.ParseReference(destination)
+	if err != nil {
+		return nil, errors.Wrapf(err, fmt.Sprintf("Cannot obtain a valid image reference for transport %q and reference %q", imageTransport.Name(), destination))
+	}
+
+	return destRef, nil
+}
+
+// getImageTags retrieves all the tags associated to an image hosted on a
+// container registry.
+// It returns a string slice of tags and any error encountered.
+func getImageTags(ctx context.Context, sysCtx *types.SystemContext, imgRef types.ImageReference) ([]string, error) {
+	name := imgRef.DockerReference().Name()
+	logrus.WithFields(logrus.Fields{
+		"image": name,
+	}).Info("Getting tags")
+	tags, err := docker.GetRepositoryTags(ctx, sysCtx, imgRef)
+
+	switch err := err.(type) {
+	case nil:
+		break
+	case docker.ErrUnauthorizedForCredentials:
+		// Some registries may decide to block the "list all tags" endpoint.
+		// Gracefully allow the sync to continue in this case.
+		logrus.Warnf("Registry disallows tag list retrieval: %s", err)
+		break
+	default:
+		return tags, errors.Wrapf(err, fmt.Sprintf("Error determining repository tags for image %s", name))
+	}
+
+	return tags, nil
+}
+
+// isTagSpecified checks if an image name includes a tag and returns any errors
+// encountered.
+func isTagSpecified(imageName string) (bool, error) {
+	normNamed, err := reference.ParseNormalizedNamed(imageName)
+	if err != nil {
+		return false, err
+	}
+
+	tagged := !reference.IsNameOnly(normNamed)
+	logrus.WithFields(logrus.Fields{
+		"imagename": imageName,
+		"tagged":    tagged,
+	}).Info("Tag presence check")
+	return tagged, nil
+}
+
+// imagesTopCopyFromRepo builds a list of image references from the tags
+// found in the source repository.
+// It returns an image reference slice with as many elements as the tags found
+// and any error encountered.
+func imagesToCopyFromRepo(repoReference types.ImageReference, repoName string, sourceCtx *types.SystemContext) ([]types.ImageReference, error) {
+	var sourceReferences []types.ImageReference
+	tags, err := getImageTags(context.Background(), sourceCtx, repoReference)
+	if err != nil {
+		return sourceReferences, err
+	}
+
+	for _, tag := range tags {
+		imageAndTag := fmt.Sprintf("%s:%s", repoName, tag)
+		ref, err := docker.ParseReference(imageAndTag)
+		if err != nil {
+			return nil, errors.Wrapf(err, fmt.Sprintf("Cannot obtain a valid image reference for transport %q and reference %q", docker.Transport.Name(), imageAndTag))
+		}
+		sourceReferences = append(sourceReferences, ref)
+	}
+	return sourceReferences, nil
+}
+
+// imagesTopCopyFromDir builds a list of image references from the images found
+// in the source directory.
+// It returns an image reference slice with as many elements as the images found
+// and any error encountered.
+func imagesToCopyFromDir(dirPath string) ([]types.ImageReference, error) {
+	var sourceReferences []types.ImageReference
+	err := filepath.Walk(dirPath, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if !info.IsDir() && info.Name() == "manifest.json" {
+			dirname := filepath.Dir(path)
+			ref, err := directory.Transport.ParseReference(dirname)
+			if err != nil {
+				return errors.Wrapf(err, fmt.Sprintf("Cannot obtain a valid image reference for transport %q and reference %q", directory.Transport.Name(), dirname))
+			}
+			sourceReferences = append(sourceReferences, ref)
+			return filepath.SkipDir
+		}
+		return nil
+	})
+
+	if err != nil {
+		return sourceReferences,
+			errors.Wrapf(err, fmt.Sprintf("Error walking the path %q", dirPath))
+	}
+
+	return sourceReferences, nil
+}
+
+// imagesTopCopyFromDir builds a list of repository descriptors from the images
+// in a registry configuration.
+// It returns a repository descriptors slice with as many elements as the images
+// found and any error encountered. Each element of the slice is a list of
+// tagged image references, to be used as sync source.
+func imagesToCopyFromRegistry(registryName string, cfg registrySyncConfig, sourceCtx types.SystemContext) ([]repoDescriptor, error) {
+	var repoDescList []repoDescriptor
+	for imageName, tags := range cfg.Images {
+		repoName := fmt.Sprintf("//%s", path.Join(registryName, imageName))
+		logrus.WithFields(logrus.Fields{
+			"repo":     imageName,
+			"registry": registryName,
+		}).Info("Processing repo")
+
+		serverCtx := &sourceCtx
+		// override ctx with per-registryName options
+		serverCtx.DockerCertPath = cfg.CertDir
+		serverCtx.DockerDaemonCertPath = cfg.CertDir
+		serverCtx.DockerDaemonInsecureSkipTLSVerify = (cfg.TLSVerify.skip == types.OptionalBoolTrue)
+		serverCtx.DockerInsecureSkipTLSVerify = cfg.TLSVerify.skip
+		serverCtx.DockerAuthConfig = &cfg.Credentials
+
+		var sourceReferences []types.ImageReference
+		for _, tag := range tags {
+			source := fmt.Sprintf("%s:%s", repoName, tag)
+
+			imageRef, err := docker.ParseReference(source)
+			if err != nil {
+				logrus.WithFields(logrus.Fields{
+					"tag": source,
+				}).Error("Error processing tag, skipping")
+				logrus.Errorf("Error getting image reference: %s", err)
+				continue
+			}
+			sourceReferences = append(sourceReferences, imageRef)
+		}
+
+		if len(tags) == 0 {
+			logrus.WithFields(logrus.Fields{
+				"repo":     imageName,
+				"registry": registryName,
+			}).Info("Querying registry for image tags")
+
+			imageRef, err := docker.ParseReference(repoName)
+			if err != nil {
+				logrus.WithFields(logrus.Fields{
+					"repo":     imageName,
+					"registry": registryName,
+				}).Error("Error processing repo, skipping")
+				logrus.Error(err)
+				continue
+			}
+
+			sourceReferences, err = imagesToCopyFromRepo(imageRef, repoName, serverCtx)
+			if err != nil {
+				logrus.WithFields(logrus.Fields{
+					"repo":     imageName,
+					"registry": registryName,
+				}).Error("Error processing repo, skipping")
+				logrus.Error(err)
+				continue
+			}
+		}
+
+		if len(sourceReferences) == 0 {
+			logrus.WithFields(logrus.Fields{
+				"repo":     imageName,
+				"registry": registryName,
+			}).Warnf("No tags to sync found")
+			continue
+		}
+		repoDescList = append(repoDescList, repoDescriptor{
+			TaggedImages: sourceReferences,
+			Context:      serverCtx})
+	}
+
+	return repoDescList, nil
+}
+
+// imagesToCopy retrieves all the images to copy from a specified sync source
+// and transport.
+// It returns a slice of repository descriptors, where each descriptor is a
+// list of tagged image references to be used as sync source, and any error
+// encountered.
+func imagesToCopy(source string, transport string, sourceCtx *types.SystemContext) ([]repoDescriptor, error) {
+	var descriptors []repoDescriptor
+
+	switch transport {
+	case docker.Transport.Name():
+		desc := repoDescriptor{
+			Context: sourceCtx,
+		}
+		refName := fmt.Sprintf("//%s", source)
+		srcRef, err := docker.ParseReference(refName)
+		if err != nil {
+			return nil, errors.Wrapf(err, fmt.Sprintf("Cannot obtain a valid image reference for transport %q and reference %q", docker.Transport.Name(), refName))
+		}
+		imageTagged, err := isTagSpecified(source)
+		if err != nil {
+			return descriptors, err
+		}
+
+		if imageTagged {
+			desc.TaggedImages = append(desc.TaggedImages, srcRef)
+			descriptors = append(descriptors, desc)
+			break
+		}
+
+		desc.TaggedImages, err = imagesToCopyFromRepo(
+			srcRef,
+			fmt.Sprintf("//%s", source),
+			sourceCtx)
+
+		if err != nil {
+			return descriptors, err
+		}
+		if len(desc.TaggedImages) == 0 {
+			return descriptors, errors.Errorf("No images to sync found in %q", source)
+		}
+		descriptors = append(descriptors, desc)
+
+	case directory.Transport.Name():
+		desc := repoDescriptor{
+			Context: sourceCtx,
+		}
+
+		if _, err := os.Stat(source); err != nil {
+			return descriptors, errors.Wrap(err, "Invalid source directory specified")
+		}
+		desc.DirBasePath = source
+		var err error
+		desc.TaggedImages, err = imagesToCopyFromDir(source)
+		if err != nil {
+			return descriptors, err
+		}
+		if len(desc.TaggedImages) == 0 {
+			return descriptors, errors.Errorf("No images to sync found in %q", source)
+		}
+		descriptors = append(descriptors, desc)
+
+	case "yaml":
+		cfg, err := newSourceConfig(source)
+		if err != nil {
+			return descriptors, err
+		}
+		for registryName, registryConfig := range cfg {
+			if len(registryConfig.Images) == 0 {
+				logrus.WithFields(logrus.Fields{
+					"registry": registryName,
+				}).Warn("No images specified for registry")
+				continue
+			}
+
+			descs, err := imagesToCopyFromRegistry(registryName, registryConfig, *sourceCtx)
+			if err != nil {
+				return descriptors, errors.Wrapf(err, "Failed to retrieve list of images from registry %q", registryName)
+			}
+			descriptors = append(descriptors, descs...)
+		}
+	}
+
+	return descriptors, nil
+}
+
+func (opts *syncOptions) run(args []string, stdout io.Writer) error {
+	if len(args) != 2 {
+		return errorShouldDisplayUsage{errors.New("Exactly two arguments expected")}
+	}
+
+	policyContext, err := opts.global.getPolicyContext()
+	if err != nil {
+		return errors.Wrapf(err, "Error loading trust policy")
+	}
+	defer policyContext.Destroy()
+
+	// validate source and destination options
+	contains := func(val string, list []string) (_ bool) {
+		for _, l := range list {
+			if l == val {
+				return true
+			}
+		}
+		return
+	}
+
+	if len(opts.source) == 0 {
+		return errors.New("A source transport must be specified")
+	}
+	if !contains(opts.source, []string{docker.Transport.Name(), directory.Transport.Name(), "yaml"}) {
+		return errors.Errorf("%q is not a valid source transport", opts.source)
+	}
+
+	if len(opts.destination) == 0 {
+		return errors.New("A destination transport must be specified")
+	}
+	if !contains(opts.destination, []string{docker.Transport.Name(), directory.Transport.Name()}) {
+		return errors.Errorf("%q is not a valid destination transport", opts.destination)
+	}
+
+	if opts.source == opts.destination && opts.source == directory.Transport.Name() {
+		return errors.New("sync from 'dir' to 'dir' not implemented, consider using rsync instead")
+	}
+
+	sourceCtx, err := opts.srcImage.newSystemContext()
+	if err != nil {
+		return err
+	}
+
+	sourceArg := args[0]
+	srcRepoList, err := imagesToCopy(sourceArg, opts.source, sourceCtx)
+	if err != nil {
+		return err
+	}
+
+	destination := args[1]
+	destinationCtx, err := opts.destImage.newSystemContext()
+	if err != nil {
+		return err
+	}
+
+	ctx, cancel := opts.global.commandTimeoutContext()
+	defer cancel()
+
+	imagesNumber := 0
+	options := copy.Options{
+		RemoveSignatures: opts.removeSignatures,
+		SignBy:           opts.signByFingerprint,
+		ReportWriter:     os.Stdout,
+		DestinationCtx:   destinationCtx,
+	}
+
+	for _, srcRepo := range srcRepoList {
+		options.SourceCtx = srcRepo.Context
+		for counter, ref := range srcRepo.TaggedImages {
+			var destSuffix string
+			switch ref.Transport() {
+			case docker.Transport:
+				// docker -> dir or docker -> docker
+				destSuffix = ref.DockerReference().String()
+			case directory.Transport:
+				// dir -> docker (we don't allow `dir` -> `dir` sync operations)
+				destSuffix = strings.TrimPrefix(ref.StringWithinTransport(), srcRepo.DirBasePath)
+				if destSuffix == "" {
+					// if source is a full path to an image, have destPath scoped to repo:tag
+					destSuffix = path.Base(srcRepo.DirBasePath)
+				}
+			}
+
+			if !opts.scoped {
+				destSuffix = path.Base(destSuffix)
+			}
+
+			destRef, err := destinationReference(path.Join(destination, destSuffix), opts.destination)
+			if err != nil {
+				return err
+			}
+
+			logrus.WithFields(logrus.Fields{
+				"from": transports.ImageName(ref),
+				"to":   transports.ImageName(destRef),
+			}).Infof("Copying image tag %d/%d", counter+1, len(srcRepo.TaggedImages))
+
+			_, err = copy.Image(ctx, policyContext, destRef, ref, &options)
+			if err != nil {
+				return errors.Wrapf(err, fmt.Sprintf("Error copying tag %q", transports.ImageName(ref)))
+			}
+			imagesNumber++
+		}
+	}
+
+	logrus.Infof("Synced %d images from %d sources", imagesNumber, len(srcRepoList))
+	return nil
+}
--- a/cmd/skopeo/utils.go
+++ b/cmd/skopeo/utils.go
@@ -2,13 +2,13 @@ package main
 
 import (
 	"context"
-	"errors"
 	"io"
 	"strings"
 
 	"github.com/containers/image/v5/pkg/compression"
 	"github.com/containers/image/v5/transports/alltransports"
 	"github.com/containers/image/v5/types"
+	"github.com/pkg/errors"
 	"github.com/urfave/cli"
 )
 
@@ -50,24 +50,34 @@ func sharedImageFlags() ([]cli.Flag, *sharedImageOptions) {
 	}, &opts
 }
 
+// imageOptions collects CLI flags specific to the "docker" transport, which are
+// the same across subcommands, but may be different for each image
+// (e.g. may differ between the source and destination of a copy)
+type dockerImageOptions struct {
+	global         *globalOptions      // May be shared across several imageOptions instances.
+	shared         *sharedImageOptions // May be shared across several imageOptions instances.
+	credsOption    optionalString      // username[:password] for accessing a registry
+	dockerCertPath string              // A directory using Docker-like *.{crt,cert,key} files for connecting to a registry or a daemon
+	tlsVerify      optionalBool        // Require HTTPS and verify certificates (for docker: and docker-daemon:)
+	noCreds        bool                // Access the registry anonymously
+}
+
 // imageOptions collects CLI flags which are the same across subcommands, but may be different for each image
 // (e.g. may differ between the source and destination of a copy)
 type imageOptions struct {
-	global           *globalOptions      // May be shared across several imageOptions instances.
-	shared           *sharedImageOptions // May be shared across several imageOptions instances.
-	credsOption      optionalString      // username[:password] for accessing a registry
-	dockerCertPath   string              // A directory using Docker-like *.{crt,cert,key} files for connecting to a registry or a daemon
-	tlsVerify        optionalBool        // Require HTTPS and verify certificates (for docker: and docker-daemon:)
-	sharedBlobDir    string              // A directory to use for OCI blobs, shared across repositories
-	dockerDaemonHost string              // docker-daemon: host to connect to
-	noCreds          bool                // Access the registry anonymously
+	dockerImageOptions
+	sharedBlobDir    string // A directory to use for OCI blobs, shared across repositories
+	dockerDaemonHost string // docker-daemon: host to connect to
 }
 
-// imageFlags prepares a collection of CLI flags writing into imageOptions, and the managed imageOptions structure.
-func imageFlags(global *globalOptions, shared *sharedImageOptions, flagPrefix, credsOptionAlias string) ([]cli.Flag, *imageOptions) {
+// dockerImageFlags prepares a collection of docker-transport specific CLI flags
+// writing into imageOptions, and the managed imageOptions structure.
+func dockerImageFlags(global *globalOptions, shared *sharedImageOptions, flagPrefix, credsOptionAlias string) ([]cli.Flag, *imageOptions) {
 	opts := imageOptions{
-		global: global,
-		shared: shared,
+		dockerImageOptions: dockerImageOptions{
+			global: global,
+			shared: shared,
+		},
 	}
 
 	// This is horribly ugly, but we need to support the old option forms of (skopeo copy) for compatibility.
@@ -93,6 +103,19 @@ func imageFlags(global *globalOptions, shared *sharedImageOptions, flagPrefix, c
 			Usage: "require HTTPS and verify certificates when talking to the container registry or daemon (defaults to true)",
 			Value: newOptionalBoolValue(&opts.tlsVerify),
 		},
+		cli.BoolFlag{
+			Name:        flagPrefix + "no-creds",
+			Usage:       "Access the registry anonymously",
+			Destination: &opts.noCreds,
+		},
+	}, &opts
+}
+
+// imageFlags prepares a collection of CLI flags writing into imageOptions, and the managed imageOptions structure.
+func imageFlags(global *globalOptions, shared *sharedImageOptions, flagPrefix, credsOptionAlias string) ([]cli.Flag, *imageOptions) {
+	dockerFlags, opts := dockerImageFlags(global, shared, flagPrefix, credsOptionAlias)
+
+	return append(dockerFlags, []cli.Flag{
 		cli.StringFlag{
 			Name:        flagPrefix + "shared-blob-dir",
 			Usage:       "`DIRECTORY` to use to share blobs across OCI repositories",
@@ -103,12 +126,7 @@ func imageFlags(global *globalOptions, shared *sharedImageOptions, flagPrefix, c
 			Usage:       "use docker daemon host at `HOST` (docker-daemon: only)",
 			Destination: &opts.dockerDaemonHost,
 		},
-		cli.BoolFlag{
-			Name:        flagPrefix + "no-creds",
-			Usage:       "Access the registry anonymously",
-			Destination: &opts.noCreds,
-		},
-	}, &opts
+	}...), opts
 }
 
 // newSystemContext returns a *types.SystemContext corresponding to opts.
--- a/docs/skopeo-sync.1.md
+++ b/docs/skopeo-sync.1.md
@@ -1,0 +1,142 @@
+% skopeo-sync(1)
+
+## NAME
+skopeo\-sync - Synchronize images between container registries and local directories.
+
+
+## SYNOPSIS
+**skopeo sync** --src _transport_ --dest _transport_ _source_ _destination_
+
+## DESCRIPTION
+Synchronize images between container registries and local directories.
+The synchronization is achieved by copying all the images found at _source_ to _destination_.
+
+Useful to synchronize a local container registry mirror, and to to populate registries running inside of air-gapped environments.
+
+Differently from other skopeo commands, skopeo sync requires both source and destination transports to be specified separately from _source_ and _destination_.
+One of the problems of prefixing a destination with its transport is that, the registry `docker://hostname:port` would be wrongly interpreted as an image reference at a non-fully qualified registry, with `hostname` and `port` the image name and tag.
+
+Available _source_ transports:
+ - _docker_ (i.e. `--src docker`): _source_ is a repository hosted on a container registry (e.g.: `registry.example.com/busybox`).
+ If no image tag is specified, skopeo sync copies all the tags found in that repository.
+ - _dir_ (i.e. `--src dir`): _source_ is a local directory path (e.g.: `/media/usb/`). Refer to skopeo(1) **dir:**_path_ for the local image format.
+ - _yaml_ (i.e. `--src yaml`): _source_ is local YAML file path.
+ The YAML file should specify the list of images copied from different container registries (local directories are not supported). Refer to EXAMPLES for the file format.
+
+Available _destination_ transports:
+ - _docker_ (i.e. `--dest docker`): _destination_ is a container registry (e.g.: `my-registry.local.lan`).
+ - _dir_ (i.e. `--dest dir`): _destination_ is a local directory path (e.g.: `/media/usb/`).
+ One directory per source 'image:tag' is created for each copied image.
+
+When the `--scoped` option is specified, images are prefixed with the source image path so that multiple images with the same
+name can be stored at _destination_.
+
+## OPTIONS
+**--authfile** _path_
+
+Path of the authentication file. Default is ${XDG\_RUNTIME\_DIR}/containers/auth.json, which is set using `podman login`.
+If the authorization state is not found there, $HOME/.docker/config.json is checked, which is set using `docker login`.
+
+**--src** _transport_ Transport for the source repository.
+
+**--dest** _transport_ Destination transport.
+
+**--scoped** Prefix images with the source image path, so that multiple images with the same name can be stored at _destination_.
+
+**--remove-signatures** Do not copy signatures, if any, from _source-image_. This is necessary when copying a signed image to a destination which does not support signatures.
+
+**--sign-by=**_key-id_ Add a signature using that key ID for an image name corresponding to _destination-image_.
+
+**--src-creds** _username[:password]_ for accessing the source registry.
+
+**--dest-creds** _username[:password]_ for accessing the destination registry.
+
+**--src-cert-dir** _path_ Use certificates (*.crt, *.cert, *.key) at _path_ to connect to the source registry or daemon.
+
+**--src-no-creds** _bool-value_ Access the registry anonymously.
+
+**--src-tls-verify** _bool-value_ Require HTTPS and verify certificates when talking to a container source registry or daemon (defaults to true).
+
+**--dest-cert-dir** _path_ Use certificates (*.crt, *.cert, *.key) at _path_ to connect to the destination registry or daemon.
+
+**--dest-no-creds** _bool-value_  Access the registry anonymously.
+
+**--dest-tls-verify** _bool-value_ Require HTTPS and verify certificates when talking to a container destination registry or daemon (defaults to true).
+
+## EXAMPLES
+
+### Synchronizing to a local directory
+```
+$ skopeo sync --src docker --dest dir registry.example.com/busybox /media/usb
+```
+Images are located at:
+```
+/media/usb/busybox:1-glibc
+/media/usb/busybox:1-musl
+/media/usb/busybox:1-ubuntu
+...
+/media/usb/busybox:latest
+```
+
+### Synchronizing to a local directory, scoped
+```
+$ skopeo sync --src docker --dest dir --scoped registry.example.com/busybox /media/usb
+```
+Images are located at:
+```
+/media/usb/registry.example.com/busybox:1-glibc
+/media/usb/registry.example.com/busybox:1-musl
+/media/usb/registry.example.com/busybox:1-ubuntu
+...
+/media/usb/registry.example.com/busybox:latest
+```
+
+### Synchronizing to a container registry
+```
+skopeo sync --src docker --dest docker registry.example.com/busybox my-registry.local.lan
+```
+Destination registry content:
+```
+REPO                           TAGS
+registry.example.com/busybox   1-glibc, 1-musl, 1-ubuntu, ..., latest
+```
+
+### YAML file content (used _source_ for `**--src yaml**`)
+
+```yaml
+registry.example.com:
+    images:
+        busybox: []
+        redis:
+            - "1.0"
+            - "2.0"
+    credentials:
+        username: john
+        password: this is a secret
+    tls-verify: true
+    cert-dir: /home/john/certs
+quay.io:
+    tls-verify: false
+    images:
+        coreos/etcd:
+            - latest
+```
+
+This will copy the following images:
+- Repository `registry.example.com/busybox`: all images, as no tags are specified.
+- Repository `registry.example.com/redis`: images tagged "1.0" and "2.0".
+- Repository `quay.io/coreos/etcd`: images tagged "latest".
+
+For the registry `registry.example.com`, the "john"/"this is a secret" credentials are used, with server TLS certificates located at `/home/john/certs`.
+
+TLS verification is normally enabled, and it can be disabled setting `tls-verify` to `true`.
+In the above example, TLS verification is enabled for `reigstry.example.com`, while is
+disabled for `quay.io`.
+
+## SEE ALSO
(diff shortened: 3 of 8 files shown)


runcom:
@mtrmac PTAL as well

vrothberg:
@runcom @rhatdan  @mtrmac friendly ping. I'd love to see this feature in an upcoming Skopeo release.

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,509 @@
...
+	Images      map[string][]string
+	Credentials types.DockerAuthConfig
+	TLSVerify   bool   `yaml:"tls-verify"`
+	CertDir     string `yaml:"cert-dir"`
+}
+
+type sourceCfg map[string]registrySyncCfg
+
+func newSourceConfig(yamlFile string) (cfg sourceCfg, err error) {
+
```
giuseppe: extra empty line

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,509 @@
...
+	return
+}
+
+func isValidTransport(transport types.ImageTransport) (bool, error) {
+	dockerTransport := transports.Get("docker")
+	if dockerTransport == nil {
+		return false, fmt.Errorf("Cannot find 'docker' transport type")
+	}
+
+	dirTransport := transports.Get("dir")
```
giuseppe: could we add to `validTransports` only the ones that are present?  i.e. if dirTransport is missing we just skip it and not add it to `validTransports` as anyway the `docker` -> `docker` case doesn't need it.

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,509 @@
...
+
+		if _, err := os.Stat(tgtDir); err != nil {
+			if os.IsNotExist(err) {
+				if err := os.MkdirAll(tgtDir, 0755); err != nil {
+					return nil, fmt.Errorf("Error creating directory for image %s: %v",
+						tgtDir,
+						err)
+				}
+			} else {
+				return nil, fmt.Errorf("Error while checking existance of directory %s: %v",
```
giuseppe: typo here: existance -> existence

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,509 @@
...
+	'docker://' transport and no tag is specified.
+
+	DESTINATION can be either a docker registry
+	(eg: docker://my-registry.local.lan) or a local directory
+	(eg: dir:///media/usb).
+
+	When DESTINATION is a local directory one directory per 'image:tag' is going
+	to be created.
+	`),
+	ArgsUsage: "[--source SOURCE] [--source-file SOURCE-FILE] DESTINATION",
```
giuseppe: I see why we have two different options, but it looks a bit unnatural to add a mandatory argument in this way.

Could it be like "[--source-yaml] SOURCE DESTINATION"

if `--source-yaml` is specified then `SOURCE` points to a YAML file.

Not ideal either, but in the simplest case you have `skopeo sync docker://... docker://...`.

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,509 @@
...
+		if retErr != nil {
+			// some registries may decide to block the "list all tags" endpoint
+			// gracefully allow the inspect to continue in this case. Currently
+			// the IBM Bluemix container registry has this restriction.
+			if !strings.Contains(retErr.Error(), "401") {
+				return tags, fmt.Errorf("Error determining repository tags: %v", retErr)
+			}
+			logrus.Warn("Registry disallows tag list retrieval; skipping")
+		}
+	}
```
TomSweeneyRedHat: Maybe I need more tea, but should you return 'tags, nil' here and/or at line 152?

TomSweeneyRedHat: Oh shoot, forgot golang deals with bare returns.  This code is OK as is, but I'd prefer specific values being returned.  I find it more readable.  My $.02, your mileage may vary.

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,509 @@
...
+	destinationCtx, err := contextFromGlobalOptions(c, "dest-")
+	if err != nil {
+		return err
+	}
+
+	signBy := c.String("sign-by")
+	removeSignatures := c.Bool("remove-signatures")
+
+	//TODO: should we assume that's our default manifest type?
+	manifestType := manifest.DockerV2Schema2MediaType
```
TomSweeneyRedHat: I think so, but I'd prefer a way to pass in a different variant.  But I'd lean on @mtrmac or @runcom for the final word.

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,509 @@
...
+
+	Useful to keep in sync a local docker registry mirror. Can be used
+	to populate also registries running inside of air-gapped environments.
+
+	SOURCE can be either a repository hosted on a docker registry
+	(eg: docker://docker.io/busybox) or a local directory
+	(eg: dir:///media/usb/).
+
+	SOURCE-FILE is a YAML file with a set of source images from different
+	docker registry. Local directory are not supported.
```
TomSweeneyRedHat: Suggest:  
SOURCE-FILE is a YAML file with a set of source images from a different

Then either:
docker registry. Local directory is not supported.

or 
docker registry. Local directories are not supported.

I prefer the second.

docs/skopeo.1.md
```diff
@@ -192,6 +192,48 @@ Verify a signature using local files, digest will be printed on success.
 
 **Note:** If you do use this, make sure that the image can not be changed at the source location between the times of its verification and use.
 
+## skopeo sync
+**skopeo copy** [**--source** _source_] [**--source-file** _source-file_] _destination_
+
+Copy all the images from _source_ (or _source-file_) to _destination_.
+
+Useful to keep in sync a local docker registry mirror. It can be used to populate also registries running inside of air-gapped environments.
```
TomSweeneyRedHat: Suggest:
Useful to keep in sync with a local docker registry mirror. It can also be used to populate registries running inside of air-gapped environments.

cmd/skopeo/sync.go
TomSweeneyRedHat: My thinking was just to make the code more flexible if ever needed.  @mtrmac if you think that's not necessary or the cost isn't worth it, I'm fine with that.  Just preserving the manifest as you suggested makes sense to me too.

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,509 @@
...
+}
+
+var syncCmd = cli.Command{
+	Name:  "sync",
+	Usage: "Sync one or more images from one location to another",
+	Description: fmt.Sprint(`
+
+	Copy all the images from SOURCE to DESTINATION.
+
+	Useful to keep in sync a local docker registry mirror. Can be used
```
rhatdan: s'/docker/container/g'

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,509 @@
...
+	to populate also registries running inside of air-gapped environments.
+
+	SOURCE can be either a repository hosted on a docker registry
```
rhatdan: s'/docker/container/g'

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,509 @@
...
+	(eg: docker://docker.io/busybox) or a local directory
+	(eg: dir:///media/usb/).
+
+	SOURCE-FILE is a YAML file with a set of source images from different
+	docker registry. Local directory are not supported.
+
+	Skopeo sync will copy all the tags of an image when SOURCE uses the
+	'docker://' transport and no tag is specified.
+
+	DESTINATION can be either a docker registry
```
rhatdan: s'/docker/container/g'

rhatdan:
I would like to see this a little less "docker' centric.  Perhaps use some examples other then docker.io.  registry.example.com or some such.

cmd/skopeo/sync.go
mtrmac: I’m _weakly_ against exposing options just because the underlying code implements them: it takes a possibly unnecessary effort, and after the options are introduced, it’s difficult to both remove them and to gauge whether they have been ever necessary in the first place.

But in this case, I’m _primarily_ saying that the default behavior should, unless there is a specific reason, not be _hard-coding_ manifest types at all. The existing autodetection is expected to just work, and it is future-proof, unlike hard-coding a specific default which is certain to become obsolete at some point.

docs/skopeo.1.md
```diff
@@ -192,6 +192,48 @@ Verify a signature using local files, digest will be printed on success.
...
+
+_source_ can be either a repository hosted on a docker registry (eg: docker://docker.io/busybox) or a local directory (eg: dir:///media/usb/).
+
+_source-file_ is a YAML file with a set of source images from different docker registry. Local directory are not supported.
+
+**skopeo sync** will copy all the tags of an image when _source_ uses the docker://' transport and no tag is specified.
+
+_destination_ can be either a docker registry (eg: docker://my-registry.local.lan) or a local directory (eg: dir:///media/usb).
```
mtrmac: The PR introduction says

> The image namespace is changed during the `docker://` to `docker://` or `dir:` copy. The FQDN of the registry hosting the image will be added as new root namespace of the image. For example, the image `registry.example.com/busybox:latest` will be copied to `registry.local.lan/registry.example.com/busybox:latest`

This inflexibly hard-codes policy, and seems entirely unnecessary: whatever script is used for the regular sync can just specify `registry.local.lan/registry.example.com/busybox` as the destination.

Also, this policy forces `sync` to invent an entirely new `docker://hostname` syntax, which is ambiguous with the image name syntax used everywhere else, and has a _different meaning_ (the image `docker://hostname` == `docker://docker.io/library/hostname:latest`).

… and AFAICT it s not documented in this man page.

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,509 @@
...
+	// Normalize the image name, this will automatically add
+	// the `latest` tag when no tag has been specified
+	normName, err := reference.ParseNormalizedNamed(image)
+	if err != nil {
+		return true, err
+	}
+
+	// if the tag is `latest` -> the tag has been automatically added -> no tag
+	// was specified by the user
+	return !strings.HasSuffix(reference.TagNameOnly(normName).String(), ":latest"), nil
```
mtrmac: (I’m sorry about not looking at this for so long. I do hope to make a proper review soonish.)

One general comment: podman and buildah have similarly built fair amount of code around manipulating the string forms of image names, and have been rather struggling to correctly handle the various corner cases and ambiguous forms of input.

I _very strongly_ want to avoid introducing such fragility to skopeo.

As a very local case, this should not be using `.String()` and then `HasSuffix` to parse it, but an interface conformance test against `reference.NamedTagged`. [Or maybe something else entirely, to also handle digested references.] 

_Most_ of the string parsing and formatting in here looks, at a first glance, like code that should not be using strings (though I haven’t read enough of the PR to be know that it is actually avoidable, to be fair.)

docs/skopeo.1.md
marcov: Some kind of scoping is necessary when running `sync` from a YAML file, as you may be pulling the same image name from different namespaces or domains.

The reason to have this kind of scoping is that the original image location can be easily extrapolated, and this information is useful when running a server that is configured to mirror multiple registries.

Maybe the requirement can be dropped when source references to a `docker://` location (and a flag can be added to make the current behaviour optional)

cmd/skopeo/sync.go
marcov: Thanks, it turned out that most the raw strings manipulation could be avoided by rewriting some parts of the code and making use of [containers/image](https://github.com/containers/image) libraries.

rhatdan:
@mtrmac @runcom Can we merge this?

chilicat:
> The image namespace is changed during the docker:// to docker:// or dir: copy. The FQDN of the registry hosting the image will be added as new root namespace of the image. For example, the image 
> registry.example.com/busybox:latest will be copied to 
> registry.local.lan/registry.example.com/busybox:latest.

I would prefer to not have the original registry name as namespace in the destination registry. The source registry might be a internal registry where the name changes over time or where we want to hide the actual name on the customer side.

Maybe that is something which could be configurable?

flavio:
> I would prefer to not have the original registry name as namespace in the destination registry. The
> source registry might be a internal registry where the name changes over time or where we want to 
> hide the actual name on the customer side.
>
> Maybe that is something which could be configurable?

We could add a `--dest-prefix` flag to allow something like that:

```
skopeo sync --dest-prefix acme-inc docker://wip-registry.lan/busybox docker://example.com 
```

That would copy all the `busybox` images from `wip-registry.lan/busybox` to `example.com/acme-inc/busybox`.

We should also cover this option inside of the source yaml file. Let's take this `source.yaml` as an example:

```yaml
docker.io:
    images:
        busybox: []
wip-registry:
    dest-prefix: acme-inc
    images:
        coreos/etcd:
            - latest
```

Let's assume we run this command:

```
skopeo sync --source-yaml source.yaml docker://target.com
```

By this point, inside of the `target.com` registry, we would have the following images:

  * `docker.io/busybox` -> `docker pull target.com/docker.io/busybox`
  * `acme-inc/coreos/etcd` -> `docker pull target.com/acme-inc/coreos/etcd`


Would that work with you?

How should we proceed if everybody agrees with this design? Should we get this PR merged and then implement this new flag via a new PR (to limit the amount of code to be reviewed)?

marcov:
Rebased and added a few integration tests for `sync`. Run with: `make test-integration TESTFLAGS="-check.f Sync"`

vrothberg:
@marcov, could you rebase this change on master?

@containers/image-maintainers, can we revive this PR? We're forced to run this downstream but would prefer to have a non-patched skopeo :)

docs/skopeo.1.md
```diff
@@ -192,6 +192,45 @@ Verify a signature using local files, digest will be printed on success.
...
+
+Copy all the images from _source_ to _destination_.
+
+Useful to keep in sync with a local container registry mirror. It can also be used to populate registries running inside of air-gapped environments.
+
+_source_ can be either a repository hosted on a container registry (eg: docker://registry.example.com/busybox) or a local directory (eg: dir:/media/usb/).
+
+If **--source-yaml** is specified, then _source_ points to a YAML file with a list of source images from different container registries (local directories are not supported).
+
+When the source location is a container registry and no tags are specifieds,  **skopeo sync** will copy all the tags associated to the source image.
```
muff1nman: Small typo here `specifieds`

marcov: @muff1nman  thanks for the review!

marcov:
@vrothberg Rebased, and also (hopefully) fixed a problem with integration tests failing when repeating the setup of a openshift cluster.

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,505 @@
...
+			Name:  "remove-signatures",
+			Usage: "Do not copy signatures from SOURCE images",
+		},
+		cli.StringFlag{
+			Name:  "sign-by",
+			Usage: "Sign the image using a GPG key with the specified `FINGERPRINT`",
+		},
+		cli.BoolFlag{
+			Name:  "source-yaml",
+			Usage: "Interpret SOURCE as a YAML file with a list of images from different container regitries",
```
muff1nman: Typo with `regitries`

Vad1mo:
@flavio @marcov Could you rebase and update to the latest https://github.com/containers/image

marcov:
@Vad1mo yes, and thanks for the ping.

Vad1mo:
The automated prefix of the whole domain to the namespace like `
from="docker://vad1mo/zportal:1.0.142" to="docker://docker.8gears.com/vad1mo/docker.io/vad1mo/zportal:1.0.142"
` makes it unpredictable in cases when you want to move data from one registry to another.  

Im my case I want to migrate data from one registry to another, and the namespaces should stay the same even if the domain changes.

marcov:
@Vad1mo we are aware of that, we also encounter this situation in our use cases. We plan to add the option to modify this behaviour.

As a temporary workaround, you can do:
- `sync docker://vad1mo/zportal dir:localdir`
and 
- `sinc dir:localdir/docker.io/ docker://docker.8gears.com/`

marcov:
Rebased on master. Heads up: I realized that TLS verification is OFF by default  when source is a YAML file, I need to change the mapping key from `tls-verify` to `skip-tls` to avoid this.

rhatdan:
@mtrmac Needs your review, but I like the idea.

marcov:
@luisdavim copying from Bittorrent would be trivial to support, as *sync* is copying images using `copy.Image` (same as *copy*).

As for the current status of this patch, we are waiting for any feedback from the maintainers.

rhatdan:
@vrothberg @mtrmac @runcom Any movement on this patch.  I would like to get this into skopeo and cut a new release.

@flavio Needs a rebase for the man page change.

marcov:
Rebased. 
I still need to add a couple of changes:
- turn `tls-verify` on by default when reading sources from a yaml file
- filter out not applicable src and dest flags, this is needed since flags/options processing has been unified in `utils.go`

marcov:
I have addressed the changes detailed in my previous comment with a new push.

vrothberg:
Thanks a lot, @marcov! @mtrmac, let's have a look at this PR as it's warming up again. If the approaching cold doesn't strike me, I will start tomorrow.

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,594 @@
...
+	if dirURL.Scheme != directory.Transport.Name() {
+		return "", fmt.Errorf("Not an dir URL: %v", dirURL)
+	}
+
+	if dirURL.Opaque != "" {
+		// relative dir path, e.g. dir:localdir (without "/" or "//")
+		dirPath = dirURL.Opaque
+	} else {
+		dirPath = path.Join("/", dirURL.Host, dirURL.Path)
+	}
```
mtrmac: The _transport_`:`_image_ syntax really is not URLs (e.g. no percent-escaping) ; overall I don’t think it’s reasonable to mix the two.

I realize the code needs to detect whether a `docker://` reference was specified with a tag, and the current reference API makes that impossible, so _something_ needs to change (should the sources/destinations even look similar when the `:latest` tag defaulting is not there?); still, using the `URL` parser is not making that problem any easier AFAICT.

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,594 @@
...
+			}
+			if !found {
+				filtered = append(filtered, f)
+			}
+		}
+		return filtered
+	}
+
+	srcFlags = filterFlags(srcFlags, "src-")
+	destFlags = filterFlags(destFlags, "dest-")
```
mtrmac: This feels unclean, I’ll take another look.

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,594 @@
...
+	if len(repoDesc.TaggedImages) == 0 {
+		return repoDesc, errors.New("No images to sync found in SOURCE")
+	}
+
+	return repoDesc, nil
+}
+
+// Given a yaml file and a source context, returns a list of repository descriptors,
+// each containing a list of tagged image references, to be used as sync source.
+func syncFromYaml(yamlFile string, sourceCtx *types.SystemContext) (repoDescList []repoDescriptor, err error) {
```
mtrmac: I’m still unsure about this — what does the `.yaml` format do that couldn’t just as well be done with an ordinary shell script calling `skopeo`? Allow recording credentials in a file so that they don’t show up in **ps**(1)?

marcov:
Thanks for starting reviewing @mtrmac!

>I also still feel very hesitant about hard-coding the use of the source name as a namespace in the destination.

We can switch this in an option (off by default) to make the behaviour of `sync` consistent with `copy`. We explained above the need for the namespace scoping when dealing with homonymous repositories on different registries.

cmd/skopeo/sync.go
marcov: I opted for `url.URL` because in multiple occasions I needed to extract {transport, registry hostname, image path} and generate new URLs with them.
Doing it using low level `string` required more code and replicating in part what `net/url` is already doing.

I did not thought about handling escaping correctly, thanks for raising this. Looks like you can deal with by using the value of [`URL.RawPath`](https://golang.org/pkg/net/url/#URL).

Overall I feel using URL is adequate to parsing _transport_`:`_image_, if you take into account things like escaping and `dir:` non-absolute paths.

cmd/skopeo/sync.go
marcov: This is required because some of the `imageOptions` and `destImageOptions` are not meaningful to the sync command. 
So I am filtering those options out. I agree this is sub-optimal, we could find a better way.

cmd/skopeo/sync.go
marcov: The use case is the ability to specify multiple images (and any tags combination) hosted on multiple registries (each one with its own auth method) in a yaml file. 

You then call `sync` one time only with:
`skopeo sync --source yaml images.yaml dir:syncedImages`

Instead of:
```
skopeo sync [opts] docker://registry1/image1 dir:destDir
skopeo sync [opts] docker://registry1/image2 dir:destDir
...
skopeo sync [opts] docker://registry1/imageN dir:destDir
skopeo sync [opts] docker://registry2/image1 dir:destDir
...
skopeo sync [opts] docker://registryN/imageN dir:destDir
```

docs/skopeo-sync.1.md
```diff
@@ -0,0 +1,80 @@
...
+        coreos/etcd:
+            - latest
+```
+
+## SEE ALSO
+skopeo(1), podman-login(1), docker-login(1)
+
+## AUTHORS
+
+Antonio Murdaca <runcom@redhat.com>, Miloslav Trmac <mitr@redhat.com>, Jhon Honce <jhonce@redhat.com>
```
vrothberg: Just starting the review now. I think the real authors are Flavio and Marco :^)

docs/skopeo-sync.1.md
```diff
@@ -0,0 +1,80 @@
+% skopeo-sync(1)
+
+## NAME
+skopeo\-sync - Copy images from one or more repositories to a user specified destination.
+
+## SYNOPSIS
+**skopeo sync** [**--source-yaml**] _source_ _destination_
+
+## DESCRIPTION
+Copy images from one or more repositories to a user specified destination.
```
vrothberg: Please give some more details about the internals. Which images are copied? Will they be updated if the destination has an image with the same name already? Is there something like path traversal/walks?

docs/skopeo-sync.1.md
```diff
@@ -0,0 +1,80 @@
...
+
+Useful to keep in sync with a local container registry mirror. It can also be used to populate registries running inside of air-gapped environments.
+
+_source_ can be:
+ - a repository hosted on a container registry (eg: docker://registry.example.com/busybox)
```
vrothberg: Missing dot at the end.

docs/skopeo-sync.1.md
```diff
@@ -0,0 +1,80 @@
...
+ - a local directory (eg: dir:/media/usb/).
+ - (when **--source-yaml** is specified) a YAML file with a list of source images from different container registries.
+
+When the source location is a container repository and no tags are specified,  **skopeo sync** will copy all the tags associated to the source image.
+
+_destination_ can be either a container registry (eg: docker://my-registry.local.lan) or a local directory (eg: dir:/media/usb).
+
+When _destination_ is a local directory, one directory per 'image:tag' will be created.
```
vrothberg: Can we give an example?

docs/skopeo-sync.1.md
```diff
@@ -0,0 +1,80 @@
...
+**skopeo sync** [**--source-yaml**] _source_ _destination_
+
+## DESCRIPTION
+Copy images from one or more repositories to a user specified destination.
+
+Useful to keep in sync with a local container registry mirror. It can also be used to populate registries running inside of air-gapped environments.
+
+_source_ can be:
+ - a repository hosted on a container registry (eg: docker://registry.example.com/busybox)
+ - a local directory (eg: dir:/media/usb/).
```
vrothberg: Must the local directory have a specific structure?

docs/skopeo-sync.1.md
```diff
@@ -0,0 +1,80 @@
...
+When _destination_ is a local directory, one directory per 'image:tag' will be created.
+
+## OPTIONS
+
+**--authfile** _path_
+
+Path of the authentication file. Default is ${XDG_RUNTIME\_DIR}/containers/auth.json, which is set using `podman login`.
+If the authorization state is not found there, $HOME/.docker/config.json is checked, which is set using `docker login`.
+
+**--remove-signatures** do not copy signatures, if any, from _source-image_. Necessary when copying a signed image to a destination which does not support signatures.
```
vrothberg: s/do not/Do not/

docs/skopeo-sync.1.md
```diff
@@ -0,0 +1,80 @@
...
+
+**--src-tls-verify** _bool-value_ Require HTTPS and verify certificates when talking to container source registry or daemon (defaults to true)
+
+**--dest-cert-dir** _path_ Use certificates at _path_ (*.crt, *.cert, *.key) to connect to the destination registry or daemon
+
+**--dest-tls-verify** _bool-value_ Require HTTPS and verify certificates when talking to container destination registry or daemon (defaults to true)
+
+## EXAMPLES
+
+Content of a YAML file to be used with **--source-yaml**:
```
vrothberg: Please explain which images will be copied based on the example file.

docs/skopeo-sync.1.md
```diff
@@ -0,0 +1,80 @@
...
+
+## DESCRIPTION
+Copy images from one or more repositories to a user specified destination.
+
+Useful to keep in sync with a local container registry mirror. It can also be used to populate registries running inside of air-gapped environments.
+
+_source_ can be:
+ - a repository hosted on a container registry (eg: docker://registry.example.com/busybox)
+ - a local directory (eg: dir:/media/usb/).
+ - (when **--source-yaml** is specified) a YAML file with a list of source images from different container registries.
```
vrothberg: I think we should be very verbose that the yaml file works only with the `docker://` transport.

docs/skopeo-sync.1.md
```diff
@@ -0,0 +1,80 @@
...
+**--authfile** _path_
+
+Path of the authentication file. Default is ${XDG_RUNTIME\_DIR}/containers/auth.json, which is set using `podman login`.
+If the authorization state is not found there, $HOME/.docker/config.json is checked, which is set using `docker login`.
+
+**--remove-signatures** do not copy signatures, if any, from _source-image_. Necessary when copying a signed image to a destination which does not support signatures.
+
+**--sign-by=**_key-id_ add a signature using that key ID for an image name corresponding to _destination-image_
+
+**--source-yaml** Interpret _source_ as a YAML file with a list of images from different container registries
```
vrothberg: Please mention that it only works with the `docker://` transport. Might need to be updated in the help text as well.

cmd/skopeo/utils.go
```diff
@@ -228,3 +231,15 @@ func parseImageSource(ctx context.Context, opts *imageOptions, name string) (typ
...
 }
+
+func yamlUnmarshal(yamlFile string, cfg interface{}) error {
+	source, err := ioutil.ReadFile(yamlFile)
+	if err != nil {
+		return err
+	}
+	err = yaml.Unmarshal(source, cfg)
+	if err != nil {
+		return errors.WithMessage(err, fmt.Sprintf("cannot unmarshal %s", yamlFile))
```
vrothberg: The rest of Skopeo usually uses `error.Wrapf(...)`. Can we stick to that? In this specific case, `errors.WithMessagef(...)` would avoid the call to `Sprintf(...)`.

Same applies to other uses of `WithMessage(...)`.

cmd/skopeo/sync.go
vrothberg: I share @mtrmac thoughts. `net/url.Parse(...)` is full of very odd behaviors. We tried fighting it multiple times and ended up avoiding it entirely.

Can't we use "github.com/containers/image/transports/alltransports" instead of using "net/url.URL"?

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,594 @@
...
+		return true
+	case directory.Transport:
+		return true
+	}
+
+	return false
+}
+
+// Return a URL object from an input string
+func parseURL(urlString string) (*url.URL, error) {
```
vrothberg: I would really prefer to use "github.com/containers/image/transports/alltransports" instead of that (see another comment).

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,594 @@
...
+	if !valid {
+		return parsedURL, errors.New("Invalid transport")
+	}
+
+	return parsedURL, nil
+}
+
+// Given a tranport and an image name (without the transport), returns an
+// ImageReference.
+func getImageReference(transport types.ImageTransport, imgName string) (types.ImageReference, error) {
```
vrothberg: If we use `alltransports.ParseImageName(...)` the reference would point to the transport.

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,594 @@
...
+		Description: fmt.Sprint(`
+
+	Copy all the images from SOURCE to DESTINATION.
+
+	Useful to keep in sync a local container registry mirror. Can be used
+	to populate also registries running inside of air-gapped environments.
+
+	SOURCE can be either a repository hosted on a container registry
+	(eg: docker://registry.example.com/busybox) or a local directory
+	(eg: dir:/media/usb/).
```
vrothberg: s/eg:/e.g.,/

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,594 @@
...
+	If --source-yaml is specified, then SOURCE points to a YAML file with
+	a list of source images from different container registries
+	(local directories are not supported).
+
+	When syncing from a repository where and no tags are specified, skopeo
+	sync will copy all the tags contained in that repository.
+
+	DESTINATION can be either a container registry
+	(eg: docker://my-registry.local.lan) or a local directory
+	(eg: dir:/media/usb).
```
vrothberg: s/eg:/e.g.,/

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,594 @@
...
+		return filtered
+	}
+
+	srcFlags = filterFlags(srcFlags, "src-")
+	destFlags = filterFlags(destFlags, "dest-")
+
+	return cli.Command{
+		Name:  "sync",
+		Usage: "Sync one or more images from one location to another",
+		Description: fmt.Sprint(`
```
vrothberg: @TomSweeneyRedHat can you have a look at the text?

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,594 @@
...
+	(eg: dir:/media/usb/).
+
+	If --source-yaml is specified, then SOURCE points to a YAML file with
+	a list of source images from different container registries
+	(local directories are not supported).
+
+	When syncing from a repository where and no tags are specified, skopeo
+	sync will copy all the tags contained in that repository.
+
+	DESTINATION can be either a container registry
```
vrothberg: This seems redundant to SOURCE and could get merged.

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,594 @@
...
+		for counter, ref := range srcRepo.TaggedImages {
+			destRef, err := buildFinalDestination(ref, destinationURL, srcRepo.DirBasePath)
+			if err != nil {
+				return err
+			}
+
+			logrus.WithFields(logrus.Fields{
+				"from": transports.ImageName(ref),
+				"to":   transports.ImageName(destRef),
+			}).Infof("Copying image tag %d/%d", counter+1, len(srcRepo.TaggedImages))
```
vrothberg: All Infof logs are printed by default which does not look right to me. I suggest printing those logs on stdout.

mtrmac:
> ```shell
> [root@nebuchadnezzar skopeo]$ ./skopeo sync docker://localhost:5000 dir:`pwd`/sync-test
> …
> FATA[0002] Error copying tag 'docker://localhost:5000': Error determining manifest MIME type for docker://localhost:5000: Error reading manifest 5000 in docker.io/library/localhost: errors:
> ```
>
> That's something we must address and something that must be clear documented in manpage and the help text. The big question is: _How do we translate source and destination without breaking the reference parsing?_

As the example above shows, the `docker://` string format, as is, just does not leave any “room” for specifying namespace prefixes; it was originally designed to support short names for `docker.io` images instead.

We _can_ write a different parser that supports prefixes (by requiring a host name to always be present — recall the earlier discussion about such a parser for `sysregistriesv2` names, although they don’t have the `docker://` prefix), but then we would end up with two different `docker://` syntaxes that have different rules.

In retrospect, accepting hostname-unqualified names in `docker://` at all was probably a mistake, but a ~necessary one at the time, and anyway here we are.

---

I don’t have a _good_ suggestion how to solve this; my best guess is not using the `alltransports.ParseImageName` format at all, and have `sync` explicitly say `--docker-source` `--dir-source` / `--docker-dest` / `--dir-dest`, or maybe `--source-type docker` / `--dest-type docker`, or something like that, and modulo all the bikeshedding about spelling and ordering and explicit options vs. positional parameters (vs. YAML); and then construct the `docker.Reference` / `dir.Reference` values with explicit knowledge of those particular transports.

Better ideas are certainly welcome.

cmd/skopeo/sync.go
mtrmac: > I opted for `url.URL` because in multiple occasions I needed to extract {transport, registry hostname, image path} and generate new URLs with them.

None of those values are URLs, using the proper definition.

> Doing it using low level `string` required more code and replicating in part what `net/url` is already doing.

Almost nothing `net/url` is doing is applicable or correct for those values. At best `net/url.getScheme` is a very complex way to do `strings.SplitN(… ":", 2)` — with extra logic that either does not matter or is incompatible with the `alltransports.ParseImageName` parser.

And this code ends up having to correct for the mismatches, like the `if` quoted above, or `fmt.Sprintf("//%s%s", sourceURL.Host, sourceURL.Path)` elsewhere.

The use of `url.URL` is _already_ adding more code than it saves, and that’s even before getting the corner cases correct.

(I do realize I was speaking against hand-rolled parsers in 
https://github.com/containers/skopeo/pull/524#discussion_r210649873 , but using a mismatching parser is not better.)

> Overall I feel using URL is adequate to parsing _transport_`:`_image_, if you take into account things like escaping and `dir:` non-absolute paths.

Well, but `dir:` _does not do escaping_ of any kind, so using a parser that does would interpret the input differently.

(Arguably `oci:` and `oci-archive:` _should_ do some escaping because the `oci:`_path_`:`_imageName` format can’t represent _path_ values that contain a `:`, but it’s quite unclear how to map that to an URL in a way that makes sense while still using the URL reserved characters vs. encoded characters to represent this.)

cmd/skopeo/sync.go
mtrmac: > The use case is the ability to specify multiple images (and any tags combination) hosted on multiple registries (each one with its own auth method) in a yaml file.

Sure, the config file allows doing that, but in what way is it materially different from writing a shell script just like your second example? Or, really,

```sh
for source in \
    docker://registry1/image1 \
    docker://registry1/image2 \
...
    docker://registry1/imageN \
    docker://registry2/image1 \
...
    docker://registryN/imageN \
    ; do
        skopeo sync [opts] "$source" dir:destDir
done
```

(I’m not saying the `yaml` config definitely does not make sense, just that it’s not obviously clear to me that it is worth defining and maintaining another stable file format.)

rhatdan:
Seems like this PR Has stalled again?

marcov:
@rhatdan from my side I'll plan to progress on this next week when I'll be back from sick leave.

docs/skopeo.1.md
mtrmac: > Some kind of scoping is necessary when running sync from a YAML file, as you may be pulling the same image name from different namespaces or domains.

My first instinct is to specify the destination in the YAML as well, then, not to hard-code the naming in software.

It’s not obvious to me why `sync --source-yaml $src $dest` would have the sources specified in a (presumably mostly fixed) config file but the destination in a (presumably mostly changing) destination.

If a user sets up a regular (e.g. weekly) sync process, won’t both the source and the destination the same for all runs (every week the same servers and same destinations)? (My best guess is that when the destination is on a removable medium, the physical device, and thus presumably the mount path, may change each time — but aren’t mounts often named by the volume label, which is always the same?)

mtrmac:
> We explained above the need for the namespace scoping when dealing with homonymous repositories on different registries.

(To make sure this is not lost) https://github.com/containers/skopeo/pull/524#discussion_r274606257 .

cmd/skopeo/sync.go
mtrmac: We could split `imageOptions` by transports, or pass something like `skipTransports []string` to `imageFlags`—but it’s very tempting to just not leave the options in and not do any filtering.

Let’s first figure out how to specify/parse the hierarchy roots; that might change how we refer to transports.

docs/skopeo-sync.1.md
mtrmac: Starting with a single sentence is good; most of this (like the “path traversal/walk” question) is better placed in the description of individual source/destination formats.

“What does ‘sync’ mean” should be explained somewhere around here, though.

marcov:
> I don’t have a good suggestion how to solve this; my best guess is not using the alltransports.ParseImageName format at all, and have sync explicitly say --docker-source --dir-source / --docker-dest / --dir-dest, or maybe --source-type docker / --dest-type docker, or something like that, and modulo all the bikeshedding about spelling and ordering and explicit options vs. positional parameters (vs. YAML); and then construct the docker.Reference / dir.Reference values with explicit knowledge of those particular transports.
>
> Better ideas are certainly welcome.

I'd like to propose something different, that would allow to keep using `docker` and `dir` transports to specify source and destinations (with one exception). 

| Source | Destination | Notes |
| --- | --- | --- |
| Repository<br>e.g.<br>`docker://busybox` | Repository<br>e.g.<br>`docker://localhost:5000/busybox` | Source and destination are parsed as docker references <br><br>(both of them must be untagged)|
| Repository<br>e.g.<br>`docker://busybox` | Directory<br>e.g.<br>`dir:$(pwd)/syncImages`| Source is a docker reference, destination a `dir:path` <br><br> (inside `path` there will be one directory per image tag)|
| Directory<br>e.g.<br>`dir:$(pwd)/syncImages`  | **Register path**<br>e.g.<br> `localhost:5000[/namespace]` | Destination is not a docker reference |

------------------

The only thing "special" here is that, when syncing from a directory, the destination is not a repository but a register path. 
This means the destination will not be specified with the `docker` transport prefix, and will not be parsed with `alltransports.ParseImageName`. It will be parsed as a URL (if that's ok for you), or any other way that fits better.

Let me know what you think.

cmd/skopeo/sync.go
marcov: > Sure, the config file allows doing that, but in what way is it materially different from writing a shell script just like your second example?

One advantage is that yaml can be generated from some other tool using common libraries, while the same is not so obvious for bash. 

E.g. we have an [Helm plugin to mirror Chart repositories](https://github.com/openSUSE/helm-mirror#helm-mirror-plugin). It generates yaml files that can be consumed by `skopeo sync`, so that also the images specified in the charts can be mirrored.

We thought it's more convenient to do it in this way, instead of having another tool just to read the list of images/ registers configuration and run `skopeo sync` from it.

marcov:
Hi @vrothberg, @mtrmac , 
I'd like to get a feedback on the proposal in my previous comment to advance with PR.

vrothberg:
I am getting back to this now. Sorry for the long silence, @marcov! My goal is to avoid a birthday party for this PR, so I raise the priority on my side.

@mtrmac @rhatdan PTAL as well

docs/skopeo-sync.1.md
vrothberg: This valid, please fix.

docs/skopeo-sync.1.md
vrothberg: This valid.

docs/skopeo-sync.1.md
vrothberg: Still valid.

docs/skopeo-sync.1.md
vrothberg: Ideally combine that with an exemplary `skopeo sync --source-yaml $file $registry` call to explain what's happening.

docs/skopeo-sync.1.md
```diff
@@ -0,0 +1,80 @@
...
+ - a local directory (eg: dir:/media/usb/).
+ - (when **--source-yaml** is specified) a YAML file with a list of source images from different container registries.
+
+When the source location is a container repository and no tags are specified,  **skopeo sync** will copy all the tags associated to the source image.
+
+_destination_ can be either a container registry (eg: docker://my-registry.local.lan) or a local directory (eg: dir:/media/usb).
+
+When _destination_ is a local directory, one directory per 'image:tag' will be created.
+
+## OPTIONS
```
vrothberg: Please make sure that the capitalization of the list is consistent. Some descriptions start with caps and some don't.

cmd/skopeo/sync.go
vrothberg: still valid: `e.g.,`

cmd/skopeo/sync.go
vrothberg: ping @TomSweeneyRedHat  :)

cmd/skopeo/utils.go
vrothberg: Still applies...

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,594 @@
...
+}
+
+type tlsVerifyConfig struct {
+	skip bool
+}
+
+type registrySyncCfg struct {
+	Images      map[string][]string
+	Credentials types.DockerAuthConfig
+	TLSVerify   tlsVerifyConfig `yaml:"tls-verify"`
```
vrothberg: I'd prefer making this `TLSSkipVerify` as security should something for opt-out rather for opt-in. Is a dedicated struct needed for that or can we just make it a bool?

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,594 @@
...
+	DirBasePath  string // base path when source is 'dir'
+	TaggedImages []types.ImageReference
+	Context      *types.SystemContext
+}
+
+type tlsVerifyConfig struct {
+	skip bool
+}
+
+type registrySyncCfg struct {
```
vrothberg: s/Cfg/Config/ to be consistent

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,594 @@
...
+	Images      map[string][]string
+	Credentials types.DockerAuthConfig
+	TLSVerify   tlsVerifyConfig `yaml:"tls-verify"`
+	CertDir     string          `yaml:"cert-dir"`
+}
+
+type sourceCfg map[string]registrySyncCfg
```
vrothberg: s/Cfg/Config/ to be consistent

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,594 @@
...
+	"github.com/containers/image/docker"
+	"github.com/containers/image/docker/reference"
+	"github.com/containers/image/transports"
+	"github.com/containers/image/types"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"github.com/urfave/cli"
+)
+
+type syncOptions struct {
```
vrothberg: Although not strictly needed for go-docs, I'd love to have comments for all the types and their fields where the use and intention is not obvious. It'll help with future maintenance.

marcov:
Hi @vrothberg, I first wanted to know what do you think about [my proposition](#issuecomment-484800652) of avoiding the (ab)use of `docker://` transport to identify a registry URL.
Once we agree on that, I'll update the PR to include fixes to the other details.

mtrmac:
> (inside `path` there will be one directory per image tag)
>
> | Source | Destination | Notes |
> | --- | --- | --- |
> | Directory e.g. `dir:$(pwd)/syncImages` | **Register path** e.g. `localhost:5000[/namespace]` | Destination is not a docker reference |
>
> The only thing "special" here is that, when syncing from a directory, the destination is not a repository but a register path.
> This means the destination will not be specified with the `docker` transport prefix, and will not be parsed with `alltransports.ParseImageName`. It will be parsed as a URL (if that's ok for you), or any other way that fits better.

1) It’s surprising that a source and destination for docker repositories use different syntaxes, all the more so when about every use of a docker repository as a source should eventually be parsed with use of a different repository as a destination.
2) This does not really say how the parser decides that the destination should be interpreted as a “registry path”. Is `dir:5000/foo` a “registry path”? Would this be done specifically by checking for `dir:` prefix as a special case? (if so, how does that evolve long term?) Would there be a `--destination-is-docker` option?

vrothberg:
@mtrmac, maybe we could add an `alltransports.ParseImageNameOffline` (naming to be improved) that would not ping the registry?

mtrmac:
`ParseImageName` does not ping anything (at least for `docker://`; it does do some FS operations on local FS paths, and various non-trivial lookups for `containers-storage`). The concern is short names vs. registry prefixes, and the inability to say “root of this namespace on a registry” using the `docker://` syntax, see your https://github.com/containers/skopeo/pull/524#pullrequestreview-210800367

vrothberg:
> `ParseImageName` does not ping anything (at least for `docker://`; it does do some FS operations on local FS paths, and various non-trivial lookups for `containers-storage`). The concern is short names vs. registry prefixes, and the inability to say “root of this namespace on a registry” using the `docker://` syntax, see your [#524 (review)](https://github.com/containers/skopeo/pull/524#pullrequestreview-210800367)

Let's say the unbearable heat impacting my concentration today, apologies. Thanks for helping me out!

Regarding the CLI, I prefer @mtrmac's proposal:

> --source-type docker / --dest-type docker [...] and then construct the docker.Reference / dir.Reference values with explicit knowledge of those particular transports.

I find it cleaner as there are no special cases which are always hard to understand and hard document. The heat is killing me but I am not sure that I could come up with a better proposal even if would snow.

mtrmac:
FWIW I would slightly prefer `--docker-source …`/`--dir-source`/… over `--source-type` because
1) it’s one less argument to pass every time
2) it would be much easier to write completions / validation / documentation for `--docker-source` than for a generic source (when `--source-type` might not yet have been provided by the user on the command line)
3) it makes it a bit more explicit that adding supports to new transports involves new design work

OTOH it is clearly less structurally elegant, e.g. verifying that the user has specified exactly one source and exactly one destination would be more annoying.

Ultimately, this is of course the same layer of indirection however we slice it, it’s “only” how we expose that indirection to users.

marcov:
Thanks for the suggestions @mtrmac, @vrothberg. Are you then implying to drop the `transport://` prefix altogether?
That's because  ` --docker-source …/--dir-source` still leaves the ambiguity of interpreting `docker://localhost:5000` as repo `localhost` with tag `5000`, instead of registry at domain `localhost` and port `5000`.

vrothberg:
> Thanks for the suggestions @mtrmac, @vrothberg. Are you then implying to drop the `transport://` prefix altogether?
> That's because ` --docker-source …/--dir-source` still leaves the ambiguity of interpreting `docker://localhost:5000` as repo `localhost` with tag `5000`, instead of registry at domain `localhost` and port `5000`.

Right, it being a source is still an issue. For destinations we expect it to be a registry or namespace, so we have the knowledge there. This means we need a way to make the difference between an image and a registry/namespace, explicit. How can we make this implicit? I see two options:

* We add yet another CLI-flag and yaml option. Personally, I'd like to avoid bloating the sync command further.
* We add a dedicated syntax for that. `localhost:5000/*`, for instance, should be universally understood.

marcov:
Sources are always repositories, so we should be fine (syncing registry to registry may be something worth adding later).

About destinations, that's correct what you say, it's either registry or registry + namespace. In both cases, we can't use `alltransports.ParseImageName`, so parsing will be done using strings tokenization.

mtrmac:
> Thanks for the suggestions @mtrmac, @vrothberg. Are you then implying to drop the `transport://` prefix altogether?

Yes; the ~whole point is to avoid the existing fixed `docker://` semantics, so that `skopeo sync` can define some other format.

> That's because ` --docker-source …/--dir-source` still leaves the ambiguity of interpreting `docker://localhost:5000` as repo `localhost` with tag `5000`, instead of registry at domain `localhost` and port `5000`.

The way I think about it, `--docker-source`/`--docker-destination` would (both using the same semantics, so that we don’t have three!) define an input format that is a good match for the `sync` operation, probably removing the `docker.io[/library]` normalization and requiring that all input explicitly starts with a host[:port], making this unambiguous.

marcov:
I force pushed a new version based on what we discussed until now. I have removed considerable parts of code (related to parsing URLs) and also simplified the logic.

I recognize it's quite a bit of code / docs to review, so I don't expect this done immediately.

Some insights:
- I introduced the `--src` and `--dest` flags that accept as values `docker`, `dir` and `yaml`.
- I introduced the `--scoped` flag, to include full source namespace when writing to destination.
- Source repository are still parsed with `alltransports.ParseImageName`, the way it works now is just fine.
- I introduced a new `dockerImageOptions` set of options, that are only applicable to the `docker` transport. In this way I don't need to filter out unneeded flags from `imageOptions`.

prashant182:
Is this PR in the pipeline to get merged? We have a situation where this PR can significantly help our use case.

vrothberg:
> Is this PR in the pipeline to get merged? We have a situation where this PR can significantly help our use case.

@prashant182, at some point it will. It's a rather big PR and we're currently under load but will review asap. Happy to see the sync functionality would help your use case!

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,545 @@
...
+
+	if len(opts.destination) == 0 {
+		return errors.New("A destination location must be specified")
+	}
+	if !contains(opts.destination, []string{docker.Transport.Name(), directory.Transport.Name()}) {
+		return errors.Errorf("%q is not a valid destination location", opts.destination)
+	}
+
+	if opts.source == opts.destination && opts.source == directory.Transport.Name() {
+		return errors.New("sync from 'dir' to 'dir' not implemented, use something like rsync instead")
```
vrothberg: "[...], consider using rsync instead"

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,545 @@
...
+	destinationCtx, err := opts.destImage.newSystemContext()
+	if err != nil {
+		return err
+	}
+
+	ctx, cancel := opts.global.commandTimeoutContext()
+	defer cancel()
+	var imagesNumber int
+	for _, srcRepo := range srcRepoList {
+		options := copy.Options{
```
vrothberg: This can be moved outside the loop, so we only create it once.

docs/skopeo-sync.1.md
```diff
@@ -0,0 +1,131 @@
...
+```
+$ skopeo sync --src docker --dest dir registry.example.com/busybox /media/usb
+```
+Images are located at:
+```
+/media/busybox:1-glibc
+/media/busybox:1-musl
+/media/busybox:1-ubuntu
+...
+/media/busybox:latest
```
vrothberg: I think this should start with `/media/usb/` (same above).

docs/skopeo-sync.1.md
```diff
@@ -0,0 +1,131 @@
...
+
+**--src-no-creds** _bool-value_ Access the registry anonymously.
+
+**--src-tls-verify** _bool-value_ Require HTTPS and verify certificates when talking to container source registry or daemon (defaults to true)
+
+**--dest-cert-dir** _path_ Use certificates at _path_ (*.crt, *.cert, *.key) to connect to the destination registry or daemon
+
+**--dest-no-creds** _bool-value_  Access the registry anonymously.
+
+**--dest-tls-verify** _bool-value_ Require HTTPS and verify certificates when talking to container destination registry or daemon (defaults to true)
```
vrothberg: `--scoped` is missing here.

docs/skopeo-sync.1.md
```diff
@@ -0,0 +1,131 @@
...
+ - _yaml_ (i.e. `--src yaml`): _source_ is local YAML file path.
+ The YAML file should specify the list of images copied from different container registries (local directories are not supported). Refer to EXAMPLES for the file format.
+
+Available _destination_ locations:
+ - _docker_ (i.e. `--src docker`): _destination_ is a container registry (e.g.: `my-registry.local.lan`).
+ - _dir_ (i.e. `--src dir`): _destination_ is a local directory path (e.g.: `/media/usb/`).
+ One directory per source 'image:tag' is created for each copied image.
+
+When the `--scoped` option is specified, images are prefixed at _destination_ with the source image path, so that mutiple images with the same
+name can be contained at _destination_.
```
vrothberg: Assuming `--scoped` makes it into the PR: we need to document some shortcoming when using a docker destination. Again the reference parsing is biting us. Assuming the source has a `:port`, it cannot be used with `--scoped`.

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,545 @@
...
+
+	opts := syncOptions{
+		global:    global,
+		srcImage:  srcOpts,
+		destImage: &imageDestOptions{imageOptions: destOpts},
+	}
+
+	return cli.Command{
+		Name:  "sync",
+		Usage: "Sync one or more images from one location to another",
```
TomSweeneyRedHat: suggest  s/"Sync/"Synchronize/

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,545 @@
...
+		Description: fmt.Sprint(`
+
+	Copy all the images from a SOURCE to a DESTINATION.
+
+	Useful to keep in sync a local container registry mirror, and to to populate
```
TomSweeneyRedHat: suggest s/keep in sync/synchronize/

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,545 @@
...
+// It returns a new unmarshaled sourceCfg object and any error encountered.
+func newSourceConfig(yamlFile string) (cfg sourceCfg, err error) {
+	err = yamlUnmarshal(yamlFile, &cfg)
+	return
+}
+
+// destinationReference creates an image reference using the provided location.
+// It returns a image reference to be used as destination of an image copy and
+// any error encountered.
+func destinationReference(destination string, location string) (types.ImageReference, error) {
```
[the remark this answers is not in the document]
TomSweeneyRedHat: and touch up the comment when the variable name changes.

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,545 @@
...
+	case docker.Transport.Name():
+		destination = fmt.Sprintf("//%s", destination)
+		transport = docker.Transport
+	case directory.Transport.Name():
+		_, err := os.Stat(destination)
+		if err == nil {
+			return nil, errors.Errorf(fmt.Sprintf("Refusing to overwrite destination directory %q", destination))
+		}
+		if !os.IsNotExist(err) {
+			return nil, errors.Wrap(err, "Destination directory could not be used")
```
TomSweeneyRedHat: s/not be used/not be found/

cmd/skopeo/sync.go
```diff
@@ -0,0 +1,545 @@
...
+		if !strings.Contains(err.Error(), "401") {
+			return tags, errors.Wrapf(err, fmt.Sprintf("Error determining repository tags for image %s", name))
+		}
+		logrus.Warnf("Registry disallows tag list retrieval: %s", err)
+	}
+
+	return tags, nil
+}
+
+// isTagspecified checks if an image name includes a tag and returns any errors
```
TomSweeneyRedHat: nit s/isTagspecified/isTagSpecified/

docs/skopeo-sync.1.md
```diff
@@ -0,0 +1,131 @@
+% skopeo-sync(1)
+
+## NAME
+skopeo\-sync - Synchronize images between container registries and local directories.
+
+
+## SYNOPSIS
+**skopeo sync** [**sync-options**] --src _source-location_ --dest _destination-location_ _source_ _destination_
```
TomSweeneyRedHat: I'd remove "[**sync-options**]"  Most of our pages don't add that, it's implied that the options go there.

docs/skopeo-sync.1.md
TomSweeneyRedHat: As is --src and --destination.  I'd also alphabetize the options.

docs/skopeo-sync.1.md
```diff
@@ -0,0 +1,131 @@
...
+
+## DESCRIPTION
+Synchronize images between container registries and local directories.
+The synchronization is obtained by copying all the images found at _source_ to _destination_.
```
TomSweeneyRedHat: s/obtained/achieved/

docs/skopeo-sync.1.md
TomSweeneyRedHat: suggest `s/prefixed at _desitnation_ with the source image path, /with the source image path /` (removed comma too)

docs/skopeo-sync.1.md
```diff
@@ -0,0 +1,131 @@
...
+ - _dir_ (i.e. `--src dir`): _destination_ is a local directory path (e.g.: `/media/usb/`).
+ One directory per source 'image:tag' is created for each copied image.
+
+When the `--scoped` option is specified, images are prefixed at _destination_ with the source image path, so that mutiple images with the same
+name can be contained at _destination_.
+
+## OPTIONS
+**--authfile** _path_
+
+Path of the authentication file. Default is ${XDG_RUNTIME\_DIR}/containers/auth.json, which is set using `podman login`.
```
TomSweeneyRedHat: `s/${XDG_RUNTIME\_DIR}/${XDG\_RUNTIME\_DIR}/`

docs/skopeo-sync.1.md
```diff
@@ -0,0 +1,131 @@
...
+If the authorization state is not found there, $HOME/.docker/config.json is checked, which is set using `docker login`.
+
+**--remove-signatures** Do not copy signatures, if any, from _source-image_. Necessary when copying a signed image to a destination which does not support signatures.
```
TomSweeneyRedHat: s/Necessary/This is necessary/

docs/skopeo-sync.1.md
```diff
@@ -0,0 +1,131 @@
...
+
+**--sign-by=**_key-id_ Add a signature using that key ID for an image name corresponding to _destination-image_.
+
+**--src-creds** _username[:password]_ for accessing the source registry
+
+**--dest-creds** _username[:password]_ for accessing the destination registry
+
+**--src-cert-dir** _path_ Use certificates at _path_ (*.crt, *.cert, *.key) to connect to the source registry or daemon
```
TomSweeneyRedHat: missing end period for three lines above

docs/skopeo-sync.1.md
TomSweeneyRedHat: Half the lines have an ending period, half don't.  I'd add one to all.

docs/skopeo-sync.1.md
```diff
@@ -0,0 +1,131 @@
...
+
+**--src-no-creds** _bool-value_ Access the registry anonymously.
+
+**--src-tls-verify** _bool-value_ Require HTTPS and verify certificates when talking to container source registry or daemon (defaults to true)
```
TomSweeneyRedHat: s/verify certificates when talking to/verifies certificates when talking to a/

docs/skopeo-sync.1.md
TomSweeneyRedHat: s/verify certificates when talking to/verifies certificates when talking to a/

docs/skopeo-sync.1.md
```diff
@@ -0,0 +1,131 @@
...
+
+**--dest-cert-dir** _path_ Use certificates at _path_ (*.crt, *.cert, *.key) to connect to the destination registry or daemon
```
TomSweeneyRedHat: suggest `s/Use certificates at _path_ (*.crt, *.cert, *.key)/Use certificates (*.crt, *.cert, *.key) at _path_/`

docs/skopeo-sync.1.md
TomSweeneyRedHat: suggest `s/Use certificates at _path_ (*.crt, *.cert, *.key)/Use certificates (*.crt, *.cert, *.key) at _path_/`

docs/skopeo-sync.1.md
TomSweeneyRedHat: s/verify certificates when talking to/verifies certificates when talking to a/

docs/skopeo-sync.1.md
```diff
@@ -0,0 +1,131 @@
...
+    cert-dir: /home/john/certs
+quay.io:
+    tls-verify: false
+    images:
+        coreos/etcd:
+            - latest
+```
+
+This will copy the following images:
+- Repository `registry.example.com/busybox`: all images.
```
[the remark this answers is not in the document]
TomSweeneyRedHat: nit to previous comment s/as not/as no/

edsantiago:
All my comments have already been stated except one: would you consider adding a short sentence or two to the man page `DESCRIPTION` summarizing the October comments in this thread, explaining why it's necessary to use `--src/--dest` options instead of the more conventional prefixes? I find those awkward, and believe that others may appreciate a little insight into the problem they solve. Thank you.

cmd/skopeo/utils.go
marcov: Repetita iuvant :-)

cmd/skopeo/sync.go
marcov: "not be used" is correct, mind the double negation :-)

marcov:
Thanks for the reviews @vrothberg @TomSweeneyRedHat @edsantiago

I've addressed your comments, they're all valid.

marcov:
@mtrmac did you have a chance to review these latest changes?

sagikazarmark:
Thanks for the PR and the [blog post](https://flavio.castelli.me/2018/07/18/hackweek-project-docker-registry-mirror/)! It's great and fits our use case perfectly.

Is there an ETA for merging this PR? Is it stable enough to use a custom build until it gets merged?

thkukuk:
At least the openSUSE/SUSE registries don't have different names for every architecture, which can lead to the fact, that not every tag is available for all architectures. skopeo sync aborts in such a case. Is there any way to tell skopeo to ignore this tag in this case?

FATA[0001] Error copying tag "docker://registry.opensuse.org/opensuse/busybox:1.30.1-4.67": Error choosing an image from manifest list docker://registry.opensuse.org/opensuse/busybox:1.30.1-4.67: no image found in manifest list for architecture amd64, OS linux

vrothberg:
> At least the openSUSE/SUSE registries don't have different names for every architecture, which can lead to the fact, that not every tag is available for all architectures. skopeo sync aborts in such a case. Is there any way to tell skopeo to ignore this tag in this case?

I guess that's something `skopeo-sync` has to do. I guess that @nalind's PR to properly support manifest lists (i.e., https://github.com/containers/image/pull/400) could help.

nalind:
Yeah, I think setting `ImageListSelection: copy.CopyAllImages` in the `copy.Options` created in `syncOptions.Run` would be enough to copy tags that point to lists as lists+images, and non-lists as non-lists.  The `docker` and `dir` transports don't return errors indicating that they can't handle lists, so you shouldn't need to add any logic to decide what to do in that situation.

rhatdan:
@flavio Needs a rebase.

flavio:
Sorry for the delay. I'll take care of that.

cc @vrothberg and @rhatdan

rhatdan:
@flavio Still needs a rebase.

flavio:
@rhatdan and @vrothberg the rebase is done now :)

vrothberg:
@flavio, can you rerun `make vendor` and commit the change?
```
tree is dirty, please commit all changes and sync the vendor.conf

 M go.mod
```

flavio:
@vrothberg done, sorry about the hiccup

vrothberg:
Thanks, @flavio! No need to update the branch, we can rebase and merge via GitHub.

vrothberg:
@flavio, one last time, now it's `vendor/modules.txt`. Are you using go 1.13 for vendoring? In case not, there's a `make vendor-in-container` now that uses `golang:1.13` for vendoring.

flavio:
@vrothberg you're right. I was using go 1.12. I've followed your advice, hopefully everything is good now :crossed_fingers:

vrothberg:
@flavio not yet. I assume that encryption PR caused a conflict with the go.{mod,sum}. Sorry for that!

flavio:
@vrothberg fixed again. I :heart: go modules ;)

Let me know if you want me to squash some commits together to clean up the history

vrothberg:
> @vrothberg fixed again. I heart go modules wink

It's a very bumpy ride with go but I hope things will calm down in 2020 :D

> Let me know if you want me to squash some commits together to clean up the history

Good point, sorry for missing that. Yes, please squash them into one commit. One commit can avoid some noise during bisects.

flavio:
@vrothberg I squashed all the commits into a single one, except for https://github.com/containers/skopeo/pull/524/commits/544fd2ee3b820b9dd074ab755eef43ac0decdbef

I think that was done to make the test pass, but it doesn't seem strictly related with the `sync` command. I think it would be better to not squash it. What do you think?

vrothberg:
SGTM

vrothberg:
Test failures look legit. Feel free to ping me if you want me to dig deeper.

flavio:
Don't worry, I'm looking into them

flavio:
Updated to fix the broken integration tests. Everything should be green, at least it looks like that when I run the checks locally :crossed_fingers:

rhatdan:
@vrothberg Is this ready to go in?

vrothberg:
Thanks to @flavio and @marcov for the long breath and the great work!

rhatdan:
Thanks @flavio and @marcov for getting this in.

marcov:
Happy to see this merged at last! 🎉

flavio:
Thanks @vrothberg and @rhatdan for the assistance!

saschagrunert:
Do we want to release a new version including this change :innocent: ?

vrothberg:
> Do we want to release a new version including this change innocent ?

Maybe in January. We wanted the new changes to cook a bit longer in master until releasing it.

vrothberg:
There's a couple of other things (e.g., login/logout support) that would be good to get in as well.

PR #535, opened 2018-08-13
Add support for using podman in testing skopeo

rhatdan:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,7 @@ SIGSTOREDIR=${DESTDIR}/var/lib/atomic/sigstore
 BASHINSTALLDIR=${PREFIX}/share/bash-completion/completions
 GO_MD2MAN ?= go-md2man
 GO ?= go
+CONTAINER_RUNTIME := $(shell command -v podman 2> /dev/null | echo docker)
 
 ifeq ($(DEBUG), 1)
   override GOGCFLAGS += -N -l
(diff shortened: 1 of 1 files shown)


rhatdan:
@mtrmac @runcom PTAL

rhatdan:
Ok how about this version which will use podman if it exists and then uses docker if it does not.

rhatdan:
@mtrmac PTAL

Makefile
```diff
@@ -65,27 +66,27 @@ endif
...
 
 binary-static: cmd/skopeo
-	docker build ${DOCKER_BUILD_ARGS} -f Dockerfile.build -t skopeobuildimage .
-	docker run --rm --security-opt label:disable -v $$(pwd):/src/github.com/containers/skopeo \
+	${CONTAINER_RUNTIME} build ${CONTAINER_BUILD_ARGS} -f Containerfile.build -t skopeobuildimage .
+	${CONTAINER_RUNTIME} run --rm --security-opt label:disable -v $$(pwd):/src/github.com/containers/skopeo \
 		skopeobuildimage make binary-local-static $(if $(DEBUG),DEBUG=$(DEBUG)) BUILDTAGS='$(BUILDTAGS)'
 
-# Build w/o using Docker containers
+# Build w/o using Container containers
```
mtrmac: This redundantly repeats itself :)

Makefile
```diff
@@ -65,27 +66,27 @@ endif
 #           use source debugging tools like delve.
 all: binary docs
 
-# Build a docker image (skopeobuild) that has everything we need to build.
+# Build a container image (skopeobuild) that has everything we need to build.
 # Then do the build and the output (skopeo) should appear in current dir
 binary: cmd/skopeo
-	docker build ${DOCKER_BUILD_ARGS} -f Dockerfile.build -t skopeobuildimage .
-	docker run --rm --security-opt label:disable -v $$(pwd):/src/github.com/containers/skopeo \
+	${CONTAINER_RUNTIME} build ${CONTAINER_BUILD_ARGS} -f Containerfile.build -t skopeobuildimage .
```
mtrmac: `Containerfile.build`? Heh. I can live with that, but this PR needs to actually rename the file or these targets would break.

OTOH there is also a plain `Dockerfile`, used by the `build-container` rule, and renaming that one is a bit more awkward. Sure, not impossible — either way, the two inputs to `$CONTAINER_RUNTIME build` should probably be named consistently.

Makefile
rhatdan: Fixed.

Makefile
rhatdan: These should still be Dockerfile.  A little too liberal with find/replace

Fixed

rhatdan: We can look into changing the Dockerfile afterwards to make sure podman will work inside of the container.

mtrmac:
👍 , please rebase.

rhatdan:
@mtrmac @runcom ready to merge.

PR #537, opened 2018-08-16
Makefile: make vendor a .PHONY target

giuseppe:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: all binary build-container build-local clean install install-binary install-completions shell test-integration
+.PHONY: all binary build-container docs build-local clean install install-binary install-completions shell test-integration vendor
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -146,5 +145,6 @@ validate-local:
 test-unit-local:
 	$(GPGME_ENV) $(GO) test -tags "$(BUILDTAGS)" $$($(GO) list -tags "$(BUILDTAGS)" -e ./... | grep -v '^github\.com/containers/skopeo/\(integration\|vendor/.*\)$$')
 
+.PHONY: vendor
```
rhatdan: Why add this in two places?

giuseppe: no one is enough :-) I refactored it but forgot to remove the old one

mtrmac:
👍

PR #538, opened 2018-08-20
Allow specifying fine grained install targets

CameronNemo:
One may wish to omit the bash completions, docs, or even the binary. For example if these are installed at a different point in the build process, not using the Makefile.

I specifically want this because I am seeing the following error while building a package for Void Linux.

    make: *** No rule to make target 'skopeo', needed by 'install-binary'.  Stop.

The binary is built in a different location (GOPATH) and installed before `make install` is invoked.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: all binary build-container docs build-local clean install install-binary install-completions shell test-integration vendor
+.PHONY: all binary build-container docs build-local clean install install-defaults install-binary install-docs install-completions shell test-integration vendor
 
 export GO15VENDOREXPERIMENT=1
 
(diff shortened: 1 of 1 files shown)


rhatdan:
@CameronNemo You should sign your commit, but this LGTM
@runcom @mtrmac PTAL

TomSweeneyRedHat:
LGTM

mtrmac:
Thanks for the PR.

What are you _ultimately_ trying to do? There might be another way to do it.

> The binary is built in a different location (GOPATH) and installed before `make install` is invoked.

- The repository already only works when checked out in `$GOPATH`, so I’m not sure what “a different location `(GOPATH)`” means.
- If you are building the binary in some non-default way, I’d expect that you are already patching `Makefile`, to keep up with the not-entirely-trivial and possibly changing commands used in `make binary`* — and at that point it seems reasonable to patch the installation path equivalently. Or are you running the compilation completely out of control of the `Makefile`, and if so, why?

CameronNemo:
The latter. It runs go install and the binary ends up at bin/skopeo.

PR #539, opened 2018-08-21
Add tls-verify to copy to toggle both src and dest

TomSweeneyRedHat:

Add tls-verify to the copy command so the end user doesn't have to specify both a destination and source tls if they both have the same setting.


mtrmac:
Not sure about this… it really _should_ be more difficult to disable TLS for _two_ servers than for one, IMHO; otherwise the easiest thing would be to disable both when the user really needs to only disable one.

Meanwhile, the deprecated `skopeo --tls-verify=false copy` still works and is not likely to go away.

PR #540, opened 2018-08-23
[WIP][RFC] skopeo: add support for pulling/seeding images through BitTorrent

giuseppe:
Add support for pulling and seeding images through BitTorrent.

Multiple images can be seed at the same time, for each of them it is necessary to specify the remote location and the local storage where they are stored.  It keeps seeding until SIGTERM is received.

```console
$ skopeo copy docker://quay.io/giuseppe/hello-world dir:/tmp/seed
$ skopeo seed --torrent-trackers udp4://192.168.125.222:6969 \
    docker://quay.io/giuseppe/hello-world dir:/tmp/seed

```

For pulling an image using BitTorrent.

```console
$ skopeo copy --torrent-trackers udp4://192.168.125.222:6969 \
   --try-torrent docker://quay.io/giuseppe/hello-world dir:/tmp/test
```

I am using an internal BitTorrent tracker as I am hitting an issue with the quay.io tracker, as I always get " "ERROR 503: Service Unavailable: Back-end server is at capacity" /cc @jzelinskie


cmd/skopeo/seed.go
```diff
@@ -0,0 +1,91 @@
...
+		if err != nil {
+			return fmt.Errorf("Invalid source image %s: %v", c.Args()[off], err)
+		}
+
+		refSrc, err := alltransports.ParseImageName(c.Args()[off+1])
+		if err != nil {
+			return fmt.Errorf("Invalid storage image %s: %v", c.Args()[off+1], err)
+		}
+
+		if err := client.Seed(context.Background(), ctx, ref, refSrc); err != nil {
```
mtrmac: If I am reading the instructions right:
> ```console
> $ skopeo copy docker://quay.io/giuseppe/hello-world dir:/tmp/seed
> ```

This copies the image to a local directory.

> ```console
> $ skopeo seed --torrent-trackers udp4://192.168.125.222:6969
> ```

… and this makes _yet another_ copy of the blobs inside `torrent.Client.dir`. Is that really necessary to do both?

It does sound useful to seed an image from a local source for the benefit of a remote destination… maybe it’s just the instructions that could be updated to show that the `dir:` copy is optional; or am I misunderstanding this?

giuseppe: no another copy is done.  It probably makes no sense to try to support other transports, and we can force `dir:` is used?  In this case I can try to tweak the torrent client to read directly from the `dir:` storage

cmd/skopeo/seed.go
```diff
@@ -0,0 +1,91 @@
...
+			}
+		}()
+	}
+
+	s := make(chan os.Signal)
+	signal.Notify(s, os.Interrupt, syscall.SIGTERM)
+	<-s
+	logrus.Info("got signal, shut down")
+	client.Close()
+	os.Exit(1)
```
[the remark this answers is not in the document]
giuseppe: I am using this so that we don't leave temporary files around, as `skopeo seed` never exits, but if we can read directly from the `dir:` storage so that we don't need another copy, then it should be fine to not handle the signal here

cmd/skopeo/seed.go
mtrmac: I’m not quite thrilled about code reaching into the internals of `dir:` (I assume the seed code requires random access, so `GetBlob` is insufficient); if this works just as well with

```console
dockerImageName=docker://quay.io/giuseppe/hello-world
skopeo seed --torrent-trackers udp4://192.168.125.222:6969 \
    $dockerImageName $dockerImageName
```
, with only the initial copy taking some time, then it’s just a matter of documentation for me.

cmd/skopeo/seed.go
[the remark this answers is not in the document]
mtrmac: I am perfectly fine with doing all necessary cleanup; all I wanted to suggest is that this should just return `nil`, let all other cleanups, if any, happen, and have the process return the default exit status 0.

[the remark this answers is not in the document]
mtrmac: > I am perfectly fine with doing all necessary cleanup

… and waiting for a signal.

cmd/skopeo/seed.go
giuseppe: what I don't really like is how to specify the fully qualified image, as we need to access the tracker there or at least download the .torrent file (e.g. http://quay.io/c1/torrent/giuseppe/hello-world/blobs/sha256:a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4)

So we need two arguments for each image, the fully qualified name and the local storage.

Is there a better way for doing it?

giuseppe: /cc @rhatdan

mtrmac: \*shrug\* I’m fine with specifying two locations for source/destination (after all `skopeo copy` requires two locations as well, and this is conceptually fairly similar), I was really only speaking up against the _example_ causing two copies when one would do.

giuseppe:
@jzelinskie do you know who I can ping for help with the quay.io registry and the "ERROR 503: Service Unavailable: Back-end server is at capacity" error I am seeing there?

jzelinskie:
@giuseppe I'm back from vacation and can take a look. The Quay.io instance of Chihaya has been firing off alarms recently, so I need to investigate our production stack.

jzelinskie:
@giuseppe got everything fixed with our deployment

giuseppe:
@jzelinskie thanks a lot.  It works now.  If anyone is interested to play with it, there is no need to setup a different tracker and it is enough to:

```console
$ skopeo copy --try-torrent docker://quay.io/giuseppe/hello-world dir:/tmp/seed
```

and to seed the layers:
```
$ skopeo seed docker://quay.io/giuseppe/hello-world dir:/tmp/seed
```

rhatdan:
How much did the skopeo package grow in size based off this patch?

giuseppe:
> How much did the skopeo package grow in size based off this patch?

this has a big impact on the binary size: from ~26Mb to ~38Mb

giuseppe:
without debug info the difference is ~16Mb vs ~21Mb

giuseppe:
I'm still having problems with the quay tracker, can't start any pull, I'll need to investigate it more but from a quick check I am having the same problem using `quayctl`

giuseppe:
@jzelinskie not sure what I am doing wrong but I've tried setting up two DigitalOcean droplets one for seeding and one for pulling using quayctl but the pull always fall back to HTTP, it works if I use a custom tracker.

rhatdan:
@giuseppe Needs rebase.  Are you still working on this?  Seems like this would be a nice feature,

giuseppe:
I am still blocked with quay.io, last time I tried torrents were not working

jzelinskie:
@giuseppe I can take a look at this on Monday to unblock you. Do you have steps to reproduce?

giuseppe:
> @giuseppe I can take a look at this on Monday to unblock you. Do you have steps to reproduce?

@jzelinskie thanks.  What I am trying to manually do is, fetch the torrent for a layer, e.g.
https://quay.io/c1/torrent/giuseppe/hello-world/blobs/sha256:e71c36a80ba912dd7a5a9f2f2d6136c148afa19bc7d024bd616b74a0bc7a2774

then tried a few different clients for downloading it.  I'm also seeding at the same time from another client, but still the download doesn't start.

jzelinskie:
@giuseppe There was a misconfiguration in the tracker deployment that made the tracker fail authentication. I tweaked the config and things appear to be working for me now.

giuseppe:
thanks, it seems to work now.  I've fixed a couple of issues in the `containers/image` implementation and pushed a new version.

rhatdan:
I was talking to a customer about this issue at Kubecon last week, and they were very interested in this feature.  They were looking at this from very large scale.  Basically wanting to seed 1000 machines simultaneously.  Does it become more valuable for that use case?

jzelinskie:
@rhatdan Apologies. I saw you out of the corner of my eye at Kubecon, but totally missed a chance to chat with you about this. There really isn't any specification around this work, so a client will pretty much only work with Quay. We've separated every layer into its own individual BitTorrent swarm, which could be different from how others would like to implement it.

That being said, it would be awesome to get this client work merged into Skopeo so that we can deprecate [quayctl](https://github.com/coreos/quayctl) which exists solely to facilitate hackery loading BitTorrent images into the docker CAS.

rhatdan:
@jzelinskie Cool, if we can get to the point that Skopeo with particular Registries like Quay work well, we might be able to convince others to support it.   
@giuseppe Needs a rebase.

rhatdan:
@giuseppe @vrothberg @mtrmac What is the state of this PR?  Do we want to move forward with it?

PR #541, opened 2018-08-27
Allow passing TESTFLAGS env to hack/make.sh from make

marcov:
Minor Makefile change to allow passing the env TESTFLAGS to make. That's pretty
convenient to filter what tests to run.
E.g. run integration tests containing the substring `Copy`:
make test-integration TESTFLAGS="-check.f Copy"


mtrmac:
👍 Thanks!

PR #552, opened 2018-09-17
Fix (make DISABLE_CGO=1)

### LINKED ISSUE #546: make binary-static DISABLE_CGO=1 fails
I tried to make a static binary per docs and failed: 
`taras@dev-taras:~/skopeo$  make binary-static DISABLE_CGO=1`

```docker run --rm --security-opt label:disable -v $(pwd):/src/github.com/containers/skopeo \
        golang:1.10 make binary-local-static  BUILDTAGS='containers_image_ostree_stub exclude_graphdriver_devicemapper exclude_graphdriver_btrfs containers_image_openpgp'
make: *** No rule to make target 'binary-local-static'.  Stop.
Makefile:83: recipe for target 'binary-static' failed
make: *** [binary-static] Error 2
```

mtrmac:
... which has, apparently, never worked, because the `golang` image has neither the `GOPATH` nor the working directory the `Makefile` expects.

Rather than move all this configuration into the `Makefile` to be able to work with the `golang` images, just always use the `skopeobuildimage` path, and only override the tags, to minimize divergence.

Fixes #546.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,9 +55,7 @@ LIBDM_BUILD_TAG = $(shell hack/libdm_tag.sh)
 LOCAL_BUILD_TAGS = $(BTRFS_BUILD_TAG) $(LIBDM_BUILD_TAG) $(DARWIN_BUILD_TAG)
 BUILDTAGS += $(LOCAL_BUILD_TAGS)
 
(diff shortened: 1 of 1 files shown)


runcom:
LGTM

PR #558, opened 2018-10-11
DO NOT MERGE: update containers/image

nalind:
`github.com/containers/image/copy.Image()` now returns the copied manifest (as of https://github.com/containers/image/pull/519), so we at least need to ignore it.


mtrmac:
👍 , thanks!

Let’s do the rebase/merge tomorrow to minimize the CI breakage time.

PR #565, opened 2018-10-31
add timeout support for image copy

armstrongli:
the copy sub command may stuck if there are some network issue(e.g. slow network cause of network flapping in infrastructure), skopeo stuck there and application cannot do anything. It results in all the workers hang and cannot continue(retry, or report failure). 

Add the command-timeout option to make it possible to terminate the command to avoid this case.

Test result:
```
./skopeo copy --command-timeout=2s  --src-tls-verify=true --src-creds ****:***** docker://ubuntu:14.04 dir:/tmp/*****/raw
Copying blob sha256:*******
 4.53 MB / 48.98 MB [=====>-------------------------------------------------] 1s
FATA[0002] Error writing blob: context deadline exceeded
```

### DIFF
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"context"
 	"errors"
 	"fmt"
 	"os"
@@ -88,7 +87,10 @@ func copyHandler(c *cli.Context) error {
 		}
 	}
 
-	_, err = copy.Image(context.Background(), policyContext, destRef, srcRef, &copy.Options{
+	ctx, cancel := commandTimeoutContextFromGlobalOptions(c)
+	defer cancel()
+
+	_, err = copy.Image(ctx, policyContext, destRef, srcRef, &copy.Options{
 		RemoveSignatures:      removeSignatures,
 		SignBy:                signBy,
 		ReportWriter:          os.Stdout,
--- a/cmd/skopeo/delete.go
+++ b/cmd/skopeo/delete.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"context"
 	"errors"
 	"fmt"
 	"strings"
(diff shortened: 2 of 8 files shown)


cmd/skopeo/copy.go
```diff
@@ -192,5 +201,10 @@ var copyCmd = cli.Command{
 			Value: "",
 			Usage: "use docker daemon host at `HOST` (docker-daemon destinations only)",
 		},
+		cli.DurationFlag{
+			Name:  "command-timeout",
+			Value: 30 * time.Minute,
```
rhatdan: I don't think this is ever used?

rhatdan:
This would require modifications of man pages, command completions.

cmd/skopeo/copy.go
armstrongli: the command is used by the jenkins to do image validation and other scan work. The command hangs and cause all the jenkins jobs stuck. 

For general command line usage, it is not very often used.

armstrongli:
@rhatdan comments addressed. please take a look

armstrongli:
@mtrmac 
Thank you for your feedback. Agree that there are other options to get the hang issue resolved, it would be nicer to get `skopeo` client support `command-timeout`.   With the option, operator can directly get to know what is the problem and resolve the issue with less effort. 

like the etcd client - `etcdctl` - has the command option. it allows operator to get the command to run with regular options.

mtrmac:
Fair enough.  Then, two design questions:

- Is 30 minutes (or any timeout at all, really) a reasonable default? It’s not obviously unreasonable, at least.
- Should this be an option specific to `skopeo copy`, or a global option applicable to all subcommands? The latter seems a bit preferable.

@rhatdan , others, WDYT?

armstrongli:
@mtrmac . make sense. will update the code to add it into global command option.

rhatdan:
Yes I think the default should be infinite.
LGTM

rhatdan:
Thanks @armstrongli

PR #584, opened 2019-01-07
[DO NOT MERGE] test new c/image dependency

Silvanoc:
This PR adds the dependency github.com/coreos/pkg needed by https://github.com/containers/image/pull/546 to get the tests passed according to [the documentation](https://github.com/containers/skopeo/blob/master/README.md#dependencies-management).


mtrmac:
ACK on the implementation (though I don’t really love the approach — but let’s discuss that in #546). Thanks!

Silvanoc:
Not needed after closing containers/image#546.

PR #596, opened 2019-01-23
completions: Fix bash completions when a option requires a argument

eramoto:
Since the string of options variable as pattern in the case statement has not been delimited and it does not match the value of prev variable, bash completions tries to complement any option even when a specified option requires a argument.
This fix stops complementing options when a option requires a argument.


rhatdan:
LGTM

rhatdan:
@vrothberg @TomSweeneyRedHat @runcom @mtrmac PTAL

vrothberg:
Thanks for the contribution, @eramoto! It looks like commit 47f24b4 slipped in by accident. Can you clean this up? The code looks good to me, but I want to test it quickly on my machine.

vrothberg:
Works fine :+1:

PR #600, opened 2019-02-07
Vendor latest github.com/containers/storage

nalind:
Update github.com/containers/storage to master(be846769b4977f3ae0c0256d50827440e72fc0b0).  This is largely about getting the changes from https://github.com/containers/storage/pull/283 here so that when we introduce a patch to containers/image that depends on the newer fields that PR introduced, its tests will pass.


rhatdan:
LGTM, although we should probably move to a new release of containers/storage.

nalind:
Hmm, I think we're going to need https://github.com/containers/storage/pull/284, too.

nalind:
Updated to 06b6c2e4cf254f5922a79da058c94ac2a65bb92f.

mtrmac:
LGTM pending tests. Thanks!

PR #602, opened 2019-02-09
update containers/image

vrothberg:
* new progress bar library


vrothberg:
That's the CI PR for https://github.com/containers/image/pull/581. @mtrmac @rhatdan please take a look and please also play with it to make sure the output satisfies your expectations.

giuseppe:
I've tried to pull `jboss/wildfly` with skopeo.  After the first layers are copied in parallel, there is some flickering between the last layers and the config file, with the progress bars writing on each other

vrothberg:
> I've tried to pull `jboss/wildfly` with skopeo. After the first layers are copied in parallel, there is some flickering between the last layers and the config file, with the progress bars writing on each other

Thank you so much for checking, @giuseppe. I don't see those flakes on my machine but believe to have an idea how to avoid that :) Currently, we create two progress objects (one for blobs, one for the config). I guess having one will solve your issue.

giuseppe:
it works now!

giuseppe:
LGTM

vrothberg:
Eugh. I need a beer after that :-D

vrothberg:
@rhatdan @giuseppe CI is green again, and the output works on my machine. Please check on yours and let's get it merged after that.

The multi-bar usage is quite challenging, not only because the libraries have certain expectations (some documented, some not) but also our code is a bit spaghetti, so it's hard to let the progress bars know when things are finished (see the `bar.SetTotal(...)` calls).

vrothberg:
Update: we need to merge https://github.com/containers/image/pull/581, then revendor, and then we can merge this PR.

vrothberg:
The PR has been merged into containers/image, so we can merge here to reenable CI over at c/image.

PR #607, opened 2019-02-21
[DO NOT MERGE] Default ostree stub

Silvanoc:
This PR is only to test the integration of the changes of [PR 587 of c/image](https://github.com/containers/image/pull/587).


hack/ostree_tag.sh
```diff
@@ -1,6 +1,4 @@
 #!/bin/bash
 if pkg-config ostree-1 2> /dev/null ; then
-	echo ostree
-else
-	echo containers_image_ostree_stub
+	echo ostree ostree_repos
```
mtrmac: This will need to be updated to use `containers_image_ostree` instead of `ostree_repos`.

giuseppe:
just to make it sure, by default skopeo will still need the "ostree:" backend as it is used by the atomic CLI

Silvanoc:
>  just to make it sure, by default skopeo will still need the "ostree:" backend as it is used by the atomic CLI

I WANT to keep that backend on Skopeo, so I wouldn't try on purpose to provide a default configuration removing it from Skopeo.

I'm trying to make sure that I'm not introducing these kind of undesired side-effects with https://github.com/containers/image/pull/587.

rhatdan:
Replaced by https://github.com/containers/skopeo/pull/751

PR #611, opened 2019-03-06
completions: Fix completions with a global option and indentation

eramoto:
This PR fixes the following problems: 
* bash completions does not work after specifying a global option.
* both of tabs and spaces in indentation are used.


vrothberg:
I tested the patch locally. Thanks a lot for fixing the issue, @eramoto!

PR #618, opened 2019-03-14
Add skopeo registry mirror integration tests

saschagrunert:
This PR adds the registry mirror integration test to skopeo.

- Update toml to latest release
- Update containers/image

### DIFF
--- a/integration/copy_test.go
+++ b/integration/copy_test.go
@@ -662,3 +662,37 @@ func verifyManifestMIMEType(c *check.C, dir string, expectedMIMEType string) {
 	mimeType := manifest.GuessMIMEType(manifestBlob)
 	c.Assert(mimeType, check.Equals, expectedMIMEType)
 }
+
+const regConfFixture = "./fixtures/registries.conf"
+
+func (s *SkopeoSuite) TestSuccessCopySrcWithMirror(c *check.C) {
+	dir, err := ioutil.TempDir("", "copy-mirror")
+	c.Assert(err, check.IsNil)
+
+	assertSkopeoSucceeds(c, "", "--registries-conf="+regConfFixture, "copy",
+		"docker://mirror.invalid/busybox", "dir:"+dir)
+}
+
+func (s *SkopeoSuite) TestFailureCopySrcWithMirrorsUnavailable(c *check.C) {
+	dir, err := ioutil.TempDir("", "copy-mirror")
+	c.Assert(err, check.IsNil)
+
+	assertSkopeoFails(c, ".*no such host.*", "--registries-conf="+regConfFixture, "copy",
+		"docker://invalid.invalid/busybox", "dir:"+dir)
+}
+
+func (s *SkopeoSuite) TestSuccessCopySrcWithMirrorAndPrefix(c *check.C) {
+	dir, err := ioutil.TempDir("", "copy-mirror")
+	c.Assert(err, check.IsNil)
+
+	assertSkopeoSucceeds(c, "", "--registries-conf="+regConfFixture, "copy",
+		"docker://gcr.invalid/foo/bar/busybox", "dir:"+dir)
+}
+
+func (s *SkopeoSuite) TestFailureCopySrcWithMirrorAndPrefixUnavailable(c *check.C) {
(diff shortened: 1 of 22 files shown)


integration/copy_test.go
```diff
@@ -662,3 +662,66 @@ func verifyManifestMIMEType(c *check.C, dir string, expectedMIMEType string) {
 	mimeType := manifest.GuessMIMEType(manifestBlob)
 	c.Assert(mimeType, check.Equals, expectedMIMEType)
 }
+
+const (
+	regConf    = "/etc/containers/registries.conf"
+	regConfBak = "/etc/containers/registries.conf.bak"
+)
+
+func (s *SkopeoSuite) injectRegistriesConf(c *check.C) string {
```
mtrmac: I’d prefer adding a `--registries.conf` option that sets `SystemRegistriesConfPath`, and calling `skopeo --registries.conf $fixture`. Surprisingly little of the existing code requires changing system-wide files the way this code does.

integration/fixtures/registries.conf
```diff
@@ -0,0 +1,26 @@
+[[registry]]
+url = "mirror.local"
+mirror = [
+    { url = "mirror-0.local" },
+    { url = "mirror-1.local" },
```
mtrmac: All the `.local` host names could in principle exist. Maybe `*.invalid`, per RFC 2606?

integration/fixtures/registries.conf
```diff
@@ -0,0 +1,26 @@
...
+    { url = "gcr.io/google-containers" },
+]
+
+[[registry]]
+url = "invalid.local"
+mirror = [
+    { url = "invliad-mirror-0.local" },
+    { url = "invliad-mirror-1.local" },
```
mtrmac: Is the typo essential here?

integration/fixtures/registries.conf
```diff
@@ -0,0 +1,26 @@
...
+url = "gcr.local"
+prefix = "gcr.local/foo/bar"
+mirror = [
+    { url = "wrong-mirror-0.local" },
+    { url = "gcr.io/google-containers" },
+]
+
+[[registry]]
+url = "gcr.io"
+prefix = "gcr.io/google-containers"
```
mtrmac: Is this used at all? Does this maybe exist to ensure that the result of `mirror.local/busybox` is not rewritten twice? If so, using a clearly-invalid `url` might be useful, and there definitely should be a comment.

integration/copy_test.go
rhatdan: Buildah has --registries-conf, although I prefer to keep these  hidden in those tools.

integration/fixtures/registries.conf
saschagrunert: Alright, I changed them.

integration/fixtures/registries.conf
saschagrunert: Nope

integration/fixtures/registries.conf
saschagrunert: I adapted it as suggested.

integration/copy_test.go
mtrmac: Sure, the dash works just as well and consistency is good.

saschagrunert: I'm not sure if adding a new command line argument is nice just for testing purposes. @vrothberg WDYT?

saschagrunert:
@mtrmac @vrothberg PTAL

integration/copy_test.go
vrothberg: I think that it can be a nice addition (also for debugging). But as Dan mentioned, it should be hidden to avoid too much creativity.

Note that there is some selfish motivation behind: having such a flag saves me time as I don't need to copy the registries.conf around :^)

saschagrunert:
@vrothberg I added the hidden `--registry-conf` command line argument and adapted the tests. PTAL.

vrothberg:
@saschagrunert, thanks, LGTM! Would you rebase? @mtrmac PTAL

rhatdan:
LGTM

PR #624, opened 2019-03-29
skopeo: add rootless support

giuseppe:
still it depends on some changes in containers/storage and containers/image that were not merged.

Opening the PR so we can start playing with it


cmd/skopeo/main.go
```diff
@@ -99,13 +100,18 @@ func createApp() (*cli.App, *globalOptions) {
...
 		logrus.SetLevel(logrus.DebugLevel)
 	}
 	if opts.tlsVerify.present {
 		logrus.Warn("'--tls-verify' is deprecated, please set this on the specific subcommand")
 	}
+	switch ctx.Args().First() {
+	case "", "help", "version":
+		return nil
+	}
+	unshare.MaybeReexecUsingUserNamespace(false)
```
mtrmac: It’s great to have an one-liner 🎉 

Hard-coding the list of commands to ignore in here is acceptable but not ideal.  I wonder whether sticking this into maybe `opts.newSystemContext` would be a bit better — it would certainly not be _clean_ either (notably because that function can be called multiple times per process, and not necessarily at the immediate start of an operation), but it would be better targeted even with the current codebase (`manifest-digest`, `standalone-sign`, `standalone-verify` and `untrusted-signature-dump-without-verification` would be excluded).

Or, well, add a mandatory `shouldRunInUserNamespaces` return value from each of the `*Cmd` functions currently returning `cli.Command` and be explicit about it.

(BTW, `cli.App.Before` is not called for the built-in `help` and `version` subcommands anyway. But that’s undocumented AFAICS, so we should not rely on it.)

vrothberg:
@giuseppe can you rerun `make vendor` and add the changes?

cmd/skopeo/main.go
giuseppe: I've whitelisted the commands that need a namespace using `cli.Command.Before`

mtrmac: Nice idea. Can the code in `globalOptions.before` can be removed now?

giuseppe: yes sure, forgot to amend it.  Done now

mtrmac:
Non-blocking: an integration test would be nice.

mtrmac:
LGTM (needs eventual re-vendoring to c/image master).

Do you plan to work on an integration test, or should we merge https://github.com/containers/image/pull/613 + this now?

giuseppe:
@mtrmac let me check how easy it is.  The main difficulty is that the test needs to run with uid!=0, so we need to setup the environment for the user

Makefile
```diff
@@ -37,7 +37,7 @@ endif
 GIT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD 2>/dev/null)
 IMAGE := skopeo-dev$(if $(GIT_BRANCH),:$(GIT_BRANCH))
 # set env like gobuildtag?
-CONTAINER_CMD := ${CONTAINER_RUNTIME} run --rm -i -e TESTFLAGS="$(TESTFLAGS)" #$(CONTAINER_ENVS)
+CONTAINER_CMD := ${CONTAINER_RUNTIME} run --privileged --rm -i -e TESTFLAGS="$(TESTFLAGS)" #$(CONTAINER_ENVS)
```
mtrmac: Do the existing root-run tests really need `--privileged` now?

giuseppe: no, I can tune it later if this works, probably I'll need to enable some more syscalls in the seccomp profile that are used when running as rootless

mtrmac: My first impression is that the `--privileged` should only be necessary in `CONTAINER_RUN_ROOTLESS`, not in `CONTAINER_CMD`.

Sure, if rootless tests need extra privileges, fine.

giuseppe:
I forgot we cannot use user namespaces on the Travis kernel, the kernel there is too old

mtrmac:
@giuseppe Please rebase on top of `master`, and re-vendor from the c/image `master` branch now that https://github.com/containers/image/pull/613 was merged.

giuseppe:
revendored and ready to review

cmd/skopeo/signing.go
```diff
@@ -18,6 +18,7 @@ type standaloneSignOptions struct {
 func standaloneSignCmd() cli.Command {
 	opts := standaloneSignOptions{}
 	return cli.Command{
+		Before:    needsRexec,
```
mtrmac: Oops, sorry for missing this - `standalone`* should not need reexec ; OTOH `inspect` and `layers` do need it.

giuseppe: fixed now

rhatdan:
LGTM
@vrothberg @mtrmac PTAL
Once we get this in, I think it needs a new release.  Do we need updates to the man pages?

mtrmac:
LGTM. Thanks!

PR #626, opened 2019-04-07
Update broken links in info docs

grdryn:


TomSweeneyRedHat:
@grdryn Thank you very much, looks like those 5's ran away at some point.  Thanks for spotting and putting up the PR!
LGTM

PR #629, opened 2019-04-15
added help to Makefile

chuanchang:
$ make help
Usage: make <target>

 * 'install' - Install binaries and documents to system locations
 * 'binary' - Build skopeo with a container
 * 'binary-local' - Build skopeo locally
 * 'test-unit' - Execute unit tests
 * 'test-integration' - Execute integration tests
 * 'validate' - Verify whether there is no conflict and all Go source files have been formatted, linted and vetted
 * 'check' - Including above validate, test-integration and test-unit
 * 'shell' - Run the built image and attach to a shell
 * 'clean' - Clean artifacts


vrothberg:
Thanks for the contribution, @chuanchang! The code LGTM but you need to rebase ontop of master before we can merge.

@mtrmac, @rhatdan PTAL

Makefile
```diff
@@ -67,6 +67,19 @@ endif
...
+	@echo "Usage: make <target>"
+	@echo
+	@echo " * 'install' - Install binaries and documents to system locations"
+	@echo " * 'binary' - Build skopeo with container"
+	@echo " * 'binary-local' - Build skopeo on the local and without container"
+	@echo " * 'test-unit' - Execute unit tests"
+	@echo " * 'test-integration' - Execute integration tests"
+	@echo " * 'validate' - Verify whether there is no conflict and all Go source files have been formatted, linted and vetted"
+	@echo " * 'check' - Including above validate, test-integration and test-unit"
+	@echo " * 'shell' - Run the built image and attach a shell"
```
vrothberg: attach *to* a shell

Makefile
```diff
@@ -67,6 +67,19 @@ endif
 #           use source debugging tools like delve.
 all: binary docs-in-container
 
+help:
+	@echo "Usage: make <target>"
+	@echo
+	@echo " * 'install' - Install binaries and documents to system locations"
+	@echo " * 'binary' - Build skopeo with container"
```
vrothberg: with *a*

Makefile
```diff
@@ -67,6 +67,19 @@ endif
...
+	@echo " * 'binary-local' - Build skopeo on the local and without container"
```
vrothberg: Just "Build skopeo locally" should be fine

mtrmac:
The text is fine, but I don’t quite see the point. `make help` is not a common convention, and if the user has to read the `Makefile` to find out about `make help`, the user can just as well find out about the commands by reading the file.

What difficulty did you actually run against that motivated this? It’s quite likely that either our end-user or contributors’ documentation needs to be improved, or needs to mention some of these targets.

chuanchang:
> The text is fine, but I don’t quite see the point. `make help` is not a common convention, and if the user has to read the `Makefile` to find out about `make help`, the user can just as well find out about the commands by reading the file.
> 
> What difficulty did you actually run against that motivated this? It’s quite likely that either our end-user or contributors’ documentation needs to be improved, or needs to mention some of these targets.

Miloslav, thank you for comments, although it's not a common convention, you know there are some upstream projects provides 'Makefile help' to list supported items with users, for instance, k8s, containerd, libpod and cri-o projects, these users will expect skopeo also providing the same method, in order words, they are used to run 'make help' and get details from Makefile rather than using cat/vim+Makefile to see details.

rhatdan:
I am fine with this change, and see it as possibly useful, with no down sides.

LGTM

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
Looks like a rebase may be required.

chuanchang:
@TomSweeneyRedHat @rhatdan thank you for reviewing.

chuanchang:
> Looks like a rebase may be required.

@TomSweeneyRedHat this PR is based on latest commit fee5981, so it should be okay I think.

rhatdan:
Thanks @chuanchang

PR #635, opened 2019-04-23
Vendor the latest buildah master

saschagrunert:
This commit contains the necessary split-up between buildah/pkg and
buildah/util to avoid dependency breaks.

Needed for https://github.com/containers/skopeo/pull/618 and https://github.com/containers/image/pull/564


vendor.conf
```diff
@@ -3,7 +3,7 @@ github.com/urfave/cli v1.20.0
 github.com/kr/pretty v0.1.0
 github.com/kr/text v0.1.0
 github.com/containers/image e9c3d17ddb8cb5d48450bc0588525b17b181c049
-github.com/containers/buildah 25b7c1164a262934a6526ded1115dfacfc51f8ec
+github.com/containers/buildah 810efa340ab43753034e2ed08ec290e4abab7e72
```
rhatdan: We should declare a buildah release rather then doing this via UUID.

saschagrunert: I agree, may I ask you to create a new release for me? :innocent:

mtrmac: This is not making things any worse, and buildah will need another change just after https://github.com/containers/image/pull/564 is merged.

mtrmac:
LGTM. Thanks!

Issue #637, opened 2019-04-24
Unable to use Skopeo in a non-privileged container after #624


ananace:
```
$ docker run --rm -it --entrypoint sh skopeo-build
/ # /skopeo copy docker://a docker://b
Error during unshare(CLONE_NEWUSER): Operation not permitted
ERRO[0000] error parsing PID "": strconv.Atoi: parsing "": invalid syntax 
ERRO[0000] (unable to determine exit status)
$ docker run --privileged --rm -it --entrypoint sh skopeo-build
/ # /skopeo copy docker://a docker://b
FATA[0001] Error determining manifest MIME type for docker://a:latest: Error reading manifest latest in docker.io/library/a: errors:
denied: requested access to the resource is denied
unauthorized: authentication required
```

vrothberg:
@giuseppe PTAL

giuseppe:
`unshare` might be blocked by seccomp.  Could you try with --security-opt seccomp=unconfined?

ananace:
Setting seccomp=undefined indeed seems to work as well.
Though it's not something that helps overly much in the case of our CI/CD pipelines, as there's - by design - no way to request modified permissions there.

giuseppe:
@ananace could you check if this PR fixes the issue for you https://github.com/containers/skopeo/pull/638 ?

ananace:
@giuseppe Seems to solve the issue, yes.

PR #638, opened 2019-04-24
rootless: do not create a user namespace if not needed

giuseppe:
do not create a user namespace if we already have the capabilities we
need for pulling and storing an image.

Closes: https://github.com/containers/skopeo/issues/637

### DIFF
--- a/cmd/skopeo/unshare.go
+++ b/cmd/skopeo/unshare.go
@@ -2,5 +2,6 @@
 
 package main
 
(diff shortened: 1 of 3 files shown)


TomSweeneyRedHat:
LGTM

rhatdan:
@mtrmac I believe an hope that we will only attempt to use the UserNS if the user is non-root and they are attempting to use containers/storage.

PR #639, opened 2019-04-24
Remove references to atomic

rhatdan:
Move signature yaml file to point at /var/lib/containers/sigstore.

Change skopeo-copy.1 to use containers-storage and docker transports
rather then atomic.


rhatdan:
@giuseppe @vrothberg @mtrmac PTAL

rhatdan:
The postinstall of the skopeo package will have to copy any signature files in /var/lib/atomic into /var/lib/containers/sigstore

rhatdan:
Also needs to add a symbolic link between /var/lib/atomic -> /var/lib/containers

mtrmac:
> The postinstall of the skopeo package will have to copy any signature files in /var/lib/atomic into /var/lib/containers/sigstore

> Also needs to add a symbolic link between /var/lib/atomic -> /var/lib/containers

IIRC RPM has (had?) big difficulties dealing with a path changing between a directory/non-directory on package updates. Is this known to be possible? We need to be certain before merging this.

(Or, well, we could introduce a `/var/lib/containers/sigstore`→`/var/lib/atomic/sigstore` symbolic link instead… which makes the benefit of this PR smaller.)

rhatdan:
@lsm5 Do you know?

lsm5:
I'm not certain, but we can try scratch builds for this.

giuseppe:
LGTM

rhatdan:
@lsm5 Did you ever try it out?

rhatdan:
@jnovy @lsm5 Would like to move this forward.

rhatdan:
@mtrmac @lsm5 @vrothberg Lets more this forward.  We should not be shipping anything referencing atomic at this point.

rhatdan:
@jnovy in containers-common could we just add a ghost on /var/lib/atomic, and in the post install
mv files from /var/lib/atomic/sigstore to /var/lib/containers/sigstore. On a single trigger update.

rhatdan:
I have pushed a new version of this to make sure we don't have any regresssions.  I am going to merge this and then the packagers can deal with the change.  Upstream should no longer be using this name.

mtrmac:
I’d rather prefer to see a proof of concept for the packaging, but if you want to merge it anyway, I guess I can live with that.

PR #640, opened 2019-04-25
 Vendor update container/storage

rhatdan:
overlay: propagate errors from mountProgram
utils: root in a userns uses global conf file
Fix handling of additional stores
Correctly check permissions on rootless directory
Fix possible integer overflow on 32bit builds
Evaluate device path for lvm
lockfile test: make concurrent RW test determinisitc
lockfile test: make concurrent read tests deterministic


vrothberg:
LGTM. I'll just merge it as it's a simple vendor update.

PR #644, opened 2019-04-29
migrate to go modules

vrothberg:
Note that there is no way to control which files are being preserved in the `./vendor` directory anymore. Hence, the man pages and the registries.conf from `./vendor/github.com/containers/image` are deleted.

@lsm5, afair those are needed for the Ubuntu package. Can we find an alternative for the packaging?


vrothberg:
Closing in favor of https://github.com/containers/skopeo/pull/681.

Issue #225, opened 2016-10-05
Build upstream man pages from containers/image/docs directory


aweiteka:
Per discussion https://github.com/projectatomic/atomic/issues/616, we want to include the man pages from https://github.com/containers/image and deliver them with skopeo.

@mtrmac I was unable to get this working in the Makefile.

rhatdan:
We deliver man pages now.

mtrmac:
`MANPAGES` in `Makefile` only includes `docs/*.md`; the man pages in `vendor/github.com/containers/image/docs` are included by `make vendor` but not installed.

mtrmac:
Per the conversation around https://github.com/containers/skopeo/pull/645#issuecomment-489127187 , this seems unnecessary for RPM packaging, in the Fedora ecosystem at least.

It would _still_ be useful, for users building Skopeo from source — but that should be a comparatively rare case (contributors to Skopeo or testers of a recent version) and such users presumably know enough about the implementation to read the right version of c/image documentation on-line.

PR #645, opened 2019-04-29
Install man pages from containers image as well as skopeo

rhatdan:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
-.PHONY: all binary build-container docs docs-in-container build-local clean install install-binary install-completions shell test-integration .install.vndr vendor
+.PHONY: all binary build-container docs docs-in-container build-local clean install install-binary install-completions shell test-integration .install.vndr vendor image.docs
+
 
 export GO15VENDOREXPERIMENT=1
 
@@ -25,6 +26,7 @@ BASHINSTALLDIR=${PREFIX}/share/bash-completion/completions
 GO ?= go
 CONTAINER_RUNTIME := $(shell command -v podman 2> /dev/null || echo docker)
 GOMD2MAN ?= $(shell command -v go-md2man || echo '$(GOBIN)/go-md2man')
+IMAGEVERSION := $(shell grep ^github.com/containers/image vendor.conf  | cut -f2 -d ' ')
 
 ifeq ($(DEBUG), 1)
   override GOGCFLAGS += -N -l
@@ -106,7 +108,13 @@ build-container:
 $(MANPAGES): %: %.md
 	@sed -e 's/\((skopeo.*\.md)\)//' -e 's/\[\(skopeo.*\)\]/\1/' $<  | $(GOMD2MAN) -in /dev/stdin -out $@
 
-docs: $(MANPAGES)
+image.docs: vendor.conf
+	@git clone https://github.com/containers/image.git image.docs 2>/dev/null || true
+	@(cd image.docs; \
+	  git checkout ${IMAGEVERSION} 2>/dev/null || git checkout ${IMAGEVERSION} -b ${IMAGEVERSION}; \
+	for i in docs/*.md; do $(GOMD2MAN) -in $$i -out $${i//.md}; done)
+
+docs: $(MANPAGES) image.docs
 
 docs-in-container:
 	${CONTAINER_RUNTIME} build ${BUILD_ARGS} -f Dockerfile.build -t skopeobuildimage .
@@ -114,7 +122,7 @@ docs-in-container:
 		skopeobuildimage make docs $(if $(DEBUG),DEBUG=$(DEBUG)) BUILDTAGS='$(BUILDTAGS)'
 
 clean:
-	rm -f skopeo docs/*.1
+	rm -rf skopeo docs/*.1 image.docs
 
 install: install-binary install-docs install-completions
 	install -d -m 755 ${SIGSTOREDIR}
@@ -130,6 +138,8 @@ install-binary: ./skopeo
 install-docs: docs
 	install -d -m 755 ${MANINSTALLDIR}/man1
 	install -m 644 docs/*.1 ${MANINSTALLDIR}/man1/
+	install -d -m 755 ${MANINSTALLDIR}/man5
+	install -m 644 image.docs/docs/*.5 ${MANINSTALLDIR}/man5/
 
 install-completions:
 	install -m 755 -d ${BASHINSTALLDIR}
--- a/vendor/github.com/containers/image/internal/testing/explicitfilepath-tmpdir/tmpdir.go
+++ b/vendor/github.com/containers/image/internal/testing/explicitfilepath-tmpdir/tmpdir.go
@@ -1,0 +1,29 @@
+// Package tmpdir is a TESTING-ONLY utility.
+//
+// Some tests directly or indirectly exercising the directory/explicitfilepath
+// subpackage expect the path returned by ioutil.TempDir to be canonical in the
+// directory/explicitfilepath sense (absolute, no symlinks, cleaned up).
+//
+// ioutil.TempDir uses $TMPDIR by default, and on macOS, $TMPDIR is by
+// default set to /var/folders/…, with /var a symlink to /private/var ,
+// which does not match our expectations.  So, tests which want to use
+// ioutil.TempDir that way, can
+// import _ "github.com/containesr/image/internal/testing/explicitfilepath-tmpdir"
+// to ensure that $TMPDIR is canonical and usable as a base for testing
+// path canonicalization in its subdirectories.
+//
+// NEVER use this in non-testing subpackages!
+package tmpdir
+
+import (
+	"os"
+	"path/filepath"
+)
+
+func init() {
+	tmpDir := os.Getenv("TMPDIR")
+	explicitTmpDir, err := filepath.EvalSymlinks(tmpDir)
+	if err == nil {
+		os.Setenv("TMPDIR", explicitTmpDir)
+	}
+}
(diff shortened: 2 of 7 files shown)


rhatdan:
Closes: https://github.com/containers/skopeo/issues/225

rhatdan:
@mtrmac @aweiteka @vrothberg PTAL

Makefile
```diff
@@ -104,7 +104,7 @@ build-container:
 	${CONTAINER_RUNTIME} build ${BUILD_ARGS} -t "$(IMAGE)" .
 
 $(MANPAGES): %: %.md
-	@sed -e 's/\((skopeo.*\.md)\)//' -e 's/\[\(skopeo.*\)\]/\1/' $<  | $(GOMD2MAN) -in /dev/stdin -out $@
+	@sed -e 's/\((skopeo.*\.md)\)//' -e 's/\[\(skopeo.*\)\]/\1/' $<  | $(GOMD2MAN) -in /dev/stdin -out docs/$(shell basename $@)
```
mtrmac: I don’t think this creates the man pages in the expected location `$(MANPAGES)`.

rhatdan: Ok went back to original, now everything will work.

TomSweeneyRedHat:
Tests aren't hip, complaining of a dirty tree.

Makefile
```diff
@@ -49,7 +49,7 @@ CONTAINER_RUN := $(CONTAINER_CMD) "$(IMAGE)"
 
 GIT_COMMIT := $(shell git rev-parse HEAD 2> /dev/null || true)
 
-MANPAGES_MD = $(wildcard docs/*.md)
+MANPAGES_MD = $(wildcard docs/*.md) $(wildcard vendor/github.com/containers/image/docs/*.md)
```
TomSweeneyRedHat: Interesting.  Hadn't thought about this, but this is a neat advantage to vendoring...

TomSweeneyRedHat:
Changes LGTM if you can figure out the git unhappiness.

vrothberg:
Please don't merge the PR. The problem is that this only works with `vndr` but we're about to move to `go mod` where whitelisting paths is not possible. In fact, I opened a PR yesterday https://github.com/containers/skopeo/pull/644.

rhatdan:
@vrothberg Have to think how we can fix this as we go forward.  Ways to install these files?

vrothberg:
> @vrothberg Have to think how we can fix this as we go forward. Ways to install these files?

For the packages, we could make them multi-source packages.

rhatdan:
@vrothberg Yes although we then need to make sure the spec file pulls the matching containers/image with the skopeo package.

rhatdan:
Also @lsm5 Is not a big fan of doing this.
Containers-common,  should probably become its own package and grab the latest released version of containers/storage and containers/image man pages and config files.

mtrmac:
> @vrothberg Have to think how we can fix this as we go forward. Ways to install these files?

Well, there is still a `Makefile`. We could add a `make dist` that downloads the _correct_ corresponding version of c/image, extracts the image, and builds a tarball that we can use as a source in the RPM.

Doing this in the spec file would be difficult because spec files don’t have network access.  Of course the RPM could have two source items, but ensuring c/image matches the vendored version would be manual work; or externally-scripted work that could be overlooked.

Up to @lsm5 really.

rhatdan:
@vrothberg @mtrmac I reworked this pr to pull directly from github the man pages and install them
PTAL

lsm5:
@rhatdan @vrothberg so in the ubuntu package, I install `registries.conf` from containers/image vendored inside skopeo. I believe that file has been whitelisted somewhere. Can we do the same thing for the docs? So, as long as the right version is vendored, we can still keep it a single source rpm.

vrothberg:
@lsm5 we will soon use `go mod` instead of `vndr` for dependency management. `go mod` does not support whitelisting, so need to find an alternative approach.

Makefile
```diff
@@ -106,15 +107,20 @@ build-container:
 $(MANPAGES): %: %.md
 	@sed -e 's/\((skopeo.*\.md)\)//' -e 's/\[\(skopeo.*\)\]/\1/' $<  | $(GOMD2MAN) -in /dev/stdin -out $@
 
-docs: $(MANPAGES)
+.PHONY: image.docs
+image.docs:
+	@git clone https://github.com/containers/image.git image.docs 2>/dev/null || true
+	@(cd image.docs; git checkout ${IMAGEVERSION} -b ${IMAGEVERSION}; git filter-branch --prune-empty --subdirectory-filter docs -- --all 2>/dev/null;for i in *.md; do $(GOMD2MAN) -in $$i -out $${i//.md}; done) || true
```
mtrmac: Could this clone only the `IMAGEVERSION` commit (`-depth 1`) instead of all of `master`? Or is that not possible with non-tag commit references?

Why the `filter-branch` instead of just using `docs/*.md` or something like that?

Makefile
```diff
@@ -106,15 +107,20 @@ build-container:
 $(MANPAGES): %: %.md
 	@sed -e 's/\((skopeo.*\.md)\)//' -e 's/\[\(skopeo.*\)\]/\1/' $<  | $(GOMD2MAN) -in /dev/stdin -out $@
 
-docs: $(MANPAGES)
+.PHONY: image.docs
+image.docs:
```
mtrmac: This should probably have a dependency on `vendor.conf` so that it does not grow stale.

Makefile
```diff
@@ -106,15 +107,20 @@ build-container:
...
+docs: $(MANPAGES) image.docs
 
 docs-in-container:
 	${CONTAINER_RUNTIME} build ${BUILD_ARGS} -f Dockerfile.build -t skopeobuildimage .
 	${CONTAINER_RUNTIME} run --rm --security-opt label=disable -v $$(pwd):/src/github.com/containers/skopeo \
 		skopeobuildimage make docs $(if $(DEBUG),DEBUG=$(DEBUG)) BUILDTAGS='$(BUILDTAGS)'
 
 clean:
-	rm -f skopeo docs/*.1
+	rm -rf image.docs
```
mtrmac: Why does this remove the pre-existing cleanups?

Makefile
rhatdan: Because that is what google found... :^)

rhatdan: --depth=1 will not work,because of the IMAGEVERSION stuff.
I just aded the filter-branch to make it easier to debug.  But I could remove it if you don't like it.

mtrmac: \*shrug\* I can live with all this, primarily I’d like @lsm5 to confirm that this is useful for packaging.

mtrmac: WRT `filter-branch`, it seems unnecessarily expensive (O(repo history) compared to just using an explicit path at ~zero cost).

OTOH I _never_ run `make docs` so I am definitely not going to block this PR on performance optimizations.

Makefile
rhatdan: added

Makefile
rhatdan: Bug, fixed

Makefile
rhatdan: filter-branch removed

vrothberg:
> This makes sense stand-alone, but how would we make a RPM out of this? (OTOH, do we need to, or would that happen somewhere outside both this `Makefile` and the RPM `.spec`? @lsm5 ?)

When working on the packages for openSUSE, all this went into the [libcontainers-common package](https://build.opensuse.org/package/show/devel:kubic/libcontainers-common), which is a multi-source package including image, storage, and libpod along with a modified mounts.conf, registries.conf, storage.conf etc. I am not saying we have to do the same for Fedora but

> @vrothberg , can you confirm that this is compatible enough with the `go mod` approach, please?

Yes, that would work. To be on the very safe side, we could explicitly exclude `github.com/containers/image/image.docs` in the `go.mod` file to prevent confusing go modules.

To find out which version or commit is specified in the `go.mod` file we need a more sophisticated regex to catch the two cases below:
```bash
$ go list -m -v github.com/containers/image
github.com/containers/image v0.0.0-20190424091117-ff926d3c7968
$ go list -m -v go.etcd.io/bbolt
go.etcd.io/bbolt v1.3.2
```

rhatdan:
I actually do not think this will work in Fedora builds.  I believe Brew is cut off from the internet, to make sure nothing gets pulled down during the build that does not get put in the lookaside cache.  @lsm5 Could you confirm?

vrothberg:
Same applies to openSUSE as well. It would require to ship the source tarballs.

rhatdan:
Then I think we need to make this optional?  Or the packagers do not do make install-docs.

mtrmac:
> I actually do not think this will work in Fedora builds. I believe Brew is cut off from the internet, to make sure nothing gets pulled down during the build that does not get put in the lookaside cache. 

Yes.

> Then I think we need to make this optional? Or the packagers do not do make install-docs.

That’s not quite the issue; omitting `make install-docs` would remove the internet dependency, but it would also remove the man pages from the package.

_Installing_ the docs is not something the RPM packaging strictly needs help with, we can always add extra shell code into the spec file.

We need a way to _deliver_ the man pages into the RPM.  That could be a `make dist` in this repo that creates a tarball, it could be a separate update-Fedora-RPM script that downloads the man pages and commits them into the Fedora package repo directly, it could be something else.

Neither `make image.docs` nor `make install` is sufficient to handle the delivery process.

lsm5:
How about we just create a separate `containers-image` package, get the docs installed in there, and maybe have containers-common require it? Shouldn't take that long to get done, and we need not have more than 1 source tarball per rpm, I really don't wanna go our prior docker rpm route :D . Jindrich and I can get that done.

rhatdan:
If you are going to do this, you need to also do it with container-storage, which also has man pages.

lsm5:
> If you are going to do this, you need to also do it with container-storage, which also has man pages.

I'm cool with that.

rhatdan:
Moving this effort to https://github.com/containers/image/pull/661

PR #646, opened 2019-05-02
Add --no-creds flag to skopeo inspect

### LINKED ISSUE #421: RFE: Allow force anonymous inspect
By default skopeo tries to use ~/.docker/config.json to authenticate to an external registry unless username and password are provided using --creds flag

It would be handy to allow to force no token to use skopeo as a debug tool

QiWang19:
Follow PR #433 
Close #421 
Currently skopeo inspect allows to:

- Use the default credentials in $HOME/.docker.config
- Explicitly define credentials via de --creds flag
This implements a --no-creds flag which will query docker registries anonymously.

### DIFF
--- a/cmd/skopeo/utils.go
+++ b/cmd/skopeo/utils.go
@@ -63,6 +63,7 @@ type imageOptions struct {
 	tlsVerify        optionalBool        // Require HTTPS and verify certificates (for docker: and docker-daemon:)
 	sharedBlobDir    string              // A directory to use for OCI blobs, shared across repositories
 	dockerDaemonHost string              // docker-daemon: host to connect to
+	noCreds          bool                // Access the registry anonymously
 }
 
 // imageFlags prepares a collection of CLI flags writing into imageOptions, and the managed imageOptions structure.
@@ -105,6 +106,11 @@ func imageFlags(global *globalOptions, shared *sharedImageOptions, flagPrefix, c
 			Usage:       "use docker daemon host at `HOST` (docker-daemon: only)",
 			Destination: &opts.dockerDaemonHost,
 		},
+		cli.BoolFlag{
+			Name:        flagPrefix + "no-creds",
+			Usage:       "Access the registry anonymously",
+			Destination: &opts.noCreds,
+		},
 	}, &opts
 }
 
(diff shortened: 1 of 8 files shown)


cmd/skopeo/copy.go
```diff
@@ -24,7 +24,8 @@ type copyOptions struct {
 	signByFingerprint string          // Sign the image using a GPG key with the specified fingerprint
 	format            optionalString  // Force conversion of the image to a specified format
 	quiet             bool            // Suppress output information when copying images
-
+	srcNoCreds        bool
+	destNoCreds       bool
```
mtrmac: This seems unused, along with all the other variables added outside of `utils.go`.

cmd/skopeo/inspect.go
```diff
@@ -58,7 +59,7 @@ func inspectCmd(global *globalOptions) cli.Command {
 		Flags: append(append([]cli.Flag{
 			cli.BoolFlag{
 				Name:        "raw",
-				Usage:       "output raw manifest",
+				Usage:       "Output raw manifest",
```
mtrmac: If not a separate PR, make this at least a separate commit, please.

TomSweeneyRedHat:
OK, I might not be remembering correctly, but I thought that `skopeo inspect` worked against a registry that didn't require creds already?  If so, why do we need the `--no-creds` flag?

mtrmac:
> why do we need the `--no-creds` flag?

See https://github.com/containers/skopeo/pull/433#issuecomment-337062649 .

TomSweeneyRedHat:
Thanks @mtrmac.  Prior to this patch, could they just `docker/podman logout` and then pull successfully?

cmd/skopeo/inspect.go
QiWang19: maybe in another PR.

QiWang19:
to document `c/image/types.DockerAuthConfig` can be used with empty fields, just change the comment in `c/image/types/types.go` ?

mtrmac:
> to document `c/image/types.DockerAuthConfig` can be used with empty fields, just change the comment in `c/image/types/types.go` ?

Yes; we just need such a PR to exist in c/image, this PR does not need to block on it nor vendor an updated version of c/image just for the comment.

vendor/github.com/containers/image/docs/atomic-signature-embedded-json.json
```diff
@@ -1,66 +0,0 @@
-{
```
mtrmac: Why does this PR remove the various files in `vendor`?

(Guessing, run `make vendor` instead of `vndr` directly.)

QiWang19:
can this get in?

mtrmac:
LGTM.  Thanks!

PR #647, opened 2019-05-08
inspect: add a --config flag

nalind:
Add a --config option to "skopeo inspect" to dump an image's configuration blob in the OCI format, or the original format if --config and --raw are specified.  It's handy for me, and it might be useful for other people.


mtrmac:
LGTM. Thanks!

Issue #652, opened 2019-05-15
skopeo inspect command errors out with 0.1.36-dev for xenial release


khandelwalankit04:
Hi Team,

We are seeing error in inspecting image with latest build of skopeo : 0.1.36-dev. Here is the screenshot : 

![image](https://user-images.githubusercontent.com/36901094/57799109-f16c4780-7713-11e9-9010-fd574d28186d.png)

This works with 0.1.35-dev , but since ppa:projectatomic/ppa only contains the latest version of skopeo we are unable to downgrade it to previous release.  Can you provide the fix as well as update the previous stable releases for skopeo : 

![image](https://user-images.githubusercontent.com/36901094/57799298-63449100-7714-11e9-9cf8-6960bbd8cc1c.png)

Let me know if you need any additional info

rhatdan:
If you install newuid package on Ubuntu, it should fix the issue, but this definitely a bug.

khandelwalankit04:
@rhatdan : Is there a plan to also package and keep older stable version in ppa:projectatomic/ppa ?

rhatdan:
@giuseppe We should check to see if newuidmap and newgidmap exists if not then just don't enter the user namespace.  It should only care if the user is using containers/storage anyways.

rhatdan:
@khandelwalankit04 I think we are planning on shipping skopeo as a snap.  Which would only be the stable versions.

@lsm5 @jnovy Please comment.

mtrmac:
@khandelwalankit04 Is this still an issue for you? It should have been fixed in Skopeo 0.1.36 and later.

khandelwalankit04:
@mtrmac : This is no longer the issue. Thanks for fixing

mtrmac:
Thanks, closing.

PR #653, opened 2019-05-16
rootless: don't create a namespace unless for containers-storage

### LINKED ISSUE #649: Skopeo now requires capabilities when executed as regular user
It uses to work fine in bubblewrap to copy image between registry but it no longer do:

```
$ ./skopeo --version
skopeo version 0.1.36-dev commit: 904b064da44eecfeea57f2c935e230fd41574ac8
$ bwrap  --unshare-all --bind / / --proc /proc --dev dev /bin/bash
$ ./skopeo copy docker-daemon:test:latest docker-daemon:test:test
Error: error writing "allow" to setgroups: write /proc/33/setgroups: operation not permitted
ERRO[0000] error opening "allow" to /proc/33/setgroups: write /proc/33/setgroups: operation not permitted 
ERRO[0000] (unable to determine exit status)   

$ bwrap --cap-add CAP_CHOWN --cap-add CAP_DAC_OVERRIDE --cap-add CAP_FOWNER --cap-add CAP_FSETID --cap-add CAP_MKNOD --cap-add CAP_SETFCAP --unshare-all --bind / / --proc /proc --dev dev /bin/bash
$ ./skopeo copy docker-daemon:test:latest docker-daemon:test:test
OK
```

Shouldn't the tool checks if the capabilities are really needed before requiring them (e.g. when doing remote copies)?

TristanCacqueray:
This change fixes skopeo usage in restricted environment such as
bubblewrap where it doesn't need extra capabilities or user namespace
to perform its action.

Close #649

Signed-off-by: Tristan Cacqueray <tdecacqu@redhat.com>

### DIFF
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -50,7 +50,6 @@ func copyCmd(global *globalOptions) cli.Command {
 	`, strings.Join(transports.ListNames(), ", ")),
 		ArgsUsage: "SOURCE-IMAGE DESTINATION-IMAGE",
 		Action:    commandAction(opts.run),
-		Before:    needsRexec,
 		// FIXME: Do we need to namespace the GPG aspect?
 		Flags: append(append(append([]cli.Flag{
 			cli.StringSliceFlag{
@@ -86,6 +85,11 @@ func (opts *copyOptions) run(args []string, stdout io.Writer) error {
 	if len(args) != 2 {
 		return errorShouldDisplayUsage{errors.New("Exactly two arguments expected")}
 	}
+	imageNames := args
+
+	if err := reexecIfNecessaryForImages(imageNames...); err != nil {
+		return err
+	}
 
 	policyContext, err := opts.global.getPolicyContext()
 	if err != nil {
@@ -93,13 +97,13 @@ func (opts *copyOptions) run(args []string, stdout io.Writer) error {
 	}
 	defer policyContext.Destroy()
 
-	srcRef, err := alltransports.ParseImageName(args[0])
+	srcRef, err := alltransports.ParseImageName(imageNames[0])
 	if err != nil {
-		return fmt.Errorf("Invalid source name %s: %v", args[0], err)
+		return fmt.Errorf("Invalid source name %s: %v", imageNames[0], err)
 	}
-	destRef, err := alltransports.ParseImageName(args[1])
+	destRef, err := alltransports.ParseImageName(imageNames[1])
 	if err != nil {
-		return fmt.Errorf("Invalid destination name %s: %v", args[1], err)
+		return fmt.Errorf("Invalid destination name %s: %v", imageNames[1], err)
 	}
 
 	sourceCtx, err := opts.srcImage.newSystemContext()
--- a/cmd/skopeo/layers.go
+++ b/cmd/skopeo/layers.go
@@ -32,7 +32,6 @@ func layersCmd(global *globalOptions) cli.Command {
 		Name:      "layers",
 		Usage:     "Get layers of IMAGE-NAME",
 		ArgsUsage: "IMAGE-NAME [LAYER...]",
-		Before:    needsRexec,
 		Hidden:    true,
 		Action:    commandAction(opts.run),
 		Flags:     append(sharedFlags, imageFlags...),
@@ -44,6 +43,11 @@ func (opts *layersOptions) run(args []string, stdout io.Writer) (retErr error) {
 	if len(args) == 0 {
 		return errors.New("Usage: layers imageReference [layer...]")
 	}
+	imageName := args[0]
+
+	if err := reexecIfNecessaryForImages(imageName); err != nil {
+		return err
+	}
 
 	ctx, cancel := opts.global.commandTimeoutContext()
 	defer cancel()
@@ -53,7 +57,7 @@ func (opts *layersOptions) run(args []string, stdout io.Writer) (retErr error) {
 		return err
 	}
 	cache := blobinfocache.DefaultCache(sys)
-	rawSource, err := parseImageSource(ctx, opts.image, args[0])
+	rawSource, err := parseImageSource(ctx, opts.image, imageName)
 	if err != nil {
 		return err
 	}
--- a/cmd/skopeo/unshare_linux.go
+++ b/cmd/skopeo/unshare_linux.go
@@ -2,6 +2,8 @@ package main
 
 import (
 	"github.com/containers/buildah/pkg/unshare"
+	"github.com/containers/image/storage"
+	"github.com/containers/image/transports/alltransports"
 	"github.com/pkg/errors"
 	"github.com/syndtr/gocapability/capability"
 )
@@ -28,6 +30,16 @@ func maybeReexec() error {
 			// We miss a capability we need, create a user namespaces
 			unshare.MaybeReexecUsingUserNamespace(true)
 			return nil
+		}
+	}
+	return nil
+}
+
+func reexecIfNecessaryForImages(imageNames ...string) error {
+	// Check if container-storage are used before doing unshare
+	for _, imageName := range imageNames {
+		if alltransports.TransportFromImageName(imageName).Name() == storage.Transport.Name() {
+			return maybeReexec()
 		}
 	}
 	return nil
--- a/cmd/skopeo/utils.go
+++ b/cmd/skopeo/utils.go
@@ -16,10 +16,6 @@ type errorShouldDisplayUsage struct {
 	error
 }
 
-func needsRexec(c *cli.Context) error {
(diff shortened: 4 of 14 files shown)


vrothberg:
This looks like a clever solution to me. @giuseppe PTAL

giuseppe:
thanks for the patch!  Yes, I think it is a good fix, and should also address https://github.com/containers/skopeo/issues/652

TristanCacqueray:
You're welcome, thank you for the prompt review. FWIW I tried using a cleaner method (such as alltransports.ParseImageName) but I couldn't find the right predicate to test for containers-storage type. The proposed solution assumes all the containers-storage image reference start with the `containers-storage:` prefix.

giuseppe:
@mtrmac is there any cleaner way to test that?  Are you fine with this approach?

vrothberg:
One way could be to use `ref, err := ParseImageName(...)` and if `err == nil` to compare `ref.Transport().Name() == storage.Transport.Name()`.

TristanCacqueray:
The issue with ParseImageName is that it may fail early with `chown /home/fedora/.local/share/containers/storage/vfs: operation not permitted`

vrothberg:
Ouch, sorry for missing that. In this case, we can mimic what `ParseImageName(...)` effectively does:
```golang
parts := strings.SplitN(imgName, ":", 2)
if len(parts) != 2 {
   continue
}
if parts[0] == storage.Transport.Name() { // we have storage reference (but can't validate it yet) }
```

TristanCacqueray:
Done, thank you for the suggestion.

rhatdan:
LGTM
Although I still want @mtrmac to chime in.

mtrmac:
> The issue with ParseImageName is that it may fail early with `chown /home/fedora/.local/share/containers/storage/vfs: operation not permitted`

Oh, wow. That’s… a mess.

@nalind AFAICS this is essentially because `storageTransport.ParseStoreReference` needs a working `store` so that it can support image ID prefixes:
```go
if img, err := store.Image(possibleID); err == nil && img != nil && len(possibleID) >= minimumTruncatedIDLength && strings.HasPrefix(img.ID, possibleID)
```
and the like.  Is there any chance we could drop that functionality? (I guess not, but it would make dealing with this in callers so much simpler.)

Alternatively, would it be reasonably possible for `c/storage.GetStore` to return a typed error indicating that “this would work in a user namespace”?

cmd/skopeo/utils.go
```diff
@@ -16,8 +17,18 @@ type errorShouldDisplayUsage struct {
 	error
 }
 
+// Check if container-storage are used before doing the maybeReexec
 func needsRexec(c *cli.Context) error {
-	return maybeReexec()
+	for _, arg := range c.Args() {
+		parts := strings.SplitN(arg, ":", 2)
+		if len(parts) != 2 {
```
mtrmac: If we are going to do this, let’s add a `c/image/transports.TransportFromImageName` (which returns a `types.ImageTransport`) instead of hard-coding it in skopeo.

cmd/skopeo/utils.go
```diff
@@ -16,8 +17,18 @@ type errorShouldDisplayUsage struct {
 	error
 }
 
+// Check if container-storage are used before doing the maybeReexec
 func needsRexec(c *cli.Context) error {
-	return maybeReexec()
+	for _, arg := range c.Args() {
```
mtrmac: This is _way_ too imprecise.  Instead:
- In the various commands, drop `Before: needsReexec`
- In each of the command handlers, run something like `maybeReexec(inputParameter)` _only for the relevant parameters_, something like:
    ```go
        inputImageName := image.args[0]
        if err := maybeReexec(inputImageName); err != nil { 
           return err
        }
    	… := parseImageSource(ctx, opts.image, inputImage) // or parseImage or alltransports.ParseImageName
    ```
    possibly rearranging this so that the `maybeReexec`s happen at the start of the command handlers, before any other real work gets done.

cmd/skopeo/utils.go
```diff
@@ -16,8 +17,18 @@ type errorShouldDisplayUsage struct {
...
+		parts := strings.SplitN(arg, ":", 2)
+		if len(parts) != 2 {
+			continue
+		}
+		if parts[0] == storage.Transport.Name() {
```
mtrmac: (This could, in principle, be a `transport.RequiresReexecInUserNamespace()` or something like that, but ultimately we are the use of `github.com/containers/`**`storage`**`/pkg/reexec`, so that wouldn’t make the code any more general.)

fungi:
Just for another data point, we're temporarily running with a16489e applied in production and it seems to have solved #649 for us. Thanks @TristanCacqueray!

TristanCacqueray:
@mtrmac Thank you for the suggestions, I think the last change implements them along with https://github.com/containers/image/pull/631 for the TransportFromImageName procedure. Not sure what is the policy around vendoring thus I inlined the c/image change in that PR.

As @fungi mentioned, the opendev.org infrastructure is using a continuously built version of skopeo which triggered the issue of using reexec in unprivileged environment like bubblewrap.

cmd/skopeo/copy.go
```diff
@@ -87,6 +86,10 @@ func (opts *copyOptions) run(args []string, stdout io.Writer) error {
 		return errorShouldDisplayUsage{errors.New("Exactly two arguments expected")}
 	}
 
+	if err := needReexec(args); err != nil {
```
mtrmac: This works, but it’s still non-obvious that this is tied with the `args[…]` uses elsewhere. Please use actual variables,  like in https://github.com/containers/skopeo/pull/653#discussion_r284781828 , so that the `args`→image name “semantic conversion” happens at exactly one place. (IOW, so that we can, e.g. rip out and replace the CLI framework without changing the logic at all.)

cmd/skopeo/layers.go
```diff
@@ -45,6 +44,12 @@ func (opts *layersOptions) run(args []string, stdout io.Writer) (retErr error) {
 		return errors.New("Usage: layers imageReference [layer...]")
 	}
 
+	imageNames := make([]string, 1)
+	imageNames[0] = args[0]
+	if err := needReexec(imageNames); err != nil {
```
mtrmac: This would be easier to read/write with a
```go
func needReexec(imageNames ...string)
```
; no need to have an array, or artificially manufacture it, in the callers.

cmd/skopeo/unshare_linux.go
```diff
@@ -32,3 +34,13 @@ func maybeReexec() error {
 	}
 	return nil
 }
+
+func needReexec(imageNames []string) error {
```
mtrmac: (Now that this is being revisited: `needReexec` sounds as if it is returning a boolean.  I can’t think of a _good_ replacement but at least a cumbersome `reexecIfNecessaryForImages` would 1) clearly indicate an action, and 2) explain the obscure thing we are dealing with, for future readers who will have no idea that this is necessary and why. Or maybe some simpler name, with a doc string that clearly describes what is going on.)

cmd/skopeo/copy.go
TristanCacqueray: Done

cmd/skopeo/layers.go
TristanCacqueray: Thanks for the suggestion (i didn't know golang had varargs... :).

cmd/skopeo/unshare_linux.go
TristanCacqueray: Next commit uses the cumbersome name so that it embed the action purpose for the caller.

cmd/skopeo/copy.go
```diff
@@ -86,20 +85,25 @@ func (opts *copyOptions) run(args []string, stdout io.Writer) error {
 	if len(args) != 2 {
 		return errorShouldDisplayUsage{errors.New("Exactly two arguments expected")}
 	}
+	imageNames := args
+
+	if err := reexecIfNecessaryForImages(imageNames...); err != nil {
```
mtrmac: (Sigh… fine :) )

vendor.conf
```diff
@@ -2,7 +2,7 @@
 github.com/urfave/cli v1.20.0
 github.com/kr/pretty v0.1.0
 github.com/kr/text v0.1.0
-github.com/containers/image ff926d3c79684793a2135666a2cb738f44ba33dc
+github.com/containers/image 2d92e2533d517b6db4ae7d085a332915fec57867
```
mtrmac: That commit does not actually exist yet, which would make `make vendor` not reproducible.  After https://github.com/containers/image/pull/631 is merged, please update this with the actual merge commit and re-vendor.

TristanCacqueray: done

mtrmac:
LGTM. Thanks!

PR #654, opened 2019-05-18
Update release to v0.1.36

rhatdan:


mtrmac:
https://github.com/containers/image/pull/632#issuecomment-493677291 … this includes unfinished mirror support. At this point we can only hope no-one will rely on the parts that will still change.

rhatdan:
@mtrmac I found a big bug in skopeo, so will not actually be releasing this release.  But I would like to get a release out soon to fix the rootless bug, which fails if you don't have /etc/subuid setup.

PR #656, opened 2019-05-18
Skopeo crashes on any invalid transport

rhatdan:
We need to verfy that the user entered a valid transport before attempting
to see if the transport exists,  otherwise skopeo segfaults.


cmd/skopeo/unshare_linux.go
```diff
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"fmt"
```
TomSweeneyRedHat: Need to remove this, it's causing gofmt issues as it's unused.

TomSweeneyRedHat:
Changes LGTM once gofmt is happy.

mtrmac:
LGTM.

Oops. Thanks.

PR #659, opened 2019-05-20
systemtest - new set of BATS tests for RHEL8 gating

edsantiago:

### DIFF
--- a/systemtest/020-copy.bats
+++ b/systemtest/020-copy.bats
@@ -1,0 +1,79 @@
+#!/usr/bin/env bats
+#
+# Copy tests
+#
+
+load helpers
+
+function setup() {
+    standard_setup
+
+    start_registry reg
+}
+
+# From remote, to dir1, to local, to dir2;
+# compare dir1 and dir2, expect no changes
+@test "copy: dir, round trip" {
+    local remote_image=docker://busybox:latest
+    local localimg=docker://localhost:5000/busybox:unsigned
+
+    local dir1=$TESTDIR/dir1
+    local dir2=$TESTDIR/dir2
+
+    run_skopeo copy          $remote_image  dir:$dir1
+    run_skopeo copy --dest-tls-verify=false dir:$dir1  $localimg
+    run_skopeo copy  --src-tls-verify=false            $localimg  dir:$dir2
+
+    # Both extracted copies must be identical
+    diff -urN $dir1 $dir2
+}
+
+# Same as above, but using 'oci:' instead of 'dir:' and with a :latest tag
+@test "copy: oci, round trip" {
+    local remote_image=docker://busybox:latest
+    local localimg=docker://localhost:5000/busybox:unsigned
+
+    local dir1=$TESTDIR/oci1
+    local dir2=$TESTDIR/oci2
+
+    run_skopeo copy          $remote_image  oci:$dir1:latest
+    run_skopeo copy --dest-tls-verify=false oci:$dir1:latest  $localimg
+    run_skopeo copy  --src-tls-verify=false                   $localimg  oci:$dir2:latest
+
+    # Both extracted copies must be identical
+    diff -urN $dir1 $dir2
+}
+
+# Same image, extracted once with :tag and once without
+@test "copy: oci w/ and w/o tags" {
+    local remote_image=docker://busybox:latest
+
+    local dir1=$TESTDIR/dir1
+    local dir2=$TESTDIR/dir2
+
+    run_skopeo copy $remote_image oci:$dir1
+    run_skopeo copy $remote_image oci:$dir2:withtag
+
+    # Both extracted copies must be identical, except for index.json
+    diff -urN --exclude=index.json $dir1 $dir2
+
+    # ...which should differ only in the tag. (But that's too hard to check)
+    grep '"org.opencontainers.image.ref.name":"withtag"' $dir2/index.json
+}
+
+# This one seems unlikely to get fixed
+@test "copy: bug 651" {
+    skip "Enable this once skopeo issue #651 has been fixed"
+
+    run_skopeo copy --dest-tls-verify=false \
+               docker://quay.io/libpod/alpine_labels:latest \
+               docker://localhost:5000/foo
+}
+
+teardown() {
+    podman rm -f reg
+
+    standard_teardown
+}
+
+# vim: filetype=sh
--- a/systemtest/030-local-registry-tls.bats
+++ b/systemtest/030-local-registry-tls.bats
@@ -1,16 +1,7 @@
 #!/usr/bin/env bats
 #
-# This is probably a never-mind.
-#
-# The idea is to set up a local registry with locally generated certs,
-# using --dest-cert-dir to tell skopeo how to check. But no, it fails with
-#
-#        x509: certificate signed by unknown authority
-#
-# Perhaps I'm missing something? Maybe I need to add something into
-# /etc/pki/somewhere? If this is truly not possible to test without
-# a real signature, then let's just delete this test.
-#
+# Confirm that skopeo will push to and pull from a local
+# registry with locally-created TLS certificates.
 #
 load helpers
 
@@ -21,15 +12,15 @@ function setup() {
 }
 
 @test "local registry, with cert" {
-    skip "doesn't work as expected"
-
-    local remote_image=docker://busybox:latest
-    local localimg=docker://localhost:5000/busybox:unsigned
-
-    # Fails with: x509: certificate signed by unknown authority
-    run_skopeo --debug copy --dest-cert-dir=$TESTDIR/auth \
+    # Push to local registry...
+    run_skopeo copy --dest-cert-dir=$TESTDIR/client-auth \
                docker://busybox:latest \
                docker://localhost:5000/busybox:unsigned
+
+    # ...and pull it back out
+    run_skopeo copy --src-cert-dir=$TESTDIR/client-auth \
+               docker://localhost:5000/busybox:unsigned \
+               dir:$TESTDIR/extracted
 }
 
 teardown() {
--- a/systemtest/030-local-registry-tls.bats
+++ b/systemtest/030-local-registry-tls.bats
@@ -1,0 +1,41 @@
+#!/usr/bin/env bats
+#
+# This is probably a never-mind.
+#
+# The idea is to set up a local registry with locally generated certs,
+# using --dest-cert-dir to tell skopeo how to check. But no, it fails with
+#
+#        x509: certificate signed by unknown authority
+#
+# Perhaps I'm missing something? Maybe I need to add something into
+# /etc/pki/somewhere? If this is truly not possible to test without
+# a real signature, then let's just delete this test.
+#
+#
+load helpers
+
+function setup() {
+    standard_setup
+
+    start_registry --with-cert reg
+}
+
+@test "local registry, with cert" {
+    skip "doesn't work as expected"
+
+    local remote_image=docker://busybox:latest
+    local localimg=docker://localhost:5000/busybox:unsigned
+
+    # Fails with: x509: certificate signed by unknown authority
+    run_skopeo --debug copy --dest-cert-dir=$TESTDIR/auth \
+               docker://busybox:latest \
+               docker://localhost:5000/busybox:unsigned
+}
+
+teardown() {
+    podman rm -f reg
+
+    standard_teardown
+}
+
+# vim: filetype=sh
--- a/systemtest/040-local-registry-auth.bats
+++ b/systemtest/040-local-registry-auth.bats
@@ -1,0 +1,76 @@
+#!/usr/bin/env bats
+#
+# Tests with a local registry with auth
+#
+
+load helpers
+
+function setup() {
+    standard_setup
+
+    # Remove old/stale cred file
+    _cred_file=${XDG_RUNTIME_DIR:-/run/user/$(id -u)}/containers/auth.json
+    rm -f $_cred_file
+
+    # Start authenticated registry with random password
+    testuser=testuser
+    testpassword=$(random_string 15)
+
+    start_registry --testuser=testuser --testpassword=$testpassword reg
+}
+
+@test "auth: credentials on command line" {
+    # No creds
+    run_skopeo 1 inspect --tls-verify=false docker://localhost:5000/nonesuch
+    expect_output --substring "unauthorized: authentication required"
+
+    # Wrong user
+    run_skopeo 1 inspect --tls-verify=false --creds=baduser:badpassword \
+               docker://localhost:5000/nonesuch
+    expect_output --substring "unauthorized: authentication required"
+
+    # Wrong password
+    run_skopeo 1 inspect --tls-verify=false --creds=$testuser:badpassword \
+               docker://localhost:5000/nonesuch
+    expect_output --substring "unauthorized: authentication required"
+
+    # Correct creds, but no such image
+    run_skopeo 1 inspect --tls-verify=false --creds=$testuser:$testpassword \
+               docker://localhost:5000/nonesuch
+    expect_output --substring "manifest unknown: manifest unknown"
+
+    # These should pass
+    run_skopeo copy --dest-tls-verify=false --dcreds=$testuser:$testpassword \
+               docker://busybox:latest docker://localhost:5000/busybox:mine
+    run_skopeo inspect --tls-verify=false --creds=$testuser:$testpassword \
+               docker://localhost:5000/busybox:mine
+    expect_output --substring "localhost:5000/busybox"
+}
+
+@test "auth: credentials via podman login" {
+    # Logged in: skopeo should work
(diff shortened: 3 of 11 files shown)


rhatdan:
LGTM

systemtest/020-copy.bats
```diff
@@ -0,0 +1,79 @@
...
+
+# Same as above, but using 'oci:' instead of 'dir:' and with a :latest tag
+@test "copy: oci, round trip" {
+    local remote_image=docker://busybox:latest
+    local localimg=docker://localhost:5000/busybox:unsigned
+
+    local dir1=$TESTDIR/oci1
+    local dir2=$TESTDIR/oci2
+
+    run_skopeo copy          $remote_image  oci:$dir1:latest
```
TomSweeneyRedHat: nit of a nit, did you mean for so much space/tabs between copy and $remote_image?

systemtest/helpers.bash
```diff
@@ -0,0 +1,318 @@
...
+    # Called with --testuser? Create an htpasswd file
+    if [[ -n $testuser ]]; then
+        if [[ -z $testpassword ]]; then
+            die "start_registry() invoked with testuser but no testpassword"
+        fi
+
+        if ! egrep -q "^$testuser:" $AUTHDIR/htpasswd; then
+            podman run --rm --entrypoint htpasswd registry:2 \
+                   -Bbn $testuser $testpassword >> $AUTHDIR/htpasswd
+        fi
```
TomSweeneyRedHat: I've used this to setup a registry, IDK if it will solve your issues, but might spur thoughts.

```
docker run -d -p 5000:5000 --name registry -v /root/auth:/root/auth:Z -e "REGISTRY_AUTH=htpasswd" -e "REGISTRY_AUTH_HTPASSWD_REALM=Registry Realm" -e REGISTRY_AUTH_HTPASSWD_PATH=/root/auth/htpasswd -e REGISTRY_HTTP_TLS_CERTIFICATE=/root/auth/domain.crt -e REGISTRY_HTTP_TLS_KEY=/root/auth/domain.key registry:2
```

TomSweeneyRedHat:
Excellent first step @edsantiago !  I'd be fine with merging and adjusting with follow up PR's if the test pass and @mtrmac is hip.

systemtest/020-copy.bats
edsantiago: Yes, that's deliberate: that way `oci` and `localimg` line up in subsequent lines, which greatly helps my eye understand the sequence of copies.

systemtest/helpers.bash
edsantiago: Isn't that exactly what I'm doing here?

TomSweeneyRedHat: I wasn't 100% sure as it was split across several lines, so thought I'd throw out what works for me in case it helped.

edsantiago: I split it so as to avoid unnecessary arguments. If the helper is invoked with testuser/password, it creates a `htpasswd` file (lines 260-261) and adds the auth args to the registry command (line 287). If invoked with `--with-cert`, it creates the certs and adds those args to the registry command. I just didn't want to use unnecessary args; hope that makes sense.

systemtest/050-signing.bats
```diff
@@ -0,0 +1,137 @@
...
+            expect_output --substring "Source image rejected: $expected_error"
+        fi
+    done <<END_TESTS
+/myns/alice:signed
+/myns/bob:signedbyalice    Invalid GPG signature
+/myns/alice:unsigned       Signature for identity localhost:5000/myns/alice:signed is not accepted
+/myns/carol:latest         Running image docker://localhost:5000/myns/carol:latest is rejected by policy.
+/open/forall:latest
+END_TESTS
+    # FIXME: why does the message for alice:unsigned say ':signed' ?
```
mtrmac: Signatures are stored in `sigstore/`_repo_`@sha256=`_digest_`, i.e. the tag is not included (so that tag and digest references to the same image see the same signatures).

Because `alice:signed` and `alice:unsigned` are actually the same image with the same manifest digest, reading of `alice:unsigned` returns the signatures for the other image.  You should see “signature is required, but does not exist” if `alice:signed` is not created / not signed, or if the two are created from different images.

systemtest/run-tests
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+#
+# run-tests - simple wrapper allowing shortcuts on invocation
+#
+
+# FIXME
+export SKOPEO_BINARY=${SKOPEO_BINARY:-/usr/bin/skopeo}
```
mtrmac: (This bypasses the `SKOPEO_BINARY` logic in `helpers.bash`.)

systemtest/helpers.bash
```diff
@@ -0,0 +1,318 @@
...
+###############################################################################
+# BEGIN helpers for starting/stopping registries
+
+start_registry() {
+    local port=5000
+    local testuser=
+    local testpassword=
+    local create_cert=
+
+    # option processing: recognize --auth
```
mtrmac: It doesn’t AFAICS.

systemtest/helpers.bash
```diff
@@ -0,0 +1,318 @@
...
+        )
+    fi
+
+    podman run -d --name $name "${reg_args[@]}" registry:2
+}
+
+
+stop_registries() {
+    if [[ -z $SKOPEO_DEBUG_REGISTRIES ]]; then
+        podman rm -a -f
```
mtrmac: (Is this safe in the expected environment?)

systemtest/030-local-registry-tls.bats
```diff
@@ -0,0 +1,41 @@
...
+# This is probably a never-mind.
+#
+# The idea is to set up a local registry with locally generated certs,
+# using --dest-cert-dir to tell skopeo how to check. But no, it fails with
+#
+#        x509: certificate signed by unknown authority
+#
+# Perhaps I'm missing something? Maybe I need to add something into
+# /etc/pki/somewhere? If this is truly not possible to test without
+# a real signature, then let's just delete this test.
```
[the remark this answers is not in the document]
edsantiago: Interesting. The original revision of this test, copied from ancient docker tests, created a `.crt` file (not `.cert`). That led to this failure:
```
$ skopeo --debug copy --dest-cert-dir=/tmp/skopeo_bats.aLn6kG/auth docker://busybox:latest docker://localhost:5000/busybox:unsigned
...
time="2019-05-20T19:19:19Z" level=debug msg="Looking for TLS certificates and private keys in /tmp/skopeo_bats.aLn6kG/auth" 
time="2019-05-20T19:19:19Z" level=debug msg=" crt: /tmp/skopeo_bats.aLn6kG/auth/domain.crt" 
time="2019-05-20T19:19:19Z" level=debug msg=" key: /tmp/skopeo_bats.aLn6kG/auth/domain.key" 
time="2019-05-20T19:19:19Z" level=fatal msg="Error initializing destination docker://localhost:5000/busybox:unsigned: missing client certificate domain.cert for key domain.key" 
```
I was hoping to just get something working without actually having to understand this part of PKI. That was optimistic.

systemtest/040-local-registry-auth.bats
```diff
@@ -0,0 +1,76 @@
...
+# Tests with a local registry with auth
+#
+
+load helpers
+
+function setup() {
+    standard_setup
+
+    # Remove old/stale cred file
+    _cred_file=${XDG_RUNTIME_DIR:-/run/user/$(id -u)}/containers/auth.json
```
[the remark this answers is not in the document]
edsantiago: Good point! My focus had been entirely on unattended testing on a single-purpose machine; but clearly there's a need to run locally and it's rude to clobber credentials. I need to think this through better, thank you.

systemtest/050-signing.bats
edsantiago: Thank you, that makes sense. I'll remove the comment and will see if it's worthwhile to add a new test.

systemtest/030-local-registry-tls.bats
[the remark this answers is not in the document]
mtrmac: A realistic configuration would contain _only_ the CA (`.crt`) in the directory, not the registry’s private key (`.key`)

A `.key` (TLS private key to be used by the client) requires a corresponding certificate (`.cert`), so leaving around the registry _server_ private key in the clients configuration causes that error to be reported.

systemtest/helpers.bash
edsantiago: Oops. Missing documentation; fixed (not yet pushed). Thanks.

systemtest/helpers.bash
edsantiago: Good point: I wasn't thinking broadly enough. This needs some thought, because I envisioned future tests calling `start_registry()` more than once. Thanks again.

systemtest/030-local-registry-tls.bats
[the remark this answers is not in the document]
edsantiago: Thank you! That was the trick. New fixup commit pushed.

systemtest/050-signing.bats
edsantiago: Comments removed. I tried to add a new `/myns/bob:signed` image, using `busybox:1.30`, but that ended up breaking the `/myns/bob:signedbyalice` test so I think i'll just leave this set of tests for now.

systemtest/run-tests
edsantiago: Removed - it was a holdover from earlier testing that I hadn't removed. Thanks for catching it.

systemtest/helpers.bash
edsantiago: `stop_registries()` was never used anywhere, and it's much easier to just have individual tests run `podman rm -f REGISTRY` in their own `teardown()`. Removed `stop_registries()`.

edsantiago:
In response to your wrapper questions:

>  Is anything in this repo going to invoke the tests? I’m worried that they will get silently out of sync / broken if not regularly exercised.

Complete agreement: we have a mess with podman right now because the gating tests don't run in CI, so things break often and I have to go back and run damage control. 

I would like to run these tests as part of CI. Do you know how/where I could add them?

> What is the intended relationship between this and the tests in integration/? Should this eventually replace them? Is there a reason to independently keep both long-term?

I honestly don't know. The integration tests currently (seem to?) cover more than these do, but I honestly just can't grok them nor can I figure out how to run them, let alone package them in an rpm. I think if someone finds them useful, they should remain. My goal in these tests was to have something readable, simple, so that if they fail a monkey like me can read the test, understand it, and find a clean reproducer for filing bugs. From what I see of the ginkgo tests, diagnosis can only be done by experts.

Many thanks for your thorough and helpful review.

mtrmac:
> I would like to run these tests as part of CI. Do you know how/where I could add them?

The CI starts at `.travis.yml`, so either there, or inside the existing container (`make check` / `make test-integration`).
 
> > What is the intended relationship between this and the tests in integration/? Should this eventually replace them? Is there a reason to independently keep both long-term?
> 
> I honestly don't know. The integration tests currently (seem to?) cover more than these do, but I honestly just can't grok them nor can I figure out how to run them, let alone package them in an rpm. 

Yeah, they currently pretty strongly expect to be running inside a container (which should be reasonably possible to fix) and include building several registry servers and a two-year-old OpenShift (which _could_ be fixed but maybe shouldn’t :), for RPM tests; ideally we would maintain pre-built containers for this, and get the CI time from 20 minutes to… 5?)

As for Go vs. BATS, that rather depends on the complexity of the test code — the Go `assertSkopeoSucceeds` and the like is suspiciously close to a shell script already, with extra punctuation.

> I think if someone finds them useful, they should remain. My goal in these tests was to have something readable, simple, so that if they fail a monkey like me can read the test, understand it, and find a clean reproducer for filing bugs. From what I see of the ginkgo tests, diagnosis can only be done by experts.

I thought that that code was reasonably clear about what it was executing … though setting up a reproducer environment, even with the available `make shell`, is certainly a lot of pain.

edsantiago:
>The CI starts at .travis.yml

AFAICT the Travis scripts only build containerized skopeo, never actually `binary-local` or any equivalent. I have no idea how to invoke the BATS tests in this scenario, so for the time being I'm giving up. Any objection to merging this as it is?

mtrmac:
> > The CI starts at .travis.yml
> 
> AFAICT the Travis scripts only build containerized skopeo, never actually `binary-local` or any equivalent. I have no idea how to invoke the BATS tests in this scenario, so for the time being I'm giving up.

`.travis.yml`→`make check`→`make test-integration`→`make build-container`&`$(CONTAINER_RUN) hack/make.sh test-integration`→`hack/make/test-integration`→`make binary-local`&`make install`&`go test ./integration`.  Yes, it would be great to streamline this.

edsantiago:
I don't think I spoke clearly, I just don't have the right vocabulary to do so. What I was trying to say was that the CI process never actually builds a skopeo binary on any actual host VM, it only does so inside a container. The BATS tests use podman to run a registry, and running podman-within-podman is fragile: I did not want to spend time on that. 

But I just did, spent the morning trying to get a `test-system` target running inside the container, and cannot. The most recent failure, the one I'm now giving up on, is:
```
Error: could not get runtime: kernel does not support overlay fs: 'overlay' is not supported over <unknown> at "/var/lib/containers/storage/overlay": backing file system is unsupported for this graph driver
```

rhatdan:
@edsantiago volume mount an empty directory on /var/lib/containers.

podman run -v /tmp/tempdir123444:/var/lib/containers:Z ...

edsantiago:
@rhatdan that then yields (after much fussing):
```
Error: could not get runtime: kernel does not support overlay fs: 'overlay' is not supported over tmpfs at "/var/lib/containers/storage/overlay": backing file system is unsupported for this graph driver
```
Changing `/etc/containers/storage.conf:driver` from overlay to vfs (before any podman ops) yields:
```
ERRO[0000] error creating network namespace for container 4cbf2226c0afad24f35c08b9581039eb1620cefe05ad8a62139f86c6a461e0bd: mount --make-rshared /var/run/netns failed: "operation not permitted" 
Error: failed to mount shm tmpfs "/var/lib/containers/storage/vfs-containers/4cbf2226c0afad24f35c08b9581039eb1620cefe05ad8a62139f86c6a461e0bd/userdata/shm": operation not permitted
```
(adding `--net=host` resolves the first ERRO message but not the failed-to-mount one).

This is why I was reluctant to dive into podman-in-podman...

rhatdan:
@edsantiago Use /var/tmp not /tmp.  But this would require a --privileged container anyways.  Podman will not be able to run within a locked down podman container.

edsantiago:
Using `/var/tmp`, running as root with `--privileged`, hand-editing `/etc/containers/storage.conf` to change overlay to vfs, removing the `mountopt` flag, adding `--cgroup-manager=cgroupfs` to all podman invocations. With that I can finally get podman-in-podman to work without error, on simple stuff, but networking doesn't seem to work: all skopeo copy commands fail with:
```
time="2019-05-28T14:22:45Z" level=fatal msg="Error creating an updated image manifest: Get https://production.cloudflare.docker.com/registry-v2/docker/registry/v2/blobs/sha256/64/64f5d945efcc0f39ab11b3cd4ba403cc9fefe1fa3613123ca016cf3708e8cafb/data?verify=1559056335-YjM3xM0QCt3jPGWIaQQz%2FthEZTQ%3D: dial tcp: i/o timeout"
```
or
```
time="2019-05-28T14:22:07Z" level=fatal msg="Error writing blob: read tcp 10.88.9.30:35970->104.18.121.25:443: read: connection reset by peer" 
```
or
```
time="2019-05-28T14:23:20Z" level=fatal msg="Error initializing source docker://busybox:latest: Get https://auth.docker.io/token?scope=repository%3Alibrary%2Fbusybox%3Apull&service=registry.docker.io: net/http: TLS handshake timeout"
```
I feel like I'm in diminishing-returns territory: spending days of effort without any actual progress. Is there any possible way to get this CI to run without the multiple levels of podman-in-podman?

edsantiago:
Rebased, and added one more commit which (on my setup) addresses the podman-in-podman problems while still allowing it to work on a regular setup (i.e. for gating tests). The tests should now run as part of the in-progress-as-i-write-this CI.

edsantiago:
CI ran new tests, and they passed, but output looks weird; new force-push invokes bats with `--tap` to see if that fixes the output. There's also this weirdness when cleaning up the tempdir:
```
rm: cannot remove ‘/var/tmp/podman-tmp.eVP6m2/storage’: Permission denied
rm: cannot remove ‘/var/tmp/podman-tmp.eVP6m2/cache’: Permission denied
```

edsantiago:
`--tap` fixed the broken output. EPERM issue remains; I'm not going to spend any time on it unless requested to.

edsantiago:
Ping, is there anything more needed on this? I'd like to move on to the actual gating-test aspect.

rhatdan:
@vrothberg @mtrmac Can we merge this?

mtrmac:
Merging, sorry for the delay.

PR #665, opened 2019-05-29
Add ability to override the authfile location using env variables

randomswdev:
While the --authfile option allows to override the location where skopeo search for the docker authentication file, when building a container it would be nice to allow specifying a custom path using an environment variable.

This is mainly useful if merging skopeo in a container including also kaniko; kaniko is able to build images, while skopeo can supprot registry manipulation functions, thus providing capabilities equivalent to docker without requiring dind.

Kaniko allows to customize the search path for config.json using the variable DOCKER_CONFIG; this ccode change would allow supporting it also for skopeo.

### DIFF
--- a/vendor/github.com/containers/image/pkg/docker/config/config.go
+++ b/vendor/github.com/containers/image/pkg/docker/config/config.go
@@ -31,6 +31,7 @@ var (
 	xdgRuntimeDirPath       = filepath.FromSlash("containers/auth.json")
 	dockerHomePath          = filepath.FromSlash(".docker/config.json")
 	dockerLegacyHomePath    = ".dockercfg"
+	dockerConfigEnvVar      = "DOCKER_CONFIG"
 
 	// ErrNotLoggedIn is returned for users not logged into a registry
 	// that they are trying to logout of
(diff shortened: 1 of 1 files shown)


mtrmac:
Thanks for the PR.

In general, reading environment variables in libraries (like containers/image in this case) is always a bit problematic, because it’s completely non-obvious to the caller, and more or less out of its control; explicit parameters (like `types.SystemContext` in the case of c/image) are much clearer. , This is especially the case when added later to an existing functionality; the caller can’t reasonably be expected to notice that the behavior of the called code now depends on a new environment variable that the caller must clear in order to get full control of the configuration.

The same argument basically equally applies to the `skopeo` executable itself; explicit is better than implicit, and using explicit parameters prevents random environment settings from breaking scripts that don’t expect that environment setting to be present.

(That said, there _are_ a few cases where c/image is reading environment variables, for better interoperability; but they exist primarily for historical reasons = compatibility with previous mainstream implementations, not added to existing code as new features; or for extremely-widely supported variables like `HOME`.)

(BTW, if this change should happen, it should happen in the c/image repo, definitely not by patching vendored code. But the arguments against it apply equally.)

For the most part, it seems to me that
```sh
export DOCKER_CONFIG=/some/path
skopeo copy …
```
and
```
skopeo_copy='skopeo copy --authfile /some/path'
$skopeo_copy …
```
are pretty close in effect.  Sure, the exported variable applies to any random subprocess automatically, but that’s exactly the undesirable property in other situations.

randomswdev:
Ok, I agree with your point. I'm closing the merge request.

PR #668, opened 2019-06-10
Explicitly disable encrypting the test GPG key

mtrmac:
Since GPG 2.1, GPG asks for a passphrase by default; opt out when generating the test key, to fix
```
gpg: agent_genkey failed: No pinentry
gpg: key generation failed: No pinentry
```
which happens otherwise.


mtrmac:
(Merging immediately to unblock CI.)

PR #670, opened 2019-06-11
Fixing static binary build using container

vijaySamanuri:
1. LookupId panics on Linux+glibc static build (https://github.com/golang/go/issues/24787)
 2. Updated go version to 1.11.0 in Dockerfile.build
 3. added go build tags netgo osusergo in static binary

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,7 @@ LOCAL_BUILD_TAGS = $(BTRFS_BUILD_TAG) $(LIBDM_BUILD_TAG) $(OSTREE_BUILD_TAG) $(D
 BUILDTAGS += $(LOCAL_BUILD_TAGS)
 
 ifeq ($(DISABLE_CGO), 1)
-	override BUILDTAGS = containers_image_ostree_stub exclude_graphdriver_devicemapper exclude_graphdriver_btrfs containers_image_openpgp
+	override BUILDTAGS = netgo osusergo containers_image_ostree_stub exclude_graphdriver_devicemapper exclude_graphdriver_btrfs containers_image_openpgp
 endif
 
 #   make all DEBUG=1
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -1,8 +1,14 @@
 FROM ubuntu:18.10
 
-RUN apt-get update && apt-get install -y \
-    golang \
-    libbtrfs-dev \
+RUN apt-get update && apt-get install -y wget
+RUN wget -P /tmp/ https://dl.google.com/go/go1.11.linux-amd64.tar.gz && \
+    tar -xvf /tmp/go1.11.linux-amd64.tar.gz -C /usr/local/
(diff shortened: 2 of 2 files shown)


rhatdan:
LGTM

rhatdan:
@vijaySamanuri You need to sign your commit.

git commit -a -s --amend
git push --force

Makefile
```diff
@@ -59,7 +59,7 @@ LOCAL_BUILD_TAGS = $(BTRFS_BUILD_TAG) $(LIBDM_BUILD_TAG) $(OSTREE_BUILD_TAG) $(D
 BUILDTAGS += $(LOCAL_BUILD_TAGS)
 
 ifeq ($(DISABLE_CGO), 1)
-	override BUILDTAGS = containers_image_ostree_stub exclude_graphdriver_devicemapper exclude_graphdriver_btrfs containers_image_openpgp
+	override BUILDTAGS = netgo osusergo containers_image_ostree_stub exclude_graphdriver_devicemapper exclude_graphdriver_btrfs containers_image_openpgp
```
mtrmac: Thanks for the PR.

- `netgo` bypasses NSS, including the glibc resolver (per https://golang.org/pkg/net/ , e.g. breaking `.local` names.
- `osusergo` bypasses NSS as well, breaking e.g. LDAP uses.

Yes, I realize that these downsides are fundamentally necessary to make static builds work at all, but they are _significant_ downsides.

I’m sorely tempted to just remove any mention of statically-linked or non-CGo builds from the repository (@rhatdan WDYT?).

If not that, at the very least `README` should document these downsides of a static build as well.

rhatdan: I would hate to remove them, since it seems many in the GO world seem to want/require them.  I guess Documenting the shortcomings is the best solution.

The NSS stuff is only important for rootless container/storage I believe?

rhatdan:
@vijaySamanuri Are you still interested in completing this?

rhatdan:
I think we should move forward with this and just document the shortcomings of a static built version of skopeo.

rhatdan:
@vrothberg Worth taking this over, or should we just close it?

vrothberg:
> @vrothberg Worth taking this over, or should we just close it?

I have a tendency to close. I'm skeptical toward static when it's not fully supported - shall we drop the make target?

mtrmac:
Ideally, I’d prefer dropping _both_ statically-linked builds for Linux (which are unreliable and functionally restricted) and in-container builds (which are a good idea in theory, but keep breaking because we don’t use them, and don’t work with dynamically-linked builds).

Dropping the statically-linked targets would be a good first step — though this PR suggests that the in-container builds might also be broken, independently.

vrothberg:
+1, I'll create an issue for it.

vrothberg:
https://github.com/containers/skopeo/issues/755

PR #671, opened 2019-06-11
copy: add --dest-compress-format and --dest-compress-level

giuseppe:
add the possibility to specify the format and the level to use when
compressing blobs.


vendor.conf
```diff
@@ -2,14 +2,14 @@
...
-github.com/containers/image 2c0349c99af7d90694b3faa0e9bde404d407b145
 github.com/containers/buildah 810efa340ab43753034e2ed08ec290e4abab7e72
 github.com/vbauerster/mpb v3.3.4
 github.com/mattn/go-isatty v0.0.4
 github.com/VividCortex/ewma v1.1.1
 golang.org/x/sync 42b317875d0fa942474b76e1b46a6060d720ae6e
 github.com/opencontainers/go-digest c9281466c8b2f606084ac71339773efd177436e7
-github.com/containers/storage v1.12.3
+github.com/containers/storage zstd https://github.com/giuseppe/storage
+github.com/containers/image zstd https://github.com/giuseppe/image
```
mtrmac: (I think originally the idea was to list c/image, and then all it's dependencies. OTOH `go mod` is going to disrupt that anyway.)

docs/skopeo-copy.1.md
```diff
@@ -56,6 +56,10 @@ If the authorization state is not found there, $HOME/.docker/config.json is chec
 
 Existing signatures, if any, are preserved as well.
 
+**--dest-compress-format** _format_ Specifies the compression format to use.
+
+**--dest-compress-level** _format_ Specifies the compression level to use.
```
mtrmac: This needs to be much clearer about what values are acceptable (and useful), and maybe what the default is.  (Consider improving the `--help` output similarly as well.)

mtrmac: Also needs updated bash completions, I guess.

giuseppe:
rebased

giuseppe:
> @giuseppe, can we add some tests on top?

I've added a test to verify the compressed layer is a .zstd file

cmd/skopeo/flag.go
```diff
@@ -73,3 +74,37 @@ func (ob *optionalStringValue) String() string {
...
+	ob.value = v
+	ob.present = true
+	return nil
+}
+
+func (ob *optionalIntValue) String() string {
+	if !ob.present {
+		return "" // This is, sadly, not round-trip safe: --flag= is interpreted as {present:true, value:""}
+	}
+	return fmt.Sprintf("%d", ob.value)
```
rhatdan: Isn't strconv better here then using fmt?

cmd/skopeo/utils.go
```diff
@@ -193,6 +206,17 @@ func (opts *imageDestOptions) newSystemContext() (*types.SystemContext, error) {
 	ctx.OSTreeTmpDirPath = opts.osTreeTmpDir
 	ctx.DirForceCompress = opts.dirForceCompression
 	ctx.OCIAcceptUncompressedLayers = opts.ociAcceptUncompressedLayers
+	if opts.compressionFormat == "" {
+		opts.compressionFormat = "gzip"
```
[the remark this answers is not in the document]
giuseppe: yes, I've fixed it in the new version

cmd/skopeo/flag.go
giuseppe: fixed now

cmd/skopeo/utils.go
```diff
@@ -147,15 +148,23 @@ func (opts *imageOptions) newSystemContext() (*types.SystemContext, error) {
 	if opts.noCreds {
 		ctx.DockerAuthConfig = &types.DockerAuthConfig{}
 	}
+	cf, err := compression.AlgorithmByName("gzip")
```
mtrmac: This shouldn’t hard-code a Skopeo-specific default; the default in c/image should be used.

cmd/skopeo/utils.go
```diff
@@ -179,6 +188,17 @@ func imageDestFlags(global *globalOptions, shared *sharedImageOptions, flagPrefi
 			Usage:       "Allow uncompressed image layers when saving to an OCI image using the 'oci' transport. (default is to compress things that aren't compressed)",
 			Destination: &opts.ociAcceptUncompressedLayers,
 		},
+		cli.StringFlag{
+			Name:        flagPrefix + "compress-format",
+			Usage:       "`FORMAT` to use for the compression",
+			Destination: &opts.compressionFormat,
+			Value:       "gzip",
```
mtrmac: Skopeo shouldn’t hard-code a default; and it definitely should not hard-code it in two places at once.

cmd/skopeo/utils_test.go
```diff
@@ -45,6 +45,8 @@ func TestImageOptionsNewSystemContext(t *testing.T) {
 	opts := fakeImageOptions(t, "dest-", []string{}, []string{})
 	res, err := opts.newSystemContext()
 	require.NoError(t, err)
+
+	res.CompressionFormat = nil // reflect doesn't work well with functions
```
mtrmac: I can only vaguely guess what this refers to. And if the comparison using the existent `assert.Equal` does not work for that value, I’d expect that assertion to be made in some alternate way before making it invisible to the existing `assert.Equal`.

(Applies to all instances; maybe there should be a shared helper, no idea.)

docs/skopeo-copy.1.md
```diff
@@ -58,6 +58,10 @@ If the authorization state is not found there, $HOME/.docker/config.json is chec
 
 Existing signatures, if any, are preserved as well.
 
+**--dest-compress-format** _format_ Specifies the compression format to use.
```
mtrmac: Still needs to document the available values, or maybe how to discover them.

cmd/skopeo/flag.go
```diff
@@ -73,3 +73,37 @@ func (ob *optionalStringValue) String() string {
...
+// optionalInt is a int with a separate presence flag.
+type optionalInt struct {
+	present bool
+	value   int
+}
+
+// optionalInt is a cli.Generic == flag.Value implementation equivalent to
+// the one underlying flag.Int, except that it records whether the flag has been set.
+// This is distinct from optionalInt to (pretend to) force callers to use
+// newoptionalInt
```
mtrmac: “`newOptionalIntValue`”

cmd/skopeo/flag.go
```diff
@@ -73,3 +73,37 @@ func (ob *optionalStringValue) String() string {
...
+		return err
+	}
+	ob.value = v
+	ob.present = true
+	return nil
+}
+
+func (ob *optionalIntValue) String() string {
+	if !ob.present {
+		return "" // This is, sadly, not round-trip safe: --flag= is interpreted as {present:true, value:""}
```
mtrmac: `""` is not a valid value of `ob.value` in this case. What actually happens?

cmd/skopeo/flag.go
```diff
@@ -73,3 +73,37 @@ func (ob *optionalStringValue) String() string {
...
+// newoptionalInt
+type optionalIntValue optionalInt
+
+func newOptionalIntValue(p *optionalInt) cli.Generic {
+	p.present = false
+	return (*optionalIntValue)(p)
+}
+
+func (ob *optionalIntValue) Set(s string) error {
+	v, err := strconv.Atoi(s)
```
mtrmac: `flag.Int` seems to use something else; if this promises to be equivalent, it must be equivalent.

cmd/skopeo/utils.go
vrothberg: We have a chicken-egg problem now as we agreed to add the constants with https://github.com/containers/image/pull/563.

mtrmac: - Skopeo shouldn’t hard-code _any_ default; so it should not need the constants (except, maybe, I suppose, to list the right names of algorithms in a help message, if one wanted to be extremely pedantic about not hard-coding that).
- Even if this did need to use the constants, how is that a chicken-egg problem? It seems to be just a simple one-directional dependency; add constants, add options to skopeo. Or is the CI now broken because c/image has merged a change without having a skopeo PR actually ready, or something? (I‘m afraid I’ve been in meetings for the past two days and I’m a bit out of the loop.)

cmd/skopeo/utils_test.go
giuseppe: now that we have a default in containers/image, there is no need of the special handling, going to drop these

docs/skopeo-copy.1.md
giuseppe: I'll list the supported algorithms

cmd/skopeo/flag.go
giuseppe: the only user of the function is currently checking if the value is present, so this case should never happen.

To make it more generic, should I add a new argument to `newOptionalIntValue` to accept a default value to use when `!ob.present`?

mtrmac: Just fix the comment to be accurate as to what happens; no need to add dead code.

giuseppe: Thanks.  I've changed the comment

giuseppe:
@mtrmac does it look fine now?

giuseppe:
can this be reviewed?

docs/skopeo-copy.1.md
```diff
@@ -58,6 +58,10 @@ If the authorization state is not found there, $HOME/.docker/config.json is chec
 
 Existing signatures, if any, are preserved as well.
 
+**--dest-compress-format** _format_ Specifies the compression format to use.  Supported values are: `gzip`, `bzip2`, `xz`, `zstd`.
```
vrothberg: A problem with using xz and bzip2 is that those are not reflected in the manifest. Once https://github.com/containers/image/pull/563 is merged, it will actually fail during manifest update/conversion. I would prefer to not mention them.

giuseppe: psuhed a new version without `xz` and `bzip2`

docs/skopeo-copy.1.md
```diff
@@ -58,6 +58,10 @@ If the authorization state is not found there, $HOME/.docker/config.json is chec
 
 Existing signatures, if any, are preserved as well.
 
+**--dest-compress-format** _format_ Specifies the compression format to use.  Supported values are: `gzip` and `zstd`.
+
+**--dest-compress-level** _format_ Specifies the compression level to use.  The value is specific to the compression algorithm used, e.g. for zstd the accepted values are in the range 1-20 (inclusive), while for gzip it is 1-9 (inclusive).  Algorithms that don't support the setting, will ignore it.
```
vrothberg: The last sentence is now redundant since we only support gzip and zstd.

rhatdan:
LGTM

PR #673, opened 2019-06-12
Skip systemtest/050-signing.bats if skopeo can't create signatures

mtrmac:
This does not happen in this repo's tests, but containers/image's `make test-skopeo` fails in the `containers_image_openpgp` configuration with

```
not ok 10 signing
...
# time="2019-06-11T20:59:32Z" level=fatal msg="Signing not supported: signing is not supported in github.com/containers/image built with the containers_image_openpgp build tag"
```

To reproduce/test this:
```sh
make test-system BUILDTAGS='ostree containers_image_openpgp'
```

Cc: @edsantiago

### DIFF
--- a/systemtest/050-signing.bats
+++ b/systemtest/050-signing.bats
@@ -82,6 +82,15 @@ END_POLICY_JSON
 }
 
 @test "signing" {
+    run_skopeo '?' standalone-sign /dev/null busybox alice@test.redhat.com -o /dev/null
+    if [[ "$output" =~ 'signing is not supported' ]]; then
+        skip "skopeo built without support for creating signatures"
+        return 1
+    fi
+    if [ "$status" -ne 0 ]; then
+        die "exit code is $status; expected $expected_rc"
+    fi
+
     # Cache local copy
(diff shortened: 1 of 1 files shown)


edsantiago:
Nice catch. Is there any other way to catch this situation earlier? `skopeo --version | grep something`? I think it'd be much nicer if we could do:
```
    if SOME CONDITION WE CAN TEST FOR; then
        skip "this skopeo doesn't support signing"
    fi
```

mtrmac:
Yeah. `skopeo standalone-sign` with two `/dev/null`s and the like is not great, but it _is_ better than aborting the loop in the middle. Thanks for kicking me in the right direction, PTAL.

umohnani8:
LGTM

edsantiago:
Much nicer, thank you! My one minor suggestion might be to be more flexible in the error message you check for:
```
- if [[ "$output" =~ 'signing is not supported in github.com/containers/image built with the ...
+ if [[ "$output" =~ 'signing is not supported` ]]; then
```
This allows the `containers/image` message to change (slightly) without impacting us; it also allows other TBD conditions to use 'signing is not supported'. But I'll accept this too.

Issue #675, opened 2019-06-13
systemtest flakes


mtrmac:
Not always, but often enough, system tests fails with something like
```
not ok 8 auth: credentials on command line
# (from function `expect_output' in file systemtest/helpers.bash, line 189,
#  in test file systemtest/040-local-registry-auth.bats, line 27)
#   `expect_output --substring "unauthorized: authentication required"' failed
# grep: /tmp/skopeo_bats.WTZxVj/auth/htpasswd: No such file or directory
# 66f1545ce2662b1957fd55ff3ee1540af40fbc10bbbd462caa2e444805ad788b
# $ /usr/bin/skopeo inspect --tls-verify=false docker://localhost:5000/nonesuch
# time="2019-06-13T12:57:38Z" level=fatal msg="pinging docker registry returned: Get http://localhost:5000/v2/: dial tcp 127.0.0.1:5000: connect: connection refused" 
# [ rc=1 (expected) ]
# #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
# #|     FAIL: skopeo inspect --tls-verify=false docker://localhost:5000/nonesuch
# #| expected: 'unauthorized: authentication required'
# #|   actual: 'time="2019-06-13T12:57:38Z" level=fatal msg="pinging docker registry returned: Get http://localhost:5000/v2/: dial tcp 127.0.0.1:5000: connect: connection refused" '
# #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
# 66f1545ce2662b1957fd55ff3ee1540af40fbc10bbbd462caa2e444805ad788b
```
(omitting many instances of `level=error msg="unable to write pod event: "journal error: could not connect to journald socket""`

Restarting the build usually lets it proceed.

Not sure how to best fix this (or even, for sure, what the cause actually is) – maybe `start_registry` just needs to wait for the registry to start up (some random time, or until the port is known to be working).

Cc: @edsantiago

PR #676, opened 2019-06-13
DO NOT MERGE (prefer 677): Wait for the registry container to start up in systemtests

mtrmac:
... to attempt to avoid
```shell
$ for i in $(seq 1 10); do echo ====$i; systemtest/run-tests 040 || break; done
====1
 ✓ auth: credentials on command line
 ✗ auth: credentials via podman login
 (in test file systemtest/040-local-registry-auth.bats, line 54)
     `podman login --tls-verify=false -u $testuser -p $testpassword localhost:5000' failed with status 125
   grep: /tmp/skopeo_bats.ydmiLB/auth/htpasswd: No such file or directory
   f55c04a80cf4f9dba02cea68980c0c5d2493d7c2cf3f631c7ab7ed9dd65d0ca0
   Error: error authenticating creds for "localhost:5000": pinging docker registry returned: Get http://localhost:5000/v2/: dial tcp [::1]:5000: connect: connection refused
   f55c04a80cf4f9dba02cea68980c0c5d2493d7c2cf3f631c7ab7ed9dd65d0ca0
```

Fixes #675 , hopefully.

Cc: @edsantiago

PR #677, opened 2019-06-13
start_registry: wait for registry to be ready

edsantiago:
The usual 'podman run -d' race condition: we've been forking
off the container but not actually making sure it's up; this
leads to flakes in which we try (and fail) to access it.

Solution: use curl to check the port; we will expect a zero
exit status once we can connect. Time out at ten seconds.

Resolves: #675

### DIFF
--- a/systemtest/helpers.bash
+++ b/systemtest/helpers.bash
@@ -310,6 +310,18 @@ start_registry() {
     fi
 
     $PODMAN run -d --name $name "${reg_args[@]}" registry:2
+
+    # Wait for registry to actually come up
+    timeout=10
+    while [[ $timeout -ge 1 ]]; do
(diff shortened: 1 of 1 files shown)


edsantiago:
@mtrmac this one avoids the 10-second wait; ptal

mtrmac:
Thanks!

Yes, that’s much nicer than #676. LGTM pending tests.

edsantiago:
Side note, simply for the record: I considered adding `&>/dev/null` to the curl invocation, because it's possible that a test will fail and the 'connection refused' in the logs might lead the troubleshooter down the wrong path. I decided to leave the output just in case there's an error other than 'connection refused', and I decided not to add a 'echo registry container is up' message before `return` because, noisy.

PR #681, opened 2019-06-18
migrate to go modules

vrothberg:


vrothberg:
@mtrmac @rhatdan PTAL. I got upset by the `vndr` hiccups and wanted to that for a longer time. CI is now green.

rhatdan:
LGTM, But we need  a way to extract the man pages out of containers/image for distribution in the rpm spec.

vrothberg:
> LGTM, But we need a way to extract the man pages out of containers/image for distribution in the rpm spec.

That's not possible with `go mod`. IMHO this should be solved on the packaging level (e.g., having multiple sources in the `.spec`).

vrothberg:
@nalind, this smells like a storage issue. Once I ran into the issue, I can't even do `buildah from scratch` anymore:

```bash
./buildah --debug from scratch
[...]
DEBU[0000] [graphdriver] trying provided driver "overlay"                                        
DEBU[0000] overlay: mount_program=/usr/bin/fuse-overlayfs                                        
DEBU[0000] backingFs=extfs, projectQuotaSupported=false, useNativeDiff=false, usingMetacopy=false
# no output anymore
```

rhatdan:
@mtrmac @runcom @saschagrunert PTAL

mtrmac:
> > LGTM, But we need a way to extract the man pages out of containers/image for distribution in the rpm spec.
> 
> That's not possible with `go mod`. IMHO this should be solved on the packaging level (e.g., having multiple sources in the `.spec`).

The last conversation around that has been in #645 ; I can’t see any progress in the last month.

Dockerfile
```diff
@@ -11,6 +11,8 @@ RUN dnf -y update && dnf install -y make git golang golang-github-cpuguy83-go-md
 	# OpenShift deps
 	which tar wget hostname util-linux bsdtar socat ethtool device-mapper iptables tree findutils nmap-ncat e2fsprogs xfsprogs lsof docker iproute \
         bats jq podman \
+	# golint from packages instead of upstream
```
mtrmac: This is not quite a meaningful comment at this point; what’s the reader supposed to do or not do based on it?

Maybe just the commit message is enough. Or this should say something like ”don’t use (go get) to install this because…”, and it’s not clear to me why it would apply specially to `golint` and not to other packages like `podman` just above.

go.mod
```diff
@@ -0,0 +1,66 @@
+module github.com/containers/skopeo
+
+go 1.12
+
+require (
```
mtrmac: (I didn’t compare this line by line with the old `vendor.conf`)

Makefile
```diff
@@ -171,10 +171,8 @@ validate-local:
...
-
-vendor: vendor.conf .install.vndr
-	$(GOPATH)/bin/vndr \
-		-whitelist '^github.com/containers/image/docs/.*' \
-		-whitelist '^github.com/containers/image/registries.conf'
+vendor:
+	export GO111MODULE=on \
+		$(GO) mod tidy && \
+		$(GO) mod vendor && \
+		$(GO) mod verify
```
saschagrunert: The missing newline will trigger unnecessary whitespace changes with some editors.

Dockerfile
vrothberg: I'll remove the line to avoid confusion.

vrothberg:
Updated, @mtrmac @saschagrunert, PTAL.

mtrmac:
@lsm5 FYI, `skopeo/vendor` no longer contains `registries.conf` and c/image man pages.

PR #682, opened 2019-06-20
[WIP] List all repos with tags

danielorf:
Add the ability to list all repos and tags in a registry with the `list` subcommand.  This could also be used as a precursor for registry search.

Some items still in progress:
- Documentation
- Error handling
- Formatting of output - long repo names and tag slices can overwhelm Stdout as is
- Architecture of helper struct/methods currently in registry_utils.go - I chose to break this out from utils.go due to the size of the added content and specific application to registries
- Consider utilizing docker libs for auth - though it appears easier to re-write as I have done, but still prefer to rely on existing libs if possible

### DIFF
--- a/cmd/skopeo/list.go
+++ b/cmd/skopeo/list.go
@@ -1,0 +1,63 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"io"
+
+	"github.com/urfave/cli"
+)
+
+type listOptions struct {
+	global      *globalOptions
+	image       *imageOptions
+	regUsername string
+	regPassword string
+}
+
+func listCmd(global *globalOptions) cli.Command {
+	sharedFlags, sharedOpts := sharedImageFlags()
+	imageFlags, imageOpts := imageFlags(global, sharedOpts, "", "")
+	opts := listOptions{
+		global: global,
+		image:  imageOpts,
+	}
+	return cli.Command{
+		Name:  "list",
+		Usage: "Get list of repos of REGISTRY-NAME",
+		Description: fmt.Sprint(`
+		Get list of repos of REGISTRY-NAME
+
+		Supports Docker v2 registries only.
+
+		See skopeo(1) section "REGISTRY-NAMES" for the expected format
+		`),
+		ArgsUsage: "REGISTRY-NAME",
+		Before:    needsRexec,
+		Action:    commandAction(opts.run),
+		// Flags:     append(sharedFlags, imageFlags...),
+		Flags: append(append([]cli.Flag{
+			cli.StringFlag{
+				Name:        "username, u",
+				Usage:       "Registry username, defaults to empty string",
+				Destination: &opts.regUsername,
+			},
+			cli.StringFlag{
+				Name:        "password, p",
+				Usage:       "Registry password, defaults to empty string",
+				Destination: &opts.regPassword,
+			},
+		}, sharedFlags...), imageFlags...),
+	}
+}
+
+func (opts *listOptions) run(args []string, stdout io.Writer) error {
+	if len(args) != 1 {
+		return errors.New("Usage: list registryURL")
+	}
+
+	reg := NewRegistry(args[0])
+	reg.getAllReposWithTags()
+
+	return nil
+}
--- a/cmd/skopeo/list.go
+++ b/cmd/skopeo/list.go
@@ -33,7 +33,6 @@ func listCmd(global *globalOptions) cli.Command {
 		See skopeo(1) section "REGISTRY-NAMES" for the expected format
 		`),
 		ArgsUsage: "REGISTRY-NAME",
-		Before:    needsRexec,
 		Action:    commandAction(opts.run),
 		// Flags:     append(sharedFlags, imageFlags...),
 		Flags: append(append([]cli.Flag{
@@ -56,7 +55,10 @@ func (opts *listOptions) run(args []string, stdout io.Writer) error {
 		return errors.New("Usage: list registryURL")
 	}
 
-	reg := NewRegistry(args[0])
+	reg, err := NewRegistry(args[0])
+	if err != nil {
+		return err
+	}
 	reg.getAllReposWithTags()
 
 	return nil
--- a/cmd/skopeo/registry_utils.go
+++ b/cmd/skopeo/registry_utils.go
@@ -1,0 +1,251 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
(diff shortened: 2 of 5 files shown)


vrothberg:
@danielorf did you check containers/image? It already support searching repositories, which podman-search is using. I am not sure if the proposed list command is something for skopeo given search is implemented in Podman.

@rhatdan @mtrmac @baude @mheon WDYT?

danielorf:
@vrothberg I wasn't aware of this feature in containers/image.  The skopeo [TODO list](https://github.com/containers/skopeo#todo) also adds to the confusion since it lists this as a wanted feature.

vrothberg:
@danielorf, good point! I am open to add search and list to Skopeo. Some users might use Skopeo in isolation without Podman and supporting that would give flexibility.

Regarding this particular PR: the low-level plumbing should happen in c/image, and much is already in place. A good starting point could be [libpod's search implementation](https://github.com/containers/libpod/blob/master/libpod/image/search.go).

rhatdan:
I think it makes sense to add SEARCH/List to skopeo.  Any interactions with a container registry would make sense.

cmd/skopeo/registry_utils.go
```diff
@@ -0,0 +1,303 @@
...
+			reposWithTags[repo] = tags
+			lock.Unlock()
+		}(repo)
+	}
+	wg.Wait()
+
+	// Display repos with tags - format to be optimized later
+	// Consider using https://github.com/olekukonko/tablewriter
+	fmt.Println()
+	w := tabwriter.NewWriter(os.Stdout, 10, 0, 1, ' ', tabwriter.TabIndent|tabwriter.Debug)
```
mtrmac: I’d really start with JSON, like `skopeo inspect` does; it’s trivial to parse / search / re-format in consumers, and I don’t think we realistically expect the full output of this to be read by humans at all.

cmd/skopeo/list.go
```diff
@@ -0,0 +1,65 @@
...
+		global: global,
+		image:  imageOpts,
+	}
+	return cli.Command{
+		Name:  "list",
+		Usage: "Get list of repos of REGISTRY-NAME",
+		Description: fmt.Sprint(`
+		Get list of repos of REGISTRY-NAME
+
+		Supports Docker v2 registries only.
```
mtrmac: So should this subcommand be named `docker-list`? `registry-list`? `registry-catalog`?

cmd/skopeo/registry_utils.go
```diff
@@ -0,0 +1,303 @@
...
+	return nil
+}
+
+// Prints list of registry repos with all available tags
+func (r *Registry) getAllReposWithTags() error {
+	err := r.getAuthType()
+	if err != nil {
+		return err
+	}
+	r.getV2Catalog()
```
mtrmac: The full catalog may not be available at all, or it may return only a few entries. So, careful… I suppose it’s fine to expose this functionality if the registry does provide it, but we need to be be very careful in documenting this.

cmd/skopeo/registry_utils.go
```diff
@@ -0,0 +1,303 @@
...
+	// Consider using https://github.com/olekukonko/tablewriter
+	fmt.Println()
+	w := tabwriter.NewWriter(os.Stdout, 10, 0, 1, ' ', tabwriter.TabIndent|tabwriter.Debug)
+	fmt.Fprintln(w, "\t REPO \t TAGS \t")
+	fmt.Fprintln(w, "\t - - - - - - - - - - \t - - - - - - - - - - \t")
+	for _, repo := range r.repos {
+		tags := strings.Join(reposWithTags[repo], "  ")
+		fmt.Fprintf(w, "\t %s\t %v\t\n", repo, tags)
+	}
+	w.Flush()
```
mtrmac: All of this is expected to write to `stdout` of `listOptions.run`, and it idiomatically should happen in that function or at least file; helpers can return Go data structures to be formatted by the CLI front-end.

cmd/skopeo/registry_utils.go
```diff
@@ -0,0 +1,303 @@
...
+	wg.Add(len(r.repos))
+	reposWithTags := make(map[string][]string)
+
+	// Alphabetize repos
+	sort.Strings(r.repos)
+
+	for _, repo := range r.repos {
+		go func(repo string) {
+			defer wg.Done()
+			tags, _ := r.getRepoTags(repo)
```
mtrmac: It might be worth considering making the tag lookup optional; it’s likely to take most of the time for large registries, and the caller might not care.  But that really depends on what you need this command for, and is not a blocker.

cmd/skopeo/list.go
danielorf: While this is for a Docker registry spec, it makes the most sense to stick with the vendor agnostic naming scheme.   Probably `registry-list` but `registry-catalog` would work equally well.

cmd/skopeo/registry_utils.go
danielorf: I like this idea.  Tag collection consumes the vast majority of time and http calls.

cmd/skopeo/registry_utils.go
danielorf: This definitely makes sense if moving to containers/image.

cmd/skopeo/list.go
mtrmac: - I’m not at all convinced this is even a bit easy to generalize:
  - The idea of a complete catalog per server clearly does not scale and it’s a wonder it’s been created at all, ever.
  - Already several transports in c/image don’t have repos/tags, or have them with very different syntax
  - Let alone the concept of a “registry” is something that just does not apply to local files or c/storage or BitTorrent-based transports
  So far c/image has not even tried to make sets of images / repositories / servers enumerable / generalizable beyond the docker/distribution protocol.
- Fine, supposing that this _could_ be generalizable: in that case the syntax of this command must be general and admit the various containers/image transports, or something like that; it can’t just take over the full set of strings for the first argument and interpret them as docker/distribution registries. So we would have `skopeo list --docker REGISTRY_NAME` or, maybe `skopeo list docker://REGISTRY_NAME` (but that runs into **huge difficulties** with consistency, because `docker://example.com` already has a meaning define to be equivalent to `docker://docker.io/library/example.com`, so that’s best avoided).

rhatdan:
@danielorf Are you still working on this one?

danielorf:
Yes I am still working on the registry listing feature, but I'm going to approach this from containers/image as @mtrmac suggests.  So It makes sense to close this PR.

PR #684, opened 2019-06-21
copy: add a CLI flag for OCIAcceptUncompressedLayers

tych0:
There are cases where we want to pass this flag to the actual copy engine,
so let's add a CLI flag for it.

Signed-off-by: Tycho Andersen <tycho@tycho.ws>


mtrmac:
Thanks! LGTM.

rhatdan:
Need man page changes as well.

rhatdan:
Bash completions also.

PR #686, opened 2019-06-27
Makefile: set GO111MODULE=off

vrothberg:
Turn of go modules to avoid breaking build environments to accidentally
try pulling the dependencies instead of using the ./vendor directory.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 .PHONY: all binary build-container docs docs-in-container build-local clean install install-binary install-completions shell test-integration .install.vndr vendor
 
 export GO15VENDOREXPERIMENT=1
+export GO111MODULE=off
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,6 +1,7 @@
 .PHONY: all binary build-container docs docs-in-container build-local clean install install-binary install-completions shell test-integration .install.vndr vendor
 
 export GO15VENDOREXPERIMENT=1
```
saschagrunert: I guess this variable can be removed, right? :)

saschagrunert:
One comment :point_up: , the rest looks good.

rhatdan:
LGTM
I will allow @vrothberg to address your comment in a different PR if necessary.

PR #690, opened 2019-07-18
enforce blocking of registries

vrothberg:
Vendor in the latest c/image to enforce blocking of registries when
creating a c/image/docker.dockerClient.  Add integration tests to
avoid regressions.

---

~~Do not merge: that's a pet PR for https://github.com/containers/image/pull/658.~~

### DIFF
--- a/integration/blocked_test.go
+++ b/integration/blocked_test.go
@@ -1,0 +1,34 @@
+package main
+
+import (
+	"github.com/go-check/check"
+)
+
+const blockedRegistriesConf = "./fixtures/blocked-registries.conf"
+const blockedErrorRegex = `.*registry registry-blocked.com is blocked in .*`
+
+func (s *SkopeoSuite) TestCopyBlockedSource(c *check.C) {
+	assertSkopeoFails(c, blockedErrorRegex,
+		"--registries-conf", blockedRegistriesConf, "copy",
+		"docker://registry-blocked.com/image:test",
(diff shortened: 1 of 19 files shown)


integration/blocked_test.go
```diff
@@ -0,0 +1,29 @@
+package main
+
+import (
+	"github.com/go-check/check"
+)
+
+const blockedRegistriesConf="./fixtures/blocked-registries.conf"
+
+func (s *SkopeoSuite) TestCopyBlockedSource(c *check.C) {
+	assertSkopeoFails(c, "", "--registries-conf="+blockedRegistriesConf, "copy",
```
mtrmac: I would expect a non-empty regex here, to distinguish between uses of a blocked registry and accesses to nonexistent `registry-unblocked.com` failing.

Or, maybe more realistic (although slower), actually use a real registry with accessible images.

(Absolutely non-blocking: Can’t this use `"--registries-conf", blockedRegistriesConf` instead of the concatenation?)

vrothberg:
@mtrmac PTAL

mtrmac:
ACK, thanks.

vrothberg:
Oups... please don't merge. We need to merge c/image first.

vrothberg:
@mtrmac was faster :)

vrothberg:
Ready to merge now.

mtrmac:
LGTM, thanks.

PR #692, opened 2019-07-28
systemtest: don't copy the same image twice

chuanchang:
[root@dhcp-14-124 skopeo]# bats -f "copy: don't copy the same image twice" systemtest/020-copy.bats 
 ✓ copy: don't copy the same image twice

1 test, 0 failures

### DIFF
--- a/systemtest/020-copy.bats
+++ b/systemtest/020-copy.bats
@@ -70,6 +70,17 @@ function setup() {
                docker://localhost:5000/foo
 }
 
+# Don't copy the image again if the image already exists in the destination registry
+@test "copy: don't copy the same image twice" {
+    local output
+    local remote_image=docker://busybox:latest
+    local localimg=docker://localhost:5000/busybox:unsigned
+
+    run_skopeo copy --dest-tls-verify=false $remote_image $localimg
+    output=$(run_skopeo --debug copy --dest-tls-verify=false $remote_image $localimg)
+    echo $output | grep "Using blob info cache"
+}
+
 teardown() {
     podman rm -f reg
 


TomSweeneyRedHat:
LGTM

systemtest/020-copy.bats
```diff
@@ -70,6 +70,17 @@ function setup() {
...
 
+# Don't copy the image again if the image already exists in the destination registry
+@test "copy: don't copy the same image twice" {
+    local output
+    local remote_image=docker://busybox:latest
+    local localimg=docker://localhost:5000/busybox:unsigned
+
+    run_skopeo copy --dest-tls-verify=false $remote_image $localimg
+    output=$(run_skopeo --debug copy --dest-tls-verify=false $remote_image $localimg)
+    echo $output | grep "Using blob info cache"
```
mtrmac: Every skopeo run will say `Using blob info cache …` unless the path is not accessible (e.g. the first copy above would say the same thing!).  This does not in any way check that the second command does not copy data.

I guess this will need to watch the per-blob progress reports (and check that all relevant blobs, not just one of them, are properly skipped.)

chuanchang: @mtrmac, you're right, I haven't seen obvious difference between first time copy and second time copy in debug info via diff command, perhaps, skopeo should write some useful information into debug log.

mtrmac: > perhaps, skopeo should write some useful information into debug log.

WRT copying blobs vs. noticing that they already exist? Um… skopeo does write quite different debug data in those cases.

rhatdan:
@chuanchang You still interested in moving this one forward?

rhatdan:
Since there was never any response, I am closing.  Reopen if there will be further work.

PR #680, opened 2019-06-17
delete: support OCI images

vrothberg:
...implicitely by updating containers/image.

Fixes: #667
Signed-off-by: Valentin Rothberg <rothberg@redhat.com>

Issue #667, opened 2019-06-07
Unable to delete from local docker registry


deefin:
I have reviewed similar tickets in this repo but unfortunately have not been able to resolve my issue. It may be PBKC but i would like to confirm... :)

I am currently unable to delete a repo from my local registry. 

**inspect**

```
$ skopeo inspect docker://registry.local:5000/el7-base       
{
    "Name": "registry.local:5000/el7-base",
    "Digest": "sha256:2c9af36296b684757875b9a8e564861d0e4780caa4385ee64ab5a503a1e68c9f",
    "RepoTags": [
        "latest"
    ],
    "Created": "2019-06-06T02:04:12.567754296Z",
    "DockerVersion": "",
    "Labels": {
        "name": "el7-base"
    },
    "Architecture": "amd64",
    "Os": "linux",
    "Layers": [
        "sha256:677db987b43893305542e8385b60c5aae2a7f0a0ee86fd09f44e8d2082c1bce4"
    ]
}
```

**delete**

```
$ skopeo --debug delete --tls-verify=false docker://registry.local:5000/el7-base       
DEBU[0000] Using registries.d directory /etc/containers/registries.d for sigstore configuration 
DEBU[0000]  Using "default-docker" configuration        
DEBU[0000]   Using file:///var/lib/atomic/sigstore      
DEBU[0000] Looking for TLS certificates and private keys in /etc/docker/certs.d/registry.local:5000 
DEBU[0000] GET https://registry.local:5000/v2/          
DEBU[0000] Ping https://registry.local:5000/v2/ err Get https://registry.local:5000/v2/: http: server gave HTTP response to HTTPS client (&url.Error{Op:"Get", URL:"https://registry.local:5000/v2/", Err:(*errors.errorString)(0xc000450890)}) 
DEBU[0000] GET http://registry.local:5000/v2/           
DEBU[0000] Ping http://registry.local:5000/v2/ status 200 
DEBU[0000] GET http://registry.local:5000/v2/el7-base/manifests/latest 
FATA[0000] Unable to delete registry.local:5000/el7-base:latest. Image may not exist or is not stored with a v2 Schema in a v2 registry
```

**skopeo info**

```
$ skopeo --version
skopeo version 0.1.36-dev
```

* Fedora 30, installed with `dnf`

rhatdan:
The registry needs to be configured to support remote delete.
@jwhonce Do you remember how to configure this?

The man page is supposed to say>

[...]

PR #694, opened 2019-07-28
systemtest: support deleting image from registry

chuanchang:
[root@dhcp-14-124 skopeo]# bats systemtest/060-delete.bats 
 ✓ delete: remove image from registry

1 test, 0 failures

### DIFF
--- a/systemtest/060-delete.bats
+++ b/systemtest/060-delete.bats
@@ -1,0 +1,37 @@
+#!/usr/bin/env bats
+#
+# Copy tests
+#
+
+load helpers
+
+function setup() {
+    standard_setup
+
+    start_registry --enable-delete=true reg
+}
+
+# delete image from registry
+@test "delete: remove image from registry" {
+    local remote_image=docker://busybox:latest
+    local localimg=docker://localhost:5000/busybox:unsigned
+    local output=
+
+    run_skopeo copy --dest-tls-verify=false $remote_image $localimg
+    output=$(run_skopeo inspect --tls-verify=false --raw $localimg)
+    echo $output | grep "vnd.docker.distribution.manifest.v2+json"
+
+    run_skopeo delete --tls-verify=false $localimg
+
+    # make sure image is removed from registry
+    expected_rc=1
+    run_skopeo $expected_rc inspect --tls-verify=false $localimg
+}
+
+teardown() {
+    podman rm -f reg
+
+    standard_teardown
+}
+
+# vim: filetype=sh
--- a/systemtest/helpers.bash
+++ b/systemtest/helpers.bash
@@ -231,6 +231,7 @@ function expect_line_count() {
 #       --testuser=XXX      Require authentication; this is the username
 #       --testpassword=XXX  ...and the password (these two go together)
 #       --with-cert         Create a cert for running with TLS (not working)
+#       --enable-delete     Set allowing registry deletions (default: false)
 #
 #   NAME is the container name to assign.
 #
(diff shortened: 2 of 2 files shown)


TomSweeneyRedHat:
LGTM

systemtest/060-delete.bats
```diff
@@ -0,0 +1,34 @@
...
+# Copy tests
+#
+
+load helpers
+
+function setup() {
+    standard_setup
+
+    enable_delete=true
+    start_registry --enable-delete=$enable_delete reg
```
mtrmac: (Non-blocking nit: I don’t think the `enable_delete` variable here helps anything. `--enable-delete=true` would be shorter.)

systemtest/060-delete.bats
```diff
@@ -0,0 +1,34 @@
...
+
+# delete image from registry
+@test "delete: remove image from registry" {
+    local remote_image=docker://busybox:latest
+    local localimg=docker://localhost:5000/busybox:unsigned
+    local output=
+
+    run_skopeo copy --dest-tls-verify=false $remote_image $localimg
+    output=$(run_skopeo inspect --tls-verify=false --raw $localimg)
+    echo $output | grep "vnd.docker.distribution.manifest.v2+json"
```
mtrmac: What purpose does this inspect+grep serve?

systemtest/060-delete.bats
```diff
@@ -0,0 +1,34 @@
...
+
+    run_skopeo delete --tls-verify=false $localimg
```
mtrmac: Non-blocking: Might be worth also checking that the image is really no longer accessible after deleting it.

systemtest/helpers.bash
```diff
@@ -263,7 +266,12 @@ start_registry() {
     AUTHDIR=$TESTDIR/auth
     mkdir -p $AUTHDIR
 
-    local -a reg_args=(-v $AUTHDIR:/auth:Z -p $port:5000)
+    if [[ "$enable_delete" == "true" ]]; then
+        local -a reg_args=(-v $AUTHDIR:/auth:Z -p $port:5000 \
+	    -e REGISTRY_STORAGE_DELETE_ENABLED=true)
+    else
+        local -a reg_args=(-v $AUTHDIR:/auth:Z -p $port:5000)
```
mtrmac: The code below uses `reg_args+=` in conditionals; using it for the `enable_delete` case as well would avoid duplication of the `-v … -p …` parameters.

systemtest/060-delete.bats
chuanchang: sure, I will delete useless variable, thanks.

systemtest/060-delete.bats
chuanchang: it's a known issue #667 , so I gave a extra check in here

systemtest/060-delete.bats
chuanchang: Yes, skopeo list command is a PRE feature now, I will use skopeo inspect command instead

systemtest/helpers.bash
chuanchang: sure

systemtest/060-delete.bats
mtrmac: 1. But #667 was (supposedly?) fixed in #680, so this check should not be necessary any more.
2. Even if we did need to have this safety check here, it would need an explicit comment explaining the purpose.

rhatdan:
LGTM

PR #695, opened 2019-08-01
go build: use `-mod=vendor` for go >= 1.11.x

vrothberg:
Go 1.13.x isn't sensitive to the `GO111MODULE` environment variable
causing `make binary-local` to not use the vendored sources in
`./vendor`.  Force builds of module-supporting go versions to use the
vendored sources by setting `-mod=vendor`.

Verified in a `fedora:rawhide` container.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,5 @@
 .PHONY: all binary build-container docs docs-in-container build-local clean install install-binary install-completions shell test-integration .install.vndr vendor
 
-export GO111MODULE=off
-
 ifeq ($(shell uname),Darwin)
 PREFIX ?= ${DESTDIR}/usr/local
 DARWIN_BUILD_TAG=containers_image_ostree_stub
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

vrothberg:
@rhatdan, I'll open similar PRs for our other projects.

vrothberg:
@mtrmac PTAL

rhatdan:
I am going to just merge so we can fix the rawhide build.

Makefile
```diff
@@ -26,6 +24,12 @@ GO ?= go
 CONTAINER_RUNTIME := $(shell command -v podman 2> /dev/null || echo docker)
 GOMD2MAN ?= $(shell command -v go-md2man || echo '$(GOBIN)/go-md2man')
 
+GO_BUILD=$(GO) build
+# Go module support: set `-mod=vendor` to use the vendored sources
+ifeq ($(shell go help mod >/dev/null 2>&1 && echo true), true)
+  GO_BUILD=GO111MODULE=on $(GO) build -mod=vendor
```
saschagrunert: @vrothberg What is the reason to still keep `GO111MODULE=on`?

vrothberg: The Makefile still turns it off to avoid any potential side-effect of other tools `go get`-ing around, so we need to explicitly turn it on.

saschagrunert: I see, thanks for the clarification. Some dot releases of go1.11 will not work with that but I think we can live with that (the latest one should work definitely). :+1:

PR #697, opened 2019-08-02
Revert "Travis: use go 1.12.x"

vrothberg:
This reverts commit d6270f4691e6f83179ebefe445a199ee756fe159.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,6 @@ matrix:
        - docker
   -  os: osx
 
-go:
-        - 1.12.x
-
 notifications:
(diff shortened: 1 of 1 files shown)


vrothberg:
Not that commit d6270f4 had any impact, but I still prefer to clean it up.

mtrmac:
(Guessing that this is because the CI builds run in container, so what Travis install does not matter — in the future, it would be nice to always have documented a “why is this needed” and “how it works”, even for one-liners. I appreciate that sometimes the time to do that may be lacking.)

PR #700, opened 2019-08-06
ostree: use the correct build tags

marcov:
Starting from 9b902d0, the ostree transport is disabled by default,
and ostree is enabled with the tag containers_image_ostree.

/cc @vrothberg

### DIFF
--- a/hack/ostree_tag.sh
+++ b/hack/ostree_tag.sh
@@ -1,6 +1,4 @@
 #!/bin/bash
 if pkg-config ostree-1 2> /dev/null ; then
-	echo ostree
-else
-	echo containers_image_ostree_stub
+	echo containers_image_ostree
 fi


rhatdan:
LGTM

hack/ostree_tag.sh
```diff
@@ -1,6 +1,4 @@
 #!/bin/bash
 if pkg-config ostree-1 2> /dev/null ; then
-	echo ostree
```
mtrmac: With this, `containers/storage/pkg/ostree/` will now be disabled, affecting the `vfs` and `overlay` drivers.
- Is that intentional? I can’t see any explicit reference to this in the discussion.
- Independently, is it desirable? (@nalind ?)

marcov:
I think that's an error, I mistakenly grepped for ` "^// *\+build.+ostree.+"` instead of ` "^// *\+build.+ostree.*"`. Should both `ostree` and `containers_image_ostree` be used?

hack/ostree_tag.sh
nalind: The `github.com/containers/storage` `ostree` package also links with the native libraries, so if they're not present, and we don't want the build to fail because of that, the build tag would need to be removed.

For the case where detecting that the libraries are present means we enable the feature in the image library, I'd expect us to want to define the `ostree` build tag as well.

PR #703, opened 2019-08-07
ostree: use both image and & storage buildtags

marcov:
The PR #700 replaced ostree buildtag with containers_image_ostree.
    However specifying the ostree buildtag is needed by containers/storage
    vfs driver.
    
/cc @mtrmac @vrothberg

### DIFF
--- a/hack/ostree_tag.sh
+++ b/hack/ostree_tag.sh
@@ -1,4 +1,11 @@
 #!/bin/bash
-if pkg-config ostree-1 2> /dev/null ; then
-	echo containers_image_ostree
+
+if test $(${GO:-go} env GOOS) != "linux" ; then
+	exit 0
+fi
+
+if pkg-config ostree-1 &> /dev/null ; then
+	# ostree: used by containers/storage
+	# containers_image_ostree: used by containers/image
+	echo "ostree containers_image_ostree"
 fi
--- a/cmd/skopeo/main.go
+++ b/cmd/skopeo/main.go
@@ -22,7 +22,7 @@ type globalOptions struct {
 	tlsVerify          optionalBool  // Require HTTPS and verify certificates (for docker: and docker-daemon:)
 	policyPath         string        // Path to a signature verification policy file
 	insecurePolicy     bool          // Use an "allow everything" signature verification policy
-	registriesDirPath  string        // Path to a "registries.d" registry configuratio directory
+	registriesDirPath  string        // Path to a "registries.d" registry configuration directory
 	overrideArch       string        // Architecture to use for choosing images, instead of the runtime one
 	overrideOS         string        // OS to use for choosing images, instead of the runtime one
 	commandTimeout     time.Duration // Timeout for the command execution


rhatdan:
LGTM

hack/ostree_tag.sh
```diff
@@ -1,4 +1,17 @@
...
+	exit 0
+fi
+cc -E - $(pkg-config --cflags ostree-1) > /dev/null 2> /dev/null << EOF
+#include <ostree-1/ostree.h>
+EOF
+if test $? -eq 0 ; then
+	tags+=(ostree)
+fi
+
+echo "${tags[@]}"
```
mtrmac: (This test seems to come from c/storage.)

Is there a reason for the `ostree` and `containers_image_ostree` logic to differ? Both seem to have the same dependencies (CGo builds using `libostree-1`), so it seems surprising to make the decision logic different.

The new code is more comprehensive, so I have no objections to using that one, but having two different paths in the check for an unknown reason just seems surprising.

hack/ostree_tag.sh
```diff
@@ -1,4 +1,17 @@
...
 if pkg-config ostree-1 2> /dev/null ; then
-	echo containers_image_ostree
+	tags+=(containers_image_ostree)
 fi
+
+if test $(${GO:-go} env GOOS) != "linux" ; then
+	exit 0
+fi
+cc -E - $(pkg-config --cflags ostree-1) > /dev/null 2> /dev/null << EOF
+#include <ostree-1/ostree.h>
```
mtrmac: This does not seem actually representative; `pkg-config --cflags ostree-1` returns something like
```
-I/usr/include/ostree-1 -I/usr/include/gio-unix-2.0/ -I/usr/include/glib-2.0 -I/usr/lib64/glib-2.0/include -pthread 
```
and there is an `/usr/include/ostree-1/ostree.h` file. So, on a Fedora Linux, this check works whether or not `pkg-config` recognizes the package, and it _only_ works if `libostree` has headers in `/usr/include`, defeating the point of `pkg-config`.

Note that both CGo users use `#include <ostree.h>`.

hack/ostree_tag.sh
marcov: Actually I just copied here the ostree_tag.sh script from c/storage, as you realized.
Honestly I did not think about the differences, thank you for pointing it out.

(On openSUSE) I see that `pkg-config` succeeds if and only if the `libostree-devel` package is installed (the one including C headers). Because of this, checking the result of `pkg-config ostree-1` seems to be more strict than `cc -E`, that's only checking the presence of the headers files, as you also pointed out.

So, in the end, a single check on `pkg-config ostree-1` is adequate for both tags.

I'm updating the PR, thanks!

rhatdan:
@mtrmac Ok to merge?

mtrmac:
Thanks, this is clearly an improvement, merging.

Still, notes to self:
- `DARWIN_BUILD_TAG` in `Makefile` uses a no-longer-valid tag
- The CI does install ostree but apparently it is not detected (neither before nor after this PR)?!

I’ll hopefully look into those momentarily.

PR #707, opened 2019-08-26
update OCI image-spec to 775207bd45b6cb8153ce218cc59351799217451f

vrothberg:
This mainly pulls in the latest support for zstd-compressed layers and
eases testing of containers/image.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,7 +24,7 @@ require (
 	github.com/mattn/go-isatty v0.0.4 // indirect
 	github.com/mtrmac/gpgme v0.0.0-20170102180018-b2432428689c // indirect
 	github.com/opencontainers/go-digest v1.0.0-rc1
-	github.com/opencontainers/image-spec v0.0.0-20180918080442-7b1e489870ac
+	github.com/opencontainers/image-spec v1.0.0-rc6.0.20190823105129-775207bd45b6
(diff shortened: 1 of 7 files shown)


vrothberg:
@mtrmac, PTAL. This is really just a way to simplify testing https://github.com/containers/image/pull/563 - it avoids pushing first to c/image and then revendoring here.

TomSweeneyRedHat:
The changes LGTM, but I'm not understanding how they ease testing?

vrothberg:
> The changes LGTM, but I'm not understanding how they ease testing?

The CI in containers/image `git clone`s Skopeo master to run the PR against the Skopeo tests. But it only "injects" containers/image in `./vendor`. If the PR of containers/image requires to change other dependencies, then the CI run fails because we cannot compile Skopeo.

That's the case in https://github.com/containers/image/pull/563. Adding the dependency here will make https://github.com/containers/image/pull/563 compile again. Usually we have a separate PR open at Skopeo to let the c/image PR run through the CI but in this case I figured we can just do the vendor PR in advance - it'll spare some cycles/energy as we I don't have to push twice but only once.

TomSweeneyRedHat:
Verstanden, danke für die Information @vrothberg

vrothberg:
:rofl: gern geschehen, lieber Tom!

mtrmac:
Sure.

PR #708, opened 2019-08-28
inspect --raw: skip the NewImage() step

nalind:
Skip the `NewImage()` step if we're just inspecting the raw manifest, so that if the image tag or digest being inspected resolves to a manifest list, the local arch/OS combination don't need to be found in it to avoid an error.

### DIFF
--- a/cmd/skopeo/inspect.go
+++ b/cmd/skopeo/inspect.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/containers/image/docker"
+	"github.com/containers/image/image"
 	"github.com/containers/image/manifest"
 	"github.com/containers/image/transports"
 	"github.com/opencontainers/go-digest"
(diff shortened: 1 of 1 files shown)


cmd/skopeo/inspect.go
```diff
@@ -86,21 +90,37 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
...
 		}
 	}()
 
-	rawManifest, _, err := img.Manifest(ctx)
+	rawManifest, _, err := src.GetManifest(ctx, nil)
 	if err != nil {
 		return err
 	}
+
+	if opts.config || !opts.raw {
```
mtrmac: Please simplify this by moving the `opts.raw && !opts.config` code path here (and exiting early in that case).

Then `image.FromSource` can happen on the main path, and nothing below will need to worry about `img` possibly not being set.

cmd/skopeo/inspect.go
```diff
@@ -86,21 +90,37 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
...
+		img, err = image.FromSource(ctx, sys, src)
+		if err != nil {
+			return err
```
mtrmac: (This is all top-level CLI; please wrap errors in user-relevant explanations about the context, like the code below does.)

cmd/skopeo/inspect.go
```diff
@@ -86,21 +90,37 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
...
+	if err != nil {
+		return err
+	}
+
+	src, err := parseImageSource(ctx, opts.image, imageName)
 	if err != nil {
 		return err
 	}
 
+	closer := io.Closer(src)
```
mtrmac: At least, use `imgCloser` or something like that, please; to make it clear what it is for.

Or, maybe (weakly preferred?), use `image.FromUnparsedImage(…, UnparsedInstance(…))` instead of `image.FromSource`. `image.FromSource` more or less exists only to preserve the older design where `types.Image` has a `Close` method (so that callers of `types.ImageReference.NewImage` don’t have to deal with closing an `ImageSource`); like other places that need both an `ImageSource` and `Image`, that design no longer results in simpler code here. (An explicit `UnparsedInstance(src, nil)` would also be a tiny bit more explicit about what is happening WRT multi-arch.)

mtrmac:
Thanks! LGTM pending tests.

PR #710, opened 2019-08-28
Add a couple of tests for copying to/from containers-storage

nalind:
Add a couple of tests to verify that we can copy an image into and then back out of containers-storage.  The contents of an image that has been copied out of containers-storage need a bit of tweaking to compensate for containers-storage's habit of returning uncompressed versions of the
layer blobs that were originally written to it, in order to be comparable to the image as it was when it was pulled from a registry.


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

nalind:
Dropping in favor of #741.

PR #712, opened 2019-08-30
Makefile: use go proxy

vrothberg:
Use GOPROXY=https://proxy.golang.org to speed up fetching dependencies.
Setting it makes `make vendor` six times faster in my local env.

For details please refer to https://proxy.golang.org/.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 .PHONY: all binary build-container docs docs-in-container build-local clean install install-binary install-completions shell test-integration .install.vndr vendor
 
+export GOPROXY=https://proxy.golang.org
+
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

mtrmac:
Do I understand correctly that this will become the default in Go 1.13, and we are merely opting in early? If so, LGTM.

vrothberg:
> Do I understand correctly that this will become the default in Go 1.13, and we are merely opting in early? If so, LGTM.

That's correct. Go < 1.13 will follow us a while (at least in CI), so I found it nice to speed up things so cheaply.

Merging as we have 2 x LGTM

PR #713, opened 2019-09-02
Shellcheck fixes

vrothberg:
Fixes for https://bugzilla.redhat.com/show_bug.cgi?id=1701373.

Ran locally with `shellcheck -a`.

Tested the updated script locally and it continues working as before.

@rhatdan @mtrmac PTAL

### DIFF
--- a/completions/bash/skopeo
+++ b/completions/bash/skopeo
@@ -203,7 +203,6 @@ _cli_bash_autocomplete() {
      local completions_func=_skopeo_${command}
      declare -F $completions_func >/dev/null && $completions_func
 
-     eval "$previous_extglob_setting"
      return 0
 }
 
--- a/completions/bash/skopeo
+++ b/completions/bash/skopeo
@@ -174,7 +174,7 @@ _skopeo_skopeo() {
 }
 
 _cli_bash_autocomplete() {
-     local cur opts base
+     local cur opts
 
      COMPREPLY=()
      cur="${COMP_WORDS[COMP_CWORD]}"
--- a/completions/bash/skopeo
+++ b/completions/bash/skopeo
@@ -159,10 +159,9 @@ _skopeo_skopeo() {
        --version -v
        --help -h
      "
-     commands=$( "${COMP_WORDS[@]:0:$COMP_CWORD}" --generate-bash-completion )
 
      case "$prev" in
-         $main_options_with_args_glob )
+         --policy|--registries.d|--override-arch|--override-os|--command-timeout)
              return
              ;;
      esac
@@ -172,6 +171,7 @@ _skopeo_skopeo() {
              COMPREPLY=( $( compgen -W "$boolean_options $options_with_args" -- "$cur" ) )
              ;;
          *)
+             commands=$( "${COMP_WORDS[@]:0:$COMP_CWORD}" --generate-bash-completion )
              COMPREPLY=( $( compgen -W "${commands[*]} help" -- "$cur" ) )
              ;;
      esac
@@ -189,12 +189,9 @@ _cli_bash_autocomplete() {
 
      local command=${PROG} cpos=0
      local counter=1
-     counter=1
      while [ $counter -lt "$cword" ]; do
          case "${words[$counter]}" in
-             -*)
-                ;;
-             *)
+             skopeo|copy|inspect|delete|manifest-digest|standalone-sign|standalone-verify|help|h)
                  command="${words[$counter]//-/_}"
                  cpos=$counter
                  (( cpos++ ))
--- a/completions/bash/skopeo
+++ b/completions/bash/skopeo
@@ -10,7 +10,7 @@ _complete_() {
     local option_with_args
     for option_with_args in $options_with_args $transports
     do
-        if [ "$option_with_args" == "$prev" -o "$option_with_args" == "$cur" ]
+        if [ "$option_with_args" == "$prev" ] || [ "$option_with_args" == "$cur" ]
         then
             return
         fi
--- a/completions/bash/skopeo
+++ b/completions/bash/skopeo
@@ -1,6 +1,6 @@
 #! /bin/bash
 
-: ${PROG:=$(basename ${BASH_SOURCE})}
+PROG="skopeo"
 
 _complete_() {
     local options_with_args=$1
--- a/completions/bash/skopeo
+++ b/completions/bash/skopeo
@@ -177,7 +177,7 @@ _skopeo_skopeo() {
 }
 
 _cli_bash_autocomplete() {
-     local cur opts
+     local cur
 
      COMPREPLY=()
      cur="${COMP_WORDS[COMP_CWORD]}"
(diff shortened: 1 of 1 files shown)


vrothberg:
```
Failed to download metadata for repo 'updates'
Error: Failed to download metadata for repo 'updates'
```
... will try again later

completions/bash/skopeo
```diff
@@ -155,13 +159,7 @@ _skopeo_skopeo() {
        --version -v
        --help -h
      "
-     commands=$( ${COMP_WORDS[@]:0:$COMP_CWORD} --generate-bash-completion )
-
-     case "$prev" in
-         $main_options_with_args_glob )
```
mtrmac: Sure, this is not making things _worse_, but it also doesn’t seem quite the right thing.

Consider `skopeo --policy `<completion>: After this change, the code will fall through the `*)` case below and try to complete a command name; it should either use the `-*)` case or exit here, which, in both cases, results in no completions.

I think this should somehow use the `$options_with_args` instead. (But then again, the use of `$options_in_args` in  `_complete_` does not work either, AFAICS, because it is not the right syntax for alternatives in a `case`. Randomly checking some examples in `/usr/share/bash-completion`, `_init_completion -s` looks like a possible option, but also quite a bit more work than locally fixing warnings.)

completions/bash/skopeo
```diff
@@ -1,6 +1,7 @@
 #! /bin/bash
+# shellcheck shell=bash disable=SC2207
 
-: ${PROG:=$(basename ${BASH_SOURCE})}
+PROG="skopeo"
```
mtrmac: It would be cleaner to just remove the variable at this point; we rely on `PROG` being `skopeo` for the `_skopeo_skopeo` function to be called. (Or, well, we could not do that, and handle that special case cleanly as a true special case.)

(Alternatively, to just fix the warning without changing the rest, `${BASH_SOURCE[0]}` might work.)

completions/bash/skopeo
```diff
@@ -1,6 +1,7 @@
 #! /bin/bash
+# shellcheck shell=bash disable=SC2207
```
mtrmac: I’m not quite happy about this global override… Do I understand correctly that this is all for the
```sh
COMPREPLY=( $( compgen -W …) )
```
instances?

(Note that the example of `compgen` in `(info bash)` changes `IFS` to parse the output correctly before running the command.)

If these are the only causes of the warning, couldn’t they all call a single helper function that either uses one of the supported replacements (to be explicitly and fully safe), or uses only a _local_ override (to continue to rely on the fact that the sets of options don’t contain any other elements that should/could be split)?

completions/bash/skopeo
vrothberg: I updated it now to check for the options. Just using the variable doesn't work without some glob magic, so I opted for maintaining the list manually. I also had to adapt the logic for selecting the function to avoid the script believing that there's a function for the value of `--policy $value` for instance.

completions/bash/skopeo
vrothberg: Done

completions/bash/skopeo
vrothberg: Yes, it's for `compgen` instances. I changed them to use `read -r` now.

vrothberg:
@rhatdan, we need similar fixes for the other projects as well. Especially the changes in https://github.com/containers/skopeo/pull/713/commits/b31f0da5c6327b6ab75a03cbecef7701fbae76bd are required to support `podman --root /path [TAB]` (note that no completions occur any more).

vrothberg:
> @rhatdan, we need similar fixes for the other projects as well. Especially the changes in [b31f0da](https://github.com/containers/skopeo/commit/b31f0da5c6327b6ab75a03cbecef7701fbae76bd) are required to support `podman --root /path [TAB]` (note that no completions occur any more).

FYI @baude @mheon @TomSweeneyRedHat

rhatdan:
LGTM
@mtrmac Can you merge?

completions/bash/skopeo
mtrmac: (Absolutely non-blocking: This works, but it’s also non-obvious enough that a helper might be easier to maintain (or at least comment on) than copy&pasting it over the place.)

completions/bash/skopeo
```diff
@@ -183,15 +185,12 @@ _cli_bash_autocomplete() {
...
      local counter=1
-     counter=1
-     while [ $counter -lt $cword ]; do
+     while [ $counter -lt "$cword" ]; do
          case "${words[$counter]}" in
-             -*)
-                ;;
-             *)
-                 command=$(echo "${words[$counter]}" | sed 's/-/_/g')
+             skopeo|copy|inspect|delete|manifest-digest|standalone-sign|standalone-verify|help|h)
```
mtrmac: Hum… this is not really correct, consider `skopeo --policy copy inspect …` and the like. I appreciate that in practice, it’s clearly better than the current code for both filenames and parameter values, but it’s likely to confuse future readers into thinking that this code is correct; maybe worth at least a comment or filing an issue.

Also, this is yet another place to maintain this list… could that do with some of the “glob magic” and based on a helper that calls `--generate-bash-completion`?

(Really the trick of `_skopeo_skopeo` is, in hindsight, a mistake.)

completions/bash/skopeo
```diff
@@ -155,26 +156,27 @@ _skopeo_skopeo() {
        --version -v
        --help -h
      "
-     commands=$( ${COMP_WORDS[@]:0:$COMP_CWORD} --generate-bash-completion )
 
      case "$prev" in
-         $main_options_with_args_glob )
+         --policy|--registries.d|--override-arch|--override-os|--command-timeout)
```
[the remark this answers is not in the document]
vrothberg: Added a comment.

completions/bash/skopeo
vrothberg: It's really nasty to get it correct since the entire logic is split. I don't want to rewrite the script but merely wanted to address the big shellcheck complaints and address the completion bug after --flags.

vrothberg:
> The more I look at this, the more incorrect logic I find…

Same here :see_no_evil: 

> Maybe it would be better, after all, to keep the scope of this PR to only shellcheck fixes, and not try to improve the logic on `--option param`? I.e. go back to your original idea of just giving up on `main_options_with_args_glob`.

I prefer the current version of the previous one. The logic can be tricked into reporting bogus but at least it's reporting something.

mtrmac:
LGTM, let’s get this in.

(For future reference, if anyone finds this PR to see what the commits were about:

`/usr/share/bash-completion/completions/pkg-config` could be a good starting point: `_init_completion -s` to handle `--option=param` and `--option param` uniformly, a simple `case` to add completions for individual options — and even an automated way to extract available options out of `--help`. It would probably need fair amount of tweaking (to handle the individual subcommands), but it seems much closer to being actually correct.)

Issue #276, opened 2016-12-16
Can not list Tags on Repo when there is no latest tag


plumcraft:
When i'm trying to list tags on specific repo, i got this error message :

- Erreur : manifest unknown: manifest unknown

Thanks

runcom:
@plumcraft could you try with `skopeo inspect yourimage:atagwhichexist`?

plumcraft:
@runcom : Thanks for your reply. 

The issue is i need to list tags on specific repo which does not have a tag named "latest", i have  "manifest unknown: manifest unknown". Do you have a filter where there is no latest tag ?
Next i will rely on that to clean the docker registry.

When i execute skopeo inspect on a specific tag, it's working.

Thanks

runcom:
I see, unfortunately skopeo does work on a specific tag (e.g. latest) when inspecting an image. The tag list it returns it's fetched when inspecting _a_ particular tag. The way it works, if you don't specify a tag it uses latest, if you don't have a latest tag, skopeo is unable show image information, thus it doesn't make sense to retrieve the tag list. 

I feel we should remove the tag list from inspect and have a dedicated command to show the tag list for a given repository.

plumcraft:
That make sense. It would be awesome having a dedicated command for that.

runcom:
> That make sense. It would be awesome having a dedicated command for that.

Great, time to contribute? :) 

I'd like to hear if this makes sense by @mtrmac also

mtrmac:
> I feel we should remove the tag list from inspect and have a dedicated command to show the tag list for a given repository.

I’m not sure about breaking existing users of `skopeo inspect`, but having a recommended command which is clearly Docker-specific and does not need the magic handling of `docker.Image` in `inspect.go` would be nice.

rhatdan:
@runcom @plumcraft @mtrmac @vrothberg What is the latest on this issue?  Is this still something we are interested in?

mtrmac:
At the API level, there is now `c/image/docker.GetRepositoryTags`, which does not require an existing tag.

Adding this as a separate command to `skopeo`, and ideally deprecating the list of tags in `skopeo inspect` (but that would break users) is still worth considering.

zhill:
This hasn't been updated in a while, but I'm interested in providing a command to list tags from a repository. It is implemented as a distinct top-level command, e.g. `skopeo list-tags docker://myregistry.com/myrepo/`.

[...]

PR #716, opened 2019-09-10
Adds "docker-list-tags" command to list tags

zhill:
Adds new command `docker-list-tags` to list the tags in a docker V2 registry's repository.

Discussion in #276. I expect some changes will be needed in the naming so opening this early to facilitate discussion.

If there is any additional testing or docs required I will gladly update.


cmd/skopeo/tags.go
```diff
@@ -0,0 +1,96 @@
...
+type tagListOutput struct {
+	DockerRepository string   `json:",omitempty"`
+	Tags             []string `json:",omitempty"`
+}
+
+type tagsOptions struct {
+	global *globalOptions
+	image  *imageOptions
+	raw    bool // Output the raw manifest instead of parsing information about the image
+	config bool // Output the raw config blob instead of parsing information about the image
```
mtrmac: These two members are unused AFAICS.

cmd/skopeo/tags.go
```diff
@@ -0,0 +1,96 @@
...
+	if len(args) != 1 {
+		return errors.New("Usage: skopeo docker-list-tags <registry host>:<port>/<repository name>")
+	}
+
+	// Add a leading // to make it valid docker reference, even though user doesn't specify the transport
+	repoRef := fmt.Sprintf("//%s", args[0])
+
+	repo, err := docker.ParseReference(repoRef)
+	if err != nil {
+		return err
```
mtrmac: Please add context to errors throughout,
```go
   return fmt.Errorf("What was the program trying to do that failed: %v", err)
```

cmd/skopeo/tags.go
```diff
@@ -0,0 +1,96 @@
...
+
+	sys, err := opts.image.newSystemContext()
+	if err != nil {
+		return err
+	}
+
+	outputData.Tags, err = docker.GetRepositoryTags(ctx, sys, repo)
+
+	if err != nil {
+		// some registries may decide to block the "list all tags" endpoint
```
mtrmac: If this workaround is needed in several places, it would be better to move the `GetRepositoryTags` call + the workaround to a common helper in `utils.go`, so that it can be maintained in one place.

cmd/skopeo/tags.go
```diff
@@ -0,0 +1,96 @@
...
+		// gracefully allow the inspect to continue in this case. Currently
+		// the IBM Bluemix container registry has this restriction.
+		if !strings.Contains(err.Error(), "401") {
+			return fmt.Errorf("Error determining repository tags: %v", err)
+		}
+		logrus.Warnf("Registry disallows tag list retrieval; skipping")
```
mtrmac: Shouldn’t this be a hard error for this command?

cmd/skopeo/tags.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	out, err := json.MarshalIndent(outputData, "", "    ")
```
mtrmac: (RFC: Not sure that JSON should be the only possible output format; a simple one-tag-per-line could be simpler to read and consume from shell scripts.

OTOH this works perfectly fine, and other formats can be added as options by whoever needs them.)

cmd/skopeo/tags_test.go
```diff
@@ -0,0 +1,45 @@
...
+		{},
+		{"invalid1", "invalid2"},
+	} {
+		out, err := runSkopeo(append([]string{"docker-list-tags"}, args...)...)
+		assertTestFailed(t, out, err, "Usage:")
+	}
+
+	// invalid repository name
+	out, err := runSkopeo("docker-list-tags", "/this/doesnt/exist")
+	assertTestFailed(t, out, err, "invalid reference format")
```
mtrmac: The above is great to have here; I’m not sure that the tests below, which actually verify run-time behavior against a registry, are a good fit for unit tests.

Running them in the integration test phase might be a better fit, both against `docker.io` or possibly against the test registry with precisely known contents (i.e. the tests could also include correctness of the output.)

docs/skopeo-docker-list-tags.1.md
```diff
@@ -0,0 +1,72 @@
...
+
+Return a list of tags from _repository-name_ in a registry
+
+  _repository-name_ name of repository to retrieve tag listing from
+
+  **--config** output configuration in OCI format, default is to format in JSON
+
+  _image-name_ name of image to retrieve configuration for
+
+  **--config** **--raw** output configuration in raw format
```
mtrmac: These options don’t exist.

(Note to self: I haven’t read the man page word by word yet.)

docs/skopeo.1.md
```diff
@@ -47,7 +47,12 @@ Most commands refer to container images, using a _transport_`:`_details_ format.
   **ostree:**_image_[**@**_/absolute/repo/path_]
   An image in local OSTree repository.  _/absolute/repo/path_ defaults to _/ostree/repo_.
 
-## OPTIONS
+## DOCKER REPOSITORY NAMES
```
mtrmac: Might be more appropriate to place this in the `skopeo-docker-list-tags` man page, at least until it becomes more generally applicable.

docs/skopeo.1.md
```diff
@@ -47,7 +47,12 @@ Most commands refer to container images, using a _transport_`:`_details_ format.
...
+Docker Registry v2 API-compatible image registries support references of this type to the image namespaces themselves (e.g. docker.io/library/centos)
+
+  **registryhost:port/repositoryname** (e.g. docker.io/centos or docker.io/library/centos or quay.io/coreos/etcd)
```
mtrmac: (The `:port` part might be missing, this would be nice to indicate here, and perhaps include an example like that.)

docs/skopeo-docker-list-tags.1.md
```diff
@@ -0,0 +1,72 @@
+% skopeo-docker-list-tags(1)
```
mtrmac: `completions/bash` should be eventually updated as well.

zhill:
> Thanks! This is pretty good.

Thanks for the super fast response. Will review and make changes as recommended.

cmd/skopeo/tags.go
zhill: Yes, leftovers of a previous implementation idea. Will remove

cmd/skopeo/tags.go
zhill: Ack. Will do.

cmd/skopeo/tags.go
zhill: Given this isn't part of another operation, any error response is a hard error and will cause failure, so I think this condition can be removed and no need to add it to utils.go in that case.

mtrmac: Makes sense.

cmd/skopeo/tags.go
zhill: Yes, will adjust as described in previous comment as well.

cmd/skopeo/tags_test.go
zhill: Yes agree. I'll remove from unit-test and explore the integration tests.

cmd/skopeo/tags.go
```diff
@@ -0,0 +1,96 @@
...
+	imageFlags, imageOpts := imageFlags(global, sharedOpts, "", "")
+
+	opts := tagsOptions{
+		global: global,
+		image:  imageOpts,
+	}
+
+	return cli.Command{
+		Name:  "docker-list-tags",
+		Usage: "List tags in a Docker v2 API registry and repository specified by the REPOSITORY-NAME",
```
rhatdan: Can we just call this a `container registry`?

cmd/skopeo/tags.go
```diff
@@ -0,0 +1,96 @@
...
+		Action:    commandAction(opts.run),
+	}
+}
+
+func (opts *tagsOptions) run(args []string, stdout io.Writer) (retErr error) {
+	ctx, cancel := opts.global.commandTimeoutContext()
+	defer cancel()
+
+	if len(args) != 1 {
+		return errors.New("Usage: skopeo docker-list-tags <registry host>:<port>/<repository name>")
```
rhatdan: PLEASE just call this list-tags, do not add `docker-`.  If this only works currently with `docker` API, then just document it.  Lets not make the skopeo CLI any more complicated then it already is.

cmd/skopeo/tags_test.go
```diff
@@ -0,0 +1,45 @@
...
+	for _, args := range [][]string{
+		{},
+		{"invalid1", "invalid2"},
+	} {
+		out, err := runSkopeo(append([]string{"docker-list-tags"}, args...)...)
+		assertTestFailed(t, out, err, "Usage:")
+	}
+
+	// invalid repository name
+	out, err := runSkopeo("docker-list-tags", "/this/doesnt/exist")
```
rhatdan: sed s/docker-//g

cmd/skopeo/tags_test.go
```diff
@@ -0,0 +1,45 @@
...
+	assertTestFailed(t, out, err, "invalid reference format")
+
+	// invalid repository name
+	out, err = runSkopeo("docker-list-tags", "docker.io/scratch")
```
rhatdan: ditto

rhatdan:
Need bash completions.

docs/skopeo.1.md
rhatdan: I really don't want to add the letters DOCKER any more then is necessary.
CONTAINER REPORITORY NAMES
works fine here.

zhill:
@rhatdan yes, as discussed in the issue, I don't really care about the name. The contention seems to be with others' opinions on naming scope. I'm happy to use 'list-tags', 'list-registry-tags', 'list-repository-tags', or whatever as long as its descriptive of what the command does.

How about 'list-repository-tags'? @mtrmac is that ok? I'd prefer not to make the change more than once, so if we can get agreement here first then I'll change it.

Thanks again. I very much appreciate the prompt reviews and suggestions.

mtrmac:
> PLEASE just call this list-tags, do not add `docker-`. If this only works currently with `docker` API, then just document it. Lets not make the skopeo CLI any more complicated then it already is.

Well, the thing is: Are you willing to bet that no new protocols and server kinds which could use a “list tags” operation will be added during the lifetime of Skopeo? Considering the long discussions about file-at-a-time sync, I’m really not.

And if there are going to be new protocols and server kinds, then the CLI will need to accommodate them. If we don’t add the protocol anywhere in the CLI now, we’ll end up requiring an extra option, or a non-trivial command name, in the future, for the new protocols/server kinds. If `skopeo list-tags` were created now for docker/distribution, any future protocol would have to use `skopeo list-tags --fedora` or `skopeo list-fedora-tags` in the future, no matter how prevalent the future protocol became = no matter how confusing it were to have the shorter command not apply to the prevalent protocol.

Being explicit about the protocol now allows us to symmetrically handle future protocols without expressing a preference (who knows, the next one may well turn out to be a vendor-named one as well).

Also, note that this command is pretty likely to be primarily used in scripts / automation. So, in a sense, the user-friendliness does not matter nearly as much as stable behavior. (So, I’m _absolutely_ unwilling to do, or design this to eventually require, heuristic guessing about the protocol in Skopeo — we can’t know that it will be possible to safely guess that with future unknown protocols anyway.)

(And, as far as user-friendliness goes, with completions, `list-*` are all exactly as easy to type.)

If Podman wants to provide easier-to-type code with more heuristics, it can of course do that; looking at how the semantics of the currently-existing heuristics in there turned out to be unclear and inconsistent, I wouldn’t recommend it but I can’t and won’t try to stop it.)

---

> How about 'list-repository-tags'

I’m not _happy_ about that name but I guess I could settle on that if @rhatdan can; it’s still very generic, but at least I can console myself with the thought that calling image servers “registries” does not make any semantic sense, so at least there’s _some_ chance that the new protocol won’t use this jargon and we won’t collide.

docs/skopeo.1.md
mtrmac: It doesn’t; `/var/lib/containers` is also a “container repository”, we need to explicitly tie this to the docker/distribution format.

(This is also one minor reason to move this paragraph into the subcommand-specific man page; if it is in that one page, we won’t need to have a separate section for the syntax, so we won’t need to choose a good name for the section.)

rhatdan:
I am not tighing the list-tags to only container-registries.  If other containers stores supported tags, then list-tags should show there as well.

I could see  skopeo list-tags listing tags for an images stored in container-storage and docker-daemon.  Perhaps.  

The question you raise is do I see some time in the future where list-tags would talk about something totally different on a a different type of image store, and my answer would be no.

zhill:
> Yes I agree, the container store needs to be specified.

As in: `list-tags docker://host.io:port/repo` and later `list-tags container-storage://path`?

I'm fixing up integration tests now, and will get the updates pushed here shortly

mtrmac:
> > Yes I agree, the container store needs to be specified.
> 
> As in: `list-tags docker://host.io:port/repo` and later `list-tags container-storage://path`?

`docker://host.io:port/repo` is defined to be the same as `docker://host.io:port/repo:latest`; and use of that syntax implies that other image references (like `dir:path`) can probably be accepted as well.

So, if the semantics of the parameter here is different, I think it should use a different syntax, maybe `list-docker tags` or `list-tags --docker`; of the two, the former is easier to use with shell completions.

rhatdan:
Why is the additional option needed?
Isn't this enough for the tool to interpret what the user wants?
```
skopeo list-tags  docker://host.io:port/repo
skopeo list-tags container-storage
```

cmd/skopeo/tags.go
zhill: I kept symmetry with the inspect output for now. Can add non-json output if that's a blocker

rhatdan:
@zhill @mtrmac @vrothberg What is the state of this PR?

zhill:
I have pushed updates and am ready for more feedback re:naming. I had let it sit for a while b/c my cycles to operate on feedback were limited. I'd like to get a clear direction on what to name the command and how params should work if not the way it is currently implemented. I can then rebase and get the PR ready to merge

rhatdan:
@giuseppe @mtrmac Do we need to add this as a topic for the F2F?

docs/skopeo-list-repository-tags.1.md
```diff
@@ -0,0 +1,64 @@
+% skopeo-list-repository-tags(1)
+
+## NAME
+skopeo\-list\-repository\-tags - Return a list of tags from _repository-name_ in a registry
+
+## SYNOPSIS
+**skopeo list-respository-tags** [**--config**] _repository-name_
```
mtrmac: `--config` does not exist.

completions/bash/skopeo
```diff
@@ -142,6 +142,21 @@ _skopeo_layers() {
...
 
+_skopeo_docker_list_tags() {
+     local options_with_args="
+     --authfile
+     --creds
+     --cert-dir
+     "
+
+     local boolean_options="
+     --config
```
mtrmac: This does not exist.

cmd/skopeo/tags_test.go
```diff
@@ -0,0 +1,36 @@
+package main
+
+import (
+	"github.com/containers/image/docker/reference"
+	"testing"
+)
+
+type tagListing struct {
```
mtrmac: This is not used anywhere AFAICS.

docs/skopeo.1.md
mtrmac: Still outstanding.

docs/skopeo.1.md
mtrmac: Please move this paragraph to `docs/skopeo-list-repository-tags.1.md`, or maybe just drop it.

mtrmac:
@rhatdan From my POV this is now mostly up to you to state goals and/or ask ask for a design, and if both, to make them consistent with each other.  E.g. https://github.com/containers/skopeo/pull/716#issuecomment-531382253 asks a question that I think answered in the immediately preceding comment already.

Or maybe just say that you are fine with `list-repository-tags $docker_reference` as it is now.

zhill:
Thanks, I'll get the updates done ASAP, but cycles are scarce until the end of the week. I appreciate your time reviewing and helping move this forward.

cmd/skopeo/tags.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/pkg/errors"
+	"github.com/urfave/cli"
+
+	"io"
+)
+
+// tagListOutput is the output format of (skopeo tags), primarily so that we can format it with a simple json.MarshalIndent.
+type tagListOutput struct {
+	DockerRepository string   `json:",omitempty"`
+	Tags             []string `json:",omitempty"`
```
mtrmac: Maybe drop the `omitempty` from annotation from both:
- `DockerRepository` is never going to be empty
- Empty `tags` could happen, but having it explicitly present would be more explicit (and possibly easier to parse to shell script hacks, though enabling such abuse is also somewhat a disadvantage).

cmd/skopeo/tags.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+type tagsOptions struct {
+	global *globalOptions
+	image  *imageOptions
+}
+
+func tagsCmd(global *globalOptions) cli.Command {
+	sharedFlags, sharedOpts := sharedImageFlags()
+	imageFlags, imageOpts := imageFlags(global, sharedOpts, "", "")
```
mtrmac: There is now a `dockerImageFlags()`, which drops some irrelevant options (which the documentation was already silentl about).

cmd/skopeo/tags.go
```diff
@@ -0,0 +1,90 @@
...
+		Action:    commandAction(opts.run),
+	}
+}
+
+func (opts *tagsOptions) run(args []string, stdout io.Writer) (retErr error) {
+	ctx, cancel := opts.global.commandTimeoutContext()
+	defer cancel()
+
+	if len(args) != 1 {
+		return errors.New("Usage: skopeo list-repository-tags <registry host>:<port>/<repository name>")
```
mtrmac: Maybe
```go
return errorShouldDisplayUsage{errors.New("Exactly one argument expected")}
````

File: cmd/skopeo/tags.go [hunk 21]
```diff
@@ -0,0 +1,90 @@
...
+
+	out, err := json.MarshalIndent(outputData, "", "    ")
+	if err != nil {
+		return err
+	}
+
+	_, err = fmt.Fprintf(stdout, "%s\n", string(out))
+
+	if err != nil {
+		return fmt.Errorf("Error marshalling json output from tag listing: %v", err)
```
mtrmac: I suppose this belongs to `MarshalIndent`, the `Fprintf` error should have a different description. (It wouldn’t at all be unprecedented to just silently ignore the error, so just returning `err` without bothering with extra description would still be fine — as long as the description is not _misleading_.)

integration/list_repository_tags_test.go
```diff
@@ -0,0 +1,21 @@
...
+	check.Suite(&TagListSuite{})
+}
+
+type TagListSuite struct {
+	registry *testRegistryV2
+}
+
+// Simple tag listing
+func (s *TagListSuite) TestListSimple(c *check.C) {
+	// Note: these hit dockerhub, replace with local registry
```
mtrmac: Drop this comment, I guess?

cmd/skopeo/tags.go
zhill: agree. will update

cmd/skopeo/tags.go
zhill: ack, will update here

cmd/skopeo/tags.go
zhill: ack

cmd/skopeo/tags_test.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+
+	// invalid repository name
+	out, err := runSkopeo("list-repository-tags", "/this/doesnt/exist")
+	assertTestFailed(t, out, err, "Cannot parse repository reference")
+}
+
+func TestParsing(t *testing.T) {
+	tag := "docker.io:1000/nginx:foobar"
+	r, er2 := reference.ParseNormalizedNamed(tag)
```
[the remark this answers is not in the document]
zhill: yes, the test needs a bit of work. Primarily around how to handle references that actually include a tag. In the context of a repository the tag is confusing but could just be ignored. Any opinion on if that should result in an error? Given that the command is 'list-repository-tags' if you pass a tag reference there is still a repository component that can be used, so I think silently ignoring the tag portion is viable and helps reduce the confusion of tagless docker ref implicitly being the 'latest' tag as both conditions are handled consistently.

zhill:
Thanks for the review (again). I'll get updates in today. I've been very cycle constrained lately but have time to focus on this and get it over the finish line. Thanks!

vrothberg:
> (@vrothberg WDYT?)

Thanks for the ping! I will have a look at this PR tomorrow morning with a fresh pair of eyes.

vrothberg:
@zhill, can you also rebase the PR so we can merge it more easily?

zhill:
Yes, working on that now. I'll rebase and squash it down to a single commit now that we're close (didn't want to lose history previously). Thanks again folks! I know its been a long PR due to my intermittent availability, the patience is appreciated.

cmd/skopeo/tags_test.go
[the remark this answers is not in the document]
mtrmac: I’m leaning towards refusing tagged/digested input, because it indicates some probability that the user is confused about what tags/digests are, or what format the input (passed to this subcommand) is in vs. is supposed to be in. But that’s a weak opinion, I can easily see the case for accepting input that we know how to handle and possibly simplifying shell scripts (so that they don’t do something incorrect in general, like dropping anything after the last `:`)

If the goal is to test / ensure stability of parsing the input, please
- split the `repoRef` + `docker.ParseReference` code into a separate helper function
- add full input vs. output tests for _that_ function
so that if the implementation of parsing changes, the tests continue to test the updated implementation.  This test, which verifies the behavior of `reference.ParseNormalizedNamed`, which is _not called in the actual implementation at all_, does not verify / ensure stability of what the command does.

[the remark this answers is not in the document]
mtrmac: > so that if the implementation of parsing changes, the tests continue to test the updated implementation

… and because that makes a cleanly-separated “unit” that can be cheaply unit-tested without side-effects or contacting any external registry.

cmd/skopeo/tags.go
```diff
@@ -0,0 +1,90 @@
...
+	opts := tagsOptions{
+		global: global,
+		image:  imageOpts,
+	}
+
+	return cli.Command{
+		Name:  "list-repository-tags",
+		Usage: "List tags in a Docker v2 API repository specified by the REPOSITORY-NAME",
+		Description: fmt.Sprintf(`
+	Return the list of tags in from the repository in the Docker v2 registry "REPOSITORY-NAME"
```
vrothberg: ...tags ~~in~~ from the ...

cmd/skopeo/tags.go
```diff
@@ -0,0 +1,90 @@
...
+	sharedFlags, sharedOpts := sharedImageFlags()
+	imageFlags, imageOpts := imageFlags(global, sharedOpts, "", "")
+
+	opts := tagsOptions{
+		global: global,
+		image:  imageOpts,
+	}
+
+	return cli.Command{
+		Name:  "list-repository-tags",
```
vrothberg: I have a preference to name the command `list-tags` and require the input to follow `transport:....` convention. Maybe we want to support something similar for other transports in the future. Embedding "repository" into the command name and making it docker-transport only would close that door in some way.

The code wouldn't need to change much as we can error out for other transports for now. But I am late to the game and may miss context.

@mtrmac WDYT?

cmd/skopeo/tags.go
```diff
@@ -0,0 +1,90 @@
...
+		return fmt.Errorf("Error listing repository tags: %v", err)
+	}
+
+	out, err := json.MarshalIndent(outputData, "", "    ")
+	if err != nil {
+		return err
+	}
+
+	_, err = fmt.Fprintf(stdout, "%s\n", string(out))
+
```
vrothberg: Can we delete the empty line?

cmd/skopeo/tags.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/containers/image/v5/docker"
+	"github.com/pkg/errors"
+	"github.com/urfave/cli"
+
+	"io"
+)
+
+// tagListOutput is the output format of (skopeo tags), primarily so that we can format it with a simple json.MarshalIndent.
+type tagListOutput struct {
+	DockerRepository string   `json:",omitempty"`
```
vrothberg: I prefer calling it just "Repository". The distribution work in the OCI is ongoing which will make the protocol truly independent. Same in the man page.

docs/skopeo-list-repository-tags.1.md
```diff
@@ -0,0 +1,64 @@
...
+        "31",
+        "32",
+        "branched",
+        "heisenbug",
+        "latest",
+        "modular",
+        "rawhide"
+    ]
+}
+
```
vrothberg: Drop empty line

docs/skopeo-list-repository-tags.1.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+# SEE ALSO
+skopeo(1), podman-login(1), docker-login(1)
+
+## AUTHORS
+
+Antonio Murdaca <runcom@redhat.com>, Miloslav Trmac <mitr@redhat.com>, Jhon Honce <jhonce@redhat.com>
```
vrothberg: Change author list to you :)

docs/skopeo.1.md
```diff
@@ -44,6 +44,11 @@ Most commands refer to container images, using a _transport_`:`_details_ format.
   **oci:**_path_**:**_tag_
   An image _tag_ in a directory compliant with "Open Container Image Layout Specification" at _path_.
 
+## CONTAINER REPOSITORY NAMES
+Docker Registry v2 API-compatible image registries support references of this type to the image namespaces themselves (e.g. docker.io/library/centos)
```
vrothberg: Trailing dot missing

cmd/skopeo/tags.go
zhill: done

zhill:
I've made the changes and rebased along with squashing to a single commit for easier review and merge.

cmd/skopeo/tags.go
mtrmac: My worry is that it’s not certain that things can be generalized for all transports.

It’s already _a bit_ conceptually dubious with `docker://`, where `docker://foo` means “the `docker://foo:latest` _image_”, not “the `docker://foo` _repository_”.

In general, the whole idea of “repository” and “tag” does generalize _at all_ well — compare https://github.com/containers/skopeo/pull/716#issuecomment-531382253 (besides syntax problems) suggesting that the whole local storage is a single “repository” and the hostname/[ns/…/]repo:tag value is a “tag”: it’s not at all an unreasonable position to take, but it makes the concept of “tag” meaningless and rules out the possibility of any transport-agnostic consumer of the output of this command.

(Compare https://github.com/containers/skopeo/pull/682#discussion_r295892434 and https://github.com/containers/skopeo/pull/524#issuecomment-469884710 dealing with registry names, all of this ends up with several `docker://` syntaxes with different meanings and rules.)

My default heuristic in such cases is to intentionally make the command extremely specific and limited; if we ever generalize it, we can easily add a general command, and maintain the specific one as a trivial shim. Going the other way from a prematurely-general interface to a specific one, or changing the prematurely-general interface to be general in a different way, requires better ability to forecast the future.

---

But, overall, sure:
- `list-tags docker://…` and `list-docker-tags …` are equivalent in the information that is being provided
- The “`docker://repo` means `:latest`” quibble is pedantic but does not make a difference  in practice (apart from our ability to cleanly reject tagged/digested input, but this PR does not reject it anyway)
- `list-repository-tags …` is already an specifics-losing compromise that I’m not at all strongly attached to

As long as the input is clearly namespaced and therefore unambiguous (i.e. not `skopeo list heuristically-guess-what-I-mean`), I think any of the three options works, and we’ve taken 4 months to decide and it’s not going to get any easier with 4 more months.

@vrothberg So, to settle this once for all, `skopeo list-tags docker://…`?

zhill:
Once the naming decision is made I'll update with a new commit for that change. Thanks @mtrmac and @vrothberg !

vrothberg:
> @vrothberg So, to settle this once for all, skopeo list-tags docker://…?

Ay! I am in favor of that. Thanks!

rhatdan:
I am also.

cmd/skopeo/list_tags.go
```diff
@@ -0,0 +1,142 @@
...
+// Customized version of the alltransports.ParseImageName and docker.ParseReference that does not place a default tag in the reference
+// Would really love to not have this, but needed to enforce tag-less and digest-less names
+func parseDockerRepositoryReference(refString string) (types.ImageReference, error) {
+	parts := strings.SplitN(refString, ":", 2)
+	if len(parts) != 2 {
+		return nil, errors.Errorf(`Invalid image name "%s", expected colon-separated transport:reference`, refString)
+	}
+
+	transport := transports.Get(parts[0])
+	if transport == nil || transport.Name() != docker.Transport.Name() {
```
mtrmac: `run` already did a check for the right transport; it shouldn’t be necessary to do that twice. Preferably, do the checks in this function only, so that they can be covered by the unit tests.

(`alltransports.ParseImageName`, with no manual string parsing here, would be esthetically preferable, but that would not allow rejecting tagged/digested input. I’m honestly undecided whether the extra check is worth it…)

Alternatively, just checking explicitly for the `docker.transport.Name()+"://"` prefix, and stripping exactly that prefix, would do just as many assumptions about the format, and be simpler.

cmd/skopeo/list_tags.go
```diff
@@ -0,0 +1,142 @@
...
+
+	if !reference.IsNameOnly(ref) {
+		return nil, errors.New(`No tag or digest allowed in reference`)
+	}
+
+	// Checks ok, now return a reference. This is a hack because the tag listing code expects a full image reference even though the tag is ignored
+	return docker.NewReference(reference.TagNameOnly(ref))
+}
+
+func listDockerTags(ctx context.Context, sys *types.SystemContext, referenceInput string) (string, []string, error) {
```
mtrmac: (Non-blocking: I’m not sure splitting this function is all that useful. It doesn't _hurt much_, OTOH the extra indirection makes it a bit harder to see that bot the caller of this and the helper called by this do an equivalent check for the expected transport value.)

cmd/skopeo/list_tags.go
```diff
@@ -0,0 +1,142 @@
...
+	if transport == nil {
+		return errors.New("Invalid transport")
+	}
+
+	if transport.Name() == docker.Transport.Name() {
+		repositoryName, tagListing, err = listDockerTags(ctx, sys, args[0])
+		if err != nil {
+			return err
+		}
+	} else {
```
mtrmac: Switch this to `if /* unexpected transport */ { return fmt.Errorf }`, then there’s no need for an `else` branch that somehow directly continues as the only option into the after-`if`/`else` code, and `repositoryName`/`tagListing` can use the `:=` declare+assign operator instead of a separate declaration.

cmd/skopeo/list_tags_test.go
```diff
@@ -0,0 +1,62 @@
...
+		ref, err := parseDockerRepositoryReference(test[0])
+		ref2, err2 := alltransports.ParseImageName(test[0])
+
+		if assert.NoError(t, err, "Could not parse, got error on %v", test[0]) && assert.NoError(t, err2, "Could not parse with regular parser, got error on %v", test[0]) {
+			assert.Equal(t, ref.DockerReference().String(), ref2.DockerReference().String(), "Different parsing output for input %v. Repo parse = %v, regular parser = %v", test[0], ref, ref2)
+		}
+	}
+}
+
+func TestUnsupportedRepositoryReferenceParser(t *testing.T) {
```
mtrmac: (Non-blocking: This can just as well be a part of `TestDockerRepositoryReferenceParser`.)

cmd/skopeo/list_tags_test.go
```diff
@@ -0,0 +1,62 @@
...
+			assert.Error(t, err, "Expected error in parsing but no error raised for %v", test[0])
+		} else {
+			if assert.NoError(t, err, "Could not parse, got error on %v", test[0]) {
+				assert.Equal(t, test[1], ref.DockerReference().Name(), "Mismatched parse result for input %v", test[0])
+			}
+		}
+	}
+}
+
+func TestDockerRepositoryReferenceParserDrift(t *testing.T) {
```
mtrmac: Non-blocking: An alternative to this approach would be to actually use `alltransports.ParseImageName().DockerReference()` as the primary data in `parseDockerRepositoryReference`, and do an extra `ParseNormalizedNamed` check at run-time only to check there isn’t a tag/digest.

That check can also add an “if the two don’t match, fail with an internal error” condition — that would ensure that they _always_ match even in cases that this forgets to test, and also all tests of `parseDockerRepositoryReference` would implicitly verify that the two results match in all tested cases.

cmd/skopeo/list_tags.go
zhill: > `run` already did a check for the right transport; it shouldn’t be necessary to do that twice. Preferably, do the checks in this function only, so that they can be covered by the unit tests.
> 

Ack. Yes, this was just a function-level sanity check to ensure the input is valid. A simpler check as you outline below is better since I made the change to a transport-specific parser.

> (`alltransports.ParseImageName`, with no manual string parsing here, would be esthetically preferable, but that would not allow rejecting tagged/digested input. I’m honestly undecided whether the extra check is worth it…)

Agreed. I'm of the opinion that it is needed since 'list-tags docker://myimage:latest' could be interpreted to mean "show me the other tags that also point to this image digest", but if you guys feel that docs are a sufficient means of clarification then I can remove a bunch of this redundant checking code.

> 
> Alternatively, just checking explicitly for the `docker.transport.Name()+"://"` prefix, and stripping exactly that prefix, would do just as many assumptions about the format, and be simpler.

Agree as above. I'll make this change to simplify things and remove redundant code.

cmd/skopeo/list_tags.go
zhill: Not opposed, but that structure I put there was to make it clear where additional transport support would be added. Each new one is just another condition without having to change the function flow at all.

mtrmac: Ah, OK. I’m still not seeing that at all that likely to happen :)

In that case, a
```go
switch transport.Name() {
case docker.Transport.Name():
   …
default:
  // fail
}
```
structure would be even clearer about the intent. (Non-blocking I guess.)

cmd/skopeo/list_tags.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	sys, err := opts.image.newSystemContext()
+	if err != nil {
+		return err
+	}
+
+	transport := alltransports.TransportFromImageName(args[0])
+	if transport == nil {
+		return errors.New("Invalid transport")
```
mtrmac: Something like `Invalid repository %q: does not specify a transport`, maybe, to point the user at the invalid input. “Invalid transport” alone is somewhat opaque to users not familiar with Skopeo / c/image jargon.

cmd/skopeo/list_tags.go
mtrmac: OK, looking at the `run` code as the one responsible for branching to different transports: a simple check+strip of the `docker.transport.name()+"///"` prefix, failing with an `Internal error: Docker reference expected` (to make it clear that the end-user-visible error handling happens elsewhere) would work well.

zhill:
apologies for going dark, some other stuff has come up but I'll get to this ASAP. I don't see any issues with making the suggested changes and I'll get those in as soon as I can but may be few more days.

zhill:
I'll update the branch to latest, but I think I resolved all the comments. Note that I did remove the integration test since there wasn't one for the `inspect` command and such a test was dependent on an external registry and didn't really provide a lot of value that was't covered by the unit tests. I can add it back in if you feel it provides real test value moving forward.

PR #701, opened 2019-08-06
Release v0.1.39

vrothberg:
@mtrmac @rhatdan PTAL

This mainly pulls in the latest fixes in c/image and c/storage.

PR #718, opened 2019-09-16
[DO NOT MERGE] pet PR for c/image#701

lsm5:
go mod edit -replace github.com/containers/image=github.com/lsm5/containers-image@master-
openssl-wip

make vendor

go get github.com/containers/libtrust

make vendor

pet PR as per [this comment](https://github.com/containers/libtrust/pull/3#issuecomment-531666207).
@vrothberg PTAL


lsm5:
@vrothberg if you're still around, CI is green, so hopefully c/image PR should be good for merge??

go.mod
```diff
@@ -44,3 +44,5 @@ require (
 	gopkg.in/yaml.v2 v2.0.0-20141029210843-d466437aa4ad // indirect
 	k8s.io/client-go v0.0.0-20181219152756-3dd551c0f083 // indirect
 )
+
+replace github.com/docker/libtrust => github.com/containers/libtrust v0.0.0-20190913040956-14b96171aa3b
```
mtrmac: Does this really need to be here? I would expect an update to the private branch of containers/image  (testing the code that directly refers to github.com/containers/libtrust), and that transitively brings in github.com/containers/libtrust , which is _actually vendored_ as github.com/containers/libtrust .

This PR does not actually add github.com/containers/libtrust at the expected ordinary path, and it passes pretty much only because it **doesn’t test the c/image PR**.

mtrmac: > ```
> go mod edit -replace github.com/containers/image=github.com/lsm5/containers-image@master-openssl-wip
> make vendor
> go get github.com/containers/libtrust
> make vendor
> ```

This seems plausible (maybe we should include the full commands in `CONTRIBUTING.md` as well), but the contents of the PR don’t correspond to these commands AFAICS.

vrothberg: Yes, that's what I had in mind as well -> https://github.com/containers/skopeo/pull/720 is passing the vendor checks

lsm5:
@vrothberg so with go 1.11 on ubuntu 18.04, I could see more untracked files, updated PR>

go.mod
mtrmac: @lsm5 Drop this override, please.

vrothberg:
LGTM ... once the c/image PR is merged, we need to drop the replace in this PR here and can merge

mtrmac:
(For the record, this can’t be merged right now because it does not correspond to what #701 ultimately did.)

mtrmac:
@lsm5 Assuming this is no longer necessary, please reopen if that’s not the case.

Issue #721, opened 2019-09-18
Yaml dependency version


MericCelik:
Hello,

The latest release of skopeo (0.1.39) has a dependency of a quite an old version of yaml package, which can be seen as
`gopkg.in/yaml.v2 v2.0.0-20141029210843-d466437aa4ad`. 

In this commit, the library is licensed under LGPLv3. It seems that this changed to Apache 2.0 in newer releases. Would you consider to update this dependency? 

I also want to mention that there might be a problem with the vendor version. The commit id of the dependency belongs to a much older release, rather than v2.0.0. The license file in the skopeo's yaml vendor and the license file in the v2.0.0 release in the yaml Github repo does not match as well. So if I'm not mistaken, skopeo uses a yaml package version older than 2.0.0. Maybe a problem occured during the vendoring.

Hope it helps!

vrothberg:
Hi @MericCelik, thanks for opening the issue. I've opened PR https://github.com/containers/skopeo/pull/722 to update the dependency to the latest release.

mtrmac:
#736 has updated the library. Thanks for your report!

PR #722, opened 2019-09-18
vendor gopkg.in/yaml.v2@v2.2.2

vrothberg:
Fixes: #721


mtrmac:
Sure, this works. OTOH, it’s an indirect dependency of a dependency of a dependency.

Could this happen in the c/image Go module, to cause skopeo vendor to be automatically updated?

Or, I guess, even better, update c/image to use a recent version `github.com/ghodss/yaml`, which is now a Go module, and explicitly requires `gopkg.in/yaml.v2 v2.2.2`; that way there’s no magic knowledge about this package necessary in Skopeo itself.

vrothberg:
c/image might be the better place but it has an even older version. I don't have much time to invest it further, so I am okay with merging it.
We should do some dependency house cleaning in c/image and c/storage to propagate updates more easily (and frequently).

I'll leave it up to you, @mtrmac. Feel free to close if you prefer the cleaner way.

mtrmac:
The licensing concerns apply to all users of c/image equally, so a Skopeo-only fix is not really sufficient.

rhatdan:
I opened https://github.com/containers/image/pull/707, does this help?

mtrmac:
> I opened [containers/image#707](https://github.com/containers/image/pull/707), does this help?

Not with _this_ issue (YAML.v2) but upgrading github.com/imdario/mergo removes a dependency on YAML.v1 , which has the same license.  But that fix is actually more involved, see https://github.com/containers/image/pull/714 .

BTW this will be automatically resolved next time c/image is updated in skopeo after https://github.com/containers/image/pull/714 , because that updates https://github.com/stretchr/testify to v1.4.0, which pulls in the v2.2.2 version of YAML.v2.  Still, explicitly encoding that minimum version _in c/image_ might be worthwhile.

vrothberg:
Closing. Let's get those things ironed out in c/image, where applicable. Thanks for looking into it, @mtrmac.

PR #723, opened 2019-09-19
[DO NOT MERGE] use latest c/image

lsm5:

@vrothberg PTAL


lsm5:
hmm, this one seems wrong, zero files changed, but `CopySuite.TestCopyAtomicExtension` seems to fail with the latest upstream c/image but works fine with c/image commit 0cc0e9 which is what's being used currrently.

PR #731, opened 2019-09-30
systemtest/010-inspect.bats: require only PATH

vrothberg:
The "inspect: env" test started failing since the environment in the
`fedora:latest` image has changed.  Hence, only check for `PATH` in
the image's environment which is a defacto standard.


vrothberg:
@mtrmac PTAL

PR #732, opened 2019-10-01
Image encryption/decryption support in skopeo

harche:
Add support for image encryption/decryption in `skopeo`

For this to work, `containers/image` needs to support encrypted images. There is a PR by  @lumjjb to add the support in `containers/image`, https://github.com/containers/image/pull/673. As soon as that PR is merged and is vendored into `skopeo` I will update this PR to use vendored `containers/image`. 

Signed-off-by: Harshal Patil <harshal.patil@in.ibm.com>
Signed-off-by: Brandon Lum <lumjjb@gmail.com>


cmd/skopeo/copy.go
```diff
@@ -77,6 +80,16 @@ func copyCmd(global *globalOptions) cli.Command {
 				Usage: "`MANIFEST TYPE` (oci, v2s1, or v2s2) to use when saving image to directory using the 'dir:' transport (default is manifest type of source)",
 				Value: newOptionalStringValue(&opts.format),
 			},
+			cli.StringFlag{
+				Name:        "encryption-keys",
+				Usage:       "keys (comma separated) with the encryption protocol to use needed to encrypt the image (e.g. jwe:/path/to/key.pem)",
+				Destination: &opts.encryptionKeys,
+			},
+			cli.StringFlag{
```
mtrmac: Why not use `cli.StringSliceFlag`?

lumjjb: Yea - agree we can make this change.

docs/skopeo-copy.1.md
```diff
@@ -80,6 +84,25 @@ To copy and sign an image:
...
 ```
 
+To encrypt an image:
+```sh
+./skopeo copy docker://docker.io/library/nginx:latest oci:local_nginx:latest
+
+openssl genrsa -out private.key 1024
+openssl rsa -in private.key -pubout > public.key
+
+./skopeo  copy --encryption-key jwe:./public.key oci:local_nginx:latest oci:try-encrypt:enc
```
vrothberg: Can we name the destination `oci:local_nginx:encrypted`? I think this will highlight the relationship with the source a bit more. Same for the commands below.

docs/skopeo-copy.1.md
```diff
@@ -80,6 +84,25 @@ To copy and sign an image:
 $ skopeo copy --sign-by dev@example.com atomic:example/busybox:streaming atomic:example/busybox:gold
 ```
 
+To encrypt an image:
+```sh
+./skopeo copy docker://docker.io/library/nginx:latest oci:local_nginx:latest
```
vrothberg: `s/\.skopeo/\$ skopeo/` same for the commands below (including openssl).

integration/copy_test.go
```diff
@@ -160,6 +163,56 @@ func (s *CopySuite) TestCopySimple(c *check.C) {
...
+	defer os.RemoveAll(dir1)
+	dir2, err := ioutil.TempDir("", "copy-2")
+	c.Assert(err, check.IsNil)
+	defer os.RemoveAll(dir2)
+	dir3, err := ioutil.TempDir("", "copy-3")
+	c.Assert(err, check.IsNil)
+	defer os.RemoveAll(dir3)
+	dir4, err := ioutil.TempDir("", "copy-4")
+	c.Assert(err, check.IsNil)
+	defer os.RemoveAll(dir4)
```
vrothberg: Can we name those variables as something more descriptive? E.g., `dir1` as `unencryptedDir`? It make the tests easier to understand and ultimately easier to maintain in the future.

docs/skopeo-copy.1.md
```diff
@@ -30,6 +30,10 @@ If the authorization state is not found there, $HOME/.docker/config.json is chec
 
 **--sign-by=**_key-id_ add a signature using that key ID for an image name corresponding to _destination-image_
 
+**--encryption-key** _Key_ a reference prefixed with the encryption protocol to use. e.g. jwe:/path/to/key.pem or pgp:admin@example.com or pkcs7:/path/to/x509-file
```
vrothberg: `e.g.` -> `For instance,`.

vrothberg: Which protocols are supported? Are there any other requirements? We should add a sentence (or two) about that.

vrothberg:
I added a `WIP -` to the title to prevent us from accidentally merging the c/image fork. We first need to merge https://github.com/containers/image/pull/673 and can then merge this PR here.

mtrmac:
> Since it's not yet in the OCI spec and potentially subject to change, I prefer marking the {de,en}cryption supports in the man pages and the help message as experimental to avoid users from depending on it.
> @mtrmac @mrunalp @rhatdan WDYT?

Yeah, there’s the dependency cycle between only releasing a final version of the code after a final version of an interoperable specification, and only releasing a final version of an interoperable specification after there is a production implementation (or two?).

At this point I think making it as experimental is appropriate, as well as being careful about implying any endorsement by OCI (which this PR does not do).

lumjjb:
>apart from the actual ocicrypt design/implementation.

@mtrmac there are some reviews on this but they are not part of the current repo since it was moved. "Archived" discussions on the ocicrypt library can be found in the PRs listed below. There were quite a bit of discussion on calls as well. the discussions were between IBM, our internal NIST counterpart and Docker. In addition to that, there are some notes on the discussions in the OCI weekly dev call as well in the HackMD. 

Reviews:
- https://github.com/containerd/containerd/pull/3001
- https://github.com/containerd/containerd/pull/3134

cmd/skopeo/copy.go
```diff
@@ -147,13 +160,61 @@ func (opts *copyOptions) run(args []string, stdout io.Writer) error {
...
+	if len(opts.encryptionKeys.Value()) > 0 {
+		// encryption
+		if len(opts.decryptionKeys.Value()) > 0 {
+			return fmt.Errorf("--encryption-key and --decryption-key cannot be specified together")
+		}
+
+		encryptionKeys := opts.encryptionKeys.Value()
+		ecc, err := enchelpers.CreateCryptoConfig(encryptionKeys, []string{})
+		if err != nil {
+			return err
```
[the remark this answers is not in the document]
harche: While fixing I realized that if the user specifies both `--encryption-key` and `--decryption-key` it won't be possible to tell which one is invalid. So I moved that check [outside](https://github.com/containers/skopeo/pull/732/files#diff-0afbed05940006aa6ff131fc915c90b5R164) in a separate if condition.

harche:
Test

vrothberg:
The c/image PR is merged now, so we can vendor c/image directly. The CI is currently failing with the vendor checks as we're now enforcing go 1.13. This version of golang behaves slightly different and removes some unneeded files from ./vendor. I'll open a PR to add a `make vendor-in-container` target.

If you don't have 1.13 installed, you can use a container:
```bash
$ podman run --privileged --rm --env HOME=/root -v `pwd`:/src -w /src docker.io/library/golang:1.13 make vendor
```

harche:
@vrothberg I just pushed by updating vendor with go 1.13.4. 

About directly importing `c/image`, right now in go.mod we [refer](https://github.com/containers/skopeo/blob/master/go.mod#L7) to the specific release of the `c/image`. So IMO we will have to wait for the next release of `c/image` unless we are OK with replacing `github.com/containers/image/v5 v5.0.0` with `github.com/containers/image/v5 502848a1358b079fe809f00b14df01dfd00192eb` in skopeo's go.mod.

go.mod
```diff
@@ -21,3 +22,5 @@ require (
 	github.com/urfave/cli v1.22.1
 	go4.org v0.0.0-20190218023631-ce4c26f7be8e // indirect
 )
+
+replace github.com/containers/image/v5 => github.com/lumjjb/image/v5 v5.0.0-20191125184705-a298da5c535d
```
vrothberg: Please drop the replace here and use commit `502848a1358b` for vendoring c/image/v5.

vrothberg:
> @vrothberg I just pushed by updating vendor with go 1.13.4.
> 
> About directly importing `c/image`, right now in go.mod we [refer](https://github.com/containers/skopeo/blob/master/go.mod#L7) to the specific release of the `c/image`. So IMO we will have to wait for the next release of `c/image` unless we are OK with replacing `github.com/containers/image/v5 v5.0.0` with `github.com/containers/image/v5 502848a1358b079fe809f00b14df01dfd00192eb` in skopeo's go.mod.

We need to vendor commit 502848a1358b from c/image as we don't want to depend on a private fork of c/image but to always use the latest required commit. Once we cut a new release for skopeo, we should vendor a release of c/image though.

harche:
> > @vrothberg I just pushed by updating vendor with go 1.13.4.
> > About directly importing `c/image`, right now in go.mod we [refer](https://github.com/containers/skopeo/blob/master/go.mod#L7) to the specific release of the `c/image`. So IMO we will have to wait for the next release of `c/image` unless we are OK with replacing `github.com/containers/image/v5 v5.0.0` with `github.com/containers/image/v5 502848a1358b079fe809f00b14df01dfd00192eb` in skopeo's go.mod.
> 
> We need to vendor commit 502848a1358b from c/image as we don't want to depend on a private fork of c/image but to always use the latest required commit. Once we cut a new release for skopeo, we should vendor a release of c/image though.

Done. Thanks.

cmd/skopeo/copy.go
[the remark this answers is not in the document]
mtrmac: https://github.com/containers/skopeo/pull/784

PR #733, opened 2019-10-02
DO NOT MERGE - Test containers/image/pull/721

vrothberg:


vendor/github.com/containers/image/types/types.go
```diff
@@ -6,7 +6,6 @@ import (
 	"time"
```
mtrmac: This file should have been removed; this suggests that some non-v4/-prefixed references remain.

vrothberg: Yes. I missed some. Did it manually - `gomove` did the trick now.

mtrmac:
I have used this work to update to final v4.0.1 in #736. Thanks!

PR #742, opened 2019-10-15
Temporarily work around auth.json location confusion

nalind:
Temporarily work around https://github.com/containers/libpod/issues/4227 until Fedora has an updated package that includes the fix, at which point we should revert this patch.

### DIFF
--- a/systemtest/040-local-registry-auth.bats
+++ b/systemtest/040-local-registry-auth.bats
@@ -14,6 +14,15 @@ function setup() {
     mkdir -p $_cred_dir/containers
     rm -f $_cred_dir/containers/auth.json
 
+    # TODO: This is here to work around
+    # https://github.com/containers/libpod/issues/4227 in the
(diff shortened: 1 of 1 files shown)


mtrmac:
I can live with this (keeping CI broken is a real risk), but please accompany this with a scary and clear comment about why this needs to exist, how it works and when it can be removed, to make it much less likely that will stay around indefinitely.

mtrmac:
(Basically this breaks the only test that noticed that `podman login` and `skopeo` became inconsistent, so overriding that test is Very Bad long-term.)

nalind:
Added a comment to the test script.

mtrmac:
Thanks! LGTM, please merge after tests pass.

TomSweeneyRedHat:
LGTM
One set of the tests crashed and burned with what looked to be unrelated network issues.  I've restarted it in hopes it gets lucky this time.

PR #746, opened 2019-10-21
systemtest: inpsect container image different from current platform arch

chuanchang:
When --raw is provided, can inspect show the raw manifest list, w/o
requiring any particular platform to be present, this test case is
used for make sure inspect command w/ --raw option works well for
container image is different from current platform arch.

$ bats -f "diff platform" systemtest/010-inspect.bats
 ✓ inspect: image manifest list w/ diff platform

1 test, 0 failures

### DIFF
--- a/systemtest/010-inspect.bats
+++ b/systemtest/010-inspect.bats
@@ -79,4 +79,38 @@ END_EXPECT
     done
 }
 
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@mtrmac @vrothberg @giuseppe Can we get this merged. Just adding a new test.

PR #760, opened 2019-11-05
Support running tests from c/image

mtrmac:
See https://github.com/containers/skopeo/pull/760#issuecomment-557286929 below.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,10 +26,11 @@ GO ?= go
 CONTAINER_RUNTIME := $(shell command -v podman 2> /dev/null || echo docker)
 GOMD2MAN ?= $(shell command -v go-md2man || echo '$(GOBIN)/go-md2man')
 
-GO_BUILD=$(GO) build
-# Go module support: set `-mod=vendor` to use the vendored sources
+# Go module support: set `-mod=vendor` to use the vendored sources.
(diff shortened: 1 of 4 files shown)


mtrmac:
For this PR to work, _both_ the tests in here and https://github.com/containers/image/pull/744 need to succeed, which is not currently the case.

mtrmac:
Yay, success!

This primarily modifies the test suite so that it can again succeed when run from c/image’s `make test-skopeo`. See individual commit messages for details.

The two major elements are:
- Use `-mod=vendor` everywhere, so that a vendored-in copy of c/image under test is visible to Go 1.13.
- Remove `--net=host` from the container running `make test-system`; it nowadays breaks localhost accesses (e.g. `ping 127.0.0.1` fails). I have no idea _why_ it breaks networking, but arguably `--net=host` was never necessary and having separate namespaces is cleaner anyway, and we don’t quite have the luxury to stop c/image development just to diagnose this right now.

https://github.com/containers/image/pull/744 , with
> ```
> SKOPEO_REPO = mtrmac/skopeo
> SKOPEO_BRANCH = test-failures
> ```
(which will be momentarily removed) succeeds.

@vrothberg PTAL.

rhatdan:
LGTM

PR #784, opened 2019-12-09
Add context to --encryption-key / --decryption-key processing failures

mtrmac:
... to avoid context-less errors like
```console
$ skopeo copy --decryption-key foo dir:x dir:y
FATA[0000] open foo: no such file or directory
$ skopeo copy --encryption-key /dev/null dir:x dir:y
FATA[0000] Invalid recipient format
```


rhatdan:
LGTM
@vrothberg PTAL

TomSweeneyRedHat:
LGTM

Issue #790, opened 2019-12-18
Add override-variant flag


JirkaChadima:
Hi, first of all thanks for this great piece of software.

As discussed in https://github.com/containers/skopeo/pull/420, it might be useful to have a `--override-variant` flag along with `--override-arch` and `--override-os` options. For example [redis](https://hub.docker.com/v2/repositories/library/redis/tags/?page_size=25&page=1) is differentiating between `arm/v5` and `arm/v7` via the variant option.

My use case: We are building a custom OS image for Raspberry Pi and we need to populate the OS image with `arm/v7` specific docker images from docker hub during build phase that runs on `amd64` PC. Why? Because the resulting device won't have any internet when it is fired up, so `docker pull` won't work. Without the `--override-variant` flag, I'm unable to download a redis image with `skopeo copy` from https://hub.docker.com/_/redis/ because with `--override-arch=arm`, the `v5` variant gets downloaded.

rhatdan:
Would you be interested in opening a PR for this?
@vrothberg @nalind @mtrmac WDYT?

JirkaChadima:
I can try to put it together over the holidays. I'm no go expert though, so it might be a little rough around the edges.

rhatdan:
SGTM

mtrmac:
Fixed by #793.

PR #793, opened 2020-01-01
Adds the --override-variant command line flag

JirkaChadima:
This is part of #790 and can be merged only after https://github.com/containers/image/pull/786 or https://github.com/containers/image/pull/753

I'm not sure how dependencies work exactly in Go so I did not test this locally against the patched `containers/image` version, but it should work once that dependency is updated.


rhatdan:
Man pages?

mtrmac:
ACK to the Skopeo part (but it will need to update c/image to one that actually supports that flag).

rhatdan:
@JirkaChadima @mtrmac @vrothberg Do we still need/want this PR?

JirkaChadima:
> @JirkaChadima @mtrmac @vrothberg Do we still need/want this PR?

I'm fine without it, we have changed the pipeline for other reasons. But I still think it's a valuable addition to skopeo.

rhatdan:
@sujil02  PTAL

vrothberg:
I rebased to retrigger the tests; just to be sure.

mtrmac:
@JirkaChadima Thanks again!

PR #798, opened 2020-01-10
add specific authfile options to copy (and sync) command.

### LINKED ISSUE #773: Skopeo picks wrong credentials from .docker/config.json on copy
I am trying to copy an image from one private registry to another one with 
```
skopeo copy docker://some-docker-registry.com/source-registry/my-image:1.0.0 docker://src-docker-registry.com/dest-registry/my-image:1.0.0
```
This fails because, skopeo tries to authenticate to `source-registry` with the credentials from `dest-registry`. 

The debug log shows
```
DEBU[0000] GET https://some-docker-registry.com/.../token?account=some-other-name&scope=repository%3Adest-registry%2Fimage%3Apull 
```

I think it might be important here that the registry urls differ only in the last path segment.

The `.docker/config.json` contains two entries
```json
{
  "auths": {
    "some-docker-registry.com/source-registry": {
      "username": "some-name",
      "password": "***",
      "email": "some-name@example.com",
      "auth": "***"
    },
    "some-docker-registry.com/dest-registry": {
      "username": "some-other-name",
      "password": "***",
      "email": "some-other-name@example.com",
      "auth": "***"
    }
  }
}
```
I can successfully inspect both registries individually, when heaving separate config.json files for each registry. 

I think this should either be fixed or one should be able to pass separate config files for source and destination like one can do with the cert paths. This would be something like 
```
skopeo copy --src-authfile some-config.json --dest-authfile some-other-config.json  ...
```

dastrobu:
With additional prefixed flags for authfiles, it is possible to override the shared authfile flag to use different authfiles for src and dest registries. This is an important feature if the two registries have the same domain (but different paths) and require separate credentials.

Closes #773.

Signed-off-by: Daniel Strobusch <1847260+dastrobu@users.noreply.github.com>

### DIFF
--- a/cmd/skopeo/utils.go
+++ b/cmd/skopeo/utils.go
@@ -56,6 +56,7 @@ func sharedImageFlags() ([]cli.Flag, *sharedImageOptions) {
 type dockerImageOptions struct {
 	global         *globalOptions      // May be shared across several imageOptions instances.
 	shared         *sharedImageOptions // May be shared across several imageOptions instances.
+	authFilePath   optionalString      // Path to a */containers/auth.json (prefixed version to override shared image option).
 	credsOption    optionalString      // username[:password] for accessing a registry
 	dockerCertPath string              // A directory using Docker-like *.{crt,cert,key} files for connecting to a registry or a daemon
 	tlsVerify      optionalBool        // Require HTTPS and verify certificates (for docker: and docker-daemon:)
@@ -87,7 +88,18 @@ func dockerImageFlags(global *globalOptions, shared *sharedImageOptions, flagPre
 		credsOptionExtra += "," + credsOptionAlias
 	}
 
-	return []cli.Flag{
+	var flags []cli.Flag
+	if flagPrefix != "" {
+		// the non-prefixed flag is handled by a shared flag.
+		flags = append(flags,
+			cli.GenericFlag{
+				Name:  flagPrefix + "authfile",
+				Usage: "path of the authentication file. Default is ${XDG_RUNTIME_DIR}/containers/auth.json",
+				Value: newOptionalStringValue(&opts.authFilePath),
+			},
+		)
+	}
+	flags = append(flags,
 		cli.GenericFlag{
 			Name:  flagPrefix + "creds" + credsOptionExtra,
 			Usage: "Use `USERNAME[:PASSWORD]` for accessing the registry",
@@ -108,7 +120,8 @@ func dockerImageFlags(global *globalOptions, shared *sharedImageOptions, flagPre
 			Usage:       "Access the registry anonymously",
 			Destination: &opts.noCreds,
 		},
-	}, &opts
+	)
+	return flags, &opts
 }
 
 // imageFlags prepares a collection of CLI flags writing into imageOptions, and the managed imageOptions structure.
@@ -142,6 +155,9 @@ func (opts *imageOptions) newSystemContext() (*types.SystemContext, error) {
 		DockerDaemonHost:         opts.dockerDaemonHost,
 		DockerDaemonCertPath:     opts.dockerCertPath,
 		SystemRegistriesConfPath: opts.global.registriesConfPath,
+	}
+	if opts.dockerImageOptions.authFilePath.present {
+		ctx.AuthFilePath = opts.dockerImageOptions.authFilePath.value
 	}
 	if opts.tlsVerify.present {
 		ctx.DockerDaemonInsecureSkipTLSVerify = !opts.tlsVerify.value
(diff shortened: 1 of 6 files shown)


cmd/skopeo/delete.go
```diff
@@ -17,7 +17,9 @@ type deleteOptions struct {
 }
 
 func deleteCmd(global *globalOptions) cli.Command {
-	sharedFlags, sharedOpts := sharedImageFlags()
+	// delete has no shared flags or opts
+	var sharedFlags []cli.Flag
+	sharedOpts := &sharedImageOptions{}
```
mtrmac: This is to avoid defining the `--authfile` flag twice, I expect?

If so, doing it this way is risky because adding any new flags to `sharedImageFlags` would not be reflected here (and it would not be possible to fix cleanly).

Rather, I think `sharedImageFlags` should remain as is, and `dockerImageOptions` should _optionally_ add the specific `--authfile` flag only if the prefix is non-empty.

cmd/skopeo/utils.go
```diff
@@ -56,6 +56,7 @@ func sharedImageFlags() ([]cli.Flag, *sharedImageOptions) {
 type dockerImageOptions struct {
 	global         *globalOptions      // May be shared across several imageOptions instances.
 	shared         *sharedImageOptions // May be shared across several imageOptions instances.
+	authFilePath   optionalString      // Path to a */containers/auth.json (not authFilePath is available as shared image option and as non-shared prefixed version).
```
mtrmac: I’m not sure what the “not authFilePath is available…” part means. Just guessing, something like “if set,  shared.authFilePath is ignored”, but that does not quite fit.

(One complication in explaining this is that “non-shared” does not necessary imply “prefixed”, consider `skopeo inspect --authfile`, which does not use source/destination prefixes.)

cmd/skopeo/utils.go
```diff
@@ -143,6 +149,10 @@ func (opts *imageOptions) newSystemContext() (*types.SystemContext, error) {
 		DockerDaemonCertPath:     opts.dockerCertPath,
 		SystemRegistriesConfPath: opts.global.registriesConfPath,
 	}
+	// replace the shared authfile with the specific one if present.
```
mtrmac: (Absolutely non-blocking: not sure the comment is necessary, to an extent just repeats what the code says.)

cmd/skopeo/utils_test.go
```diff
@@ -180,3 +181,50 @@ func TestImageDestOptionsNewSystemContext(t *testing.T) {
 	_, err = opts.newSystemContext()
 	assert.Error(t, err)
 }
+
+// since there is a shared authfile image option and a non-shared (prefixed) one, make sure the override logic
+// works correctly.
+func TestImageOptionsAuthfileOverride(t *testing.T) {
+	// if authfile and dest-authfile is provided, dest-authfile wins
+	{
+		// Set everything to non-default values.
```
mtrmac: This comment does not match the data (in none of the three instances), just drop it.

cmd/skopeo/utils_test.go
```diff
@@ -180,3 +181,50 @@ func TestImageDestOptionsNewSystemContext(t *testing.T) {
 	_, err = opts.newSystemContext()
 	assert.Error(t, err)
 }
+
+// since there is a shared authfile image option and a non-shared (prefixed) one, make sure the override logic
+// works correctly.
+func TestImageOptionsAuthfileOverride(t *testing.T) {
+	// if authfile and dest-authfile is provided, dest-authfile wins
```
mtrmac: This would be much shorter, and more readable, as a table-driven test (see `TestImageOptionsNewSystemContext` for an example).

It would also be nice to add a test for the `prefix == ""` case.

cmd/skopeo/delete.go
dastrobu: I thought about this as well but wanted to keep the logic for shared flags, simple. However, now code is adjusted according to your suggestions.

cmd/skopeo/utils.go
dastrobu: updated comment.

cmd/skopeo/utils.go
dastrobu: removed

cmd/skopeo/utils_test.go
dastrobu: refactored test according to your suggestions below.

cmd/skopeo/utils_test.go
dastrobu: refactored and included empty prefix case.

Issue #805, opened 2020-01-29
Skopeo support for BigFilesTemporaryDir (SystemContext)


malaiwah:
I am using Skopeo to export docker-daemon images to docker-archive format and I find the tool really useful *but* it seems there is no way to define the temporary directory (it does not honour the TMPDIR environment variable) for the temporary copies made. It defaults to /var/tmp, which on the systems I use is quite tight.

There is a way to instruct containers/image to use a different folder for temporary files, but the option is not exposed thru the skopeo cli with an --option. The SystemContext field is called BigFilesTemporaryDir .

Here is where it is set for its default: https://github.com/containers/image/blob/de5ad242e515688f69225ca6d897cfb221e0f5c7/internal/tmpdir/tmpdir.go#L18

Here is where the copy operation fails on my system: https://github.com/containers/image/blob/d13a90f787a7d3cd58e6f13c7cd1f2307b74a1ef/docker/tarfile/dest.go#L110

Here is the pull request for BigFilesTemporaryDir: https://github.com/containers/image/pull/747

malaiwah:
It might be easy to put it right here:
https://github.com/containers/skopeo/blob/4489ddd8a5caa0181575620c56e912032582374b/cmd/skopeo/utils.go#L128
and
https://github.com/containers/skopeo/blob/4489ddd8a5caa0181575620c56e912032582374b/cmd/skopeo/utils.go#L147
and I might take a try at it.

rhatdan:
We are using this on Buildah now, but we should expose this or at least allow the setting of TMPDIR to modify the setting.

mtrmac:
#868 was merged.

PR #806, opened 2020-01-29
Skopeo should support for BigFilesTemporaryDir (SystemContext)

malaiwah:
Enhancement request: https://github.com/containers/skopeo/issues/805
Tested locally and works, but as this is my first time contributing to this project I am not entirely sure about style/code practices around here.


cmd/skopeo/utils.go
```diff
@@ -47,6 +48,11 @@ func sharedImageFlags() ([]cli.Flag, *sharedImageOptions) {
 			Usage:       "path of the authentication file. Default is ${XDG_RUNTIME_DIR}/containers/auth.json",
 			Destination: &opts.authFilePath,
 		},
+		cli.StringFlag{
+			Name:        "big-files-temporary-dir",
```
rhatdan: Can we just call this --tmpdir?

mtrmac:
> Buildah and Podman man defaults to $TMPDIR which is set to /var/tmp by default.

That’s not how $TMPDIR is defined to work; it usually points to a `/tmp`-like directory (which on systemd Linuxes tends to be tmpfs, thus undesirable for this purpose). Compare https://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xbd_chap10.html and https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap01.html .

(To clarify, the “which is set to `/var/tmp` by default” means “Podman/Buildah set it that way”, not “Linux systems set it that way by default”).

Given the systemd re-definition of `/tmp` (which… happened), using TMPDIR for the “big files” directory overloads the variable with no way for the user to set a consistent value for container tools and the others.

mtrmac:
> I think this option should be on `skopeo copy`, not on `skopeo`

No, consider horrible things like `skopeo inspect docker-daemon:…`, which use a `docker save`, save the (potentially huge!) result into a temporary file, and then read that temporary file to find the relevant parts of the archive.

A global option is appropriate.

cmd/skopeo/utils.go
malaiwah: I would love this to be for the whole tmpdir, but it might lead to confusion: I believe I saw some parts of containers/image or skopeo that used ioutil.TempDir directly without overriding with the big-files-temporary-dir system context value for where to put the directory. I would assume ioutil.TempDir does the right thing by default and honours TMPDIR/TMP environment variable(s).

malaiwah:
I reformatted the code changes and it passes checks.

cmd/skopeo/utils.go
mtrmac: Most (but not all) of the `ioutil.Temp*` calls create a temporary file in some specific directory, to implement atomic replacement.

> I would assume ioutil.TempDir does the right thing by default and honours TMPDIR/TMP environment variable(s).

On Unix, yes, it does (falling back to `/tmp`).

(Overall, I don’t feel strongly about the name at all.)

cmd/skopeo/utils.go
```diff
@@ -35,7 +35,8 @@ func commandAction(handler func(args []string, stdout io.Writer) error) cli.Acti
 // sharedImageOptions collects CLI flags which are image-related, but do not change across images.
 // This really should be a part of globalOptions, but that would break existing users of (skopeo copy --authfile=).
 type sharedImageOptions struct {
-	authFilePath string // Path to a */containers/auth.json
+	authFilePath         string // Path to a */containers/auth.json
+	bigFilesTemporaryDir string // A directory to use for big temporary files
```
mtrmac: Please use `globalOptions`, and a flag in `createApp()`, for this; `sharedImageOptions` is mostly a hack for `--authfile`.

rhatdan:
I hate multi work with multi "-" options.  I think they are for people who type better then I do.  
Please just use --tmpdir and explain that it is for storing big temporary files.  Have it defaulted to /var/tmp

mtrmac:
> Have it defaulted to /var/tmp

Not that part. c/image already defaults to `/var/tmp` if the `BigFilesTemporaryDir` override is not present, we shouldn’t be maintaining the hard-coded value in two places.

rhatdan:
SGTM.

rhatdan:
@mtrmac @malaiwah What is going on with this PR?

mtrmac:
@rhatdan Have the review comments been addressed?

rhatdan:
If they haven't we should be encouraging @malaiwah to finish the PR. Not letting it languish.

malaiwah:
:-) That is on my list.. true... true..

rhatdan:
@malaiwah friendly ping.

PR #802, opened 2020-01-23
Manually update buildah to v1.13.1

mtrmac:
Should help with #791.

PR #810, opened 2020-02-04
Remove the reference to openshift/api

### LINKED ISSUE #791: Dependabot can't parse your go.mod
Dependabot couldn't parse the go.mod found at `/go.mod`.

The error Dependabot encountered was:

```
go: github.com/containers/buildah@v1.11.6 requires
	github.com/openshift/api@v3.9.1-0.20190810003144-27fb16909b15+incompatible: invalid pseudo-version: preceding tag (v3.9.0) not found
```

[View the update logs](https://app.dependabot.com/accounts/containers/update-logs/18225939).

mtrmac:
This is another attempt to fix #791 after ineffective #802.

The referenced tag has been removed, which breaks dependabot (#791).

This is another attempt to fix it, by removing an explicit reference (which was added when updating Buildah, because the version seemed newer than Buildah's v0.0.0 with a newer commit).

The referenced package is never even physically vendored in here, so remove the reference:

> ```
> go mod edit -droprequire=github.com/openshift/api
> make vendor
> ```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,6 @@ require (
 	github.com/opencontainers/image-spec v1.0.2-0.20190823105129-775207bd45b6
 	github.com/opencontainers/image-tools v0.0.0-20170926011501-6d941547fa1d
 	github.com/opencontainers/runtime-spec v1.0.0 // indirect
-	github.com/openshift/api v3.9.1-0.20190810003144-27fb16909b15+incompatible // indirect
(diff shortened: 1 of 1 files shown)


mtrmac:
(Merging without review to hopefully fix this and allow Dependabot to file PRs and run tests overnight, when Travis might be a bit less loaded.)

PR #814, opened 2020-02-06
Partial image encryption support

harche:
While https://github.com/containers/skopeo/pull/732 brought in image encryption support in skopeo, this PR adds the support for partial image encryption. 

User will be able to select which layer of the image they want to encrypt instead of encrypting an entire image. This is very useful from the layer reusability point of view. e.g. User can take a standard `ubuntu:18:04` image and add their custom layer which they will like to protect with the encryption. 
 When that image is saved in the snapshotter of the runtime, the layers that aren't encrypted end up getting reused which results in better disk space utilization as well as faster downloads (only have to download encrypted layers if the rest of the layers are already present in the snapshotter)

Signed-off-by: Harshal Patil <harshal.patil@in.ibm.com>

### DIFF
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -28,6 +28,7 @@ type copyOptions struct {
 	format            optionalString  // Force conversion of the image to a specified format
 	quiet             bool            // Suppress output information when copying images
 	all               bool            // Copy all of the images if the source is a list
+	encryptLayer      cli.IntSlice    // The list of layers to encrypt
 	encryptionKeys    cli.StringSlice // Keys needed to encrypt the image
 	decryptionKeys    cli.StringSlice // Keys needed to decrypt the image
 }
@@ -91,6 +92,11 @@ func copyCmd(global *globalOptions) cli.Command {
 				Name:  "encryption-key",
 				Usage: "*Experimental* key with the encryption protocol to use needed to encrypt the image (e.g. jwe:/path/to/key.pem)",
 				Value: &opts.encryptionKeys,
+			},
+			cli.IntSliceFlag{
+				Name:  "encrypt-layer",
+				Usage: "*Experimental* the 0-indexed layer indices, with support for negative indexing (e.g. 0 is the first layer, -1 is the last layer)",
+				Value: &opts.encryptLayer,
 			},
 			cli.StringSliceFlag{
 				Name:  "decryption-key",
@@ -180,9 +186,14 @@ func (opts *copyOptions) run(args []string, stdout io.Writer) error {
 	var encConfig *encconfig.EncryptConfig
 	var decConfig *encconfig.DecryptConfig
 
+	if len(opts.encryptLayer.Value()) > 0 && len(opts.encryptionKeys.Value()) == 0 {
+		return fmt.Errorf("--encrypt-layer can only be used with --encryption-key")
+	}
+
 	if len(opts.encryptionKeys.Value()) > 0 {
 		// encryption
-		encLayers = &[]int{}
+		p := opts.encryptLayer.Value()
+		encLayers = &p
 		encryptionKeys := opts.encryptionKeys.Value()
 		ecc, err := enchelpers.CreateCryptoConfig(encryptionKeys, []string{})
 		if err != nil {
--- a/docs/skopeo-copy.1.md
+++ b/docs/skopeo-copy.1.md
@@ -98,12 +98,12 @@ To copy and sign an image:
 
 To encrypt an image:
 ```sh
-skopeo copy docker://docker.io/library/nginx:latest oci:local_nginx:latest
+skopeo copy docker://docker.io/library/nginx:1.17.8 oci:local_nginx:1.17.8
 
 openssl genrsa -out private.key 1024
 openssl rsa -in private.key -pubout > public.key
 
-skopeo  copy --encryption-key jwe:./public.key oci:local_nginx:latest oci:try-encrypt:encrypted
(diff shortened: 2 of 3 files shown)


docs/skopeo-copy.1.md
```diff
@@ -98,7 +98,7 @@ To copy and sign an image:
 
 To encrypt an image:
 ```sh
-skopeo copy docker://docker.io/library/nginx:latest oci:local_nginx:latest
+skopeo copy docker://docker.io/library/nginx:1.17.8 oci:local_nginx:1.17.8
```
mtrmac: Why this change? (It’s not obviously _undesirable_, but it seems quite unrelated and I can’t see why it is necessary.)

integration/copy_test.go
```diff
@@ -577,7 +586,7 @@ func (s *CopySuite) TestCopyEncryption(c *check.C) {
 		"oci:"+encryptedImgDir+":encrypted", "oci:"+decryptedImgDir+":decrypted")
 
 	// Copy a standard busybox image locally
-	assertSkopeoSucceeds(c, "", "copy", "docker://busybox", "oci:"+originalImageDir+":latest")
+	assertSkopeoSucceeds(c, "", "copy", "docker://busybox:1.31.1", "oci:"+originalImageDir+":latest")
```
mtrmac: Why this change? (It’s not obviously _undesirable_, but it seems quite unrelated and I can’t see why it is necessary.)

integration/copy_test.go
```diff
@@ -553,6 +553,15 @@ func (s *CopySuite) TestCopyEncryption(c *check.C) {
 	defer os.RemoveAll(keysDir)
 	undecryptedImgDir, err := ioutil.TempDir("", "copy-5")
 	defer os.RemoveAll(undecryptedImgDir)
+	MultiLayerImageDir, err := ioutil.TempDir("", "copy-6")
```
mtrmac: (lower-case first letter?)

cmd/skopeo/copy.go
```diff
@@ -182,7 +188,8 @@ func (opts *copyOptions) run(args []string, stdout io.Writer) error {
 
 	if len(opts.encryptionKeys.Value()) > 0 {
```
mtrmac: Please add a check to reject the command if the user specifies `--encrypt-layer` without `--encryption-key`.

docs/skopeo-copy.1.md
harche: I added an addition section, https://github.com/containers/skopeo/pull/814/files#diff-30c826204072774bd69d2ae15e60aef9R128, to show the partial image encryption in that file. 

The example image, `nginx:1.17.8` has exactly 3 layers and in order to showcase partial image encryption I encrypt the 2nd layer while keep the 1st and 3rd layer as is. While it didn't matter earlier because we were encrypting all the layers, this time I wanted to make sure that the image has exactly same 3 layers. We do not have control over how nginx image is built in future, maybe they could just shrink it to just one single layer. That's why I replaced `latest` with `1.17.8` tag. 

This will keep this doc unaffected by any future changes with respect to number of layers in the nginx image.

integration/copy_test.go
harche: To accomodate partial image encryption, I changed existing test case for image encryption slightly. So now after encrypting the image we count how many layers in the target image got encrypted. If you do a full image encryption then the number of layers from the original image should match with the number of encrypted layers in the target image. While if you do partial image encryption, the number of encrypted layers in the target image should be equivalent to the number of `encrypt-layer` argument(s) while encrypting the original image. 

So the reason I replaced `latest` with `1.31.1` is to make sure our test case shouldn't fail if the number of layers in the busybox change in future.

integration/copy_test.go
harche: sorry for the typo. Fixed it.

cmd/skopeo/copy.go
harche: done.

lumjjb:
Linking https://github.com/containers/skopeo/pull/732 for context

harche:
@mtrmac Thanks for the comments.

docs/skopeo-copy.1.md
mtrmac: Ah, I didn’t realize this is directly connected to the new example. The new example does benefit from fixing a version, sure.

docs/skopeo-copy.1.md
```diff
@@ -120,6 +120,14 @@ To decrypt an image that requires more than one key:
 ```sh
 skopeo copy --decryption-key ./private1.key --decryption-key ./private2.key --decryption-key ./private3.key oci:try-encrypt:encrypted oci:try-decrypt:decrypted
 ```
+
+Container images can also be partially encrypted by specifying the index of the layer. Layers are 0-indexed indices, with support for negative indexing. i.e. 0 is the first layer, -1 is the last layer.
+
+Let's say out of 3 layers that the image `docker.io/library/nginx:1.17.8` is made up of, we only want to encrypt the 2nd layer,
+```sh
+skopeo  copy --encryption-key jwe:./public.key --encrypt-layer 1 oci:local_nginx oci:try-encrypt:encrypted
```
mtrmac: Should this be `oci:local_nginx:1.17.8`, then?

integration/copy_test.go
mtrmac: Thanks!

docs/skopeo-copy.1.md
harche: Oh yes. Fixed it. Thanks.

harche:
> Thanks! One final documentation nit.

@mtrmac fixed it. Thanks.

harche:
@TomSweeneyRedHat thanks.

Issue #822, opened 2020-02-19
RFE: support REGISTRY_AUTH_FILE for credentials


edsantiago:
[ Slightly related to #434, but unique enough to merit its own issue. ]

podman allows setting `REGISTRY_AUTH_FILE=/path/to/auth.json` to override its default. skopeo should honor that envariable.

mtrmac:
In general I’m not quite thrilled with implicit environment variable overrides (often, although certainly not always, the same thing can be achieved with a shell alias/function), but sure, it might make sense in specific situations.

mtrmac:
Support was added in #829.

PR #829, opened 2020-02-21
add support for REGISTRY_AUTH_FILE

rhatdan:
Fix cli to use REGISTRY_AUTH_FILE if set and to display the
default location to use for authfiles in the `skopeo copy --help`

Modify tests to verify the different settings.

### DIFF
--- a/cmd/skopeo/utils.go
+++ b/cmd/skopeo/utils.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"io"
+	"os"
 	"strings"
 
 	"github.com/containers/image/v5/pkg/compression"
@@ -44,7 +45,8 @@ func sharedImageFlags() ([]cli.Flag, *sharedImageOptions) {
 	return []cli.Flag{
 		cli.StringFlag{
 			Name:        "authfile",
-			Usage:       "path of the authentication file. Default is ${XDG_RUNTIME_DIR}/containers/auth.json",
+			Usage:       "path of the authentication file. Example: ${XDG_RUNTIME_DIR}/containers/auth.json",
+			Value:       os.Getenv("REGISTRY_AUTH_FILE"),
 			Destination: &opts.authFilePath,
 		},
 	}, &opts
@@ -94,7 +96,7 @@ func dockerImageFlags(global *globalOptions, shared *sharedImageOptions, flagPre
 		flags = append(flags,
 			cli.GenericFlag{
 				Name:  flagPrefix + "authfile",
-				Usage: "path of the authentication file. Default is ${XDG_RUNTIME_DIR}/containers/auth.json",
+				Usage: "path of the authentication file. Example: ${XDG_RUNTIME_DIR}/containers/auth.json",
 				Value: newOptionalStringValue(&opts.authFilePath),
 			},
 		)
--- a/cmd/skopeo/utils_test.go
+++ b/cmd/skopeo/utils_test.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"flag"
+	"os"
 	"testing"
 
 	"github.com/containers/image/v5/types"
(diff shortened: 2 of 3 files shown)


mtrmac:
Link: https://github.com/containers/skopeo/issues/822

cmd/skopeo/utils.go
```diff
@@ -44,12 +47,33 @@ func sharedImageFlags() ([]cli.Flag, *sharedImageOptions) {
 	return []cli.Flag{
 		cli.StringFlag{
 			Name:        "authfile",
-			Usage:       "path of the authentication file. Default is ${XDG_RUNTIME_DIR}/containers/auth.json",
+			Value:       defaultAuthFile(),
+			Usage:       "path of the authentication file.",
```
mtrmac: Why isn’t this `authFileUsage`?

cmd/skopeo/utils.go
```diff
@@ -44,12 +47,33 @@ func sharedImageFlags() ([]cli.Flag, *sharedImageOptions) {
...
 			Destination: &opts.authFilePath,
 		},
 	}, &opts
 }
 
+func defaultAuthFile() string {
+	authFile := os.Getenv("REGISTRY_AUTH_FILE")
+	if authFile == "" {
+		runtimeDir := os.Getenv("XDG_RUNTIME_DIR")
```
mtrmac: No; if the file should be searched for in $XDG_RUNTIME_DIR, that should be a c/image default.

… in fact, it already _is_ the default in c/image, so why is it being added here?

cmd/skopeo/utils_test.go
```diff
@@ -42,10 +44,16 @@ func fakeImageOptions(t *testing.T, flagPrefix string, globalFlags []string, cmd
 
 func TestImageOptionsNewSystemContext(t *testing.T) {
 	// Default state
+
+	// Make sure when XDG_RUNTIME_DIR is set, defaults to "$XDG_RUNTIME_DIR/containers/auth.json"
+	runtimeDir := "/run/test/0"
+	os.Setenv("XDG_RUNTIME_DIR", runtimeDir)
```
mtrmac: See how `signing_test.go` ensures the test overrides don’t leak into other tests.

Or even https://github.com/containers/image/blob/d13a90f787a7d3cd58e6f13c7cd1f2307b74a1ef/pkg/blobinfocache/default_test.go#L24-L33 .

cmd/skopeo/utils.go
rhatdan: I guess it could be.  It will do the exact same thing and look the same.

cmd/skopeo/utils.go
rhatdan: This helps the user document the behaviour.  We document this in the current code in skopeo that $XDG_RUNTIME_DIR/containers/conf, I am just expanding the path to match what the usre had.

cmd/skopeo/utils.go
rhatdan: No it would not. The issue is setting the Value field causes the CLI help information to expand to  path of the authentication file. (default: "/run/user/3267/containers/auth.json") (default: "/run/user/3267/containers/auth.json")
IE It would put the message out twice.

cmd/skopeo/utils.go
rhatdan: The problem with burying the default in containers image is that it can not be revealed to the user. If you want to add the defaultAuthFile interface to containers image then I would be fine with it.

cmd/skopeo/utils_test.go
rhatdan: Fixed

cmd/skopeo/utils.go
rhatdan: I have changed the code to show the expected default value, but iff XDG_RUNTIME_DIR is not set and the user does not specify a value, then containers/image will still get "" so it can figure out its own default value.

cmd/skopeo/utils.go
mtrmac: If you want to show the path c/image would use, then call _exactly_ the code in c/image that determines that path, don’t make a copy, please.

(Sure, there might be some conversation about making that path public through an API, although I don’t expect that to be a major concern anyway — but just copy&pasting the logic with no process to keep it the same is long-term worse than not showing the path.)

rhatdan:
I simplified this and fell back to only using REGISTRY_AUTH_FILE.  No longer changing the default, unless this environment variable is set.
@mtrmac @TomSweeneyRedHat @vrothberg PTAL

rhatdan:
@mtrmac @TomSweeneyRedHat @vrothberg friendly ping

rhatdan:
@mtrmac @vrothberg Can we get this one in?

Issue #832, opened 2020-02-24
Couldn't install skopeo v0.1.41


ashoks27:
Hi All,

I am new a skopeo and wanted to install it. I have downloaded the source code and built an docker image using the dockerfile which I got from source code. But am unable to create a container using that Image. 

Please advise on how to build a container from here . Basicly I wanted to copy a images from one registry 

[root@serverA ~]# docker images |head
REPOSITORY                      TAG        IMAGE ID            CREATED             SIZE
container/skopeo                latest     3166d8202a8a        3 days ago          2.49 GB
skopeo-test                        latest     3166d8202a8a        3 days ago          2.49 GB
docker.io/fedora                latest     536f3995adeb        3 days ago          193 MB

root@serverA skopeo-0.1.41]# docker run skopeo --image=skopeo
Unable to find image 'skopeo:latest' locally
Trying to pull repository registry.access.redhat.com/skopeo ...
Pulling repository registry.access.redhat.com/skopeo
Trying to pull repository docker.io/library/skopeo ...
Trying to pull repository registry.fedoraproject.org/skopeo ...
Pulling repository registry.fedoraproject.org/skopeo
Trying to pull repository quay.io/skopeo ...
Trying to pull repository registry.centos.org/skopeo ...
Pulling repository registry.centos.org/skopeo
Trying to pull repository docker.pkg.github.com/skopeo ...
Pulling repository docker.pkg.github.com/skopeo
Trying to pull repository docker.io/library/skopeo ...
**/usr/bin/docker-current: repository docker.io/skopeo not found: does not exist or no pull access.**
See '/usr/bin/docker-current run --help'.
[root@serverA skopeo-0.1.41]#

mtrmac:
Thanks for your report.

Some ways to install `Skopeo` are available at https://github.com/containers/skopeo#obtaining-skopeo.  The `Dockerfile` in the repository exists for tests, it does not create an end-user-image.

One reason why Skopeo containers aren’t all that useful is that Skopeo very often needs access to credentials, which are typically set up by `podman login` / `docker login`. So, a Skopeo container could only work by passing the credential files via mounted volumes, for every single command — it’s so much simpler to have a local command and run it directly in the same user account as `podman login`.

ashoks27:
Thanks for your respose Miloslav...

[...]

PR #835, opened 2020-02-27
Mention CentOS/RHEL in installation instructions

mtrmac:
apparently the Fedora reference is not obviously relevant.

(I didn’t add the same update for the local build instructions, because that requires more information about enabling non-default repositories, and especially RHEL users should get a chance to think for a bit before overriding a system package. This is not at all clear to be the right decision — anyway, filing a PR for the easy decision for now.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,9 +149,13 @@ you'll get an error. You can fix this by either logging in (via `docker login`) 
 
 Obtaining skopeo
 -
-`skopeo` may already be packaged in your distribution, for example on Fedora 23 and later you can install it using
+`skopeo` may already be packaged in your distribution, for example on RHEL/CentOS ≥ 8 or Fedora you can install it using
 ```sh
 $ sudo dnf install skopeo
+```
+on RHEL/CentOS ≤ 7.x:
+```sh
+$ sudo yum install skopeo
 ```
 for openSUSE:
 ```sh


mtrmac:
Ref: #832.

README.md
```diff
@@ -149,7 +149,7 @@ you'll get an error. You can fix this by either logging in (via `docker login`)
 
 Obtaining skopeo
 -
-`skopeo` may already be packaged in your distribution, for example on Fedora 23 and later you can install it using
+`skopeo` may already be packaged in your distribution, for example on RHEL/CentOS/Fedora you can install it using
 ```sh
 $ sudo dnf install skopeo
```
TomSweeneyRedHat: IDK if it's worth bothering with, but dnf is only available on RHEL 8+ unless specifically installed.  Maybe add "install it using dnf or yum"

Your call.

TomSweeneyRedHat:
One question for consideration, LGTM

README.md
mtrmac: Thanks, good point. I’l try to think of a way to make this clearer.

mtrmac: I couldn’t think of anything smart, so I just described it the brute-force way.

rhatdan:
LGTM

Issue #839, opened 2020-03-04
RFE: Provide a skopeo-login/skope-logout


TomSweeneyRedHat:
It would be good to provide a skopeo-login and a skopeo-logout command that works in the same manner as podman-login and podman-logout.  That way another package would not have to be installed in order for login to work.

Bonus points:  Have the guts of the login and logout code live in containers/common so that only minimal implementations would need to be made inside each of the three (Podman, Buildah, Skopeo) projects.

vrothberg:
Sharing everything in containers/common could be tough as Skopeo is not using cobra for the CLI.

TomSweeneyRedHat:
You know what I call that @vrothberg ?  More opportunities!  ;^)

mtrmac:
This was added by #865 .

PR #857, opened 2020-03-18
Use cobra in skopeo

QiWang19:
Use cobra in skopeo can help share code with podman/buildah(code for skopeo login/logout CLI).
https://github.com/containers/skopeo/issues/839


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/containers/skopeo
 go 1.12
 
 require (
+	github.com/QiWang19/image v0.0.0-20190424091117-ff926d3c7968
```
vrothberg: This seems wrong.

go.mod
```diff
@@ -18,6 +19,8 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/russross/blackfriday v2.0.0+incompatible // indirect
 	github.com/sirupsen/logrus v1.4.2
+	github.com/spf13/cobra v0.0.5
+	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.5.1
 	github.com/syndtr/gocapability v0.0.0-20180916011248-d98352740cb2
 	github.com/urfave/cli v1.22.1
```
vrothberg: We should be able to delete this dependency with this PR. Did you run `go mod tidy`?

cmd/skopeo/signing_test.go
```diff
@@ -7,7 +7,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/containers/image/v5/signature"
+	"github.com/QiWang19/image/signature"
```
vrothberg: This seems wrong.

cmd/skopeo/utils.go
```diff
@@ -158,25 +125,26 @@ func (opts *imageOptions) newSystemContext() (*types.SystemContext, error) {
 		DockerDaemonCertPath:     opts.dockerCertPath,
 		SystemRegistriesConfPath: opts.global.registriesConfPath,
 	}
-	if opts.dockerImageOptions.authFilePath.present {
-		ctx.AuthFilePath = opts.dockerImageOptions.authFilePath.value
+	if cmd.Flags().Changed("src-authfile") || cmd.Flags().Changed("dest-authfile") {
```
mtrmac: This should not be just guessing the prefix, if at all possible.

cmd/skopeo/utils.go
```diff
@@ -274,30 +226,40 @@ func getDockerAuth(creds string) (*types.DockerAuthConfig, error) {
...
 	if err != nil {
 		return nil, err
 	}
 	return ref.NewImageSource(ctx, sys)
 }
+
+// UsageTemplate returns the usage template for skopeo commands
+// This blocks the desplaying of the global options. The main skopeo
+// command should not use this.
+func UsageTemplate() string {
```
mtrmac: This can probably be private to this package.

cmd/skopeo/utils_test.go
```diff
@@ -1,68 +1,92 @@
...
+	fs := pflag.FlagSet{}
+	fs.BoolVar(&opts.debug, "debug", false, "enable debug output")
+	fs.BoolVar(&opts.tlsVerify, "tls-verify", true, "Require HTTPS and verify certificates when accessing the registry")
+	fs.StringVar(&opts.policyPath, "policy", "", "Path to a trust policy file")
+	fs.BoolVar(&opts.debug, "insecure-policy", false, "run the tool without any policy check")
+	fs.StringVar(&opts.registriesDirPath, "registries.d", "", "use registry configuration files in `DIR` (e.g. for container signature storage)")
+	fs.StringVar(&opts.overrideArch, "override-arch", "", "use `ARCH` instead of the architecture of the machine for choosing images")
+	fs.StringVar(&opts.overrideOS, "override-os", "", "use `OS` instead of the running OS for choosing images")
+	fs.DurationVar(&opts.commandTimeout, "command-timeout", 0, "timeout for the command execution")
+	fs.StringVar(&opts.registriesConfPath, "registries-conf", "", "path to the registries.conf file")
```
mtrmac: Defining a completely new set of options here makes the tests dubious useful for testing CLI option handling of the actual code.

mtrmac:
I’m also somewhat wondering about how the integration into Podman is going to work — the two have different sets of global options, and Podman’s subcommand-independent(?) infrastructure is fairly extensive. But, one step at a time.

cmd/skopeo/copy.go
```diff
@@ -3,111 +3,85 @@ package main
...
+	flags.StringVarP(&opts.format, "format", "f", "", `MANIFEST TYPE (oci, v2s1, or v2s2) to use when saving image to directory using the 'dir:' transport (default is manifest type of source)`)
+	flags.StringSliceVar(&opts.encryptionKeys, "encryption-key", []string{}, "*Experimental* key with the encryption protocol to use needed to encrypt the image (e.g. jwe:/path/to/key.pem)")
+	flags.IntSliceVar(&opts.encryptLayer, "encrypt-layer", []int{}, "*Experimental* the 0-indexed layer indices, with support for negative indexing (e.g. 0 is the first layer, -1 is the last layer)")
+	flags.StringSliceVar(&opts.decryptionKeys, "decryption-key", []string{}, "*Experimental* key needed to decrypt the image")
+	rootCmd.AddCommand(copyCommand)
 }
 
-func (opts *copyOptions) run(args []string, stdout io.Writer) error {
+func copyCmd(c *cobra.Command, args []string, opts copyOptions) error {
+	stdout := os.Stdout
```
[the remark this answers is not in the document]
mtrmac: … and existing tests do fail on this.

cmd/skopeo/layers.go
```diff
@@ -13,32 +12,38 @@ import (
...
-	imageFlags, imageOpts := imageFlags(global, sharedOpts, "", "")
-	opts := layersOptions{
-		global: global,
-		image:  imageOpts,
-	}
-	return cli.Command{
-		Name:      "layers",
-		Usage:     "Get layers of IMAGE-NAME",
-		ArgsUsage: "IMAGE-NAME [LAYER...]",
-		Hidden:    true,
```
mtrmac: `layers` should continue to be hidden.

cmd/skopeo/main.go
```diff
@@ -29,90 +26,52 @@ type globalOptions struct {
...
+)
 
-	app := cli.NewApp()
-	app.EnableBashCompletion = true
-	app.Name = "skopeo"
-	if gitCommit != "" {
-		app.Version = fmt.Sprintf("%s commit: %s", version.Version, gitCommit)
-	} else {
-		app.Version = version.Version
-	}
```
mtrmac: `skopeo --version` should be preserved.

cmd/skopeo/copy.go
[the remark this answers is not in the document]
mtrmac: This should be possible to do using `Command.OutOrStdout()` + `Command.Setout()`.

cmd/skopeo/copy.go
```diff
@@ -132,26 +106,27 @@ func (opts *copyOptions) run(args []string, stdout io.Writer) error {
...
-	destinationCtx, err := opts.destImage.newSystemContext()
+	destinationCtx, err := opts.destImage.newSystemContext(c)
 	if err != nil {
 		return err
 	}
 
 	var manifestType string
-	if opts.format.present {
-		switch opts.format.value {
+	if c.Flags().Changed("format") {
```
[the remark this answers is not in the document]
mtrmac: It seems this should be possible using `flags.Var(&opts.format)`.

cmd/skopeo/copy.go
```diff
@@ -3,111 +3,85 @@ package main
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return copyCmd(cmd, args, opts)
+		},
+		Example: `skopeo copy --sign-by dev@example.com container-storage:example/busybox:streaming docker://example/busybox:gold`,
+	}
+	copyCommand.SetUsageTemplate(UsageTemplate())
+	copyCommand.DisableFlagsInUseLine = true
+
+	flags := copyCommand.Flags()
+	flags.SetInterspersed(false)
```
mtrmac: Why? Things like `skopeo copy dir:a dir:b --format foo` do currently recognize flags.

mtrmac:
> I’m also somewhat wondering about how the integration into Podman is going to work — the two have different sets of global options

As a first guess, the old structure (`func copyCmd(global *globalOptions) Command`) seems broadly acceptable: Podman can have its own parsing of the global options, somehow get them into the `global` structure (that would be new and annoying to maintain, sure), and then invoke `Command.Execute*()` with the remaining options/arguments.

cmd/skopeo/utils.go
```diff
@@ -274,30 +209,40 @@ func getDockerAuth(creds string) (*types.DockerAuthConfig, error) {
...
+Aliases:
+  {{.NameAndAliases}}{{end}}{{if .HasExample}}
+
+Examples:
+  {{.Example}}{{end}}{{if .HasAvailableSubCommands}}
+
+Available Commands:{{range .Commands}}{{if (or .IsAvailableCommand (eq .Name "help"))}}
+  {{rpad .Name .NamePadding }} {{.Short}}{{end}}{{end}}{{end}}{{if .HasAvailableLocalFlags}}
+
+Flags:
```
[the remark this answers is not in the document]
QiWang19: This removes  "Global Flags" from command help. `c/libpod` and `c/buildah` also have a copy of this, can this be moved to c/common?

QiWang19:
@mtrmac @vrothberg PTAL

override the `IsBoolFlag() bool` seems not work

mtrmac:
> override the `IsBoolFlag() bool` seems not work

Could you elaborate, please? I have no idea what is the problem / expected / actual behavior.

mtrmac:
Is that about the test failures?

QiWang19:
> Is that about the test failures?

Yes.  That's about the test failures. For OptionalBool `--OB true`, the `true` should not be regarded as the value of the --OB.

cmd/skopeo/delete.go
```diff
@@ -3,44 +3,52 @@ package main
...
-	opts := deleteOptions{
-		global: global,
-		image:  imageOpts,
-	}
-	return cli.Command{
-		Name:  "delete",
-		Usage: "Delete image IMAGE-NAME",
-		Description: fmt.Sprintf(`
-	Delete an "IMAGE_NAME" from a transport
+func deleteCommand(globalFlagResults *globalOptions) *cobra.Command {
```
mtrmac: (Non-blocking I guess: `globalFlagResults` is a pretty unwieldy, and “flag results” is not intuitively comprehensible.

It’s also inconsistent with `sharedOpts`/`imageOpts`.

If you don’t like just `global`, which might be good enough given the context, maybe `globalOpts`?)

cmd/skopeo/delete.go
```diff
@@ -3,44 +3,52 @@ package main
...
 )
 
 type deleteOptions struct {
 	global *globalOptions
 	image  *imageOptions
 }
 
-func deleteCmd(global *globalOptions) cli.Command {
-	sharedFlags, sharedOpts := sharedImageFlags()
-	imageFlags, imageOpts := imageFlags(global, sharedOpts, "", "")
```
mtrmac: (Overall, this would be notably easier to review if it didn’t move unchanged code around, or if there were a separate commit with code changes, and then a separate commit with moves and no changes.)

cmd/skopeo/delete.go
```diff
@@ -3,44 +3,52 @@ package main
...
-		image:  imageOpts,
-	}
-	return cli.Command{
-		Name:  "delete",
-		Usage: "Delete image IMAGE-NAME",
-		Description: fmt.Sprintf(`
-	Delete an "IMAGE_NAME" from a transport
+func deleteCommand(globalFlagResults *globalOptions) *cobra.Command {
+	var (
+		opts              deleteOptions
```
mtrmac: Why was this changed from just using a struct constructor? Now the reader has to worry about uses of the uninitialized value.

cmd/skopeo/delete.go
```diff
@@ -3,44 +3,52 @@ package main
...
-		Action:    commandAction(opts.run),
-		Flags:     append(sharedFlags, imageFlags...),
+See skopeo(1) section "IMAGE NAMES" for the expected format
+`, strings.Join(transports.ListNames(), ", "))
+	)
+	deleteCommand := &cobra.Command{
+		Use:   "delete [command options] IMAGE-NAME",
+		Short: "Delete image IMAGE-NAME",
+		Long:  deleteDescription,
+		RunE: func(cmd *cobra.Command, args []string) error {
```
mtrmac: Please preserve the `commandAction` helper (possibly renaming it) to shield the actions from a `cobra.Command`.

cmd/skopeo/delete.go
```diff
@@ -3,44 +3,52 @@ package main
...
+			return deleteCmd(cmd, args, opts)
+		},
+		Example: `skopeo delete docker://registry.example.com/example/pause:latest`,
 	}
+	deleteCommand.SetUsageTemplate(usageTemplate())
+	deleteCommand.DisableFlagsInUseLine = true
```
mtrmac: If these options have to be set for every `Command`, it would be nice to move that into a helper instead of copy&pasting everywhere.

cmd/skopeo/delete.go
```diff
@@ -3,44 +3,52 @@ package main
...
+	imageFlags, imageOpts := imageFlags(globalFlagResults, sharedOpts, "", "")
+	opts.image = imageOpts
+	opts.global = globalFlagResults
+	flags.AddFlagSet(&sharedFlags)
+	flags.AddFlagSet(&imageFlags)
+	return deleteCommand
 }
 
-func (opts *deleteOptions) run(args []string, stdout io.Writer) error {
+func deleteCmd(c *cobra.Command, args []string, opts deleteOptions) error {
```
mtrmac: Why was this method turned into a global? I don’t think we need all those names separately in the package namespace. The `deleteCmd`/`deleteCommand` distinction is rather awkward, as well as calling this function `deleteCmd` (which suggests “delete a command” much more than “perform the delete operation”).

cmd/skopeo/inspect.go
```diff
@@ -39,42 +38,42 @@ type inspectOptions struct {
...
+	opts.global = globalFlagResults
+	opts.image = imageOpts
+	flags.AddFlagSet(&sharedFlags)
+	flags.AddFlagSet(&imageFlags)
+	return inspectCommand
 }
 
-func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error) {
+func inspectCmd(c *cobra.Command, args []string, opts inspectOptions) (retErr error) {
+	stdout := c.OutOrStdout()
```
mtrmac: This should be done in `commandAction` or something like that, not something every single command should have to worry about. (See the other comments about `commandAction` as well.)

cmd/skopeo/delete.go
```diff
@@ -3,44 +3,52 @@ package main
...
 
-	See skopeo(1) section "IMAGE NAMES" for the expected format
-	`, strings.Join(transports.ListNames(), ", ")),
-		ArgsUsage: "IMAGE-NAME",
-		Action:    commandAction(opts.run),
-		Flags:     append(sharedFlags, imageFlags...),
+See skopeo(1) section "IMAGE NAMES" for the expected format
+`, strings.Join(transports.ListNames(), ", "))
+	)
+	deleteCommand := &cobra.Command{
```
mtrmac: (Absolutely non-blocking: This is fine, but given the context, just `cmd` would work just as well. Especially the “delete” qualifier is strongly implied by context.)

cmd/skopeo/list_tags.go
```diff
@@ -130,7 +137,7 @@ func (opts *tagsOptions) run(args []string, stdout io.Writer) (retErr error) {
 	if err != nil {
 		return err
 	}
-	_, err = fmt.Fprintf(stdout, "%s\n", string(out))
+	_, err = fmt.Fprintf(os.Stdout, "%s\n", string(out))
```
mtrmac: `OutOrStdout`, via a shared helper, please.

cmd/skopeo/list_tags_test.go
```diff
@@ -1,9 +1,10 @@
 package main
 
 import (
+	"testing"
+
 	"github.com/containers/image/v5/transports/alltransports"
 	"github.com/stretchr/testify/assert"
-	"testing"
```
mtrmac: This seems unidiomatic.

cmd/skopeo/manifest.go
```diff
@@ -3,27 +3,31 @@ package main
...
 }
 
-func manifestDigestCmd() cli.Command {
-	opts := manifestDigestOptions{}
-	return cli.Command{
-		Name:      "manifest-digest",
-		Usage:     "Compute a manifest digest of a file",
-		ArgsUsage: "MANIFEST",
-		Action:    commandAction(opts.run),
+func manifestCommand(globalFlagResults *globalOptions) *cobra.Command {
```
mtrmac: The command is `manifest-digest`, not `manifest`.

cmd/skopeo/signing.go
```diff
@@ -4,35 +4,67 @@ import (
...
-		Action:    commandAction(opts.run),
-		Flags: []cli.Flag{
-			cli.StringFlag{
-				Name:        "output, o",
-				Usage:       "output the signature to `SIGNATURE`",
-				Destination: &opts.output,
-			},
+func standaloneSignCommand(globalFlagResults *globalOptions) *cobra.Command {
+	var standaloneSignOpts standaloneSignOptions
+	standaloneSignCommand := &cobra.Command{
```
mtrmac: The `standaloneSign` qualifiers in local variable names are such a large proportion of the text that it just makes telling the variables apart and reading the code difficult for no benefit.

cmd/skopeo/signing.go
```diff
@@ -4,35 +4,67 @@ import (
...
+	standaloneSignCommand.DisableFlagsInUseLine = true
+
+	standaloneSignflags := standaloneSignCommand.Flags()
+	standaloneSignflags.StringVarP(&standaloneSignOpts.output, "output", "o", "", "output the signature to `SIGNATURE`")
+	return standaloneSignCommand
+
 }
 
-func (opts *standaloneSignOptions) run(args []string, stdout io.Writer) error {
+func standaloneVerifyCommand(globalFlagResults *globalOptions) *cobra.Command {
```
mtrmac: Please keep the file organized by command, so that definition of flags and their use is close together.

cmd/skopeo/delete.go
QiWang19: it will be initialized in the flag definition, other fields remain zero value.

QiWang19:
@mtrmac PTAL
I haven't found the method to make OptionalBool work as bool to fix the test. 
Can it be converted to the Flag.BoolVar , like the code in podman or buildah?
```
if c.IsSet("tls-verify") {
		dockerRegistryOptions.DockerInsecureSkipTLSVerify = types.NewOptionalBool(!c.TlsVerify)
	}
```

QiWang19:
start #865, keeping optional bool and hiding `cobra.Command` behind the `RunE` seems can not make much code shareable

cmd/skopeo/delete.go
mtrmac: Sure, it is functionally equivalent; my point is that the period of time where the pointers are uninitialized is introduced unnecessarily — extra work to make the code less readable.

cmd/skopeo/utils.go
```diff
@@ -75,73 +72,43 @@ type imageOptions struct {
...
-		},
-	)
-	return flags, &opts
+	if credsOptionAlias == "" {
+		fs.Var(newOptionalStringValue(&flags.credsOption), flagPrefix+"creds", "Use `USERNAME[:PASSWORD]` for accessing the registry")
+	} else {
+		fs.Var(newOptionalStringValue(&flags.credsOption), flagPrefix+"creds", "Use `USERNAME[:PASSWORD]` for accessing the registry")
+		fs.Var(newOptionalStringValue(&flags.credsOption), credsOptionAlias, "Use `USERNAME[:PASSWORD]` for accessing the registry")
+		if err := fs.MarkHidden(credsOptionAlias); err != nil {
+			panic(fmt.Sprintf("error marking %s flag as hidden: %v", credsOptionAlias, err))
```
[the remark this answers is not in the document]
mtrmac: The cleanest way to do this is probably to use
```go
flag := vs.VarPF(…)
flag.Hidden = true
```
with no unnecessary lookup, and no error path.

mtrmac:
> @mtrmac PTAL
> I haven't found the method to make OptionalBool work as bool to fix the test.
> Can it be converted to the Flag.BoolVar , like the code in podman or buildah?
>
> ```go
> if c.IsSet("tls-verify") {
> 		dockerRegistryOptions.DockerInsecureSkipTLSVerify = types.NewOptionalBool(!c.TlsVerify)
> 	}
> ```

I don’t think that would actually be good enough: the `TestOptionalBoolIsBoolFlag` flag verifies the behavior of `optionalBool` and the test failing indicates that the value of the `--tls-verify` option would sometimes be treated as `false` when the previous behavior was `true`. That would be a security vulnerability, and a complete blocker to this PR. (_That_, in turn would not quite be a disaster: Podman could call a Skopeo login implementation by filling the `loginOptions` using its own CLI parser. But hopefully we won’t need that.)

Looking at the code, the test can be made to work by using
```go
flag.NoOptDefVal = "true"
```
see (WIP and _very_ unclean) https://github.com/mtrmac/skopeo/tree/corba .

OTOH:
- This _must_ be done in a way that can’t be forgotten by the caller, absolutely not at every call site; i.e. instead of `newOptionalBoolValue`, probably an `optionalBoolFlag(*pflag.FlagSet, …)`.
- Manually setting `NoOptDefVal` is something that can be done in the public API, but it’s pretty ugly and obscure. Still to investigate: using `PFlagFromGoFlag` to let the library do that.

mtrmac:
> > @mtrmac PTAL
> > I haven't found the method to make OptionalBool work as bool to fix the test.
> > Can it be converted to the Flag.BoolVar , like the code in podman or buildah?
> > ```go
> > if c.IsSet("tls-verify") {
> > 		dockerRegistryOptions.DockerInsecureSkipTLSVerify = types.NewOptionalBool(!c.TlsVerify)
> > 	}
> > ```
> 
> I don’t think that would actually be good enough

… thinking about it more, `flag.BoolVar` does set `NoOptDefVal = "true"` internally, so that could actually work. Still, `"tls-verify"` is _exactly_ the reason the string lookups in command implementations are unwanted (https://github.com/containers/skopeo/pull/523 ).

QiWang19:
@mtrmac PTAL. Use `optionalBoolFlag(*pflag.flagSet..)`. works for tests

QiWang19:
urfave/cli flag `--generate-bash-completion` used in `completions/bash/skopeo` would not work. Does the whole completions/bash/skopeo need rewrite?

mtrmac:
> Does the whole completions/bash/skopeo need rewrite?

Probably — OTOH Cobra has its own completion generation support, and it seems to be fairly rich in features. So the shell script might end up being much simpler.

QiWang19:
@mtrmac PTAL

QiWang19:
@mtrmac PTAL. can this PR get merged?

rhatdan:
@mtrmac @vrothberg PTAL

mtrmac:
(I apologize for the delay; I’ll review this today unless something extremely urgent comes up.)

go.mod
```diff
@@ -17,10 +17,12 @@ require (
 	github.com/opencontainers/runtime-spec v1.0.0 // indirect
 	github.com/pkg/errors v0.9.1
 	github.com/russross/blackfriday v2.0.0+incompatible // indirect
+	github.com/shurcooL/sanitized_anchor_name v1.0.0 // indirect
```
mtrmac, line 20: (This is probably no longer used, per `go mod why -m`)

cmd/skopeo/signing.go
```diff
@@ -8,28 +8,25 @@ import (
...
-				Usage:       "output the signature to `SIGNATURE`",
-				Destination: &opts.output,
-			},
-		},
-	}
+	cmd := &cobra.Command{
+		Use:     "standalone-sign [command options] MANIFEST DOCKER-REFERENCE KEY-FINGERPRINT",
+		Short:   "Create a signature using local files",
+		RunE:    commandAction(opts.run),
+		Example: `skopeo standalone-sign busybox-manifest.json registry.example.com/example/busybox `,
```
mtrmac, line 24: This example is missing the `KEY-FINGERPRINT`; but on balance I’d probably prefer not giving an example, `standalone-sign` should not be encouraged.

cmd/skopeo/signing.go
```diff
@@ -64,14 +61,16 @@ func (opts *standaloneSignOptions) run(args []string, stdout io.Writer) error {
...
-		Name:      "standalone-verify",
-		Usage:     "Verify a signature using local files",
-		ArgsUsage: "MANIFEST DOCKER-REFERENCE KEY-FINGERPRINT SIGNATURE",
-		Action:    commandAction(opts.run),
-	}
+	cmd := &cobra.Command{
+		Use:     "standalone-verify MANIFEST DOCKER-REFERENCE KEY-FINGERPRINT SIGNATURE",
+		Short:   "Verify a signature using local files",
+		RunE:    commandAction(opts.run),
+		Example: `skopeo standalone-verify busybox-manifest.json registry.example.com/example/busybox`,
```
mtrmac, line 70: This example is missing the `KEY-FINGERPRINT`; but on balance I’d probably prefer not giving an example, `standalone-verify` should be actively discouraged for most users.

cmd/skopeo/list_tags_test.go
mtrmac: I’m sorry, I don’t know what I was thinking. This is fine, of course.

QiWang19:
@mtrmac PTAL

cmd/skopeo/flag.go
```diff
@@ -18,7 +18,14 @@ type optionalBool struct {
 // newOptionalBool
```
mtrmac, line 18: This should point to `optionalBoolFlag`, not `newOptionalBool`, now.

cmd/skopeo/flag.go
```diff
@@ -18,7 +18,14 @@ type optionalBool struct {
...
 type optionalBoolValue optionalBool
 
-func newOptionalBoolValue(p *optionalBool) cli.Generic {
+func optionalBoolFlag(fs *pflag.FlagSet, p *optionalBool, name, usage string) {
+	if ret := fs.Lookup(name); ret == nil {
```
mtrmac, line 22: Why the `Lookup` here? The Cobra implementations don’t do that, and AFAICS tests pass fine without it.

cmd/skopeo/flag_test.go
```diff
@@ -152,30 +146,28 @@ func TestOptionalStringSet(t *testing.T) {
...
-	app := cli.NewApp()
-	app.EnableBashCompletion = true
-	app.Flags = []cli.Flag{
-		cli.GenericFlag{
-			Name:  "global-OS",
-			Value: newOptionalStringValue(&globalOS),
+
+	app := &cobra.Command{
+		Use: "app",
+		RunE: func(cmd *cobra.Command, args []string) error {
```
mtrmac, line 152: Setting `RunE ` seems unnecessary.

cmd/skopeo/flag_test.go
```diff
@@ -152,30 +146,28 @@ func TestOptionalStringSet(t *testing.T) {
...
-		Name: "cmd",
-		Flags: []cli.Flag{
-			cli.GenericFlag{
-				Name:  "command-OS",
-				Value: newOptionalStringValue(&commandOS),
-			},
-		},
-		Action: func(*cli.Context) error {
+
+	app.Flags().Var(newOptionalStringValue(&globalOS), "global-OS", "")
```
mtrmac, line 157: This should be `PersistentFlags`.

cmd/skopeo/flag_test.go
```diff
@@ -152,30 +146,28 @@ func TestOptionalStringSet(t *testing.T) {
...
-	app.Commands = []cli.Command{{
-		Name: "cmd",
-		Flags: []cli.Flag{
-			cli.GenericFlag{
-				Name:  "command-OS",
-				Value: newOptionalStringValue(&commandOS),
-			},
-		},
-		Action: func(*cli.Context) error {
+
```
mtrmac, line 156: (Non-blocking: If there should be an empty line here (I don’t think there has to be one), it should be one line below, separating the “global” `app` and `globalOS`, and the sub-command `cmd` and `commandOS`.)

cmd/skopeo/flag_test.go
```diff
@@ -216,23 +208,26 @@ func TestOptionalStringIsBoolFlag(t *testing.T) {
...
-			Name: "cmd",
-			Flags: []cli.Flag{
-				cli.GenericFlag{
-					Name:  "OS",
-					Value: newOptionalStringValue(&os),
-				},
+
+		app := &cobra.Command{
+			Use: "app",
+			RunE: func(cmd *cobra.Command, args []string) error {
```
mtrmac, line 214: Setting `RunE` seems unnecessary.

cmd/skopeo/main.go
```diff
@@ -31,96 +30,64 @@ type globalOptions struct {
...
+	if err := rootCommand.PersistentFlags().MarkHidden("registries-conf"); err != nil {
+		logrus.Error("unable to mark registries-conf flag as hidden")
+	}
+	rootCommand.PersistentFlags().StringVar(&opts.tmpDir, "tmpdir", "", "directory used to store temporary files")
+	rootCommand.AddCommand(
 		inspectCmd(&opts),
 		layersCmd(&opts),
-		tagsCmd(&opts),
+		copyCmd(&opts),
+		deleteCmd(&opts),
```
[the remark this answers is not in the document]
QiWang19: I should keep it alphabetical order.

QiWang19:
@mtrmac I cleaned the reviews. PTAL

QiWang19:
@mtrmac PTAL

QiWang19:
PTAL. Can this get merged or more fixes needed?

PR #862, opened 2020-03-19
Use fully-qualified image names

edsantiago:
Replace shortnames with FQINs; this should allow tests to
run regardless of the state of registries.conf.

And, fix one broken new test that invoked 'jq' (without dot).
This usage works in Fedora, but not in RHEL.

### DIFF
--- a/systemtest/020-copy.bats
+++ b/systemtest/020-copy.bats
@@ -14,7 +14,7 @@ function setup() {
 # From remote, to dir1, to local, to dir2;
 # compare dir1 and dir2, expect no changes
 @test "copy: dir, round trip" {
-    local remote_image=docker://busybox:latest
+    local remote_image=docker://docker.io/library/busybox:latest
     local localimg=docker://localhost:5000/busybox:unsigned
 
     local dir1=$TESTDIR/dir1
@@ -30,7 +30,7 @@ function setup() {
 
 # Same as above, but using 'oci:' instead of 'dir:' and with a :latest tag
 @test "copy: oci, round trip" {
-    local remote_image=docker://busybox:latest
+    local remote_image=docker://docker.io/library/busybox:latest
     local localimg=docker://localhost:5000/busybox:unsigned
 
     local dir1=$TESTDIR/oci1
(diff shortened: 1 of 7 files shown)


systemtest/020-copy.bats
```diff
@@ -14,7 +14,7 @@ function setup() {
 # From remote, to dir1, to local, to dir2;
 # compare dir1 and dir2, expect no changes
 @test "copy: dir, round trip" {
-    local remote_image=docker://busybox:latest
+    local remote_image=docker://docker.io/library/busybox:latest
```
mtrmac: (None of the `docker://` edits should be necessary: in the `docker://` syntax, short names always mean `docker.io`. But they don’t hurt at all and it’s nice to get in the habit.

Or is there a case where the `docker://` syntax triggers a search? I’d very much like to know about it, and get rid of it.)

mtrmac:
(FWIW I keep a few more edits to these tests to make them work on CentOS **7** in https://github.com/containers/skopeo/pull/46 . That PR is probably going to stay WIP for a while, are those individual parts useful anyway?)

edsantiago:
@mtrmac the `jq .` one should no longer be necessary after this merges. The `IFS` I find puzzling but can't look at in depth right now. My first guess is that your system has a too-old version of BATS? Are you using `>= 1.0` ?

mtrmac:
> @mtrmac the `jq .` one should no longer be necessary after this merges.

Yes.

> The `IFS` I find puzzling but can't look at in depth right now. My first guess is that your system has a too-old version of BATS? Are you using `>= 1.0` ?

Yes, CentOS has (had?) a rather old version, this works around a bug that has long been fixed.

> CentOS, as of bats-0.4.0-1.20141016git3b33a5a.el7.noarch, misses the fix for sstephenson/bats#89 , causing (read) not to split words at white space.

mtrmac:
I was just pointing out the other change in case it were relevant to your other work. If this is only targeting 8, it’s probably not necessary.

systemtest/020-copy.bats
edsantiago: Thank you! This ("short names always mean docker.io") explains why I couldn't get the tests to fail using a doctored `registries.conf` without docker.io! I spent a long time wondering, then finally decided to go with FQINs anyway. 

So: no, don't worry, there is no case (that I've seen) where `docker://` triggers a registry search. I was just trying to get the tests using full names consistently. Thanks again!

PR #865, opened 2020-03-26
Add skopeo Login from c/common

QiWang19:
Used c/common/pkg/auth for skopeo login
Based on PR #857


QiWang19:
@rhatdan PTAL 
This PR is on top of #857, but it's still under review.

rhatdan:
Lets move the shared code to github.com/containers/common, And not vendor buildah into skopeo.

QiWang19:
How much code should be shared i c/common? If skopeo keeps `OptionalBool`, podman and buildah CLI cannot fully fit in skopeo. They can share a subset, and skopeo should defined its own cli for `Optional..` flags.

mtrmac:
(This should have been the first design concern before writing code I think.)

We can’t trivially share global code anyway, due to things like `--debug` vs `--log-level=debug`. (Well, we _could_ work to make them all the same. But that’s non-trivial and a fair bit of work.)

My first guess would be, in Skopeo terms, to make the `loginOptions`, and `loginOptions.run` public, and let every application configure its CLI parsing. That would make all flags consistent within projects, but different across all the `login`s, and require a fair bit of code.

Alternatively, make `globalOptions` and `func loginCommand(*globalOptions) *cobra.Command` public, let every application configure CLI parsing for the global options, make the `login` options consistent between projects (but different from other subomcmands in that project).

mtrmac:
BTW making the “optional bool” flag type public and used elsewhere would make sense to me, if we can make it work with Cobra nicely.

rhatdan:
All of the handling of certs an plugins can be  isolated from the CLI Parsing can't it?

mtrmac:
> Well if we need OptionalBool for Skopeo, I do not know why we don't need it for Buidlah And Podman.

`optionalBool` and forbidding string-based parameter matching was an attempt to learn from https://github.com/containers/skopeo/pull/523 and make that kind of vulnerability impossible. Buildah/Podman could simply not make those kind of mistakes, I guess :)

---

> All of the handling of certs an plugins can be isolated from the CLI Parsing can't it?

In principle, but that kind of depends on the adjustments that are not “CLI parsing”, e.g. Podman overriding the defaults in “rootless” mode and creating the directories in advance. Those kinds of activities are not “CLI parsing” but it must be possible to inject them into the operation at some point (or, well, they could be removed and pushed down into the backend libraries, as has been somewhat happening recently).

QiWang19:
@mtrmac @rhatdan PTAL. The skopeo uses public CLI definition from c/common/auth, keep its own cli parsing `newSystemcontext`

cmd/skopeo/login.go
```diff
@@ -0,0 +1,76 @@
...
+		Use:     "login",
+		Short:   "Login to a container registry",
+		Long:    "Login to a container registry on a specified server.",
+		RunE:    commandAction(opts.run),
+		Example: `skopeo login quay.io`,
+	}
+	usageHelper(cmd)
+	flags := cmd.Flags()
+	optionalBoolFlag(flags, newOptionalBoolValue(&opts.getLogin), "get-login", "Return the current login user for the registry")
+	optionalBoolFlag(flags, newOptionalBoolValue(&opts.tlsVerify), "tls-verify", "Return the current login user for the registry")
```
mtrmac, line 36: (This is probably for the other PR: Now that there is an `optionalBoolFlag` helper, it should call `newOptionalBoolValue` internally.)

mtrmac, line 36: The option description needs updating.

cmd/skopeo/login.go
```diff
@@ -0,0 +1,76 @@
...
+	if err != nil {
+		return err
+	}
+	sys.DockerCertPath = opts.loginOpts.CertDir
+	sys.AuthFilePath = opts.loginOpts.Authfile
+	if opts.tlsVerify.present {
+		sys.DockerDaemonInsecureSkipTLSVerify = !opts.tlsVerify.value
+		sys.DockerInsecureSkipTLSVerify = types.NewOptionalBool(!opts.tlsVerify.value)
+	}
+	server := auth.GetRegistryName(args[0])
```
mtrmac, line 60: Shouldn’t the following ~15 lines be a part of the shared code as well?

cmd/skopeo/login.go
```diff
@@ -0,0 +1,76 @@
...
+	authConfig, err := config.GetCredentials(sys, server)
+	if err != nil {
+		return errors.Wrapf(err, "error reading auth file")
+	}
+	if opts.getLogin.present {
+		if authConfig.Username == "" {
+			return errors.Errorf("not logged into %s", server)
+		}
+		fmt.Printf("%s\n", authConfig.Username)
```
mtrmac, line 69: This should go to the `stdout` parameter.

cmd/skopeo/login.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/containers/common/pkg/auth"
+	"github.com/containers/image/v5/pkg/docker/config"
+	"github.com/containers/image/v5/types"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+type loginOptions struct {
+	global    *globalOptions
+	loginOpts *auth.LoginOptions
```
mtrmac, line 16: If `loginOpts` is always created, it can be an in-line (non-pointer) member.

cmd/skopeo/logout.go
```diff
@@ -0,0 +1,51 @@
...
+	"io"
+
+	"github.com/containers/common/pkg/auth"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+type logoutOptions struct {
+	global     *globalOptions
+	logoutOpts *auth.LogoutOptions
```
mtrmac, line 13: If `logoutOpts` is always created, it can be an in-line (non-pointer) member.

cmd/skopeo/login.go
```diff
@@ -0,0 +1,76 @@
...
+	ctx, cancel := opts.global.commandTimeoutContext()
+	defer cancel()
+	sys, err := opts.global.newSystemContext()
+	if err != nil {
+		return err
+	}
+	sys.DockerCertPath = opts.loginOpts.CertDir
+	sys.AuthFilePath = opts.loginOpts.Authfile
+	if opts.tlsVerify.present {
+		sys.DockerDaemonInsecureSkipTLSVerify = !opts.tlsVerify.value
```
mtrmac, line 57: `DockerDaemon*` makes no sense here.

cmd/skopeo/logout.go
```diff
@@ -0,0 +1,51 @@
...
+
+	var server string
+	if len(args) == 1 {
+		server = auth.GetRegistryName(args[0])
+	}
+	sys, err := opts.global.newSystemContext()
+	if err != nil {
+		return err
+	}
+	sys.AuthFilePath = opts.logoutOpts.Authfile
```
mtrmac, line 49: Why is the set of options in `login` and here so different?

cmd/skopeo/utils.go
```diff
@@ -202,6 +202,21 @@ func (opts *imageDestOptions) newSystemContext() (*types.SystemContext, error) {
 	return ctx, err
 }
 
+func (opts *globalOptions) newSystemContext() (*types.SystemContext, error) {
```
mtrmac, line 205: - Due to the security impact, this should have tests
- `imageOptions.newSystemContext()` should call this to ensure the two don’t drift apart from each other.

vendor/github.com/containers/common/pkg/auth/cli.go
```diff
@@ -0,0 +1,41 @@
...
+
+// GetLoginFlags defines and returns login flags for containers tools
+func GetLoginFlags(flags *LoginOptions) *pflag.FlagSet {
+	fs := pflag.FlagSet{}
+	fs.StringVar(&flags.Authfile, "authfile", GetDefaultAuthFile(), "path of the authentication file. Use REGISTRY_AUTH_FILE environment variable to override")
+	fs.StringVar(&flags.CertDir, "cert-dir", "", "use certificates at the specified path to access the registry")
+	fs.StringVarP(&flags.Password, "password", "p", "", "Password for registry")
+	fs.StringVarP(&flags.Username, "username", "u", "", "Username for registry")
+	fs.BoolVar(&flags.StdinPassword, "password-stdin", false, "Take the password from stdin")
+	fs.BoolVar(&flags.TlsVerify, "tls-verify", false, "require HTTPS and verify certificates when accessing the registry")
```
mtrmac, line 30: ⚠️ The default **must not** be, in effect, `false`. Even if this did the right thing, there should be tests that this is always interpreted correctly.

It is really possible to have the same `tis-verify` flag defined twice?!

Apparently the “second” definition, whichever that is, is silently ignored by `AddFlagSet`.

That’s just too unreliable for a security-related flag.

vendor/github.com/containers/common/pkg/auth/cli.go
```diff
@@ -0,0 +1,41 @@
...
+	Username      string
+	StdinPassword bool
+	TlsVerify     bool
+	GetLogin      bool
+}
+
+// LogoutOptions represents the results for flags in logout
+type LogoutOptions struct {
+	Authfile string
+	All      bool
```
mtrmac, line 19: Why does `logout` have so much fewer options than `login`?

vendor/github.com/containers/common/pkg/auth/cli.go
```diff
@@ -0,0 +1,41 @@
...
+
+// LoginOptions represents the results for flags in login
+type LoginOptions struct {
+	Authfile      string
+	CertDir       string
+	Password      string
+	Username      string
+	StdinPassword bool
+	TlsVerify     bool
+	GetLogin      bool
```
mtrmac, line 13: If the distinction needs to exist at all, this should at the very least clearly document which options are handled by this subpackage and which need to be handled by the caller.

vendor/github.com/containers/common/pkg/auth/auth.go
```diff
@@ -0,0 +1,167 @@
...
+
+	"github.com/containers/image/v5/docker"
+	"github.com/containers/image/v5/pkg/docker/config"
+	"github.com/containers/image/v5/types"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/crypto/ssh/terminal"
+)
+
+func GetDefaultAuthFile() string {
```
mtrmac, line 18: Does this need to be public? If it does, it needs documentation.

vendor/github.com/containers/common/pkg/auth/auth.go
```diff
@@ -0,0 +1,167 @@
...
+	return os.Getenv("REGISTRY_AUTH_FILE")
+}
+
+func CheckAuthFile(authfile string) error {
```
mtrmac, line 22: Does this need to be public? If it does, it needs documentation.

vendor/github.com/containers/common/pkg/auth/auth.go
```diff
@@ -0,0 +1,167 @@
...
+		return nil
+	}
+	if _, err := os.Stat(authfile); err != nil {
+		return errors.Wrapf(err, "error checking authfile path %s", authfile)
+	}
+	return nil
+}
+
+// GetCredsAndLogin login to the server with creds from Stdin or CLI
+func GetCredsAndLogin(ctx context.Context, systemContext *types.SystemContext, iopts *LoginOptions, server, userFromAuthFile, passFromAuthFile string) error {
```
mtrmac, line 33: `os.Stdin`/`os.Stdout` should be parameters (in `LoginOptions` probably) instead of hard-coded, so that this can be unit-tested.

mtrmac, line 33: (Non-blocking: Not sure about the name, when this is not returning any credentials.)

mtrmac, line 33: See the comment in the caller about more code to move here.

vendor/github.com/containers/common/pkg/auth/auth.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	if unauthorized, ok := err.(docker.ErrUnauthorizedForCredentials); ok {
+		logrus.Debugf("error logging into %q: %v", server, unauthorized)
+		return errors.Errorf("error logging into %q: invalid username/password", server)
+	}
+	return errors.Wrapf(err, "error authenticating creds for %q", server)
+}
+
+// GetRegistryName scrubs and parses the input to get the server name
+func GetRegistryName(server string) string {
```
mtrmac, line 83: Does this need to be public?

vendor/github.com/containers/common/pkg/auth/auth.go
```diff
@@ -0,0 +1,167 @@
...
+	return RegistryFromFullName(ScrubServer(server))
+}
+
+// ScrubServer removes 'http://' or 'https://' from the front of the
+// server/registry string if either is there.  This will be mostly used
+// for user input from 'buildah login' and 'buildah logout'.
+func ScrubServer(server string) string {
```
mtrmac, line 90: Does this need to be public?

vendor/github.com/containers/common/pkg/auth/auth.go
```diff
@@ -0,0 +1,167 @@
...
+	server = strings.TrimPrefix(server, "https://")
+	return strings.TrimPrefix(server, "http://")
+}
+
+// RegistryFromFullName gets the registry from the input. If the input is of the form
+// quay.io/myuser/myimage, it will parse it and just return quay.io
+// It also returns true if a full image name was given
+func RegistryFromFullName(input string) string {
```
mtrmac, line 98: Does this need to be public?

vendor/github.com/containers/common/pkg/auth/auth.go
```diff
@@ -0,0 +1,167 @@
...
+// server/registry string if either is there.  This will be mostly used
+// for user input from 'buildah login' and 'buildah logout'.
+func ScrubServer(server string) string {
+	server = strings.TrimPrefix(server, "https://")
+	return strings.TrimPrefix(server, "http://")
+}
+
+// RegistryFromFullName gets the registry from the input. If the input is of the form
+// quay.io/myuser/myimage, it will parse it and just return quay.io
+// It also returns true if a full image name was given
```
mtrmac, line 97: This line is incorrect.

vendor/github.com/containers/common/pkg/auth/auth.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// GetRegistryName scrubs and parses the input to get the server name
+func GetRegistryName(server string) string {
+	return RegistryFromFullName(ScrubServer(server))
+}
+
+// ScrubServer removes 'http://' or 'https://' from the front of the
+// server/registry string if either is there.  This will be mostly used
+// for user input from 'buildah login' and 'buildah logout'.
```
mtrmac, line 89: “Buildah”

vendor/github.com/containers/common/pkg/auth/auth.go
```diff
@@ -0,0 +1,167 @@
...
+		}
+		password = string(pass)
+		fmt.Println()
+	}
+	return strings.TrimSpace(username), password, err
+}
+
+// LogoutRemovesAuth removes the authentication of server from authfile
+// removes all authtication if specifies all in the options
+func LogoutRemovesAuth(systemContext *types.SystemContext, iopts *LogoutOptions, server string) error {
```
mtrmac, line 144: “removes” reads like a predicate returning true/false to me.

mtrmac, line 144: `os.Stdin`/`os.Stdout` should be parameters (in `LogoutOptions` probably) instead of hard-coded, so that this can be unit-tested.

cmd/skopeo/login.go
QiWang19: podman/buildah have their own parsing CLI functions `SystemContextFromOptions` till use `c.Flag("tls-verify").changed`
These code can be shared if fix of `optionalBoolFlag` in an API and they use optionalBool?

cmd/skopeo/logout.go
QiWang19: These options are the current design of podman/buildah logout

mtrmac: Oh, right; logout does not do anything over the network, so the TLS-related options are irrelevant.

cmd/skopeo/login.go
mtrmac: Fair enough about the CLI parsing distinction (although it would be good to share that as well, if possible), but I was talking about this line and the following ones. Nothing in the code using `GetRegistryName`/`GetCredentials`/`IdentityToken` is directly CLI-dependent AFAICS.

vendor/github.com/containers/common/pkg/auth/cli.go
mtrmac: … because it doesn’t do TLS, so it doesn’t need them.

rhatdan:
@QiWang19 Needs a rebase.

QiWang19:
can the tests for login/logout use the registry set up in integration/registry.go?

mtrmac:
> can the tests for login/logout use the registry set up in integration/registry.go?

Sure; check_test.go and other parts already set up several registries that can maybe just be reused.

QiWang19:
@mtrmac PTAL

cmd/skopeo/login.go
```diff
@@ -0,0 +1,54 @@
...
+	getLogin  optionalBool
+	tlsVerify optionalBool
+}
+
+func loginCmd(global *globalOptions) *cobra.Command {
+	opts := loginOptions{
+		global: global,
+		loginOpts: auth.LoginOptions{
+			Stdin:  os.Stdin,
+			Stdout: os.Stdout,
```
mtrmac, line 24: This should only be set from the `run` method’s `stdout` parameter.

(Eventually the same for `Stdin`, but that can wait until we have tests that actually use it. But `Stdin` should at least be set in the same place, i.e. inside `run`.)

cmd/skopeo/login.go
mtrmac: Still outstanding.

cmd/skopeo/utils_test.go
```diff
@@ -110,6 +110,29 @@ func TestImageOptionsNewSystemContext(t *testing.T) {
...
+		"--override-variant", "overridden-variant",
+		"--tmpdir", "/srv",
+	})
+	res = opts.newSystemContext()
+	assert.Equal(t, &types.SystemContext{
+		RegistriesDirPath:    "/srv/registries.d",
+		ArchitectureChoice:   "overridden-arch",
+		OSChoice:             "overridden-os",
+		VariantChoice:        "overridden-variant",
+		BigFilesTemporaryDir: "/srv",
```
mtrmac, line 132: Tests for `SystemRegistriesConfPath` and `DockerInsecureSkipTLSVerify` are missing.

cmd/skopeo/utils.go
```diff
@@ -200,6 +192,22 @@ func (opts *imageDestOptions) newSystemContext() (*types.SystemContext, error) {
...
+func (opts *globalOptions) newSystemContext() *types.SystemContext {
+	ctx := &types.SystemContext{
+		RegistriesDirPath:        opts.registriesDirPath,
+		ArchitectureChoice:       opts.overrideArch,
+		OSChoice:                 opts.overrideOS,
+		VariantChoice:            opts.overrideVariant,
+		SystemRegistriesConfPath: opts.registriesConfPath,
+		BigFilesTemporaryDir:     opts.tmpDir,
+	}
+	// DEPRECATED: We support this for backward compatibility, but override it if a per-image flag is provided.
```
mtrmac, line 204: This should be removed from `imageOptions.newSystemContext` now (and I guess replaced with a comment that explains that the `imageOptions` option overrides the `global` one if both are present.

cmd/skopeo/login.go
```diff
@@ -0,0 +1,54 @@
...
+	optionalBoolFlag(flags, &opts.tlsVerify, "tls-verify", "require HTTPS and verify certificates when accessing the registry")
+	flags.AddFlagSet(auth.GetLoginFlags(&opts.loginOpts))
+	return cmd
+}
+
+func (opts *loginOptions) run(args []string, stdout io.Writer) error {
+	ctx, cancel := opts.global.commandTimeoutContext()
+	defer cancel()
+	sys := opts.global.newSystemContext()
+	sys.DockerCertPath = opts.loginOpts.CertDir
```
mtrmac, line 46: Nothing in c/common documents that the caller is responsible for handling this. Sigh, the `LoginOpts` members are used in way too many different ways.

cmd/skopeo/login.go
```diff
@@ -0,0 +1,54 @@
...
+	sys.AuthFilePath = opts.loginOpts.AuthFile
```
mtrmac, line 47: … and the `loginOpts.AuthFile` value is used _both_ directly and from `sys` by `auth.Login`, with unexpected results if they are inconsistent.

cmd/skopeo/login.go
```diff
@@ -0,0 +1,54 @@
...
+	if opts.tlsVerify.present {
+		sys.DockerDaemonInsecureSkipTLSVerify = !opts.tlsVerify.value
+		sys.DockerInsecureSkipTLSVerify = types.NewOptionalBool(!opts.tlsVerify.value)
+	}
+	opts.loginOpts.GetLoginSet = opts.getLogin.present
```
mtrmac, line 52: This sets `GetLoginSet` to `true` if the user uses `--get-login=false`.

Was there a reason to treat `--get-login` and `--password-stdin` differently?

cmd/skopeo/utils.go
```diff
@@ -200,6 +192,22 @@ func (opts *imageDestOptions) newSystemContext() (*types.SystemContext, error) {
 	return ctx, err
 }
 
+func (opts *globalOptions) newSystemContext() *types.SystemContext {
```
mtrmac, line 195: This should have documentation, at least the “guaranteed to return a fresh instance, so it is safe…” part.

mtrmac, line 195: It’s inconvenient to have the `globalOptions` definition/use split between `main.go` and `utils.go` like this. 

The flag definitions could be split from `createApp`, but maybe it would be simplest to just move this function to `main.go`, beside the existing `globalOptions.getPolicyContext` and the like — and the same for the new test.

completions/bash/skopeo
```diff
@@ -158,6 +158,33 @@ _skopeo_list_repository_tags() {
     _complete_ "$options_with_args" "$boolean_options"
 }
 
+_skopeo_login() {
+    local options_with_args="
+    --authfile
+    --cert-dir
+    --password -p
+    --username -u
+    --password-stdin
```
mtrmac, line 167: `--password-stdin` is a boolean option.

docs/skopeo-login.1.md
```diff
@@ -0,0 +1,101 @@
...
+environment variable. `export REGISTRY_AUTH_FILE=path`
+
+**--get-login**
+
+Return the logged-in user for the registry. Return error if no login is found.
+
+**--cert-dir**=*path*
+
+Use certificates at *path* (\*.crt, \*.cert, \*.key) to connect to the registry.
+Default certificates directory is _/etc/containers/certs.d_. (Not available for remote commands)
```
mtrmac, line 44: The “remote commands” part does not apply.

integration/login_logout_test.go
```diff
@@ -0,0 +1,61 @@
...
+	_, err := exec.LookPath(skopeoBinary)
+	c.Assert(err, check.IsNil)
+}
+
+func (s *LoginLogoutSuite) TearDownSuite(c *check.C) {
+
+}
+
+func (s *LoginLogoutSuite) SetUpTest(c *check.C) {
+	s.regV2WithAuth = setupRegistryV2At(c, privateRegistryURL1, true, false)
```
mtrmac, line 28: Why not create/tear down the registry in `SetUpSuite`/`TearDownSuite` instead? The registry carries no state AFAICS, and it would be faster.

Or, preferably, change `SkopeoSuite` that way, and make these tests parts of `SkopeoSuite`; that would get rid of even more overhead. Besides, `SkopeoSuite` already deals with authentication, so it would make sense to have everything together.

(I don’t care whether the new tests are in `login_logout_test.go` or in `check_test.go`.)

integration/login_logout_test.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func (s *LoginLogoutSuite) TearDownTest(c *check.C) {
+	if s.regV2WithAuth != nil {
+		//cmd := exec.Command("docker", "logout", s.regV2WithAuth)
```
mtrmac, line 33: These commends probably shouldn’t stay, if they ever meant anything.

integration/login_logout_test.go
```diff
@@ -0,0 +1,61 @@
...
+	wanted := "^Login Succeeded!\n$"
+	assertSkopeoSucceeds(c, wanted, "login", "--tls-verify=false", "--username="+s.regV2WithAuth.username, "--password="+s.regV2WithAuth.password, s.regV2WithAuth.url)
+	// copy to private registry using local authentication
+	imageName := fmt.Sprintf("docker://%s/busybox:mine", s.regV2WithAuth.url)
+	assertSkopeoSucceeds(c, "", "copy", "--dest-tls-verify=false", "docker://docker.io/library/busybox:latest", imageName)
+	// inspec from private registry
+	assertSkopeoSucceeds(c, "", "inspect", "--tls-verify=false", imageName)
+	// logout from the registry
+	wanted = fmt.Sprintf("^Removed login credentials for %s\n$", s.regV2WithAuth.url)
+	assertSkopeoSucceeds(c, wanted, "logout", s.regV2WithAuth.url)
```
mtrmac, line 60: Non-blocking: This could have one more test here to verify that `inspect` now fails, i.e. that `logout` did work.

cmd/skopeo/login.go
mtrmac: Consider https://github.com/containers/common/pull/141 .

cmd/skopeo/login.go
QiWang19: Document doesn't describe --get-login as a bool value that can be true or false. what it should do if --get-login=false?
```
# podman login --get-login docker.io
username
# podman login --get-login=false docker.io
username
# podman login --get-login=true docker.io
username
```

mtrmac: That’s a very good point; unfortunately, unlike C’s `getopt` and the like, neither https://golang.org/pkg/flag/ nor https://godoc.org/github.com/spf13/pflag seems to actually implement parameter-less options; the best we have is `NoOptDefVal`.

I think we can either interpret `--get-login=false` as if it were missing (`GetLoginSet = opts.getLogin.value`) or outright reject it (`if opts.getLogin.present && !opts.getLogin.value { fail }`); treating `--get-login=false` as if the user did want to do `--get-login` seems extremely unintuitive to me.

I don’t see any way at all, given the current API, to define a flag that can be used without a value but that rejects any use with a value (we can se `NoOptDefVal` to anything, and reject uses with any other value, but the user can always call the binary with the same `NoOptDefVal` value explicitly).

Overall, given the way things are in Go, and the way _everything else_, AFAIK, just uses a `pflag.Bool*` and checks the value for `true`, I don’t think spending extra effort on this one boolean (but not on `StdinPassword`, for example), is really worthwhile.

QiWang19:
@mtrmac PTAL

cmd/skopeo/logout.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+	adjustUsage(cmd)
+	cmd.Flags().AddFlagSet(auth.GetLogoutFlags(&opts.logoutOpts))
+	return cmd
+}
+
+func (opts *logoutOptions) run(args []string, stdout io.Writer) error {
+	opts.logoutOpts.Stdout = stdout
+	sys := opts.global.newSystemContext()
+	sys.AuthFilePath = opts.logoutOpts.AuthFile
```
mtrmac, line 34: This should not be necessary after https://github.com/containers/common/pull/141 .

cmd/skopeo/utils.go
mtrmac: (Non-blocking: I meant a comment specific to `--tls-verify`. I guess the current generic one works.)

integration/check_test.go
```diff
@@ -91,3 +84,30 @@ func (s *SkopeoSuite) TestNoNeedAuthToPrivateRegistryV2ImageNotFound(c *check.C)
...
+	assertSkopeoSucceeds(c, wanted, "login", "--tls-verify=false", "--username="+s.regV2WithAuth.username, "--password="+s.regV2WithAuth.password, s.regV2WithAuth.url)
+	// copy to private registry using local authentication
+	imageName := fmt.Sprintf("docker://%s/busybox:mine", s.regV2WithAuth.url)
+	assertSkopeoSucceeds(c, "", "copy", "--dest-tls-verify=false", "docker://docker.io/library/busybox:latest", imageName)
+	// inspec from private registry
+	assertSkopeoSucceeds(c, "", "inspect", "--tls-verify=false", imageName)
+	// logout from the registry
+	wanted = fmt.Sprintf("^Removed login credentials for %s\n$", s.regV2WithAuth.url)
+	assertSkopeoSucceeds(c, wanted, "logout", s.regV2WithAuth.url)
+	// inspec from private registry should fail after logout
```
mtrmac, line 110:
```suggestion
	// inspect from private registry should fail after logout
```

QiWang19:
@mtrmac PTAL

cmd/skopeo/logout.go
mtrmac: Still outstanding.

mtrmac:
https://github.com/containers/skopeo/pull/865#discussion_r422537588 is still outstanding, and please rebase. Looks good otherwise.

QiWang19:
@mtrmac PTAL

Issue #796, opened 2020-01-07
Remove dependency on c/buildah


mtrmac:
After https://github.com/containers/skopeo/pull/789, we don’t need that dependency for any _code_.

But c/buildah is dragging in newer versions of several (non-module) dependencies than this repo would otherwise use, because c/image requires older versions. So, just dropping c/buildah would cause several dependencies to be downgraded.

So, we need to work on updating the c/image dependencies manually, and drop the c/buildah dependency only after new c/image is released and updated here, so that removing the c/buildah dependency does not cause downgrades.

mtrmac:
This can happen after we update to c/image with https://github.com/containers/image/pull/874 .

PR #867, opened 2020-03-28
Update containers/image to v5.3.1

rhatdan:


go.mod
```diff
@@ -3,9 +3,8 @@ module github.com/containers/skopeo
 go 1.12
 
 require (
-	github.com/containers/buildah v1.13.1 // indirect
```
mtrmac: @rhatdan This part caused downgrades.

mtrmac: https://github.com/containers/skopeo/pull/872

rhatdan: Ok I have no idea what is going on with this.  What does buildah have to do with skopeo or containers/image?

mtrmac: > What does buildah have to do with skopeo or containers/image?

https://github.com/containers/skopeo/issues/796

I’m working on making that unnecessary in the future right now, but it has to stay around at least until the next c/image release.

PR #870, opened 2020-03-30
Sync: Add example with repository

### LINKED ISSUE #854: Sync: Add option to keep full name on target
Hi,

I'm trying to sync images between two docker registries using skopeo.

**Current behavior**
Without `--scoped` a source image `registry1/first/last:tag` results in `registry2/registry1/first/last:tag`.
With `--scoped` a source image `registry1/first/last:tag` results in `registry2/last:tag`.

**Wanted behavior**
With `--whatever` a source image `registry1/first/last:tag` results in `registry2/first/last:tag`.

My current workaround/patch to do this with `--scoped`:
```
diff --git a/cmd/skopeo/sync.go b/cmd/skopeo/sync.go
index 93f6a362..f0bc9b86 100644
--- a/cmd/skopeo/sync.go
+++ b/cmd/skopeo/sync.go
@@ -512,7 +512,11 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) error {
                        switch ref.Transport() {
                        case docker.Transport:
                                // docker -> dir or docker -> docker
-                               destSuffix = ref.DockerReference().String()
+                               removeRegistry:= func(s string) string {
+                                       split := strings.Split(s, "/")
+                                       return strings.Join(split[1:], "/")
+                               }
+                               destSuffix = removeRegistry(ref.DockerReference().String())
                        case directory.Transport:
                                // dir -> docker (we don't allow `dir` -> `dir` sync operations)
                                destSuffix = strings.TrimPrefix(ref.StringWithinTransport(), srcRepo.DirBasePath)
```
What do you think?

PhilippHomann:
Fixes #854

### DIFF
--- a/docs/skopeo-sync.1.md
+++ b/docs/skopeo-sync.1.md
@@ -105,8 +105,18 @@ skopeo sync --src docker --dest docker registry.example.com/busybox my-registry.
 ```
 Destination registry content:
 ```
-REPO                           TAGS
-registry.example.com/busybox   1-glibc, 1-musl, 1-ubuntu, ..., latest
+REPO                         TAGS
+registry.local.lan/busybox   1-glibc, 1-musl, 1-ubuntu, ..., latest
+```
+
+### Synchronizing to a container registry keeping the repository
+```
+skopeo sync --src docker --dest docker registry.example.com/repo/busybox my-registry.local.lan/repo
+```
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@vrothberg @mtrmac PTAL

PR #873, opened 2020-03-30
Update on #834: force runc only when cgroupsv1

edsantiago:
PR 834 broke Fedora gating tests, because "--runtime runc"
doesn't work so well on Rawhide. Let's try to be smarter
about when we add that override.

### DIFF
--- a/systemtest/helpers.bash
+++ b/systemtest/helpers.bash
@@ -291,9 +291,21 @@ start_registry() {
         reg_args+=( -e REGISTRY_STORAGE_DELETE_ENABLED=true)
     fi
 
+    # TODO: This is TEMPORARY (as of 2020-03-30); remove once crun is fixed.
+    # Skopeo PR #836 claims there's a "regression" in crun with cgroupsv1,
+    # but offers no details about what it is (crun issue nor PR) nor when/if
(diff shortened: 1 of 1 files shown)


edsantiago:
@vrothberg PTAL. If you can offer any insights on what the crun regression is, and/or suggestions for a better conditional to enable the `--runtime` flag, please let me know and I'll be glad to update.

systemtest/helpers.bash
```diff
@@ -291,9 +291,21 @@ start_registry() {
         reg_args+=( -e REGISTRY_STORAGE_DELETE_ENABLED=true)
     fi
 
+    # FIXME! TEMPORARY! (As of 2020-03-30)
```
TomSweeneyRedHat: Could you add TODO here please?

TomSweeneyRedHat:
Neat little fix @edsantiago 
LGTM with a nit fix.

systemtest/helpers.bash
edsantiago: Done, force-pushed, and tests green. Thanks for the feedback.

rhatdan:
@giuseppe PTAL

PR #875, opened 2020-03-31
DO NOT MERGE: Test for https://github.com/containers/image/pull/879

mtrmac:
Use c/storage/pkg/unshare instead of c/common/pkg/unshare; they can't both exist in a single executable.


mtrmac:
Passed.

PR #879, opened 2020-03-31
Remove dependency on buildah

### LINKED ISSUE #796: Remove dependency on c/buildah

mtrmac:
We no longer need it only to keep other dependencies up to date.

Fixes #796 .


rhatdan:
LGTM

PR #887, opened 2020-04-07
vendor c/image v5.4.2

vrothberg:


go.mod
```diff
@@ -23,4 +23,5 @@ require (
 	github.com/urfave/cli v1.22.1
 	go4.org v0.0.0-20190218023631-ce4c26f7be8e // indirect
 	gopkg.in/yaml.v2 v2.2.8
+	k8s.io/client-go v0.18.0 // indirect
```
mtrmac, line 26: Is this still necessary?

vrothberg: Good catch! Repushed. We need to change `make vendor` to run `go mod tidy` as a last step. It's currently not cleaning up and I assume because we run it before `go mod vendor`.

vrothberg: https://github.com/containers/skopeo/pull/888

TomSweeneyRedHat:
LGTM and happy green test buttons.

PR #888, opened 2020-04-07
make vendor: run `tidy` after `vendor`

vrothberg:
Run `go mod tidy` after `go mod vendor` to properly clean up the
dependencies.

Signed-off-by: Valentin Rothberg <rothberg@redhat.com>

PR #890, opened 2020-04-08
Fix (make vendor)

mtrmac:
`(export a=b command args)` does not run `(command args)` with `a=b`, it sets `$a` to `b`, and marks variables `$a` `$command` `$args` as exported, i.e. `(command args)` is not run.

So, before https://github.com/containers/skopeo/pull/888 we were not actually running `(go mod tidy)`, and now we are not running `(go mod vendor)`.

Just use `$(GO)`, which already sets `GO111MODULE=on`, without the extra export.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -184,10 +184,9 @@ test-unit-local:
 	$(GPGME_ENV) $(GO) test $(MOD_VENDOR) -tags "$(BUILDTAGS)" $$($(GO) list $(MOD_VENDOR) -tags "$(BUILDTAGS)" -e ./... | grep -v '^github\.com/containers/skopeo/\(integration\|vendor/.*\)$$')
 
 vendor:
-	export GO111MODULE=on \
-		$(GO) mod vendor && \
-		$(GO) mod tidy && \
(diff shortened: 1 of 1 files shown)


mtrmac:
Almost all `containers/*` repositories need the same fix. Fun…

vrothberg:
> Almost all `containers/*` repositories need the same fix. Fun…

May the dependency lord have mercy on our souls...

LGTM

vrothberg:
@mtrmac, would you open PRs against libpod, buildah and storage for that as well?

mtrmac:
Luckily, in the others it’s still `go mod tidy` that is ineffective, so this does not hurt as much as if it did if it were `go mod vendor`.

mtrmac:
> @mtrmac, would you open PRs against libpod, buildah and storage for that as well?

Will do.

mtrmac:
#891 
https://github.com/containers/common/pull/112
https://github.com/containers/buildah/pull/2290
https://github.com/containers/libpod/pull/5768

PR #893, opened 2020-04-08
Fix for yet another breaking podman change

edsantiago:
'podman info' changed format, again, without preserving backward
compatibility. Basically, some keys that used to be lower-case
are now upper-case-first-letter.

These tests need to work with new podman on rawhide, and
old podman on f31/f32 and possibly RHEL. We must therefore
add a revolting workaround for the change.


rhatdan:
LGTM
@vrothberg @mtrmac PTAL

PR #900, opened 2020-04-24
Makefile: include -nobuild install targets

lsm5:
Distro packaging often uses non-default build and linker flags.
The current Makefile cannot be used in the package build process as it
will end up rebuilding the binaries with the default flags.

This commit introduces install targets which do not depend on the build
targets. Distro packages should prefer these if they want to use any
non-default flags.

NOTE: This commit effectively leaves prior targets unchanged, so users
won't notice any difference.

@rhatdan PTAL

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,20 +128,26 @@ docs-in-container:
 clean:
 	rm -f skopeo docs/*.1
 
-install: install-binary install-docs install-completions
+install-nobuild: install-binary-nobuild install-docs-nobuild install-completions
 	install -d -m 755 ${SIGSTOREDIR}
 	install -d -m 755 ${CONTAINERSSYSCONFIGDIR}
 	install -m 644 default-policy.json ${CONTAINERSSYSCONFIGDIR}/policy.json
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -128,21 +128,27 @@ docs-in-container:
...
+install: ./skopeo docs install-nobuild 
+
+install-binary-nobuild:
 	install -d -m 755 ${INSTALLDIR}
 	install -m 755 skopeo ${INSTALLDIR}/skopeo
 
-install-docs: docs
+install-binary: ./skopeo install-binary-nobuild
+
+install-docs:
```
TomSweeneyRedHat, line 146: shouldn't this be install-docs-nobuild?

rhatdan:
@lsm5 What is going on here?  Is this still needed?

rhatdan:
@lsm5 Since you never responded, I take that as a know, Closing.  Reopen if you are still interested.

lsm5:
> @lsm5 Since you never responded, I take that as a know, Closing. Reopen if you are still interested.

sgtm, thanks :)

PR #903, opened 2020-04-30
Fix TestCopyAtomicExtension

mtrmac:
Actually test that X-R-S-S signatures are usable using `atomic:`, as the comment says.


rhatdan:
Your commit message does not seem to match the PR, you are removing atomic?

mtrmac:
The signature is _created_ using the X-R-S-S extension to the docker/distribution protocol, using the `docker://` transport. The test that it can be _consumed_ using `atomic:` (as well as X-R-S-S) is just below.

Also compare the code above, which creates the signature using `atomic:` and tests that it is accessible using both methods.

rhatdan:
LGTM

PR #904, opened 2020-04-30
Test github.com/mtrmac/image:mirror-reference

mtrmac:
<s>DO NOT MERGE: This uses unmerged https://github.com/containers/image/pull/912</s>

Vendor in https://github.com/containers/image/pull/912 , and add tests that we are enforcing the correct identities when accessing images through mirrors.


mtrmac:
Re-vendored from merged https://github.com/containers/image/pull/912 .  This will cause Skopeo to use an unreleased version of c/image, but I think it’s important to have this test added and enforcing as soon as possible (even if we released c/image within a few days).

rhatdan:
LGTM
@vrothberg PTAL

rhatdan:
@vrothberg Reping

mtrmac:
Rebased, tests pass.

PR #910, opened 2020-05-11
Update containers/image to v5.4.4

mtrmac:
Only bumps the version number after the recent vendoring from master, but Dependabot seems to be confused by that; so, update to the final release to hopefully un-confuse it.


rhatdan:
LGTM

PR #916, opened 2020-05-15
Add options `tags-only` and` tags-raw` to the list-tags command

zhangguanzhang:


zhangguanzhang:
```
$ skopeo  list-tags docker://docker.io/fedora
{
    "Repository": "docker.io/library/fedora",
    "Tags": [
        "20",
        "21",
        "22",
        "23",
        "24",
        "25",
        "26-modular",
        "26",
        "27",
        "28",
        "29",
        "30",
        "31",
        "32",
        "33",
        "branched",
        "heisenbug",
        "latest",
        "modular",
        "rawhide"
    ]
}

$ skopeo  list-tags docker://docker.io/fedora --tags-only
"20"
"21"
"22"
"23"
"24"
"25"
"26-modular"
"26"
"27"
"28"
"29"
"30"
"31"
"32"
"33"
"branched"
"heisenbug"
"latest"
"modular"
"rawhide"

$ skopeo list-tags docker://docker.io/fedora --tags-only --tags-raw
20
21
22
23
24
25
26-modular
26
27
28
29
30
31
32
33
branched
heisenbug
latest
modular
rawhide
```

TomSweeneyRedHat:
you need to gofmt your go file and repush.

cmd/skopeo/list_tags.go
```diff
@@ -49,6 +52,8 @@ See skopeo-list-tags(1) section "REPOSITORY NAMES" for the expected format
 	}
 	adjustUsage(cmd)
 	flags := cmd.Flags()
+	flags.BoolVar(&opts.tagsOnly, "tags-only", false, "only output the tags")
+	flags.BoolVar(&opts.raw, "tags-raw", false, `output raw strings, not JSON texts while "--tags-only"" is true`)
```
TomSweeneyRedHat, line 56: while -> when using

docs/skopeo-list-tags.1.md
```diff
@@ -22,6 +22,10 @@ Return a list of tags from _repository-name_ in a registry.
   **--tls-verify** _bool-value_ Require HTTPS and verify certificates when talking to container registries (defaults to true)
 
   **--no-creds** _bool-value_ Access the registry anonymously.
+  
+  **--tags-only** _bool-value_ only output the tags.
+  
+  **--tags-raw** _bool-value_ output raw strings, not JSON texts while "--tags-only"" is true.
```
TomSweeneyRedHat, line 28: while -> when

TomSweeneyRedHat:
A test or two would be a welcomed addition.

zhangguanzhang:
tks, I will try this tomorrow

rhatdan:
Why wouldn't we just use jq to get this data?  Or add a --format to the command allow users to specify format, this is what Podman and Buildah do.

cmd/skopeo/list_tags.go
```diff
@@ -128,10 +133,25 @@ func (opts *tagsOptions) run(args []string, stdout io.Writer) (retErr error) {
...
+	if opts.tagsOnly {
+		if opts.raw {
+			out = []byte(strings.Join(tagListing, "\n"))
+		} else {
+			out, err = json.Marshal(tagListing)
+			if err != nil {
+				return err
+			}
+			out = out[1 : len(out)-1]
+			out = bytes.ReplaceAll(out, []byte(","), []byte("\n"))
```
mtrmac, line 146: Hacking formatted output like this is pointlessly risky.

cmd/skopeo/list_tags.go
```diff
@@ -128,10 +133,25 @@ func (opts *tagsOptions) run(args []string, stdout io.Writer) (retErr error) {
 		Tags:       tagListing,
 	}
```
mtrmac, line 134: `outputData` can be created only on the branch where it is used.

mtrmac:
> Why wouldn't we just use jq to get this data?

It might not be available, I guess. OTOH with Skopeo itself being 22 MB large, at best, the trivial size of `jq` should be easily acceptable.

> Or add a --format to the command allow users to specify format, this is what Podman and Buildah do.

That would be a once-for-ever solution, OTOH it exposes more API that has to be kept stable. I didn’t quite follow the situation with `podman inspect` breakages — are there any lessons learned  that should be applied here?

rhatdan:
The format changes in Podman were bugs in the comparison to Docker, But they were in the field and when we changed them to match Docker inspect we inadvertently broke existing users, which I don't think is an issue with skopeo.

zhangguanzhang:
I think these two options can be used to simply output tags before `--format` is implemented

rhatdan:
I don't agree.  We don't want to have to support these flags going forward.  Either use a JSON command line processor like JQ or implement --format like Buildah and Podman have.

PR #919, opened 2020-05-16
Add Skopeo Stable Image Dockerfile for Quay

TomSweeneyRedHat:
Adds the Dockerfile for building the Skopeo container
image on quay.io.  Once merged, this image will be
built automatically upon any merge into the master
branch.  The images will live at:
quay.io/containers/skopeo:latest
quay.io/skopeo/stable:latest

I've built an image using this Dockerfile and have pushed
it to both repositories if you want to play with that.

Once merged, I'll create similar Dockerfiles for
quay.io/skopeo/testing and quay.io/skopeo/upstream.

### DIFF
--- a/contrib/skopeoimage/stable/Dockerfile
+++ b/contrib/skopeoimage/stable/Dockerfile
@@ -1,0 +1,33 @@
+# stable/Dockerfile
+#
+# Build a Skopeo container image from the latest
+# stable version of Skopeo on the Fedoras Updates System.
+# https://bodhi.fedoraproject.org/updates/?search=skopeo
+# This image can be used to create a secured container
+# that runs safely with privileges within the container.
+#
+FROM registry.fedoraproject.org/fedora:32
+
+# Don't include container-selinux and remove
+# directories used by yum that are just taking
+# up space.  Also reinstall shadow-utils as without
+# doing so, the setuid/setgid bits on newuidmap
+# and newgidmap are lost in the Fedora images. 
+RUN useradd skopeo; yum -y update; yum -y reinstall shadow-utils; yum -y install skopeo fuse-overlayfs --exclude container-selinux; yum clean all; rm -rf /var/cache /var/log/dnf* /var/log/yum.*;
+
+# Adjust storage.conf to enable Fuse storage.
+RUN sed -i -e 's|^#mount_program|mount_program|g' -e '/additionalimage.*/a "/var/lib/shared",' -e 's|^mountopt[[:space:]]*=.*$|mountopt = "nodev,fsync=0"|g' /etc/containers/storage.conf
+
+# Setup the ability to use additional stores
+# with this container image.
+RUN mkdir -p /var/lib/shared/overlay-images /var/lib/shared/overlay-layers; touch /var/lib/shared/overlay-images/images.lock; touch /var/lib/shared/overlay-layers/layers.lock
+
+# Setup skopeo's uid/guid entries
+RUN echo skopeo:100000:65536 > /etc/subuid
+RUN echo skopeo:100000:65536 > /etc/subgid
+
+# Point to the Authorization file
+ENV REGISTRY_AUTH_FILE=/auth.json
+
+# Set the entrypoint
+ENTRYPOINT ["/usr/bin/skopeo"]


vrothberg:
Please wait with merging. We have a blog in the pipeline which is using a different Dockerfile. I'll ping you in the Google Doc.

contrib/stable/Dockerfile
```diff
@@ -0,0 +1,20 @@
+# stable/Dockerfile
+#
+# Build a Skopeo container image from the latest
+# stable version of Skopeo on the Fedoras Updates System.
+# https://bodhi.fedoraproject.org/updates/?search=skopeo
+# This image can be used to create a secured container
+# that runs safely with privileges within the container.
+#
+FROM registry.fedoraproject.org/fedora:latest
```
vrothberg, line 9: We should be careful with using the latest tag:
```
~ $ podman run --rm fedora:32 cat /etc/os-release|head -n2
NAME=Fedora
VERSION="32 (Container Image)"
~ $ podman run --rm fedora:latest cat /etc/os-release|head -n2
NAME=Fedora
VERSION="31 (Container Image)"
```
Latest is pointing to an F31 image.

mtrmac:
```console
% ./skopeo --override-os linux inspect  docker://registry.fedoraproject.org/fedora:latest
…
        "version": "32"
…
        "DISTTAG=f32container",
        "FGC=f32",
```
Is it possible that you have a previously-pulled `fedora:latest` on your system?

contrib/stable/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+# Don't include container-selinux and remove
+# directories used by yum that are just taking
+# up space.
+RUN useradd build; yum -y update; yum -y reinstall shadow-utils; yum -y install skopeo fuse-overlayfs --exclude container-selinux; rm -rf /var/cache /var/log/dnf* /var/log/yum.*;
```
mtrmac, line 14: What does the `reinstall` do?

contrib/stable/Dockerfile
```diff
@@ -0,0 +1,20 @@
+# stable/Dockerfile
```
mtrmac, line 1: Is the `stable` directory name some kind of shared convention across the projects? Looking just at the contents of `contrib`, it’s not clear what to expect in a subdirectory named `stable`.

contrib/stable/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+# that runs safely with privileges within the container.
+#
+FROM registry.fedoraproject.org/fedora:latest
+
+# Don't include container-selinux and remove
+# directories used by yum that are just taking
+# up space.
+RUN useradd build; yum -y update; yum -y reinstall shadow-utils; yum -y install skopeo fuse-overlayfs --exclude container-selinux; rm -rf /var/cache /var/log/dnf* /var/log/yum.*;
+
+ADD https://raw.githubusercontent.com/containers/skopeo/master/contrib/skopeoimages/stable/containers.conf /etc/containers/
```
mtrmac, line 16: That URL does not exist and this PR does not add it AFAICS.

contrib/stable/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+# Adjust storage.conf to enable Fuse storage.
+RUN chmod 644 /etc/containers/containers.conf; sed -i -e 's|^#mount_program|mount_program|g' -e '/additionalimage.*/a "/var/lib/shared",' -e 's|^mountopt[[:space:]]*=.*$|mountopt = "nodev,fsync=0"|g' /etc/containers/storage.conf
+RUN mkdir -p /var/lib/shared/overlay-images /var/lib/shared/overlay-layers; touch /var/lib/shared/overlay-images/images.lock; touch /var/lib/shared/overlay-layers/layers.lock
```
mtrmac, line 20: Are /var/lib/shared/overlay-images really a concern for Skopeo? How would one use that? A container running inside the Skopeo. container? Or is the one running this container expected to add `-v` to overlay over this directory?

contrib/stable/Dockerfile
TomSweeneyRedHat: I'll second that, latest is pulling 32 for me from `registry.fedoraproject.org/fedora`.  I'd really rather not put in a hard version into the file.

mtrmac: OTOH, looking at the `Dockerfile`, what are the chances it will continue to work, and be optimal, for any future release without changes?

(I guess it’s better to have a `:latest` base that may break the image, if we notice, than a `:32` base that may silently continue to work for years without being updated. But it’s not a slam dunk.)

contrib/stable/Dockerfile
rhatdan: For some reason fedora base images loose the setuid/setgid bits on newuidmap and newgidmap.  This makes sure the files have the proper file caps

contrib/stable/Dockerfile
TomSweeneyRedHat: Yes, `stable`, `upstream` and `testing` are the directories that live under `contrib` on the other projects.  After this PR gets merged, I plan a follow up PR with Dockerfiles for those as well.  When I do that I'll add a README.md to contrib to provide a bit more clarity.

contrib/stable/Dockerfile
TomSweeneyRedHat: Oh fudge, I've left off the skopeoimage directory (and mis-spelled it here), will touch up.

contrib/stable/Dockerfile
TomSweeneyRedHat: I'm not sure, I lifted this part from podman's Dockerfile.  Happy to remove it if you think it's not useful.

contrib/stable/Dockerfile
mtrmac: That would be useful to document in the file itself.

contrib/stable/Dockerfile
mtrmac: I really have no idea.

TomSweeneyRedHat: @rhatdan I think this line came from your noggin for Podman, needed here?

contrib/stable/Dockerfile
TomSweeneyRedHat: added doc, good thought.

contrib/stable/Dockerfile
vrothberg: > Is it possible that you have a previously-pulled fedora:latest on your system?

Nope, that's a fresh `fedora:latest` but I now saw that I was pulling from `docker.io`. The one from the Fedora registry being used here is good :+1: 

I still cringe when seeing `:latest` tags for anything other than testing. I was "raised" with the policy to never use latest in production, see the K8s docs:

> Note: You should avoid using the :latest tag when deploying containers in production as it is harder to track which version of the image is running and more difficult to roll back properly.

If something breaks, it should break for all equally. IMHO, the responsibility of testing should be on our side (maintainers) and not the users side.

contrib/skopeoimage/stable/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+FROM registry.fedoraproject.org/fedora:latest
+
+# Don't include container-selinux and remove
+# directories used by yum that are just taking
+# up space.  Also reinstall shadow-utils as without
+# doing so, the setuid/setgid bits on newuidmap
+# and newgidmap are lost in the Fedora images. 
+RUN useradd skopeo; yum -y update; yum -y reinstall shadow-utils; yum -y install skopeo fuse-overlayfs --exclude container-selinux; yum clean all; rm -rf /var/cache /var/log/dnf* /var/log/yum.*;
+
+ADD https://raw.githubusercontent.com/containers/skopeo/master/contrib/skopeoimage/stable/containers.conf /etc/containers/
```
vrothberg, line 18: Do we need the containers.conf? Skopeo is not using it afaik.

contrib/stable/Dockerfile
rhatdan: Yes no need for this in skopeo.  This is just to allow containers to use additional stores.  I don't see a real use case for skopeo and additional stores.

podman run -v /var/lib/containers/storage:/var/lib/shared:ro quay.io:skopoe/stable skopeo copy containers-storage ...

But not really likely

rhatdan: But then again no real reason not to do it, since it will not hurt it, then buildah, podman and skopeo would all have the same feature to allow their containers to run with addition-stores.

contrib/skopeoimage/stable/Dockerfile
rhatdan: Yes no need for containers.conf at this point,  Skopeo has no implementation yet.

TomSweeneyRedHat: removed.

contrib/stable/Dockerfile
TomSweeneyRedHat: UNCLE!  Switch to `:32`

contrib/stable/Dockerfile
TomSweeneyRedHat: I've left it in, but have added a comment.

contrib/skopeoimage/stable/Dockerfile
TomSweeneyRedHat: Removed

TomSweeneyRedHat:
Just pushed with changes from the review.

rhatdan:
LGTM

mtrmac:
@TomSweeneyRedHat I love the comments documenting the rationale. Thanks!

PR #930, opened 2020-05-23
Add upstream and testing container images

TomSweeneyRedHat:
This PR adds the Dockerfiles necessary to create the upstream
and testing variants of the Skopeo container images that will
reside in quay.io/skopeo/upstream and quay.io/skopeo/testing
repositories.  The only difference in the Dockerfile between
the stable and testing image is the option `--enablerepo updates-testing`
was added.  The testing variant is relatively the same, but
I'd to clone and install Skopeo in the container.

I've also added a README.md which explains all of the varities
of images and includes some sample usage.


rhatdan:
LGTM
@vrothberg @mtrmac PTAL

PR #933, opened 2020-05-28
Retry skopeo inspect command

### LINKED ISSUE #918: Support retries on commands
Sometimes we run a command and our AWS ECR returns a 504 (or similar, retryable status code). It would be good if skopeo commands supported retries with exponential backoff either by default or opt in.

QiWang19:
Enables to retry skopeo inspect. Add `--retry-times` to set the number of times to retry.  Use exponential backoff and  1s as default initial retry delay.

fix #918 

### DIFF
--- a/cmd/skopeo/inspect.go
+++ b/cmd/skopeo/inspect.go
@@ -11,7 +11,9 @@ import (
 	"github.com/containers/image/v5/image"
 	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/transports"
+	"github.com/containers/image/v5/types"
 	digest "github.com/opencontainers/go-digest"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -33,18 +35,21 @@ type inspectOutput struct {
 }
 
 type inspectOptions struct {
-	global *globalOptions
-	image  *imageOptions
-	raw    bool // Output the raw manifest instead of parsing information about the image
-	config bool // Output the raw config blob instead of parsing information about the image
+	global    *globalOptions
+	image     *imageOptions
+	retryOpts *retryOptions
+	raw       bool // Output the raw manifest instead of parsing information about the image
+	config    bool // Output the raw config blob instead of parsing information about the image
 }
 
 func inspectCmd(global *globalOptions) *cobra.Command {
 	sharedFlags, sharedOpts := sharedImageFlags()
 	imageFlags, imageOpts := imageFlags(global, sharedOpts, "", "")
+	retryFlags, retryOpts := retryFlags()
 	opts := inspectOptions{
-		global: global,
-		image:  imageOpts,
+		global:    global,
+		image:     imageOpts,
+		retryOpts: retryOpts,
 	}
 	cmd := &cobra.Command{
 		Use:   "inspect [command options] IMAGE-NAME",
@@ -64,10 +69,16 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
 	flags.BoolVar(&opts.config, "config", false, "output configuration")
 	flags.AddFlagSet(&sharedFlags)
 	flags.AddFlagSet(&imageFlags)
+	flags.AddFlagSet(&retryFlags)
 	return cmd
 }
 
 func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error) {
+	var (
+		rawManifest []byte
+		src         types.ImageSource
+		imgInspect  *types.ImageInspectInfo
+	)
 	ctx, cancel := opts.global.commandTimeoutContext()
 	defer cancel()
 
@@ -85,9 +96,11 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error) 
 		return err
 	}
 
-	src, err := parseImageSource(ctx, opts.image, imageName)
-	if err != nil {
-		return fmt.Errorf("Error parsing image name %q: %v", imageName, err)
+	if err := retryIfNecessary(ctx, func() error {
+		src, err = parseImageSource(ctx, opts.image, imageName)
+		return err
+	}, opts.retryOpts); err != nil {
+		return errors.Wrapf(err, "Error parsing image name %q", imageName)
 	}
 
 	defer func() {
@@ -96,9 +109,11 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error) 
 		}
 	}()
 
-	rawManifest, _, err := src.GetManifest(ctx, nil)
-	if err != nil {
-		return fmt.Errorf("Error retrieving manifest for image: %v", err)
+	if err := retryIfNecessary(ctx, func() error {
+		rawManifest, _, err = src.GetManifest(ctx, nil)
+		return err
+	}, opts.retryOpts); err != nil {
+		return errors.Wrapf(err, "Error retrieving manifest for image")
 	}
 
 	if opts.raw && !opts.config {
@@ -115,9 +130,12 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error) 
 	}
 
 	if opts.config && opts.raw {
-		configBlob, err := img.ConfigBlob(ctx)
-		if err != nil {
-			return fmt.Errorf("Error reading configuration blob: %v", err)
+		var configBlob []byte
+		if err := retryIfNecessary(ctx, func() error {
+			configBlob, err = img.ConfigBlob(ctx)
+			return err
+		}, opts.retryOpts); err != nil {
+			return errors.Wrapf(err, "Error reading configuration blob")
 		}
 		_, err = stdout.Write(configBlob)
 		if err != nil {
@@ -125,9 +143,12 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error) 
 		}
 		return nil
 	} else if opts.config {
-		config, err := img.OCIConfig(ctx)
-		if err != nil {
-			return fmt.Errorf("Error reading OCI-formatted configuration data: %v", err)
+		var config *v1.Image
+		if err := retryIfNecessary(ctx, func() error {
+			config, err = img.OCIConfig(ctx)
+			return err
+		}, opts.retryOpts); err != nil {
+			return errors.Wrapf(err, "Error reading OCI-formatted configuration data")
 		}
 		err = json.NewEncoder(stdout).Encode(config)
 		if err != nil {
@@ -136,10 +157,13 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error) 
 		return nil
 	}
 
-	imgInspect, err := img.Inspect(ctx)
-	if err != nil {
+	if err := retryIfNecessary(ctx, func() error {
+		imgInspect, err = img.Inspect(ctx)
+		return err
+	}, opts.retryOpts); err != nil {
 		return err
 	}
+
 	outputData := inspectOutput{
 		Name: "", // Set below if DockerReference() is known
 		Tag:  imgInspect.Tag,
--- a/cmd/skopeo/utils.go
+++ b/cmd/skopeo/utils.go
@@ -3,13 +3,22 @@ package main
 import (
 	"context"
 	"io"
+	"math"
+	"net"
+	"net/url"
 	"os"
 	"strings"
+	"syscall"
+	"time"
 
 	"github.com/containers/image/v5/pkg/compression"
 	"github.com/containers/image/v5/transports/alltransports"
 	"github.com/containers/image/v5/types"
+	"github.com/docker/distribution/registry/api/errcode"
+	errcodev2 "github.com/docker/distribution/registry/api/v2"
+	multierror "github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 )
@@ -106,6 +115,17 @@ func imageFlags(global *globalOptions, shared *sharedImageOptions, flagPrefix, c
 	fs.StringVar(&opts.dockerDaemonHost, flagPrefix+"daemon-host", "", "use docker daemon host at `HOST` (docker-daemon: only)")
 	fs.AddFlagSet(&dockerFlags)
 	return fs, opts
+}
+
+type retryOptions struct {
+	maxRetry int // The number of times to possibly retry
+}
+
+func retryFlags() (pflag.FlagSet, *retryOptions) {
+	opts := retryOptions{}
+	fs := pflag.FlagSet{}
+	fs.IntVar(&opts.maxRetry, "retry-times", 0, "the number of times to possibly retry")
+	return fs, &opts
 }
 
 // newSystemContext returns a *types.SystemContext corresponding to opts.
@@ -256,4 +276,69 @@ Flags:
 func adjustUsage(c *cobra.Command) {
 	c.SetUsageTemplate(usageTemplate)
 	c.DisableFlagsInUseLine = true
+}
+
+func isRetryable(err error) bool {
+	err = errors.Cause(err)
+
+	if err == context.Canceled || err == context.DeadlineExceeded {
+		return false
+	}
+
+	type unwrapper interface {
+		Unwrap() error
+	}
+
+	switch e := err.(type) {
+
+	case unwrapper:
+		err = e.Unwrap()
+		return isRetryable(err)
+	case errcode.Error:
+		switch e.Code {
+		case errcode.ErrorCodeUnauthorized, errcodev2.ErrorCodeNameUnknown, errcodev2.ErrorCodeManifestUnknown:
+			return false
+		}
+		return true
+	case *net.OpError:
+		return isRetryable(e.Err)
+	case *url.Error:
+		return isRetryable(e.Err)
+	case syscall.Errno:
+		return e != syscall.ECONNREFUSED
+	case errcode.Errors:
+		// if this error is a group of errors, process them all in turn
+		for i := range e {
+			if !isRetryable(e[i]) {
+				return false
+			}
+		}
+		return true
+	case *multierror.Error:
+		// if this error is a group of errors, process them all in turn
+		for i := range e.Errors {
+			if !isRetryable(e.Errors[i]) {
+				return false
+			}
+		}
+		return true
+	}
+
+	return false
+}
+
+func retryIfNecessary(ctx context.Context, operation func() error, retryOptions *retryOptions) error {
+	err := operation()
+	for attempt := 0; err != nil && isRetryable(err) && attempt < retryOptions.maxRetry; attempt++ {
+		delay := time.Duration(int(math.Pow(2, float64(attempt)))) * time.Second
+		logrus.Infof("Warning: failed, retrying in %s ... (%d/%d)", delay, attempt+1, retryOptions.maxRetry)
+		select {
+		case <-time.After(delay):
+			break
+		case <-ctx.Done():
+			return err
+		}
+		err = operation()
+	}
+	return err
 }
--- a/completions/bash/skopeo
+++ b/completions/bash/skopeo
@@ -73,6 +73,7 @@ _skopeo_inspect() {
      --authfile
      --creds
      --cert-dir
+     --retry-times
      "
      local boolean_options="
      --config
--- a/docs/skopeo-inspect.1.md
+++ b/docs/skopeo-inspect.1.md
@@ -29,6 +29,8 @@ Return low-level information about _image-name_ in a registry
 
   **--cert-dir** _path_ Use certificates at _path_ (\*.crt, \*.cert, \*.key) to connect to the registry
 
+  **--retry-times**  the number of times to retry, retry wait time will be exponentially increased based on the number of failed attempts
+
   **--tls-verify** _bool-value_ Require HTTPS and verify certificates when talking to container registries (defaults to true)
 
   **--no-creds** _bool-value_ Access the registry anonymously.
--- a/go.mod
+++ b/go.mod
@@ -7,9 +7,11 @@ require (
 	github.com/containers/image/v5 v5.5.1
 	github.com/containers/ocicrypt v1.0.2
 	github.com/containers/storage v1.20.2
+	github.com/docker/distribution v2.7.1+incompatible
 	github.com/docker/docker v1.4.2-0.20191219165747-a9416c67da9f
 	github.com/dsnet/compress v0.0.1 // indirect
 	github.com/go-check/check v0.0.0-20180628173108-788fd7840127
+	github.com/hashicorp/go-multierror v1.0.0
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/opencontainers/image-spec v1.0.2-0.20190823105129-775207bd45b6
 	github.com/opencontainers/image-tools v0.0.0-20170926011501-6d941547fa1d


QiWang19:
@rhatdan  @mtrmac PTAL

cmd/skopeo/main.go
```diff
@@ -67,6 +69,8 @@ func createApp() (*cobra.Command, *globalOptions) {
 		logrus.Fatal("unable to mark registries-conf flag as hidden")
 	}
 	rootCommand.PersistentFlags().StringVar(&opts.tmpDir, "tmpdir", "", "directory used to store temporary files")
+	rootCommand.PersistentFlags().IntVar(&opts.maxRetry, "retry-times", 0, "the number of times to retry")
+	rootCommand.PersistentFlags().DurationVar(&opts.retryDelay, "retry-delay", 5*time.Second, "time to wait before retry execution")
```
TomSweeneyRedHat, line 73: suggest
```suggestion
	rootCommand.PersistentFlags().DurationVar(&opts.retryDelay, "retry-delay", 5*time.Second, "time to wait before retrying")
```

cmd/skopeo/main.go
```diff
@@ -99,8 +103,22 @@ func main() {
...
+	rootCmd, opts := createApp()
+	retry(rootCmd, opts)
+}
+
+func retry(rootCmd *cobra.Command, opts *globalOptions) {
+	err := rootCmd.Execute()
+	maxRetry := opts.maxRetry
+	retryDelay := opts.retryDelay
+	for retries := 0; err != nil && retries < maxRetry; retries++ {
+		logrus.Infof("Warning: failed, retrying in %s ... (%d/%d)", retryDelay, retries+1, maxRetry)
```
TomSweeneyRedHat, line 115:
```suggestion
		logrus.Infof("Warning: failed, retrying in %s seconds... (%d/%d)", retryDelay, retries+1, maxRetry)
```

docs/skopeo.1.md
```diff
@@ -64,6 +64,10 @@ Most commands refer to container images, using a _transport_`:`_details_ format.
 
   **--registries.d** _dir_ use registry configuration files in _dir_ (e.g. for container signature storage), overriding the default path.
 
+  **--retry-delay** _duration_  time to wait before retry execution.  default waiting time is 5 seconds.
```
TomSweeneyRedHat, line 67:
```suggestion
  **--retry-delay** _duration_  time to wait before retrying.  The default wait time is 5 seconds.
```

TomSweeneyRedHat:
On the fence, should we retry 2 or 3 times by default?  I think we should not retry by default, but just thought I'd ask.

cmd/skopeo/main.go
```diff
@@ -99,8 +103,22 @@ func main() {
...
 		return
 	}
-	rootCmd, _ := createApp()
-	if err := rootCmd.Execute(); err != nil {
+	rootCmd, opts := createApp()
+	retry(rootCmd, opts)
+}
+
+func retry(rootCmd *cobra.Command, opts *globalOptions) {
+	err := rootCmd.Execute()
```
mtrmac, line 111: It’s not clear that `.Execute` is safe to call multiple times; at the very least this will cause the warning in `opts.before` to be printed on every call.

mtrmac:
> On the fence, should we retry 2 or 3 times by default?

With the current logic, definitely not by default.

mtrmac:
A very vague possible path forward:
- Make this all opt-in; only use it in commands that have been explicitly updated to support retries options where the code has been individually reviewed to ensure e.g. invalid arguments are not retried). (It’s perfectly fine, I think, to only start with a single command, and expand this as/if there is interest, and experience, over time.)
- Share the retry options definitions via something like `imageFlags`.
- (Magic happens here) Write some kind of detectors of errors that are “probably worth retrying for”; e.g. the Buildah retry test seems a reasonable starting point. This will definitely not be comprehensive at the start, and various situations will be quite controversial eventually — see the discussion in https://github.com/containers/libpod/issues/6190 .
- Given the retry options, and the error detector, a “try again if retriable” helper can be written, and used by the relevant commands, for the parts of the commands implementations where it makes sense.

QiWang19:
what can make the implementation eligible for retry, like the net error or syscall errors are possible to happen? Is it possible to have a general retry helper for the commands that need retry?

QiWang19:
@mtrmac PTAL

cmd/skopeo/utils.go
```diff
@@ -104,6 +113,8 @@ func imageFlags(global *globalOptions, shared *sharedImageOptions, flagPrefix, c
 	fs := pflag.FlagSet{}
 	fs.StringVar(&opts.sharedBlobDir, flagPrefix+"shared-blob-dir", "", "`DIRECTORY` to use to share blobs across OCI repositories")
 	fs.StringVar(&opts.dockerDaemonHost, flagPrefix+"daemon-host", "", "use docker daemon host at `HOST` (docker-daemon: only)")
+	fs.IntVar(&opts.maxRetry, "retry-times", 0, "the number of times to retry")
+	fs.DurationVar(&opts.retryDelay, "retry-delay", 5*time.Second, "time to wait before retry execution")
```
mtrmac, line 117: I’m sorry, by “something like `imageFlags`” I didn’t mean exactly _inside_ `imageFlags`: `imageOptions` are not great for this, see how this has to ignore `flagPrefix`.

But this can be a new `type retryOptions` with a `func retryFlags` — and the retry implementation can work with a `retryOptions` instead of individual parameters.

mtrmac, line 117: We might want to use exponential backoff — so, I’m not sure how much to commit to these exact semantics of `retry-delay`.

cmd/skopeo/inspect.go
```diff
@@ -185,3 +187,15 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
 	fmt.Fprintf(stdout, "%s\n", string(out))
 	return nil
 }
+
+func retryGetManifest(ctx context.Context, instanceDigest *digest.Digest, maxRetry int, retryDelay time.Duration, src types.ImageSource) ([]byte, error) {
```
mtrmac, line 191: This would very soon pay off writing as a generic utility,
```go
func retryIfNecessary(ctx context.Context, operation func() error, retryOptions) error
```
or something like that.

cmd/skopeo/inspect.go
```diff
@@ -185,3 +187,15 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
...
+	rawManifest, _, err := src.GetManifest(ctx, nil)
+	for retries := 0; err != nil && isRetryable(err) && retries < maxRetry; retries++ {
+		logrus.Infof("Warning: failed, retrying in %s seconds... (%d/%d)", retryDelay, retries+1, maxRetry)
+		time.Sleep(retryDelay)
```
mtrmac, line 195: This should notice cancelation of `ctx` so that `--command-timeout` still applies.

cmd/skopeo/utils.go
```diff
@@ -257,3 +268,49 @@ func adjustUsage(c *cobra.Command) {
 	c.SetUsageTemplate(usageTemplate)
 	c.DisableFlagsInUseLine = true
 }
+
+func isRetryable(err error) bool {
+	err = errors.Cause(err)
+	type unwrapper interface {
+		Unwrap() error
+	}
+	if unwrapper, ok := err.(unwrapper); ok {
```
mtrmac, line 277: This can be a bit more readable as a type switch.

cmd/skopeo/inspect.go
```diff
@@ -96,7 +98,7 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
 		}
 	}()
 
-	rawManifest, _, err := src.GetManifest(ctx, nil)
+	rawManifest, err := retryGetManifest(ctx, nil, opts.image.maxRetry, opts.image.retryDelay, src)
```
mtrmac, line 101: It is admittedly surprising, but the part that actually fetches the manifest (and thus can fail) is actually the `NewImageSource` in `parseImageSource`; this `GetManifest` is just a local data copy nowadays. I suppose the code should wrap both together.

Also, the `ConfigBlob`/`OCIConfig` calls should probably be retried, though, admittedly, that can come separately.

mtrmac: More precisely, for `docker://` _as it is implemented right now_ the fetch happens in `NewImageSource`; other transports do implement this differently.

cmd/skopeo/utils.go
QiWang19: Dose this need to be a `max delay interval`, so the delay time will be the min(`expotional wait time`, `max delay interval`)

mtrmac: I was thinking maybe “initial retry delay”, with the further retries using a longer (documented or undocumented?) delay; that way the user can choose the delay for the hopefully common case of only one retry, and the total time (via the global timeout), without being promised a specific number of retries.

But I’m not at all sure, others may come up with much better ideas.

cmd/skopeo/inspect.go
QiWang19: move context timeout error check to isRetryable

QiWang19:
@mtrmac PTAL

cmd/skopeo/inspect.go
mtrmac: Marking timeout errors as not retrial is useful, but that’s not quite the same thing:

Assume `--initial-retry-delay` is 5 (maybe by default) and `--command-timeout` is 3. In that case this `time.Sleep` should terminate after 3 seconds, not 5 (or maybe it should not even be attempted and the command should fail with a timeout early).

See the example in https://golang.org/pkg/time/#After ; this can similarly combine `time.After` and `ctx.Done()`.

mtrmac:
@rhatdan @vrothberg / others, I’d welcome a review of the general retry logic / CLI flag design.

cmd/skopeo/inspect.go
```diff
@@ -33,18 +36,21 @@ type inspectOutput struct {
...
 type inspectOptions struct {
-	global *globalOptions
-	image  *imageOptions
-	raw    bool // Output the raw manifest instead of parsing information about the image
-	config bool // Output the raw config blob instead of parsing information about the image
+	global    *globalOptions
+	image     *imageOptions
+	raw       bool // Output the raw manifest instead of parsing information about the image
+	config    bool // Output the raw config blob instead of parsing information about the image
+	retryOpts *retryOptions
```
mtrmac, line 43: (Absolutely non-blocking: Maybe move this after to the `image` field.)

cmd/skopeo/inspect.go
```diff
@@ -85,9 +96,25 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
 		return err
 	}
 
-	src, err := parseImageSource(ctx, opts.image, imageName)
-	if err != nil {
-		return fmt.Errorf("Error parsing image name %q: %v", imageName, err)
+	if _, err := alltransports.ParseImageName(imageName); err != nil {
```
mtrmac, line 99: This is to avoid retries, because `isRetryable` returns `true` on unknown errors, isn’t it?

I’d imagine `isRetryable` should not match such these errors. (Otherwise we would pretty soon have to add a a  `neverRetry struct { err error }` wrapper and start adding it all over the place.)

cmd/skopeo/inspect.go
```diff
@@ -85,9 +96,25 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
...
+		rawManifest = manifest
+		if err != nil {
+			return errors.Wrapf(err, "Error retrieving manifest for image")
+		}
+		return err
+	}, opts.retryOpts, opts.global.commandTimeout); err != nil {
+		return err
 	}
 
 	defer func() {
```
mtrmac, line 120: This should happen just after a successful `parseImageSource`, so I guess we need two `retryIfNecessary` blocks instead of the one above.

cmd/skopeo/inspect.go
```diff
@@ -115,19 +137,33 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
...
+		var config *v1.Image
+		if err := retryIfNecessary(ctx, func() error {
+			cfg, err := img.OCIConfig(ctx)
+			config = cfg
+			if err != nil {
+				return errors.Wrapf(err, "Error reading OCI-formatted configuration data")
+			}
+			return nil
+		}, opts.retryOpts, opts.global.commandTimeout); err != nil {
+			return err
```
mtrmac, line 166: Below, `img.Inspect` should also be retried.

cmd/skopeo/inspect.go
```diff
@@ -85,9 +96,25 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
...
-	if err != nil {
-		return fmt.Errorf("Error parsing image name %q: %v", imageName, err)
+	if _, err := alltransports.ParseImageName(imageName); err != nil {
+		return err
+	}
+
+	if err := retryIfNecessary(ctx, func() error {
+		source, err := parseImageSource(ctx, opts.image, imageName)
+		src = source
+		if err != nil {
```
mtrmac, line 106: (Absolutely non-blocking, just wondering and asking for opinions: Would it be more readable to have
```go
src, err = parseImageSource…
```
?

That’s one line fewer on every call, OTOH there is a risk that we’ll habitually write `:=` and break things.)

WDYallThink?

cmd/skopeo/utils.go
```diff
@@ -108,6 +117,19 @@ func imageFlags(global *globalOptions, shared *sharedImageOptions, flagPrefix, c
 	return fs, opts
 }
 
+type retryOptions struct {
+	maxRetry   int           // The number of times to retry
+	retryDelay time.Duration // Time to wait when only retry once
```
mtrmac, line 122: Maybe `initialRetryDelay`; please update the comment.

cmd/skopeo/utils.go
```diff
@@ -108,6 +117,19 @@ func imageFlags(global *globalOptions, shared *sharedImageOptions, flagPrefix, c
...
+}
+
+func retryFlags() (pflag.FlagSet, *retryOptions) {
+	opts := retryOptions{}
+	fs := pflag.FlagSet{}
+	fs.IntVar(&opts.maxRetry, "retry-times", 0, "the number of times to retry, retry wait time will be exponential backoff delay based on the time of failed attempts")
```
mtrmac, line 128: Maybe just
```suggestion
	fs.IntVar(&opts.maxRetry, "retry-times", 0, "the number of times to possibly retry")
```
? We can talk about the time in the description of the second option, and maybe say more about the kinds of errors that cause a retry in the man page.

cmd/skopeo/utils.go
```diff
@@ -108,6 +117,19 @@ func imageFlags(global *globalOptions, shared *sharedImageOptions, flagPrefix, c
...
+	fs.DurationVar(&opts.retryDelay, "initial-retry-delay", time.Second, "the initial delay peroid and is used to calculate the rest delays between reties")
```
mtrmac, line 129: Maybe
```suggestion
	fs.DurationVar(&opts.retryDelay, "initial-retry-delay", time.Second, "the initial delay period, subsequent retries use longer delays")
```
but I’ll definitely defer to native speakers and people with more experience writing documentation.

cmd/skopeo/utils.go
```diff
@@ -257,3 +279,74 @@ func adjustUsage(c *cobra.Command) {
...
+		case errcode.ErrorCodeUnauthorized, errcodev2.ErrorCodeNameUnknown, errcodev2.ErrorCodeManifestUnknown:
+			return false
+		}
+		return true
+	case *net.OpError:
+		return isRetryable(e.Err)
+	case *url.Error:
+		return isRetryable(e.Err)
+	case syscall.Errno:
+		return !(e == syscall.ECONNREFUSED)
```
mtrmac, line 310:
```suggestion
		return (e != syscall.ECONNREFUSED)
```

cmd/skopeo/utils.go
```diff
@@ -257,3 +279,74 @@ func adjustUsage(c *cobra.Command) {
...
+		// if this error is a group of errors, process them all in turn
+		for i := range e.Errors {
+			if !isRetryable(e.Errors[i]) {
+				return false
+			}
+		}
+		return true
+	}
+
+	return true
```
mtrmac, line 329: I’d prefer to retry only recognized errors, for a start… but I might well be wrong about that.

@rhatdan @vrothberg WDYT? See also the extra `alltransports.ParseImageName` just for error checking that is required with this default.

cmd/skopeo/utils.go
```diff
@@ -257,3 +279,74 @@ func adjustUsage(c *cobra.Command) {
...
+}
+
+func retryIfNecessary(ctx context.Context, operation func() error, retryOptions *retryOptions, commandTimeout time.Duration) error {
```
mtrmac, line 332: `commandTimeout` is not used in this function (it is already included in `ctx`).

cmd/skopeo/utils.go
```diff
@@ -257,3 +279,74 @@ func adjustUsage(c *cobra.Command) {
...
+	err := operation()
+	for attempt := 0; err != nil && isRetryable(err) && attempt < retryOptions.maxRetry; attempt++ {
+		logrus.Infof("Warning: failed, retrying in %s ... (%d/%d)", exponentialDelayTime(attempt, retryOptions.retryDelay), attempt+1, retryOptions.maxRetry)
+		select {
+		case <-time.After(exponentialDelayTime(attempt, retryOptions.retryDelay)):
```
mtrmac, line 337: I’d prefer for `exponentialDelayTime` to only be computed once and stored in a variable, to make sure the two don’t drift apart over time — and then `exponentialDelayTime` can be inlined into this function.

cmd/skopeo/utils.go
```diff
@@ -257,3 +279,74 @@ func adjustUsage(c *cobra.Command) {
...
+		return true
+	}
+
+	return true
+}
+
+func retryIfNecessary(ctx context.Context, operation func() error, retryOptions *retryOptions, commandTimeout time.Duration) error {
+	err := operation()
+	for attempt := 0; err != nil && isRetryable(err) && attempt < retryOptions.maxRetry; attempt++ {
+		logrus.Infof("Warning: failed, retrying in %s ... (%d/%d)", exponentialDelayTime(attempt, retryOptions.retryDelay), attempt+1, retryOptions.maxRetry)
```
mtrmac, line 335: (Eventually it might be useful to log the error here — or log it only if it is different from the error on the previous attempt. But that would require somehow dealing with the very long errors we report so that it doesn’t overwhelm the “waiting now” information, and solving that can easily wait for a future PR.)

cmd/skopeo/inspect.go
```diff
@@ -115,19 +137,33 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
...
+		if err := retryIfNecessary(ctx, func() error {
+			blob, err := img.ConfigBlob(ctx)
+			configBlob = blob
+			if err != nil {
+				return errors.Wrapf(err, "Error reading configuration blob")
+			}
+			return nil
+		}, opts.retryOpts, opts.global.commandTimeout); err != nil {
+			return err
 		}
```
mtrmac, line 150: A very tentative thought: Would it make sense to integrate the very common `errors.Wrapf` into `retryIfNecessary`, something like this:
```go
var configBlob []byte
if err := opts.retryOpts.retryIfNecessary(ctx, func() error {
    configBlob, err = img.ConfigBlob(ctx)
    return err
}, "Error reading configuration blob", /* Other format parameters for Wrapf, if any */); err != nil {
    return err
}
```

That would (along with not using `:=`) let us go from the original 4 lines to 7 instead of 11, OTOH it is also much more magic. Is that worth it or not?

cmd/skopeo/inspect.go
QiWang19: > This is to avoid retries, because `isRetryable` returns `true` on unknown errors, isn’t it?
> 

Yes. This also because I think it's `imageName` argument validation

cmd/skopeo/inspect.go
QiWang19: one retry block wrapping `parseImageSource` and defer Close(), and the other retry block for `GetManifest`?

mtrmac: `defer` only defers until the end of the function, which in this case would be just the retry block, and `src` would not be usable afterwards.

Retry { `src = parseImageSource` }`; `defer src.Close`; Retry { `GetManifest` }.

cmd/skopeo/inspect.go
QiWang19: if the retry only wraps one function, there will only one kind of error message to return, can this error be wrapped after calling `retryIfNecessary`?
```go
var configBlob []byte
if err := opts.retryOpts.retryIfNecessary(ctx, func() error {
    configBlob, err = img.ConfigBlob(ctx)
    return err
}; err != nil {
    return errors.Wrapf(err,  "Error reading configuration blob")
}
```

cmd/skopeo/utils.go
QiWang19: ok, I can solve that in future PR. Would this make `isRetryable(err)` return both bool and an error string, so inside retryIfNecessary the error will be compared?

cmd/skopeo/utils.go
```diff
@@ -257,3 +277,75 @@ func adjustUsage(c *cobra.Command) {
...
+func retryIfNecessary(ctx context.Context, operation func() error, retryOptions *retryOptions) error {
+	err := operation()
+	for attempt := 0; err != nil && isRetryable(err) && attempt < retryOptions.maxRetry; attempt++ {
+		delay := func() time.Duration {
+			const base = 2
+			const initialRetryDelay = time.Second
+			return time.Duration(int(math.Pow(base, float64(attempt)))) * initialRetryDelay
+		}()
+
+		logrus.Infof("Warning: failed, retrying in %s ... (%d/%d)", delay, attempt+1, retryOptions.maxRetry)
```
TomSweeneyRedHat, line 339:
```suggestion
		logrus.Infof("Warning: failed, retrying in %s seconds ... (%d/%d)", delay, attempt+1, retryOptions.maxRetry)
```

docs/skopeo-inspect.1.md
```diff
@@ -29,6 +29,10 @@ Return low-level information about _image-name_ in a registry
 
   **--cert-dir** _path_ Use certificates at _path_ (\*.crt, \*.cert, \*.key) to connect to the registry
 
+  **--initial-retry-delay** _duration_  the initial delay peroid and is used to calculate the rest delays between reties (defaults to 1s)
+
+  **--retry-times**  the number of times to retry, retry wait time will be exponential backoff delay based on the time of failed attempts
```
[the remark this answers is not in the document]
rhatdan: Lets just remove the flag and do the exponential increase.

cmd/skopeo/utils.go
QiWang19: `delay` variable time.Duration type already has `s` as unit. I think it should be ok to use that?
```go
INFO[0000] Warning: failed, retrying in 1s 
```

TomSweeneyRedHat: ooops, missed that.  Yes please do ignore this comment and TY!

rhatdan:
@QiWang19 What is going on with this PR?

QiWang19:
@mtrmac PTAL. Can this PR get in?

rhatdan:
@QiWang19 @mtrmac We are going to need similar support in Podman, I believe.

cmd/skopeo/inspect.go
```diff
@@ -64,10 +69,16 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
...
 	flags.AddFlagSet(&imageFlags)
+	flags.AddFlagSet(&retryFlags)
 	return cmd
 }
 
 func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error) {
+	var (
+		rawManifest []byte
+		src         types.ImageSource
+		imgInspect  = &types.ImageInspectInfo{}
```
mtrmac, line 80: Non-blocking: AFAICS this can be `imgInspect *types.ImageInspectInfo` and avoid creating the empty object.

cmd/skopeo/utils.go
```diff
@@ -257,3 +277,72 @@ func adjustUsage(c *cobra.Command) {
...
+
+	return false
+}
+
+func retryIfNecessary(ctx context.Context, operation func() error, retryOptions *retryOptions) error {
+	err := operation()
+	for attempt := 0; err != nil && isRetryable(err) && attempt < retryOptions.maxRetry; attempt++ {
+		delay := func() time.Duration {
+			return time.Duration(int(math.Pow(2, float64(attempt)))) * time.Second
+		}()
```
mtrmac, line 335: Why the `func(){…}()`?

cmd/skopeo/utils.go
mtrmac: This function can just collect the `err` value from the previous loop iteration. I wouldn’t typically expect the most-nested error used in `isRetryable` to actually make the decision to stay the same while the outermost error changes — and if that _were_ the case, that would be a good reason to report the different error (e.g. before a proxy was failing, then the registry was failing, then the underlying storage of a registry was failing, or something like that).

cmd/skopeo/inspect.go
mtrmac: Oh, right. That’s much clearer.

cmd/skopeo/utils.go
```diff
@@ -257,3 +277,72 @@ func adjustUsage(c *cobra.Command) {
...
+		logrus.Infof("Warning: failed, retrying in %s ... (%d/%d)", delay, attempt+1, retryOptions.maxRetry)
+		select {
+		case <-time.After(delay):
+			break
+		case <-ctx.Done():
+			return err
+		}
+		if err = operation(); err == nil {
```
mtrmac, line 343: This can be just `err = operation()` to be symmetrical with the first attempt - or do the check in the first attempt as well, and drop it from the loop condition.

completions/bash/skopeo
```diff
@@ -73,6 +73,8 @@ _skopeo_inspect() {
      --authfile
      --creds
      --cert-dir
+     --initial-retry-delay
```
mtrmac, line 76: This no longer exists.

mtrmac:
> @QiWang19 @mtrmac We are going to need similar support in Podman, I believe.

… and most Skopeo subcommands as well.

QiWang19:
Will apply the retry to other subcommands in another PRs

PR #941, opened 2020-06-06
Update Skopeo Image Dockerfiles to F33

TomSweeneyRedHat:
Update the version of Fedora used in the Dockerfiles
that create the Skopeo Container Images on quay.io to
Fedora 33.  This will enable our latest image in stable
and containers to contain v1.0.0 as currently there's not
a v1.0 kit for Fedora 32.

### DIFF
--- a/contrib/skopeoimage/stable/Dockerfile
+++ b/contrib/skopeoimage/stable/Dockerfile
@@ -6,7 +6,7 @@
 # This image can be used to create a secured container
 # that runs safely with privileges within the container.
 #
-FROM registry.fedoraproject.org/fedora:32
+FROM registry.fedoraproject.org/fedora:33
 
 # Don't include container-selinux and remove
 # directories used by yum that are just taking
--- a/contrib/skopeoimage/testing/Dockerfile
+++ b/contrib/skopeoimage/testing/Dockerfile
@@ -7,7 +7,7 @@
 # This image can be used to create a secured container
 # that runs safely with privileges within the container.
 #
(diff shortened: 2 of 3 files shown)


rhatdan:
The problem with this, is that you end up with an Unstable OS on the stable package.

TomSweeneyRedHat:
@rhatdan I'm a little confused.  Are you saying F33 is unstable?  The main reason I bumped this now is there doesn't appear to be a v1.0 Skopeo rpm available on bohdi for F32 so our "latest" image is still `v0.1.42-dev`.

mtrmac:
That seems like something that should be fixed in the Fedora packaging. At least as of https://src.fedoraproject.org/rpms/skopeo/c/ee64c83df3b53074026ccee9b302b6525a7af463?branch=f32 the RPM should point at 1.0 (and there never was a 0.1.42).

mtrmac:
> (and there never was a 0.1.42).

(For the record: the bits called itself that, but it was not a released version; the next the release was called 0.2.0 (don’t confuse that with 0.2 from 2017) but the RPM has apparently called itself 0.2.0 already, without pre-release versioning.)

TomSweeneyRedHat:
FWIW, on F32 the "latest" reports:
```
# podman run quay.io/containers/skopeo:latest -v
skopeo version 0.1.42-dev
```

rhatdan:
I think the f32 builds are blowing up. Testing builds locally I get the proper version now.

PR #947, opened 2020-06-11
[WIP]fix test error caused by login warning message

QiWang19:
https://github.com/containers/image/pull/959 has printed out the warning message for credential helpers, caused the original test failure.


rhatdan:
@QiWang19 Still working on this one?

rhatdan:
Why is this a WIP, it looks like it could get merged as is, once the go.sum is fixed.

mtrmac:
> Why is this a WIP, it looks like it could get merged as is, once the go.sum is fixed.

Because it uses an unreleased version of c/image, see `go.mod`. See https://github.com/containers/skopeo/blob/master/CONTRIBUTING.md#dependencies-management , an unfortunate effect of most of c/image tests requiring Skopeo.

PR #952, opened 2020-06-16
Cleanup Dockerfile builds

rhatdan:
Add `-y` options to yum clean all
Only delete below /var/cache/dnf so that I can use the
-v /var/cache/dnf:/var/cache/dnf:O option when building
to speed up builds.


rhatdan:
@TomSweeneyRedHat PTAL

TomSweeneyRedHat:
@rhatdan when are you doing those builds that you're speeding up?  Are you pulling the complete image and then buliding upon it?

TomSweeneyRedHat:
And looking at image size, it bumps the stable image from 252MB to 398MB, I'd like to keep it on the lower size, especially since both the Buildah and Podman images are smaller than 398MB.

rhatdan:
Wow, lets look at other caches besides dnf.

rhatdan:
@TomSweeneyRedHat Your size difference is compressed versus not compressed, definitely not caused by the change.

TomSweeneyRedHat:
thx @rhatdan this needs a rebase, once done we can merge this.

PR #960, opened 2020-06-19
Run htpasswd from our build-container instead of registry:2

mtrmac:
registry:2 no longer contains htpasswd.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,6 +7,8 @@ RUN dnf -y update && dnf install -y make git golang golang-github-cpuguy83-md2ma
 	# gpgme bindings deps
 	libassuan-devel gpgme-devel \
 	gnupg \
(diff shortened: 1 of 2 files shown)


mtrmac:
@edsantiago is introducing this dependency on the environment surrounding system tests  acceptable?

edsantiago:
It's fine. It concerns me that we can't rely on [docker's own documentation](https://docs.docker.com/registry/deploying/#native-basic-auth), and I wonder what else they might break without notice, but it is perfectly fine with me to move the htpasswd dependency to something more reliable.

mtrmac:
Yes, https://github.com/docker/distribution-library-image/issues/106 . It seems they didn’t want to keep updating the container for `httpd` security fixes.

Merging now to unblock CI. Thanks!

PR #965, opened 2020-06-30
Added SemVer tag compare to sync and a 'noop' option.

senax:
This PR adds support for semver constraints in the sync command using hashicorp go-version module.

It also adds a 'noop' option which skips the actual copying of the containers.

```
sync.yaml:
...
    images-by-semver-compare:
        busybox: ">=1.29, <1.35"
```

Signed-off-by: Frank Ederveen <frank.ederveen@gmail.com>


cmd/skopeo/sync.go
```diff
@@ -51,11 +53,12 @@ type tlsVerifyConfig struct {
...
 // the source YAML file
 type registrySyncConfig struct {
-	Images           map[string][]string    // Images map images name to slices with the images' tags
-	ImagesByTagRegex map[string]string      `yaml:"images-by-tag-regex"` // Images map images name to regular expression with the images' tags
-	Credentials      types.DockerAuthConfig // Username and password used to authenticate with the registry
-	TLSVerify        tlsVerifyConfig        `yaml:"tls-verify"` // TLS verification mode (enabled by default)
-	CertDir          string                 `yaml:"cert-dir"`   // Path to the TLS certificates of the registry
+	Images                map[string][]string    // Images map images name to slices with the images' tags
+	ImagesByTagRegex      map[string]string      `yaml:"images-by-tag-regex"`      // Images map images name to regular expression with the images' tags
+	ImagesBySemVerCompare map[string]string      `yaml:"images-by-semver-compare"` // Images map images name to regular expression with the images' tags
```
mtrmac, line 58: Please update the comment.

cmd/skopeo/sync.go
```diff
@@ -380,6 +384,68 @@ func imagesToCopyFromRegistry(registryName string, cfg registrySyncConfig, sourc
 			Context:      serverCtx})
 	}
 
+	for imageName, semVerCompare := range cfg.ImagesBySemVerCompare {
```
mtrmac, line 387: This function is by now way too long and redundant.

Probably the `len(tags) != 0` case should be handled specially; the other three (`len(tags) == 0`, `ImagesByTagRegex`, `ImagesBySemVerCompare` should probably use a common helper that does the query+loops, parameterized by a “filter” callback.

And the it would be useful to have a 5-10-item unit test for the semantic versioning filter.

cmd/skopeo/sync.go
```diff
@@ -380,6 +384,68 @@ func imagesToCopyFromRegistry(registryName string, cfg registrySyncConfig, sourc
...
+		}
+
+		repoLogger.Info("Processing repo")
+
+		var sourceReferences []types.ImageReference
+
+		semVerConstraint, err := version.NewConstraint(semVerCompare)
+		if err != nil {
+			repoLogger.WithFields(logrus.Fields{
+				"semVerCompare": semVerCompare,
```
mtrmac, line 407: `semVerCompare` was already added above.

integration/sync_test.go
```diff
@@ -289,6 +314,39 @@ docker.io:
...
+    busybox: "~>1.29.1"
+`
+	var nTags = 3
+	c.Assert(nTags, check.Not(check.Equals), 0)
+
+	yamlFile := path.Join(tmpDir, "registries.yaml")
+	ioutil.WriteFile(yamlFile, []byte(yamlConfig), 0644)
+	assertSkopeoSucceeds(c, "", "sync", "--scoped", "--src", "yaml", "--dest", "dir", yamlFile, dir1)
+
+	nManifests := 0
```
mtrmac, line 335: Let’s not add a third copy of this either; please factor that out into an `assertNumberOfManifests` or something like that.

Does this actually differ in effect from the `filepath.Glob` implementation? Maybe all of that could use the same helper.

integration/sync_test.go
```diff
@@ -289,6 +314,39 @@ docker.io:
...
+	defer os.RemoveAll(tmpDir)
+	dir1 := path.Join(tmpDir, "dir1")
+
+	yamlConfig := `
+docker.io:
+  images-by-semver-compare:
+    busybox: "~>1.29.1"
+`
+	var nTags = 3
+	c.Assert(nTags, check.Not(check.Equals), 0)
```
mtrmac, line 329: The tests are now running for quite a long time, and the purpose of these tests is not to test raw registry interoperability; adding three more image copies is not making that better.

Please use a repository on one of the registries created in `SyncSuite`; maybe pull one small image from `docker.io` and use different tags for that, or copy some of the images from `integration/fixtures` (or add a few more like that) to have clearly different images for the different tags.

cmd/skopeo/sync.go
senax: Done

cmd/skopeo/sync.go
senax: I'm not really a golong person but gave it a shot. I've taken out the regex and semver filtering. The plain tags/empty list one is a bit too different to easily go in  the same function.

I've also added some unit-tests for filterRegex and filterSemVer.

integration/sync_test.go
senax: It seems that all the tests (for sync at least) could do with creating a local docker image with different tags and then using that rather than downloading anything.

I can remove my added integration tests for now, rewriting how the sync tests work would be difficult to justify to my boss at the moment.

rhatdan:
@senax Needs a rebase.

rhatdan:
@senax Are you still interested in this?  @mtrmac @vrothberg Should we take this over or close if @senax is not going to continue working on it?

rhatdan:
Replaced with #996

PR #978, opened 2020-07-16
vendor golang.org/x/text@v0.3.3

vrothberg:
Fixes: CVE-2020-14040


vrothberg:
@rhatdan @mtrmac PTAL

vendor/golang.org/x/text/transform/transform.go
```diff
@@ -659,6 +660,9 @@ func String(t Transformer, s string) (result string, n int, err error) {
 				dst = grow(dst, pDst)
 			}
 		} else if err == ErrShortSrc {
+			if atEOF {
+				return string(dst[:pDst]), pSrc, err
+			}
```
mtrmac, line 665: WRT the CVE: This the only code change we actually vendor in, and the `transform.String` is AFAICS not called anywhere; so we were not vulnerable to the original report (relating to the BOM), and this update does not change behavior anyway (apart from the new Unicode tables on Go ≥ 1.14).

PR #979, opened 2020-07-16
[1.0] vendor golang.org/x/text@v0.3.3

vrothberg:
Fixes: CVE-2020-14040


vrothberg:
Closing. We're shipping 1.1 in RHEL 8.3. @mtrmac @rhatdan, can we delete the release-1.0 branch?

mtrmac:
> @mtrmac @rhatdan, can we delete the release-1.0 branch?

I don’t have a strong opinion, I can’t find a record of any decision about this branch — and a branch can always be recreated from the tag, so feel free to do whatever you think is best.

My weak suggestion is that RH backports should be driven by RH needs tracked internally, and whether upstream branches exist should not really figure in the decision — if an upstream branch is missing, we should create it, not refuse to do the backport, so correspondingly if an extra upstream branch exists, we could just ignore it.

Every old release branch will stop being updated _some_ time, so mere presence of a branch implies neither a promise nor a requirement to backport to it.

rhatdan:
Delete it.

PR #981, opened 2020-07-16
[1.1] vendor golang.org/x/text@v0.3.3

vrothberg:
Fixes: CVE-2020-14040
Signed-off-by: Valentin Rothberg <rothberg@redhat.com>

PR #983, opened 2020-07-16
[1.1] Run htpasswd from our build-container instead of registry:2

mtrmac:
This is a backport of #960 to 1.1, per the discussion in #981.


mtrmac:
Oops, wrong destination branch.

PR #990, opened 2020-07-24
Export InspectOutput

### LINKED ISSUE #989: Export inspectOutput struct /cmd/skopeo/inspect.go
If an external program needs to read output of skopeo inspect, then they have to define its own struct. Instead, if we export inspectOutput by making it InspectOutput, it will help in reading the output by using it as a library.

ashokponkumar:
If an external program needs to read output of skopeo inspect, then they have to define its own struct. Instead, if we export inspectOutput by making it InspectOutput, it will help in reading the output by using it as a library.

closes #989

### DIFF
--- a/cmd/skopeo/inspect.go
+++ b/cmd/skopeo/inspect.go
@@ -19,8 +19,8 @@ import (
 	"github.com/spf13/cobra"
 )
 
-// inspectOutput is the output format of (skopeo inspect), primarily so that we can format it with a simple json.MarshalIndent.
-type inspectOutput struct {
+// InspectOutput is the output format of (skopeo inspect), primarily so that we can format it with a simple json.MarshalIndent.
+type InspectOutput struct {
 	Name          string `json:",omitempty"`
 	Tag           string `json:",omitempty"`
 	Digest        digest.Digest
@@ -164,7 +164,7 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error) 
 		return err
 	}
 
-	outputData := inspectOutput{
+	outputData := InspectOutput{
 		Name: "", // Set below if DockerReference() is known
 		Tag:  imgInspect.Tag,
 		// Digest is set below.


rhatdan:
@ashokponkumar Need
```
If an external program needs to read output of skopeo inspect, then they have to define its own struct. Instead, if we export inspectOutput by making it InspectOutput, it will help in reading the output by using it as a library.
``` 
In you commit message.

git commit -a --amend -s
Enter the message
git push --force

ashokponkumar:
> @ashokponkumar Need
> 
> ```
> If an external program needs to read output of skopeo inspect, then they have to define its own struct. Instead, if we export inspectOutput by making it InspectOutput, it will help in reading the output by using it as a library.
> ```
> 
> In you commit message.
> 
> git commit -a --amend -s
> Enter the message
> git push --force

Thanks @rhatdan for the detailed comment. Have done the same.

rhatdan:
One more time. 
Could you add the the title back on your commit.
git commit -a --amend -s

```
Export InspectOutput

 If an external program needs to read output of skopeo inspect, then they have to define its own struct. Instead, if we export inspectOutput by making it InspectOutput, it will help in reading the output by using it as a library.

Signed-off-by: Ashok Pon Kumar <ashokponkumar@in.ibm.com>
```
The space empty line above is important.  Git requires it to format the message nicely in the git log.

ashokponkumar:
> One more time.
> Could you add the the title back on your commit.
> git commit -a --amend -s
> 
> ```
> Export InspectOutput
> 
>  If an external program needs to read output of skopeo inspect, then they have to define its own struct. Instead, if we export inspectOutput by making it InspectOutput, it will help in reading the output by using it as a library.
> 
> Signed-off-by: Ashok Pon Kumar <ashokponkumar@in.ibm.com>
> ```
> 
> The space empty line above is important. Git requires it to format the message nicely in the git log.

Thanks @rhatdan . Done.

rhatdan:
LGTM
@mtrmac @vrothberg @TomSweeneyRedHat PTAL

ashokponkumar:
Thanks for the nice catch @mtrmac . The triviality of change blind-sighted me.

I personally like using the package reference rather than copy pasting the struct, since it is easier to keep changes up to date. But I agree it is a trade off.

I am also cognizant of the versioning implications you have mentioned. If the team feels it is okay to go ahead, then I can attempt at fixing the issue by moving it to a new package. Or else, we can close this PR.

If we decide to go ahead, I might need some guidance on the right practise given skopeo's standards.

rhatdan:
I think this is a defacto API, whether we want it or not,  Users will  build scripts based on the json output.

rhatdan:
How about something like:

https://github.com/containers/skopeo/compare/master...rhatdan:inspect?expand=1

mtrmac:
Sure, that probably _works_, but it still doesn’t automatically notify the caller about breakage if we follow semantic versioning, so I’m not sure that it _helps_. Whether we export the type or not, the callers have to manually observe changes to the (currently undocumented!) format, or have an integration test.

Admittedly, there isn’t _much_ of a downside (separating the struct from the only consumer is a trivial inconvenience). Keeping the structure internal does “allow having a conversation” about just calling the library, but there can be legitimate reasons to call out to an executable from Go (if the binary has to exist anyway, why pay the code size price twice).

So I’m not going to explicitly NAK a change like that (and of course there is ample space for bikeshedding in such a small change, consider c/skopeo/{inspect,output,api}); I’m not to enthusiastic, but if others think it is useful…

rhatdan:
Well there is a version to skopeo.
skopeo --version
A bash script could execute the skopeo command to make sure it is compatible with the version of the json structure.

rhatdan:
https://github.com/containers/skopeo/pull/992

mtrmac:
> A bash script could execute the skopeo command to make sure it is compatible with the version of the json structure.

A _bash_ script is not going to use the exported Go type in the first place.

A Go caller could do so, sure, but the exported type won’t help with any new fields added in a minor version. That needs to be noticed, tracked, and implemented manually. Neither will it help with any fields removed in a major version, due to Go modules. That needs to be noticed, tracked, and implemented manually.

rhatdan:
Sure, but that fact still remains, that we are making no guarantees, but are making this easier for people who want to do it.  Famous last words, but when was the last time we changes the Skopeo output in a breaking change?

rhatdan:
@vrothberg @mtrmac PTAL

vrothberg:
As @mtrmac mentioned earlier, the change as is won't give access to users as it's in main.

I recommend using the new Go bindings of Podman for image inspection: https://godoc.org/github.com/containers/libpod/pkg/bindings/images#GetImage

Soon, there will be a new blog post on [podman.io](https://podman.io) how to use them.

mtrmac:
> As @mtrmac mentioned earlier, the change as is won't give access to users as it's in main.

This one doesn’t work. #992 does.

I don’t think it really helps the callers, but if y’all think it’s good idea, I’m not going to object.

rhatdan:
#992 is merged.

PR #993, opened 2020-07-29
v1.1.1

vrothberg:
@mtrmac @rhatdan PTAL

Mainly meant to get the x/text bump in. Although Skopeo is unlikely affected, it will still show up in static checkers etc.


rhatdan:
LGTM

PR #1004, opened 2020-08-08
Use an inter-registry copy as the example for (skopeo copy)

### LINKED ISSUE #924: Suggestion for "$skopeo  -h "
I would appreciate if the example can be provided with "$skopeo -h " on cli  simply. 
If I missed how to see example, please let me know.

mtrmac:
because that's what users are looking for, and instead of using a `containers-storage:` source, which might not even work all that well with all the automatic defaults Podman sets up.

Fixes #924 .


rhatdan:
LGTM
@vrothberg PTAL

TomSweeneyRedHat:
LGTM
and yet another rebase needed victim

PR #1008, opened 2020-08-10
fix build in docker container

### LINKED ISSUE #1007: Build in docker container not working
Hello.
I'm trying to build a skopeo binary for Linux on my Mac by running `make build`, but the build freezes to the post-install tzdata script that installs in the recommended packages. This happens due to the fact that the docker build execution is not interactive
```
debconf: unable to initialize frontend: Dialog
debconf: (TERM is not set, so the dialog frontend is not usable.)
debconf: falling back to frontend: Readline
Configuring tzdata
------------------

Please select the geographic area in which you live. Subsequent configuration
questions will narrow this down by presenting a list of cities, representing
the time zones in which they are located.

  1. Africa      4. Australia  7. Atlantic  10. Pacific  13. Etc
  2. America     5. Arctic     8. Europe    11. SystemV
  3. Antarctica  6. Asia       9. Indian    12. US
Geographic area: 6
^Cmake: *** [binary] Interrupt: 2
```

myback:
fix #1007

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -1,13 +1,9 @@
-FROM ubuntu:20.04
+FROM golang:1.14-buster
 
-RUN apt-get update && apt-get install -y \
-    golang \
-    libbtrfs-dev \
-    git-core \
-    libdevmapper-dev \
-    libgpgme11-dev \
-    go-md2man \
-    libglib2.0-dev
+RUN apt-get update && \
+    apt-get install -y \
+        libdevmapper-dev \
+        libgpgme11-dev
 
 ENV GOPATH=/
 WORKDIR /src/github.com/containers/skopeo


TomSweeneyRedHat:
Stupid question time from me, were you able to test this on your mac and did it fix the issue you ran into?

myback:
Sure. Full log in attached file.

```bash
> ~/skopeo/: git show --summary
commit 88df74f333ed562b8df5499960e778327aada00b (HEAD -> master, origin/master, origin/HEAD)
Author: myback <54638513+myback@users.noreply.github.com>
Date:   Mon Aug 10 21:51:17 2020 +0700

    fix build in docker container

> ~/skopeo/: shasum -a 256 bin/skopeo
1e48b2d73bc3f50aad88cfef34be594491b1d61ade42f3144904ddbfb551134e  bin/skopeo
```

[build.log](https://github.com/containers/skopeo/files/5055222/build.log)

TomSweeneyRedHat:
LGTM
Thanks for the follow up @myback !

Dockerfile.build
```diff
@@ -1,6 +1,7 @@
 FROM ubuntu:20.04
 
-RUN apt-get update && apt-get install -y \
+RUN apt-get update && apt-get install -y  --no-install-recommends \
+    build-essential \
```
mtrmac, line 4: The `--no-install-recommends` option only fixes this for a fairly indirect reason; it will break again if any of the truly necessary packages start asking questions.

Per https://manpages.debian.org/jessie/debconf-doc/debconf.7.en.html it seems that the way to opt out of interactivity once for all is to use `DEBIAN_FRONTEND=noninteractive apt-get install` .

myback:
@mtrmac is these edits are ok?

myback:
This gives smaller image size and faster build
```
REPOSITORY                        TAG                 IMAGE ID            CREATED             SIZE
skopeobuildimage                  no-recommends       791858611a24        22 seconds ago      729MB
skopeobuildimage                  origin              4680b5d6b71e        4 minutes ago       795MB
```

kniec:
To me `make` is still missing in the install and when using `--no-install-recommends` a run of `make binary` fails as `gcc` can not be found.
(I also recommend ARG to provide a temporary environment var, but that is more cosmetic - I'd say why not use ENV?)

myback:
@kniec `--no-install-recommends` used for [this](https://ubuntu.com/blog/we-reduced-our-docker-images-by-60-with-no-install-recommends) and does not install gcc. I used build-essential package for install gcc, make and other packages for build.

ENV for apt-get can cause unexpected side effects. [Note](https://docs.docker.com/engine/reference/builder/#env)

i think the more correct way is to use the golang base image

kniec:
@myback That is why I used ARG and not ENV. :) 
It won't carry over into another image build like ENV will do. You can even see it when using a multistage Dockerfile.

```
$ cat Dockerfile
FROM alpine AS test1
ARG KEY=value
RUN echo $KEY

FROM test1
RUN echo $KEY
$ docker build -t test --no-cache .
Sending build context to Docker daemon  3.072kB
Step 1/5 : FROM alpine AS test1
 ---> a24bb4013296
Step 2/5 : ARG KEY=value
 ---> Running in 714bba7751a3
Removing intermediate container 714bba7751a3
 ---> 2833cb022b29
Step 3/5 : RUN echo $KEY
 ---> Running in 9ea4d7148f62
value
Removing intermediate container 9ea4d7148f62
 ---> e11d56b72078
Step 4/5 : FROM test1
 ---> e11d56b72078
Step 5/5 : RUN echo $KEY
 ---> Running in 96d6c6160321

Removing intermediate container 96d6c6160321
 ---> 8848d2fe0555
Successfully built 8848d2fe0555
Successfully tagged test:latest
$
```

The second `echo $KEY` does not yield the value as it is only present in the first stage.

rhatdan:
@myback @kniec @mtrmac we have two attempts to fix this, Let's concentrate on one and get concensus and then we can merge.

kniec:
@rhatdan I like the reduced complexity using the official golang image - I'll close my PR in favour of this one.

rhatdan:
Ok @myback Can you clean up this PR?

git rebase -i origin
Fix it up removing the meres
git push --force

myback:
done

myback:
@rhatdan done

rhatdan:
LGTM

PR #1016, opened 2020-08-14
-buildmode=pie is not supported for some arch

XiaodongLoong:
Signed-off-by: Xiaodong Liu <liuxiaodong@loongson.cn>

I build skopeo on mips64le with error:
`-buildmode=pie not supported on linux/mips64le`
There are the same issue on other cpu arch. 
reference: https://github.com/containerd/containerd/blob/master/Makefile.linux#L23

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,9 @@ BASHINSTALLDIR=${PREFIX}/share/bash-completion/completions
 
 GO ?= go
 GOBIN := $(shell $(GO) env GOBIN)
+GOOS ?= $(shell go env GOOS)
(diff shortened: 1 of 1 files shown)


XiaodongLoong:
@jnovy PTAL, thanks!

jnovy:
LGTM.

Mirek, it looks ok to me when it just excludes the unsupported arches and makes PIE default for others.

The fedora packaging team seems to be dead, see my PR here which is stale: https://pagure.io/packaging-committee/pull-request/1014

rhatdan:
LGTM

PR #1032, opened 2020-09-01
Add support for creating and pushing multi-architecture manifests

flavio:
This PR introduces two new commands that can be used to create manifest of multi-architecture images.

# Why

This PR helps to "close the circle":

  * Build intel, arm, ppc, s390x images with buildah, podman
  * Push images to a registry
  * Create a multi-architecture manifest that references all these images

This PR takes care of the latest point.

# Create manifest
This is done with the `skopeo manifest create` command.

The command takes as parameter a series of container images and produces a manifest list referencing them. The manifest is printed to the stdout.

Note well:
  * The only transport format supported is the `docker://` one. I don't think it makes sense to support the other ones.
  * The images must be on the same container registry (this is explained below).

# Push manifest to remote registry

This is done with the `skopeo manifest push` command.

The command reads the manifest from the stdin (or from a file) and then pushes it to a remote registry.

Note well:
  * The only transport format supported is the `docker://` one. I don't think it makes sense to support the other ones.
  * The images referenced by the manifest must be on the registry where the manifest is pushed. Otherwise the registry will return an error.

# Missing

I still have to write unit tests and integration tests for the new commands that have been introduced.

# Need help

I create the markdown man pages for the new commands I added. I then ran `make docs-in-container` but the markdown files were not converted to man format.

# Questions

* Should we move the `skopeo manifest-digest` command to be a sub-command of `skopeo manifest`? Basically have `skopeo manifest create|push|digest`


docs/skopeo-manifest-create.1.md
```diff
@@ -0,0 +1,68 @@
...
+    }
+  ]
+}
+```
+
+In this example we create a multi-architecture manifest referncing two images that are inside of the `registry.local.lan`.
+The manifest will reference `hello:0.0.1-amd64` as the Linux x86_64 image, while it will reference `hello:0.0.1-arm64` as the Linux ARM64 image to be used.
+
+## SEE ALSO
+skopeo(1) skopeo-manifest-push(1)
```
TomSweeneyRedHat, line 63: should add skopeo-login and docker-login here too

docs/skopeo-manifest-push.1.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+In this example we create a multi-architecture manifest by leveraging the `skopeo manifest create` command.
+The output produced by this command is then consumed by the `skopeo manifest push` command.
+
+A new manifest list is created on the `registry.local.lan` that is identified as `hello:0.0.1`.
+Users can pull the `registry.local.lan/hello:0.0.1` image and have the container engine automatically obtain the right (architecture and OS wise) image to be used.
+
+## SEE ALSO
+skopeo(1) skopeo-manifest-create(1)
```
TomSweeneyRedHat, line 48: ditto skopeo/docker-login

cmd/skopeo/manifest-create.go
```diff
@@ -0,0 +1,151 @@
...
+	}, opts.retryOpts); err != nil {
+		return details, errors.Wrapf(err, "Error retrieving manifest for image")
+	}
+
+	sys, err := opts.image.newSystemContext()
+	if err != nil {
+		return details, err
+	}
+
+	img, err := image.FromUnparsedImage(ctx, sys, image.UnparsedInstance(src, nil))
```
mtrmac, line 123: This will IIRC silently allow manifest lists and choose an image matching the current architecture.

cmd/skopeo/manifest-create.go
```diff
@@ -0,0 +1,151 @@
...
+	}, opts.retryOpts); err != nil {
+		return details, err
+	}
+
+	manifestDigest, err = manifest.Digest(rawManifest)
+	if err != nil {
+		return details, fmt.Errorf("Error computing manifest digest: %v", err)
+	}
+
+	return manifest.Schema2ManifestDescriptor{
```
mtrmac, line 140: Shouldn’t this allow creating OCI indices, at least if the components are OCI?

cmd/skopeo/manifest-push.go
```diff
@@ -0,0 +1,117 @@
...
+		_, err := io.Copy(bw, os.Stdin)
+		if err != nil {
+			return errors.Wrapf(err, "Cannot read manifest from stdin")
+		}
+		manifest = buff.Bytes()
+	}
+
+	destName := args[0]
+	transport := alltransports.TransportFromImageName(destName)
+	if transport == nil || docker.Transport.Name() != transport.Name() {
```
mtrmac, line 85: This is cleaner to do on the created `destRef` below.

cmd/skopeo/manifest-push.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "Error initializing destination %s", transports.ImageName(destRef))
+	}
+	defer func() {
+		if err := dest.Close(); err != nil {
+			retErr = errors.Wrapf(retErr, " (dest: %v)", err)
+		}
+	}()
+
+	fmt.Println("Pushing manifest")
```
mtrmac, line 110: Please use the `stdout` parameter.

mtrmac (commented):
Thanks!

Overall, Skopeo is mostly a very thin layer over c/image; this is not _clearly_ over the line of introducing Skopeo-specific functionality but it’s getting close.

More importantly, there already is https://github.com/containers/buildah/blob/master/cmd/buildah/manifest.go _and_ https://github.com/containers/podman/tree/master/cmd/podman/manifest . Do we need a third one?

What’s a good workflow for creating multi-arch images, and how do the three tools differ?

RFC: @nalind @mheon @rhatdan 

---

(Review comments only from a very quick skim, the conceptual tool/workflow/triplication question is the more important one.)

mtrmac:
> * Should we move the `skopeo manifest-digest` command to be a sub-command of `skopeo manifest`? Basically have `skopeo manifest create|push|digest`

That command primarily exists for scripts, so _moving_ it and breaking them is not desirable. OTOH it’s also obscure enough that I wouldn’t bother too much with making it easy to find; so I don’t think adding an extra way to reach the functionality is very necessary, but if it can be done cheaply enough, why not.

cmd/skopeo/manifest-create.go
mtrmac: … and that gets us into the business of using a `c/image/manifest.List` abstraction — actually probably extending it first to allow adding members. I.e. the format knowledge in creating this item should be in c/image, not in Skopeo.

mheon:
I would be quite glad to retire Podman's manifest functionality in favor of re-using another implementation, assuming it meets our needs (really, I'd be happy to retire almost all of our image code in favor of a unified implementation). This doesn't really look like something we can use towards that end, though, the APIs look more internal and less reusable.

nalind:
The `manifest` command would be modeled after what we see described at https://docs.docker.com/engine/reference/commandline/manifest/.  Generally, the workflow looks like: build arch-specific images, build the manifest list or OCI index using the arch-specific images, and then push the list or index, either along with the images, or after the images have already been pushed.

Both `buildah manifest`/`podman manifest` and this implementation suffer from what I might call the "ARM is complicated" problem, in that we don't automatically populate the `variant` field in descriptors that describe `arm` or `arm64` images, as examples in https://github.com/opencontainers/image-spec/blob/master/image-index.md do, since I don't know how to get the right value from examining an image's config blob or manifest.  (Looking at it now, I guess defaulting to setting `variant` to `v8` for those arches, and trusting human beings to "know" when to override that default, would be some improvement.)

flavio:
> Overall, Skopeo is mostly a very thin layer over c/image; this is not _clearly_ over the line of introducing Skopeo-specific functionality but it’s getting close.
> 
> More importantly, there already is https://github.com/containers/buildah/blob/master/cmd/buildah/manifest.go _and_ https://github.com/containers/podman/tree/master/cmd/podman/manifest . Do we need a third one?
> 
> What’s a good workflow for creating multi-arch images, and how do the three tools differ?

Ouch, I have to admit I didn't know podman and buildah had the ability to manage manifests. This is something that somehow my mind associated to skopeo.

I think, from a workflow POV, it still makes sense to use buldah/podman to create and push the multi-architecture manifest.

> (Review comments only from a very quick skim, the conceptual tool/workflow/triplication question is the more important one.)

Yes, that's the critical question. I've no problems closing this PR to avoid confusion with the end user workflow.

PR #932, opened 2020-05-28
[nix] Add nix derivation for static builds

hswong3i:
#### What type of PR is this?

> /kind feature

#### What this PR does / why we need it:

Similar PR goes for crun/conmon/libpod/cri-o/etc, too.

Also see:
  - ~~https://github.com/containers/crun/pull/372~~
  - ~~https://github.com/containers/conmon/pull/161~~
  - https://github.com/containers/skopeo/pull/932
  - https://github.com/containers/buildah/pull/2380
  - https://github.com/containers/libpod/pull/6402
  - https://github.com/cri-o/cri-o/pull/3804

Static binaries:
  - [crun-0.13-linux-amd64](https://github.com/alvistack/crun/releases/download/0.13/crun-0.13-linux-amd64)
  - [conmon-v2.0.17-linux-amd64](https://github.com/alvistack/conmon/releases/download/v2.0.17/conmon-v2.0.17-linux-amd64)
  - [skopeo-v1.0.0-linux-amd64](https://github.com/alvistack/skopeo/releases/download/v1.0.0/skopeo-v1.0.0-linux-amd64)
  - [buildah-v1.14.9-linux-amd64](https://github.com/alvistack/buildah/releases/download/v1.14.9/buildah-v1.14.9-linux-amd64)
  - [podman-v1.9.3-linux-amd64](https://github.com/alvistack/libpod/releases/download/v1.9.3/podman-v1.9.3-linux-amd64)
  - [cri-o-v1.17.4-linux-amd64.tar.gz](https://github.com/alvistack/cri-o/releases/download/v1.17.4/cri-o-v1.17.4-linux-amd64.tar.gz)
  - [cri-o-v1.18.1-linux-amd64.tar.gz](https://github.com/alvistack/cri-o/releases/download/v1.18.1/cri-o-v1.18.1-linux-amd64.tar.gz)

Ansible Roles:
  - https://github.com/alvistack/ansible-role-crun
  - https://github.com/alvistack/ansible-role-conmon
  - https://github.com/alvistack/ansible-role-skopeo
  - https://github.com/alvistack/ansible-role-buildah 
  - https://github.com/alvistack/ansible-role-podman
  - https://github.com/alvistack/ansible-role-cri_o

#### How to verify it

```
nix build -f nix/
```

#### Special notes for your reviewer:

Here I skip the btrfs and lvm2 support for static binary, because:
1. btrfs will not support in CentOS 8
2. With skopeo experience both btrfs and lvm2 are not easy for compile as static binary

[...]

PR #1035, opened 2020-09-09
Remove an obsolete documentation of (make binary-static)

mtrmac:
... which no longer works after #932.

This does not add documentation for the current static build approach, nor does it add any other place where `DISABLE_CGO` is documented; both are not tested by CI, and discouraged due to bad integration with the rest of the system.


vrothberg (approved):
LGTM

PR #1037, opened 2020-09-10
feat: show signers info in skopeo inspect

dnwe:
If a remote image being inspected has signatures available, show them in
the inspect output under "Signers". Attempt to display the verified key
fingerprints (if a matching public key is available locally), otherwise
fallback to showing the unverified key IDs.

### DIFF
--- a/cmd/skopeo/inspect.go
+++ b/cmd/skopeo/inspect.go
@@ -10,6 +10,7 @@ import (
 	"github.com/containers/image/v5/docker"
 	"github.com/containers/image/v5/image"
 	"github.com/containers/image/v5/manifest"
+	"github.com/containers/image/v5/signature"
 	"github.com/containers/image/v5/transports"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/skopeo/cmd/skopeo/inspect"
@@ -142,6 +143,29 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error) 
 		return nil
 	}
 
+	var signers []string
+	sig, err := img.Signatures(ctx)
+	if err != nil {
+		return fmt.Errorf("Error reading image signatures: %v", err)
+	}
+	if len(sig) > 0 {
+		mech, err := signature.NewGPGSigningMechanism()
+		if err != nil {
+			return fmt.Errorf("Error preparing local GPG: %v", err)
+		}
+		for _, s := range sig {
+			_, id, err := mech.UntrustedSignatureContents(s)
+			if err == nil {
+				_, kID, err := mech.Verify(s)
+				if err == nil {
+					signers = append(signers, "Key fingerprint: "+kID+" (verified)")
+				} else {
+					signers = append(signers, "Key ID: "+id+" (unverified)")
+				}
+			}
+		}
+	}
+
 	if err := retry.RetryIfNecessary(ctx, func() error {
 		imgInspect, err = img.Inspect(ctx)
 		return err
@@ -153,6 +177,7 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error) 
 		Name: "", // Set below if DockerReference() is known
 		Tag:  imgInspect.Tag,
 		// Digest is set below.
+		Signers:       signers,
 		RepoTags:      []string{}, // Possibly overriden for docker.Transport.
 		Created:       imgInspect.Created,
 		DockerVersion: imgInspect.DockerVersion,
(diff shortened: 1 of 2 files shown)


dnwe:
Popping this up just as a first draft proposal, as I needed the functionality recently so it was quick to code it up like this. However, looking for feedback on the viability of the change being accepted into skopeo in general, and if there’s any thoughts on how the signer info should be formatted etc.

cmd/skopeo/inspect.go
```diff
@@ -142,6 +143,29 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
...
+	}
+	if len(sig) > 0 {
+		mech, err := signature.NewGPGSigningMechanism()
+		if err != nil {
+			return fmt.Errorf("Error preparing local GPG: %v", err)
+		}
+		for _, s := range sig {
+			_, id, err := mech.UntrustedSignatureContents(s)
+			if err == nil {
+				_, kID, err := mech.Verify(s)
```
mtrmac, line 159: (This verifies against `GNUPGHOME`, not the keys in `policy.json`.)

More importantly, it does not verify the signature matches the rest of the image at all! An attacker could attach any valid signature to any other image.

mtrmac (changes requested):
Thanks for the PR.

Overall, I’m _very_ worried that facilities like this will make it possible for users to trust this information when they shouldn’t. E.g. this code says “verified” and does not say what was verified, and the result is not really useful for any purpose other than maybe checking for bit rot of the signature file.

Compare the documentation of `UntrustedSignatureContents` and the comments around its existing indirect use in Skopeo — which might do what you need now.

Also, _fundamentally_, providing any information about contents of validity of signatures on a `skopeo inspect` is mostly useless for trust decisions, because the registry (which is ~by assumption untrusted if signatures are in place) can provide a different image / different signatures on any future actual use of the image (TOCTOU inconsistency). (There are some exceptions to the “useless” claim, but require users to be extremely careful.)

The _real_ signature verification must happen at the time the image is used/“accepted”, i.e. `skopeo copy` / `podman pull`, so overall I’d prefer to focus on making the verification errors, if any, to carry enough context to allow diagnosing/fixing problems. It might well be the case that the error messages are, in some cases, hard to understand; in that case I’d rather improve them than provide a separate diagnostic toolkit.

Yes, there are _some_ cases where the storage/servers are trusted, and where it may be more efficient to verify specific properties of the signatures rather than do a full copy only to discard the image (maybe as a part of a pre-verification of a signing/publishing process), and it does make sense in principle to provide some tools for that, but historically I found it _very_ hard to discourage misuse of untrusted/unverified data by users who “just wanted to make things work”, so Skopeo tends to provide only very minimal tools that allow such trusted-only automation to be build by knowledgeable users (requiring use of local files), but not tools that could be found by a quick search by someone in a hurry.

Due to the TOCTOU concern I don’t think any “verified” claim can ever be made by `skopeo inspect`; as for listing key IDs / fingerprints, why won’t users misuse that information to assume it can be relied upon as well?

dnwe:
@mtrmac thanks — everything you say makes perfect sense and I agree that the current wording and possibly even the inclusion of any key reference at all in the output could inadvertently give a false confidence to a casual observer that the image is trustworthy.

Really the simplistic use case was just to provide a convenient way to determine:

1. does a given image on a given registry (supporting the `/extensions/v2/<namespace>/<name>/signatures/<digest>` extension) have one or more signatures attached and available for verification?
2. if so, what gpg fingerprint needs to be available and locally trusted in order to verify that the remote signature is genuine

mtrmac:
> 1. does a given image on a given registry (supporting the /extensions/v2/<namespace>/<name>/signatures/<digest> extension) have one or more signatures attached and available for verification?

Makes sense; including a count of signatures does seem useful (just to verify that the signatures have been copied — especially the sigstore mechanism can easily lead to non-obvious failures), and probably safe enough.

> 2. if so, what gpg fingerprint needs to be available and locally trusted in order to verify that the remote signature is genuine

That’s, in general, the wrong way around — the user should know who is expected to sign a particular image, and only accept that key or set of keys. “Where do I get a key so that this image can be verified”, if it leads to a key server lookup were anyone can upload keys, provides no security. (Again, specific use cases like build systems may need this information for more reasonable purposes. And I suppose some organizations may want to have a delegation/CA system; ideally that would be best handled in the policy.json enforcement mechanism, although no such code currently exists.)

rhatdan:
@dnwe Are you still working on this?

dnwe:
@rhatdan I’m happy to close this for now and can always open a different PR based on the discussion above.

PR #1047, opened 2020-09-18
Add --registry-token flags to support Bearer token authentication

airadier:
This feature https://github.com/containers/image/pull/842 was merged in containers/image to support authentication using a provided Bearer token (example use case: integration with Harbor webhooks).

This PR adds a flag to the skopeo CLI to provide support for the registry Bearer token.

Signed-off-by: Alvaro Iradier <airadier@gmail.com>

### DIFF
--- a/cmd/skopeo/utils.go
+++ b/cmd/skopeo/utils.go
@@ -93,10 +93,10 @@ func dockerImageFlags(global *globalOptions, shared *sharedImageOptions, flagPre
 		f := fs.VarPF(newOptionalStringValue(&flags.credsOption), credsOptionAlias, "", "Use `USERNAME[:PASSWORD]` for accessing the registry")
 		f.Hidden = true
 	}
+	fs.Var(newOptionalStringValue(&flags.registryToken), flagPrefix+"registry-token", "Provide a Bearer token for accessing the registry")
 	fs.StringVar(&flags.dockerCertPath, flagPrefix+"cert-dir", "", "use certificates at `PATH` (*.crt, *.cert, *.key) to connect to the registry or daemon")
 	optionalBoolFlag(&fs, &flags.tlsVerify, flagPrefix+"tls-verify", "require HTTPS and verify certificates when talking to the container registry or daemon (defaults to true)")
 	fs.BoolVar(&flags.noCreds, flagPrefix+"no-creds", false, "Access the registry anonymously")
-	fs.Var(newOptionalStringValue(&flags.registryToken), flagPrefix+"registry-token", "Provide a Bearer token for accessing the registry")
 	return fs, &flags
 }
 
@@ -133,9 +133,6 @@ func (opts *imageOptions) newSystemContext() (*types.SystemContext, error) {
 	ctx.AuthFilePath = opts.shared.authFilePath
 	ctx.DockerDaemonHost = opts.dockerDaemonHost
 	ctx.DockerDaemonCertPath = opts.dockerCertPath
-	if opts.registryToken.present {
-		ctx.DockerBearerRegistryToken = opts.registryToken.value
-	}
 	if opts.dockerImageOptions.authFilePath.present {
 		ctx.AuthFilePath = opts.dockerImageOptions.authFilePath.value
 	}
@@ -154,6 +151,9 @@ func (opts *imageOptions) newSystemContext() (*types.SystemContext, error) {
 		if err != nil {
 			return nil, err
 		}
+	}
+	if opts.registryToken.present {
+		ctx.DockerBearerRegistryToken = opts.registryToken.value
 	}
 	if opts.noCreds {
 		ctx.DockerAuthConfig = &types.DockerAuthConfig{}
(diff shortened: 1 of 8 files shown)


airadier:
Working on the unittests, expect a new commit soon.

cmd/skopeo/utils_test.go
```diff
@@ -54,6 +54,7 @@ func TestImageOptionsNewSystemContext(t *testing.T) {
 		"--dest-daemon-host", "daemon-host.example.com",
 		"--dest-tls-verify=false",
 		"--dest-creds", "creds-user:creds-password",
+		"--dest-registry-token", "faketoken",
```
TomSweeneyRedHat, line 57: `--src-registry-token` tests too?

docs/skopeo-copy.1.md
```diff
@@ -84,6 +84,10 @@ Existing signatures, if any, are preserved as well.
 
 **--dest-compress-level** _format_ Specifies the compression level to use.  The value is specific to the compression algorithm used, e.g. for zstd the accepted values are in the range 1-20 (inclusive), while for gzip it is 1-9 (inclusive).
 
+**--src-registry-token** _Bearer token_ for accessing the source registry
```
TomSweeneyRedHat, line 87: Please add periods to the end for each of these new entries ala:
```suggestion
**--src-registry-token** _Bearer token_ for accessing the source registry.
```

TomSweeneyRedHat:
OK, maybe I need more tea, but I see the `--src-registry-token` and `--dest-registry-token` in the bash completion and tests, but not the code?  Am I being oblivious (again)?

docs/skopeo-copy.1.md
airadier: Thanks. Adding to some other sentences in the doc that were missing the ending period too (as just copied-paste on of them as an example entry)

airadier:
> OK, maybe I need more tea, but I see the `--src-registry-token` and `--dest-registry-token` in the bash completion and tests, but not the code? Am I being oblivious (again)?

I also noticed this, but the existing unit tests don't check anything with the "src" prefix.

In the implementation of utils.go, the flag parsing can be done without or with a prefix ("src" or "dest" currently), so I guess that testing flag parsing with prefix "whatever" and then testing the parameters `--whatever-registry-token` is enough. The same applies for `--whatever-cert-dir`, `--whatever-creds`, etc.

However, for the bash-completion we have to include both "src" and "dest" and the no prefix case.

cmd/skopeo/utils_test.go
```diff
@@ -182,6 +185,7 @@ func TestImageDestOptionsNewSystemContext(t *testing.T) {
 		DockerDaemonInsecureSkipTLSVerify: true,
 		DirForceCompress:                  true,
 		BigFilesTemporaryDir:              "/srv",
+		DockerBearerRegistryToken:         "faketoken",
```
mtrmac, line 188: Move this just below `DockerAuthConfig`.

completions/bash/skopeo
```diff
@@ -150,6 +154,7 @@ _skopeo_list_repository_tags() {
      --authfile
```
mtrmac, line 154: Shouldn’t `skopeo layers’ have the option added as well?

mtrmac (commented):
Thanks! Looks good, just a few nits to hopefully make future maintenance a bit easier.

mtrmac:
> In the implementation of `utils.go`, the flag parsing can be done without or with a prefix ("src" or "dest" currently), so I guess that testing flag parsing with prefix "whatever" and then testing the parameters `--whatever-registry-token` is enough. 

Yes, that’s the idea. “whatever” just happens to be spelled `dest` in the tests.

completions/bash/skopeo
airadier: Yes, I think I missed that one because I made a search of the "--authfile" all over the files, and it is not in the `skopeo layers` completion either. So I am adding both the missing --authfile, and also I think the --no-creds option makes sense in here too.

cmd/skopeo/utils_test.go
airadier: As commented below, testing `--src` is not necessary, as we are just doing the test with an arbitrary prefix.

cmd/skopeo/utils.go
```diff
@@ -95,6 +96,7 @@ func dockerImageFlags(global *globalOptions, shared *sharedImageOptions, flagPre
 	fs.StringVar(&flags.dockerCertPath, flagPrefix+"cert-dir", "", "use certificates at `PATH` (*.crt, *.cert, *.key) to connect to the registry or daemon")
 	optionalBoolFlag(&fs, &flags.tlsVerify, flagPrefix+"tls-verify", "require HTTPS and verify certificates when talking to the container registry or daemon (defaults to true)")
 	fs.BoolVar(&flags.noCreds, flagPrefix+"no-creds", false, "Access the registry anonymously")
+	fs.Var(newOptionalStringValue(&flags.registryToken), flagPrefix+"registry-token", "Provide a Bearer token for accessing the registry")
```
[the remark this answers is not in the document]
airadier: Agree, fixing the order in utils.go and the tests.

completions/bash/skopeo
mtrmac: Thanks for the other fixes!

mtrmac (approved):
LGTM. Thanks again.

PR #1055, opened 2020-09-24
Makefile: add a local-cross target

nalind:
Add a make target that cross-compiles for a handful of the possible targets that `go tool dist list` can tell us about, and exercise it in CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,4 +24,4 @@ install:
 
 script:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]];   then hack/travis_osx.sh ; fi
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then make vendor && ./hack/tree_status.sh && make check ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then make vendor && ./hack/tree_status.sh && make local-cross && make check ; fi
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -24,4 +24,4 @@ install:
 
 script:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]];   then hack/travis_osx.sh ; fi
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then make vendor && ./hack/tree_status.sh && make check ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then make vendor && ./hack/tree_status.sh && make local-cross check ; fi
```
TomSweeneyRedHat, line 27: IDK, should this be:
```suggestion
  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then make vendor && ./hack/tree_status.sh && make local-cross && make check ; fi
```

nalind: `make` can be told to build multiple targets, and it stops at the first error whatever it's doing, so they're effectively equivalent.

rhatdan:
LGTM
@mtrmac @vrothberg @QiWang19 PTAL

rhatdan:
LGTM
@mtrmac @vrothberg @QiWang19 @TomSweeneyRedHat PTAL

.travis.yml
TomSweeneyRedHat: ah, ty.

TomSweeneyRedHat:
LGTM
but sorry @nalind, needs a rebase.

QiWang19:
LGTM

.travis.yml
nalind: Since I'm rebasing anyway, taking this change.

nalind:
Rebased.

TomSweeneyRedHat:
LGTM

QiWang19:
LGTM

PR #1063, opened 2020-10-02
Update README.md

ironroggers:
Improved text readablity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,22 +23,22 @@ Skopeo works with API V2 container image registries such as [docker.io](https://
 
  Skopeo operates on the following image and repository types:
 
- * containers-storage:docker-reference
+ * <i>containers-storage:docker-reference</i><br>
          An image located in a local containers/storage image store.  Both the location and image store are specified in /etc/containers/storage.conf. (This is  the backend for [Podman](https://podman.io), [CRI-O](https://cri-o.io), [Buildah](https://buildah.io) and friends)
 
- * dir:path
+ * <i>dir:path</i><br>
          An existing local directory path storing the manifest, layer tarballs and signatures as individual files. This is a non-standardized format, primarily useful for debugging or noninvasive container inspection.
 
- * docker://docker-reference
+ * <i>docker://docker-reference</i><br>
          An image in a registry implementing the "Docker Registry HTTP API V2". By default, uses the authorization state in `$XDG_RUNTIME_DIR/containers/auth.json`, which is set using `skopeo login`.
 
- * docker-archive:path[:docker-reference]
+ * <i>docker-archive:path[:docker-reference]</i><br>
          An image is stored in a `docker save`-formatted file.  docker-reference is only used when creating such a file, and it must not contain a digest.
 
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@vrothberg @mtrmac @TomSweeneyRedHat @QiWang19 PTAL

README.md
```diff
@@ -23,22 +23,22 @@ Skopeo works with API V2 container image registries such as [docker.io](https://
 
  Skopeo operates on the following image and repository types:
 
- * containers-storage:docker-reference
+ * <i>containers-storage:docker-reference</i><br>
```
TomSweeneyRedHat, line 26: I like the end result, but I'd rather not add html.  Could you change this to the md equivalents please?

```suggestion
 * **containers-storage:docker-reference**  
```
two astericks on either side of the word, then two white spaces at the end of the line to force `<br>`

TomSweeneyRedHat: and ditto below.

TomSweeneyRedHat:
@ironroggers and I rudely forgot to not thank you for the PR, so THANK YOU!

mtrmac (commented):
If the intent is to give the text formatting some semantic significance, the literal parts of the syntax should use the corresponding Markdown literal syntax, and the variables should use something else (probably italic).

ironroggers:
> @ironroggers and I rudely forgot to not thank you for the PR, so THANK YOU!

Welcome!!!

README.md
```diff
@@ -130,19 +130,19 @@ $ skopeo inspect docker://registry.fedoraproject.org/fedora:latest | jq '.Digest
 ## Copying images
 
 `skopeo` can copy container images between various storage mechanisms, including:
-* Container registries
+* **Container registries**
```
mtrmac, line 133: Is there a semantic point to this? it’s just a list.

ironroggers: The main logic is to have a blog or website look better with perfect spacings and maintaining a good header-text ratio, which is what I tried. Is it good??

mtrmac: I mean, we are not in control of the CSS on this site anyway (and most of us probably wouldn’t _want_ to tinker with it if we could); so the way this content gets longevity and resilience to (GitHub’s and others’) CSS changes is by focusing on the semantics.

ironroggers: Okay!!! I got the point. My mistake

TomSweeneyRedHat:
@mtrmac my suggestion would keep the formatting as @ironroggers had it in the html initially, i.e. a bulllet and italic for `containers-storage:docker-reference` followed by a line break and the description.  

Personally I found it nicer/easier to read, YMMV.

@ironroggers did you mean to close this?

mtrmac:
Reformatting _that list of examples_ would make sense — but to something like `containers-storage:`_docker-reference_ , actually conveying something about the syntax.

Reformatting the list of destinations in the `skopeo copy` — really font weights are a negligible problem compared to the relative placement of that section and its place in the logic of the page, so optimizing font weights before wording/placement/structure would be at the very least premature optimization.

Issue #1010, opened 2020-08-11
Build quay.io/skopeo/stable as multi-arch image


barthy1:
At this moment `quay.io/skopeo/stable` images are amd64 only.
We need s390x and ppc64le image versions to run e2e tests for Tekton on corresponding platforms.
The best case would be to have multi-arch image, which all platforms can use.
The discussion about the same request for quay.io/buildah/stable https://github.com/containers/buildah/issues/2002

TomSweeneyRedHat:
@barthy1 thanks for the issue, especially so with the pointer.  I'm hopeful that we can do this, and will ping @james-crowley and @cevich to make sure they're aware of this request too.

james-crowley:
@TomSweeneyRedHat Thanks for tagging me. Taking a brief look at Skopeo, it looks they are using Travis CI at least for their Mac OS builds. 

Adding building of `s390x` and `ppc64le` images via Travis should be pretty easy. 

If the builds for the images are done via another CI/CD pipeline, we can work with the community to get them access to `s390x` and `ppc64le` support.

Additionally, Travis also had some `arm64` resources as well. Which would be cool to see support for if Skopeo is looking at doing multi-arch images.

barthy1:
Hi @TomSweeneyRedHat 
Are there any updates to get multi-arched skopeo image?
Considering Travis as a great option to run builds on real native hardware, I'd suggest to take into account also [Github actions](https://github.com/crazy-max/ghaction-docker-buildx)
with option to build multi-arch images with emulation. For instance, for skopeo upstream image I was able to verify that it works and produces multi-arch image with `amd64`, `s390x` and `ppc64le` support.
Have a look at the [code](https://github.com/barthy1/skopeo/blob/master/.github/workflows/build-and-push.yml) and [build logs](https://github.com/barthy1/skopeo/runs/1193910387).

TomSweeneyRedHat:
Hi barthy1, high on the list, but have been tending to other fires.

barthy1:
Hi @TomSweeneyRedHat sorry for bothering you again. I'd like to be more specific in my question :) Do you consider to use Github actions to build multi-arch skopeo images? if yes, I can prepare the PR.

TomSweeneyRedHat:
@barthy1 I never turn down an offer of a PR.  If you have a good thought on how to pull this off, I'd be very happy to review it.

barthy1:
[...]

PR #1066, opened 2020-10-07
add multi-arch build of upstream skopeo image via Travis

barthy1:
This PR is connected to https://github.com/containers/skopeo/issues/1010 to get skopeo multi-arch container images.
 
To build multi-arch upstream container image Travis is used, as it has native hardware to run the build for many architectures (amd64, s390x, ppc64le). Docker is used as build and manifest tool. `quay.io/skopeo/upstream:master`, `quay.io/skopeo/stable:v1.2.0` and `quay.io/containers/skopeo:v1.2.0` are specified as target multi-arch images.
Travis config file has 3 stages:
- `local-build` to do the local binary build and test for linux/amd64 and osx, as it was in
the initial code
- `image-build-push` to build and push images for specific architectures (amd64, s390x, ppc64le)
- `multi-arch-manifest-push` to create and push manifest for multi-arch image - `quay.io/skopeo/upstream:master`, `quay.io/skopeo/stable:v1.2.0` and `quay.io/containers/skopeo:v1.2.0`

2 last stages are not done for pull request event or for non-master branch.

2 env variables in Travis are expected - QUAY_USERNAME and QUAY_PASSWORD to push the images to quay.io. 
As a result multi-arch image for 3 architectures is published.

The code is tested using another quay registry to push https://travis-ci.com/github/barthy1/skopeo/builds/194396214

Signed-off-by: Yulia Gaponenko <yulia.gaponenko1@de.ibm.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,27 +1,100 @@
 language: go
 
-matrix:
-  include:
-  -  os: linux
-     sudo: required
-     services:
-       - docker
-  -  os: osx
-
 go:
   - 1.15.x
 
 notifications:
   email: false
 
-install:
-  # Ideally, the (brew update) should not be necessary and Travis would have fairly
-  # frequenstly updated OS images; that’s not been the case historically.
-  # In particular, explicitly unlink python@2, which has been removed from Homebrew
-  # since the last OS image build (as of July 2020), but the Travis OS still
-  # contains it, and it prevents updating of Python 3.
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update && brew unlink python@2 && brew install gpgme ; fi
+env:
+  global:
+    # Multiarch manifest will support architectures from this list. It should be the same architectures, as ones in image-build-push step in this Travis config
+    - MULTIARCH_MANIFEST_ARCHITECTURES="amd64 s390x ppc64le"
+    # env variables for stable image build
+    - STABLE_IMAGE=quay.io/skopeo/stable:v1.2.0
+    - EXTRA_STABLE_IMAGE=quay.io/containers/skopeo:v1.2.0
+    # env variable for upstream image build
+    - UPSTREAM_IMAGE=quay.io/skopeo/upstream:master
 
-script:
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]];   then hack/travis_osx.sh ; fi
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then make vendor && ./hack/tree_status.sh && make local-cross && make check ; fi
+# Just declaration of the image-build-push step with script actions to execute
+x_base_steps:
+  - &image-build-push
+    services:
+      - docker
+    os: linux
+    dist: focal
+    script:
+      # skopeo upstream image build
+      - make -f release/Makefile build-image/upstream
+      # Push image in case if build is started via cron job or code is pushed to master
+      - if [ "$TRAVIS_EVENT_TYPE" == "push" ] || [ "$TRAVIS_EVENT_TYPE" == "cron" ]; then
+           make -f release/Makefile push-image/upstream ;
+        fi
+
+      # skopeo stable image build
+      - make -f release/Makefile build-image/stable
+      # Push image in case if build is started via cron job or code is pushed to master
+      - if [ "$TRAVIS_EVENT_TYPE" == "push" ] || [ "$TRAVIS_EVENT_TYPE" == "cron" ]; then
+           make -f release/Makefile push-image/stable ;
+        fi
+
+# Just declaration of stage order to run
+stages:
+  # Test for local build
+  - local-build
+  # Build and push image for 1 architecture
+  - name: image-build-push
+    if: branch = master
+  # Create and push image manifest to have multiarch image on top of architecture specific images
+  - name: manifest-multiarch-push
+    if: (type IN (push, cron)) AND branch = master
+
+# Actual execution of steps
+jobs:
+  include:
+    # Run 2 local-build steps in parallel for osx and linux/amd64 platforms
+    - stage: local-build
+      <<: *local-build
+      name: local build for osx
+      os: osx
+      install:
+        # Ideally, the (brew update) should not be necessary and Travis would have fairly
+        # frequently updated OS images; that's not been the case historically.
+        # In particular, explicitly unlink python@2, which has been removed from Homebrew
+        # since the last OS image build (as of July 2020), but the Travis OS still
+        # contains it, and it prevents updating of Python 3.
+      - brew update && brew unlink python@2 && brew install gpgme
+      script:
+      - hack/travis_osx.sh
+    - stage: local-build
+      <<: *local-build
+      name: local build for linux
+      os: linux
+      services:
+      - docker
+      script: 
+      - make vendor && ./hack/tree_status.sh && make local-cross && make check
+
+    # Run 3 image-build-push tasks in parallel for linux/amd64, linux/s390x and linux/ppc64le platforms (for upstream and stable)
+    - stage: image-build-push
+      <<: *image-build-push
+      name: images for amd64
+      arch: amd64
+
+    - stage: image-build-push
+      <<: *image-build-push
+      name: images for s390x
+      arch: s390x
+
+    - stage: image-build-push
+      <<: *image-build-push
+      name: images for ppc64le
+      arch: ppc64le
+
+    # Run final task to generate multi-arch image manifests (for upstream and stable)
+    - stage: manifest-multiarch-push
+      os: linux
+      dist: focal
+      script:
+      - make -f release/Makefile push-manifest-multiarch/upstream
+      - make -f release/Makefile push-manifest-multiarch/stable
--- a/release/Makefile
+++ b/release/Makefile
@@ -1,0 +1,40 @@
+## This Makefile is used to publish skopeo container images with Travis CI ##
+## Environment variables, used in this Makefile are specified in .travis.yml
+
+export DOCKER_CLI_EXPERIMENTAL=enabled
+GOARCH ?= $(shell go env GOARCH)
+
+# Build container image of skopeo upstream based on host architecture
+build-image/upstream:
+	docker build -t "${UPSTREAM_IMAGE}-${GOARCH}" contrib/skopeoimage/upstream
+
+# Build container image of skopeo stable based on host architecture
+build-image/stable:
+	docker build -t "${STABLE_IMAGE}-${GOARCH}" contrib/skopeoimage/stable
+
+# Push container image of skopeo upstream (based on host architecture) to image repository
+push-image/upstream:
+	echo "${QUAY_PASSWORD}" | docker login quay.io -u "${QUAY_USERNAME}" --password-stdin
+	docker push "${UPSTREAM_IMAGE}-${GOARCH}"
+
+# Push container image of skopeo stable (based on host architecture) to image default and extra repositories
+push-image/stable:
+	echo "${QUAY_PASSWORD}" | docker login quay.io -u "${QUAY_USERNAME}" --password-stdin
+	docker push "${STABLE_IMAGE}-${GOARCH}"
+	docker tag "${STABLE_IMAGE}-${GOARCH}" "${EXTRA_STABLE_IMAGE}-${GOARCH}"
+	docker push "${EXTRA_STABLE_IMAGE}-${GOARCH}"
+
+# Create and push multiarch image manifest of skopeo upstream
+push-manifest-multiarch/upstream:
+	echo "${QUAY_PASSWORD}" | docker login quay.io -u "${QUAY_USERNAME}" --password-stdin
+	docker manifest create "${UPSTREAM_IMAGE}" $(foreach arch,${MULTIARCH_MANIFEST_ARCHITECTURES}, ${UPSTREAM_IMAGE}-${arch})
+	docker manifest push --purge "${UPSTREAM_IMAGE}"
+
+# Create and push multiarch image manifest of skopeo stable
+push-manifest-multiarch/stable:
+	echo "${QUAY_PASSWORD}" | docker login quay.io -u "${QUAY_USERNAME}" --password-stdin
+	docker manifest create "${STABLE_IMAGE}" $(foreach arch,${MULTIARCH_MANIFEST_ARCHITECTURES}, ${STABLE_IMAGE}-${arch})
+	docker manifest push --purge "${STABLE_IMAGE}"
+	# Push to extra repository
+	docker manifest create "${EXTRA_STABLE_IMAGE}" $(foreach arch,${MULTIARCH_MANIFEST_ARCHITECTURES}, ${EXTRA_STABLE_IMAGE}-${arch})
+	docker manifest push --purge "${EXTRA_STABLE_IMAGE}"
--- a/release/README.md
+++ b/release/README.md
@@ -1,0 +1,39 @@
+# skopeo container image build with Travis
+
+This document describes the details and requirements to build and publish skopeo container images. The images are published as several architecture specific images and multiarch images on top for upstream and stable versions.
+
+The Travis configuration is available at `.travis.yml`.
+
+The code to build and publish images is available at `release/Makefile` and should be used only via Travis.
+
+Travis workflow has 3 major pieces:
+- `local-build` - build and test source code locally on osx and linux/amd64 environments, 2 jobs are running in parallel
+- `image-build-push` - build and push container images with several Travis jobs running in parallel to build images for several architectures (linux/amd64, linux/s390x, linux/ppc64le). Build part is done for each PR, push part is executed only in case of cron job or master branch update.
+- `manifest-multiarch-push` - create and push image manifests, which consists of architecture specific images from previous step. Executed only in case of cron job or master branch update.
+
+## Ways to have full worklow run
+- [cron job](https://docs.travis-ci.com/user/cron-jobs/#adding-cron-jobs)
+- Trigger build from Travis CI
+- Update code in master branch
+
+## Environment variables
+
+Several environment variables are used to customize image names and keep private credentials to push to quay.io repositories.
+
+**Image tags** are specified in environment variable and should be manually updated in case of new release.
+
+- `QUAY_USERNAME` and `QUAY_PASSWORD` are credentials to push image to corresponding quay.io repositories, should have write permissions. These variables should be specified in [Travis](https://docs.travis-ci.com/user/environment-variables/#defining-variables-in-repository-settings).
+
+Variables in .travis.yml
+- `MULTIARCH_MANIFEST_ARCHITECTURES` is a list with architecture shortnames, to apprear in final multiarch manifest. The values should fit to architectures used in the `image-build-push` Travis step.
+- `STABLE_IMAGE`, `EXTRA_STABLE_IMAGE` are image names to publish stable skopeo
+- `UPSTREAM_IMAGE` is image name to publish upstream skopeo
+
+### Values for environment variables
+
+| Env variable                     | Value                            |
+| -------------------------------- |----------------------------------|
+| MULTIARCH_MANIFEST_ARCHITECTURES | "amd64 s390x ppc64le"            |
+| STABLE_IMAGE                     | quay.io/skopeo/stable:v1.2.0     |
+| EXTRA_STABLE_IMAGE               | quay.io/containers/skopeo:v1.2.0 |
+| UPSTREAM_IMAGE                   | quay.io/skopeo/upstream:master |


rhatdan:
@TomSweeneyRedHat @mtrmac @cevich @edsantiago @vrothberg PTAL

If this works, this is great. would love to see this for Buildah and Podman as well.

Would be nice to get something like this to work for the stable branch.

barthy1:
@rhatdan thanks for the answer. I've created this PR to get understanding if this approach is acceptable. If it's ok, I can do the same for stable branch. Just want to go step by step :)

rhatdan:
@barthy1 I want multi arch images available in quay.io and to use manifests correctly.  So I am fine with this.  I want to get feedback @mtrmac @vrothberg @TomSweeneyRedHat though.

Makefile
```diff
@@ -82,6 +82,10 @@ ifeq ($(DISABLE_CGO), 1)
 	override BUILDTAGS = exclude_graphdriver_devicemapper exclude_graphdriver_btrfs containers_image_openpgp
 endif
 
+ARCHES = amd64 s390x ppc64le
+MA_IMAGE := quay.io/skopeo/$(IMAGE_TYPE):ma
```
TomSweeneyRedHat, line 86: IDK, but we may also want to push to quay.io/containers/skopeo:ma

TomSweeneyRedHat:
I love the concept, but unfortunately, I've been buried in other stuff the past couple of days.  I do want to dig into this though.
@edsantiago @cevich if you've a spare cycle or three to eyeball, input would be appreciated.

Makefile
barthy1: I see. I believe `quay.io/containers/skopeo:ma` should be based on the `stable` version, not on the `upstream`, which is target of this PR for now?
In general I'd suggest to add one more env variable - `EXTRA_REPO` (`EXTRA_REPOS?`) and when it exists - retag and push multi-arch image also to the extra quay.io repo(s).

TomSweeneyRedHat: Yes, I'd suggest that the `quay.io/containers/skope:ma` should be based on the `stable` version.  Currently, `quay.io/containers/skopeo:latest` and `quay.io/skopeo/stable:latest` are equivalent.

barthy1:
@TomSweeneyRedHat I've updated the PR, adding image build for stable version and pushing it to `quay.io/containers/skopeo:ma` and `quay.io/skopeo/stable:ma`

barthy1:
@TomSweeneyRedHat any suggestions for this PR?

TomSweeneyRedHat:
Thanks for the ping @barthy1 .  @edsantiago or @cevich, could one of you take a looksee at this please?

edsantiago:
I will not have time to look this week. My recollection from last month is that I had a strong averse reaction to the `:ma` : I think it is not futureproof. Is there any reason not to just build and push a multiarch `skopeo:version`?

barthy1:
@edsantiago thank you for input! I can change tag to any value community prefers to have. `ma` was chosen mostly to use some free meaningful tag. As I was not sure that it's ok to reuse existing tags.. Which tags you would like to see? 
for instance:
`quay.io/skopeo/upstream:master` for upstream 
 `quay.io/skopeo/stable:v1.2.0` and `quay.io/containers/skopeo:v1.2.0` for stable ?

rhatdan:
@barthy1 That works for me.

.travis.yml
```diff
@@ -1,27 +1,82 @@
...
+  - if: arch in (s390x, ppc64le)
+  include:
+    - stage: local-build
+      <<: *local-build
+      name: local build for osx
+      env: SOURCE_TYPE=none
+      os: osx
+      install:
+        # Ideally, the (brew update) should not be necessary and Travis would have fairly
+        # frequenstly updated OS images; that�~@~Ys not been the case historically.
```
edsantiago, line 25: Typo ("frequently", no "s"). And, weird localization characters. Please make sure your editor is set to UTF-8.

edsantiago (commented):
Thank you for your patience. At first glance it LGTM, but (caveat) this was not a thorough review, and I'm not well versed in Travis anyway.

Did you look for ways to refactor the multiple `image-build-push` stanzas? Maybe using `with_items`? I worry that the duplication will become unmaintainable.

And yes, using `:master` and/or `:v-actual-version` sounds ideal to me.

.travis.yml
barthy1: hey, I just copied it from existing text :) fixed

barthy1:
@edsantiago 

> Did you look for ways to refactor the multiple image-build-push stanzas? Maybe using with_items? I worry that the duplication will become unmaintainable.

Iterating via archs is not doable in case when we need to keep order and archs should be applied only to one stage.
I just refactored a little bit the code to have `script` for image-build-push in one place.

barthy1:
@edsantiago hey! any time this week to review the PR?

cevich:
Suggestion 1:  Since both the skopeo binary build and the container images follow after package updates, it probably makes sense to enable some cron-triggered (in Travis settings) builds.  Otherwise the binaries/images could miss out on important dependency security updates if the pace of PR merges (as written) aren't rapid enough.

Suggestion 2:  (forgive me if I missed it) It would be helpful to maintainers, if just the "build" steps also ran for PRs (i.e. only skip the push).  This way any code/script changes which affect the mechanics will be noticed right away, instead of being delayed to branch-push time.

edsantiago:
Quick note first: the CI failure is not your PR's fault; it is being investigated in #1096.

I've been trying to review this today, but it's a big job for me because I don't know anything about Travis. It's now late and my brain is fried; and I'm still not much closer to understanding the full details of this PR. I'm sorry.

barthy1:
@cevich thank you for suggestions!
> Suggestion 1: Since both the skopeo binary build and the container images follow after package updates, it probably makes sense to enable some cron-triggered (in Travis settings) builds. Otherwise the binaries/images could miss out on important dependency security updates if the pace of PR merges (as written) aren't rapid enough.

Good point. You can enable Travis cron jobs in your Travis project settings - https://docs.travis-ci.com/user/cron-jobs/ Nothing is configured in the code. 

> Suggestion 2: (forgive me if I missed it) It would be helpful to maintainers, if just the "build" steps also ran for PRs (i.e. only skip the push). This way any code/script changes which affect the mechanics will be noticed right away, instead of being delayed to branch-push time.

That's just brilliant! You are absolute right, sorry that I didn't do that from the beginning. Now I've added image builds (upstream and stable) to PR pipeline. That's you can see already in Travis setup for my PR - https://travis-ci.org/github/containers/skopeo/builds/741935428 . The push part won't be done, as it's still disabled for PR case.

cevich:
> You can enable Travis cron jobs in your Travis project settings

Yep, I'm happy to set this up.

> That's just brilliant! You are absolute right, sorry that I didn't do that from the beginning.

You wouldn't be saying that if you knew how many years it took me to arrive at that :disappointed: 

Presumably w/o much/any effort, this will also be covered by cron-triggered builds.  I believe they simply execute as if branch-push happened, so the full build + push workflow should happen.

Sorry my Travis-Foo isn't what it use to be

.travis.yml
```diff
@@ -1,27 +1,91 @@
...
+  - name: image-build-push
+    if: branch = master
+  # Create and push image manifest to have multi-arch image on top of arch specific images
+  - name: multi-arch-manifest-push
+    if: type != pull_request AND branch = master
+
+# Actual execution of steps
+jobs:
+  allow_failures:
+  - if: arch in (s390x, ppc64le)
```
cevich, line 45: I'm not sure I understand why failures should be ignored here, shouldn't we want this to fail if something breaks on those arches?

barthy1: just from my experience, many communities don't want to block their x86 builds/releases if build fails for other architectures. So it's more for community comfort. Anyway, the error is visible and not hidden in Travis, so it's always an option to go and fix it.
Also based on my experience, the non-amd64 errors are quite rare. 
So it's up to reviewers, definitely I can remove this statement.

cevich: Ahh okay, that smells like sound reasoning.  Ya, I too am fine leaving it up to others to weigh in on.

mtrmac:
> Suggestion 2: (forgive me if I missed it) It would be helpful to maintainers, if just the "build" steps also ran for PRs (i.e. only skip the push). This way any code/script changes which affect the mechanics will be noticed right away, instead of being delayed to branch-push time.

That’s definitely useful, OTOH I’m worried about the update-test cycle getting longer; it’s already too long for much comfort. Is there a way to measure the time spent running the tests but not waiting for Travis to queue the first one?

Travis reports “total time”, which just seems to sump the duration of all tasks, and probably does not account for parallelism; and a “Ran for” which I can’t interpret (it is now shorter than before, at least in one instance?!), so neither seems to be a

.travis.yml
```diff
@@ -1,27 +1,91 @@
...
+  - &image-build-push
+    services:
+      - docker
+    os: linux
+    script:
+      # skopeo upstream image build and push
+      - export SOURCE_TYPE=upstream; export REPO=skopeo/upstream; export TAG=master
+      - make build-arch-container
+      # Dont's push if it's PR check
+      - if [[ "$TRAVIS_EVENT_TYPE" != "pull_request" ]]; then make push-arch-container ; fi
```
mtrmac, line 24: Could the two conditions here be positive (“is this a cron job / …”) instead of just excluding one type? This might do unexpected pushes if any new event types are added or set up.

.travis.yml
```diff
@@ -1,27 +1,91 @@
...
+      # skopeo stable image build and push
+      - export SOURCE_TYPE=stable; export REPO=skopeo/stable; export TAG=v1.2.0; export EXTRA_REPO=containers/skopeo
```
mtrmac, line 26: The tag value should, if at all possible, be dynamically computed, or it’s going to stay at v1.2.0 forever.

(Alternatively, create a new “what to do on a release” document or script, but there’s high risk it will not be always followed.)

Makefile
```diff
@@ -82,6 +82,14 @@ ifeq ($(DISABLE_CGO), 1)
 	override BUILDTAGS = exclude_graphdriver_devicemapper exclude_graphdriver_btrfs containers_image_openpgp
 endif
 
+ARCHES=amd64 s390x ppc64le
+MA_IMAGE := $(REGISTRY)/$(REPO):$(TAG)
```
mtrmac, line 86: `MA` is non-obvious. Please use something more explicit, maybe just `MULTIARCH` or `MULTI_ARCH`. (everywhere)

mtrmac, line 86: Maybe the caller could just set `MA_IMAGE` and `EXTRA_MA_IMAGE` (or whatever they end up being named); I don’t think any of the `REGISTRY`/`REPO`/`TAG` individual variables make anything simpler, it’s just more variables to track.

.travis.yml
mtrmac: How would ignoring failures actually work with the multi-arch images? If we require x86 to succeed and ignore a ppc failure, would we still run the multi-arch push and now end up with a multi-arch image that contains a fresh x86 version and an obsolete ppc version (from some previous successful ppc run)? That seems very undesirable.

.travis.yml
mtrmac, line 26: Does it make sense to run `stable` jobs on PRs at all? Almost none of their inputs come from the upstream repo (apart from the `Dockerfile` and `Makefile`, sure.)

@TomSweeneyRedHat you’re the expert on these. WDYT?

Makefile
```diff
@@ -135,6 +143,30 @@ local-cross: bin/skopeo.darwin.amd64 bin/skopeo.linux.arm bin/skopeo.linux.arm64
 build-container:
 	${CONTAINER_RUNTIME} build ${BUILD_ARGS} -t "$(IMAGE)" .
 
+# Build container specific to arch
+build-arch-container:
+	${CONTAINER_RUNTIME} build -t "${ARCH_IMAGE}" contrib/skopeoimage/${SOURCE_TYPE}
```
mtrmac, line 148: The name suggests that this is related to `build-container` above, which is very different (this creates a container for users to consume, the above creates a private utility container used for in-container builds).

I don’t have a good suggestion. `user-container:`? Or just write good comments for both of the targets? (“Build container specific to arch” does not say much that the target name doesn’t imply already)

Makefile
```diff
@@ -135,6 +143,30 @@ local-cross: bin/skopeo.darwin.amd64 bin/skopeo.linux.arm bin/skopeo.linux.arm64
...
+
+# Push container specific to arch
+push-arch-container:
+	echo "${QUAY_PASSWORD}" | ${CONTAINER_RUNTIME} login ${REGISTRY} -u "${QUAY_USERNAME}" --password-stdin
```
mtrmac, line 152: I’m not too happy about this in the `Makefile`, especially when it fails in non-obvious ways if run without all the variables set.

If it makes sense to split code between Travis and `Makefile`, the `Makefile` targets should probably be useful for interactive use, in which case the user is presumably already logged in.

Would it be better placed in the Travis automation? I’m not sure at all.

---

`QUAY_{USERNAME,PASSWORD}` vs. `REGISTRY` is inconsistent.

To me, that’s another reason to move the login into Travis, and have Travis set full `…_IMAGE` values instead of individual components.

Makefile
```diff
@@ -135,6 +143,30 @@ local-cross: bin/skopeo.darwin.amd64 bin/skopeo.linux.arm bin/skopeo.linux.arm64
...
+	${CONTAINER_RUNTIME} push "${ARCH_IMAGE}"
+ifneq ($(EXTRA_REPO),)
+	${CONTAINER_RUNTIME} tag "${ARCH_IMAGE}" "${EXTRA_ARCH_IMAGE}"
+	${CONTAINER_RUNTIME} push "${EXTRA_ARCH_IMAGE}"
+endif
+
+# Push manifest to make multi-arch image
+push-ma-manifest:
+	echo "${QUAY_PASSWORD}" | docker login ${REGISTRY} -u "${QUAY_USERNAME}" --password-stdin
+	sudo chmod 0755 /etc/docker
```
mtrmac, line 162: ??!

Why is it even necessary?

It’s really not OK to risk that this will happen without prompting and to someone who runs this more or less by accident.

Makefile
```diff
@@ -135,6 +143,30 @@ local-cross: bin/skopeo.darwin.amd64 bin/skopeo.linux.arm bin/skopeo.linux.arm64
...
+	DOCKER_CLI_EXPERIMENTAL="enabled" docker manifest create "${MA_IMAGE}" $(foreach arch,${ARCHES}, ${MA_IMAGE}-${arch})
```
mtrmac, line 163: Podman experts, should we be using Docker for all of this?

Makefile
```diff
@@ -82,6 +82,14 @@ ifeq ($(DISABLE_CGO), 1)
 	override BUILDTAGS = exclude_graphdriver_devicemapper exclude_graphdriver_btrfs containers_image_openpgp
 endif
 
+ARCHES=amd64 s390x ppc64le
```
mtrmac, line 85: For pushing, this is set here in Makefile, but for the actual builds, this is set in Travis. Could there be one source for this? Or at least write comments in both places that link the two.

.travis.yml
```diff
@@ -1,27 +1,91 @@
...
+      <<: *image-build-push
+      name: image for ppc64le
+      arch: ppc64le
+
+    # Run final task to generate multi-arch image manifests (for upstream and stable)
+    - stage: multi-arch-manifest-push
+      os: linux
+      script:
+      - SOURCE_TYPE=upstream REPO=skopeo/upstream TAG=master make push-ma-manifest
+      - SOURCE_TYPE=stable REPO=skopeo/stable TAG=v1.2.0 EXTRA_REPO=containers/skopeo make push-ma-manifest
```
mtrmac, line 91: The REPO/… values are set in two places; can that be consolidated somehow so that they can’t get out of sync?

.travis.yml
```diff
@@ -1,27 +1,91 @@
...
-  # since the last OS image build (as of July 2020), but the Travis OS still
-  # contains it, and it prevents updating of Python 3.
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update && brew unlink python@2 && brew install gpgme ; fi
-
-script:
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]];   then hack/travis_osx.sh ; fi
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then make vendor && ./hack/tree_status.sh && make local-cross && make check ; fi
+env:
+  global:
+    - REGISTRY=quay.io
```
mtrmac, line 11: The various naming decisions are now spread over this file. I suppose it’s not _too much_ to ask for anyone reading it to read all of it, but still it would be nice if it were possible to consolidate all of them in a single place — maybe here as global environment settings?

mtrmac, line 11: Also, of this script requires some secrets/whatever to be configured in Travis, could this be explicitly documented somewhere?

mtrmac (commented):
I’m afraid I don’t know much about Travis, so mostly focusing on maintainability over time.

.travis.yml
barthy1: ok, I will remove this questionable check from the pipeline

.travis.yml
barthy1: At this moment Travis has 4 cases here `push`, `pull_request`, `api`, `cron`. I can add the check for cron and push type

.travis.yml
barthy1: it's a little bit tricky. as we have several architectures (and builds) + multi-arch manifest creation after. In general it's possible for `image-build-push` step, I mean just build container image and run it to get the version, but not doable for `multi-arch-manifest-push` as no build is done there and no container image is available to check the version.
I've follow the second option - with “what to do on a release” document, thank you for input

Makefile
barthy1: yeah, probably you are right :) Will think how to reduce env variable number and make them self-understandable

Makefile
barthy1: hmmm, ok, will think how to cover in comments the differrence. And yes, you are right, names should highlight this difference

Makefile
barthy1: > Would it be better placed in the Travis automation? I’m not sure at all.

Well, Travis has just 1 `.travis.yml` file to interact with project. Best practices here is to call make targets or just scripts from it, to avoid making it too complex. However I see your point here, we can have just separate directory with multi-arch build script there to call it from Travis. What do you think?

> To me, that’s another reason to move the login into Travis, and have Travis set full …_IMAGE values instead of individual components.

Still, I'd recommend to keep login in script, but yeah, probably env variables setup can be done in Travis

Makefile
barthy1: that was a bug for docker https://github.com/moby/moby/pull/37847
it is fixed in later version, but docker `18.06.0-ce` still has it

Makefile
barthy1: docker is default container tool for travis. podman and buildah, as far as I know, are questionable to install with some hacks (especially for non amd64 archs), see https://travis-ci.community/t/podman-libpod-support/6823/6. I can play with it, if it's hard requirement, however would prefer to use docker, as it's more stable and verified approach for Travis

Makefile
barthy1: good point, I cannot have just one env variable to specify it, but definitely will add information about linkage

.travis.yml
barthy1: yep, I will move them to global variables

barthy1: OK, I guess some special doc about the release is really required, I will work on it

Makefile
mtrmac: I don’t have a strong opinion at all.

Would it make sense to split upstream build-automation scripts (compare `hack/travis_osx.sh`) from a `Makefile` targeted at all contributors running locally? I.e. move the `push-arch-container` task completely out of `Makefile` into `hack/travis_push_arch_container.sh` or something like that?

What do others prefer? @TomSweeneyRedHat ? @vrothberg ?

Makefile
mtrmac: Thanks; that _definitely_ needs a comment, and it’s again a thing I’d much prefer to push out of the `Makefile` to `.travis.yml` or a a Travis-specific script.

Makefile
barthy1: actually I was thinking about having absolutely separate `release/Makefile` only for Travis images build. And to have in `release` directory also README.md with information about env variables, how build is done, etc..

.travis.yml
barthy1: added check for push and cron

Makefile
barthy1: Used `MULTIARCH` everywhere

Makefile
barthy1: moved to travis.yaml

.travis.yml
barthy1: created separate test doc about build details

Makefile
barthy1: now separate Makefile for Travis build

Makefile
barthy1: In the last version I used not default latest Ubuntu(focal) as Travis env, and it actually resolved this bug.

Makefile
TomSweeneyRedHat: I'd prefer podman, but @barthy1 is right, travis at this point defaults pretty hard to docker from what I've seen.  @edsantiago might have other thoughts.

rhatdan:
I am going to merge, and we can look at the Podman versus Docker in a different PR.

barthy1:
@rhatdan thank you for merging! Don't forget to specify QUAY_USERNAME and QUAY_PASSWORD in Travis settings for this repo -> https://docs.travis-ci.com/user/environment-variables/#defining-variables-in-repository-settings

rhatdan:
@TomSweeneyRedHat Do you know what these are?

barthy1:
Just in case - `QUAY_USERNAME` and `QUAY_PASSWORD` are used to push container images to `quay.io` registry. Usually you can get user and password for registry login via `quay.io` robot account -> https://docs.quay.io/glossary/robot-accounts.html
For skopeo case you need robot account with write permissions to `quay.io/skopeo` and `quay.io/containers`

cevich:
> Don't forget to specify QUAY_USERNAME and QUAY_PASSWORD

@rhatdan I know what they are, I'll see if I can set them...

cevich:
...rats.  Two major problems:

1. I don't have access to either `skopeo` or `containers`, @TomSweeneyRedHat can you invite me?
2. Quay hard-codes the repository name as a prefix to the robot user name.  i.e. It's impossible to use the same account for both `skopeo` and `containers` :disappointed:

cevich:
...double checked.  Indeed, the robot accounts are all namespaced to the user or orginization account that owns them.  There's no way around this :cry:   i.e. no matter what "name" I give the robot, quay will create `skopeo+name` and `containers+name` :sob: 

I'll open up a PR to split `QUAY_USERNAME` and `QUAY_PASSWORD` across the two quay namespaces.  I'm pretty sure that's the only option.  Once I get permissions to actually create those robot accounts, I'll setup the env. vars. in travis.

cevich:
Opened https://github.com/containers/skopeo/pull/1104

TomSweeneyRedHat:
@cevich I've just added you to quay.io/containers  /buildah /podman and /skopeo.  Why 2 when I can give you 4?

I'm hoping you can manage this without dropping a password in clear text, the other issues you noted sound "fun".

@barthy1 thanks again and again for putting this all together.

cevich:
Lol @TomSweeneyRedHat okay, why not.  Yes, I'm happy to manage this stuff as I've got my fingers in all the automation pies anyway.  Yummy.

Issue #642, opened 2019-04-26
Support Go templates in 'skopeo inspect'


debarshiray:
Go templates are easier to parse from shell scripts, than JSON. Would be nice if 'skopeo inspect' grew support for them.

rhatdan:
Podman inspect has support for this now, we should port it over.

PR #1070, opened 2020-10-08
Add --format option to skopeo inspect

rhatdan:
Fixes https://github.com/containers/skopeo/issues/642

### DIFF
--- a/cmd/skopeo/inspect.go
+++ b/cmd/skopeo/inspect.go
@@ -4,8 +4,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"os"
 	"strings"
+	"text/tabwriter"
+	"text/template"
 
+	"github.com/containers/common/pkg/report"
 	"github.com/containers/common/pkg/retry"
 	"github.com/containers/image/v5/docker"
 	"github.com/containers/image/v5/image"
@@ -23,6 +27,7 @@ type inspectOptions struct {
 	global    *globalOptions
 	image     *imageOptions
 	retryOpts *retry.RetryOptions
+	format    string
 	raw       bool // Output the raw manifest instead of parsing information about the image
 	config    bool // Output the raw config blob instead of parsing information about the image
 }
@@ -45,13 +50,16 @@ Supported transports:
 
 See skopeo(1) section "IMAGE NAMES" for the expected format
 `, strings.Join(transports.ListNames(), ", ")),
-		RunE:    commandAction(opts.run),
-		Example: `skopeo inspect docker://docker.io/fedora`,
+		RunE: commandAction(opts.run),
+		Example: `skopeo inspect docker://registry.fedoraproject.org/fedora
+  skopeo inspect --config docker://docker.io/alpine
+  skopeo inspect  --format "Name: {{.Name}} Digest: {{.Digest}}" docker://registry.access.redhat.com/ubi8`,
 	}
 	adjustUsage(cmd)
 	flags := cmd.Flags()
 	flags.BoolVar(&opts.raw, "raw", false, "output raw manifest or configuration")
 	flags.BoolVar(&opts.config, "config", false, "output configuration")
+	flags.StringVarP(&opts.format, "format", "f", "", "Format the output to a Go template")
 	flags.AddFlagSet(&sharedFlags)
 	flags.AddFlagSet(&imageFlags)
 	flags.AddFlagSet(&retryFlags)
@@ -63,12 +71,16 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error) 
 		rawManifest []byte
 		src         types.ImageSource
 		imgInspect  *types.ImageInspectInfo
+		data        []interface{}
 	)
 	ctx, cancel := opts.global.commandTimeoutContext()
 	defer cancel()
 
 	if len(args) != 1 {
 		return errors.New("Exactly one argument expected")
+	}
+	if opts.raw && opts.format != "" {
+		return errors.New("raw output does not support format option")
 	}
 	imageName := args[0]
 
@@ -90,7 +102,7 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error) 
 
 	defer func() {
 		if err := src.Close(); err != nil {
-			retErr = errors.Wrapf(retErr, "could not close image")
+			retErr = errors.Wrapf(retErr, fmt.Sprintf("(could not close image: %v) ", err))
 		}
 	}()
 
@@ -104,8 +116,9 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error) 
 	if opts.raw && !opts.config {
 		_, err := stdout.Write(rawManifest)
 		if err != nil {
-			return errors.Wrapf(err, "Error writing manifest to standard output")
+			return fmt.Errorf("Error writing manifest to standard output: %v", err)
 		}
+
 		return nil
 	}
 
@@ -135,7 +148,17 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error) 
 		}, opts.retryOpts); err != nil {
 			return errors.Wrapf(err, "Error reading OCI-formatted configuration data")
 		}
-		err = json.NewEncoder(stdout).Encode(config)
+		if report.IsJSON(opts.format) || opts.format == "" {
+			var out []byte
+			out, err = json.MarshalIndent(config, "", "    ")
+			if err == nil {
+				fmt.Fprintf(stdout, "%s\n", string(out))
+			}
+		} else {
+			row := "{{range . }}" + report.NormalizeFormat(opts.format) + "{{end}}"
+			data = append(data, config)
+			err = printTmpl(row, data)
+		}
 		if err != nil {
 			return errors.Wrapf(err, "Error writing OCI-formatted configuration data to standard output")
 		}
@@ -187,10 +210,30 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error) 
 			logrus.Warnf("Registry disallows tag list retrieval; skipping")
 		}
 	}
-	out, err := json.MarshalIndent(outputData, "", "    ")
+	if report.IsJSON(opts.format) || opts.format == "" {
+		out, err := json.MarshalIndent(outputData, "", "    ")
+		if err == nil {
+			fmt.Fprintf(stdout, "%s\n", string(out))
+		}
+		return err
+	}
+	row := "{{range . }}" + report.NormalizeFormat(opts.format) + "{{end}}"
+	data = append(data, outputData)
+	return printTmpl(row, data)
+}
+
+func inspectNormalize(row string) string {
+	r := strings.NewReplacer(
+		".ImageID", ".Image",
+	)
+	return r.Replace(row)
+}
+
+func printTmpl(row string, data []interface{}) error {
+	t, err := template.New("skopeo inspect").Parse(row)
 	if err != nil {
 		return err
 	}
-	fmt.Fprintf(stdout, "%s\n", string(out))
-	return nil
+	w := tabwriter.NewWriter(os.Stdout, 8, 2, 2, ' ', 0)
+	return t.Execute(w, data)
 }
--- a/cmd/skopeo/inspect.go
+++ b/cmd/skopeo/inspect.go
@@ -90,7 +90,7 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error) 
 
 	defer func() {
 		if err := src.Close(); err != nil {
-			retErr = errors.Wrapf(retErr, fmt.Sprintf("(could not close image: %v) ", err))
+			retErr = errors.Wrapf(retErr, "could not close image")
 		}
 	}()
 
@@ -104,14 +104,14 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error) 
 	if opts.raw && !opts.config {
 		_, err := stdout.Write(rawManifest)
 		if err != nil {
-			return fmt.Errorf("Error writing manifest to standard output: %v", err)
+			return errors.Wrapf(err, "Error writing manifest to standard output")
 		}
 		return nil
 	}
 
 	img, err := image.FromUnparsedImage(ctx, sys, image.UnparsedInstance(src, nil))
 	if err != nil {
-		return fmt.Errorf("Error parsing manifest for image: %v", err)
+		return errors.Wrapf(err, "Error parsing manifest for image")
 	}
 
 	if opts.config && opts.raw {
@@ -124,7 +124,7 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error) 
 		}
 		_, err = stdout.Write(configBlob)
 		if err != nil {
-			return fmt.Errorf("Error writing configuration blob to standard output: %v", err)
+			return errors.Wrapf(err, "Error writing configuration blob to standard output")
 		}
 		return nil
 	} else if opts.config {
@@ -137,7 +137,7 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error) 
 		}
 		err = json.NewEncoder(stdout).Encode(config)
 		if err != nil {
-			return fmt.Errorf("Error writing OCI-formatted configuration data to standard output: %v", err)
+			return errors.Wrapf(err, "Error writing OCI-formatted configuration data to standard output")
 		}
 		return nil
 	}
@@ -164,7 +164,7 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error) 
 	}
 	outputData.Digest, err = manifest.Digest(rawManifest)
 	if err != nil {
-		return fmt.Errorf("Error computing manifest digest: %v", err)
+		return errors.Wrapf(err, "Error computing manifest digest")
 	}
 	if dockerRef := img.Reference().DockerReference(); dockerRef != nil {
 		outputData.Name = dockerRef.Name()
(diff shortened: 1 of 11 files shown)


rhatdan:
@mtrmac @vrothberg @QiWang19 @TomSweeneyRedHat @ashley-cui PTAL

cmd/skopeo/inspect.go
```diff
@@ -52,6 +54,7 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
 	flags := cmd.Flags()
 	flags.BoolVar(&opts.raw, "raw", false, "output raw manifest or configuration")
 	flags.BoolVar(&opts.config, "config", false, "output configuration")
+	flags.StringVarP(&opts.format, "format", "f", "", "Format the output to a Go template or json")
```
mtrmac, line 57: This does not really explain how to use the value.

cmd/skopeo/inspect.go
```diff
@@ -187,10 +201,31 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
...
 		return err
 	}
-	fmt.Fprintf(stdout, "%s\n", string(out))
-	return nil
+	return outputFormat(outputData, opts.format)
+}
+
+func outputFormat(data interface{}, format string) error {
+	var out formats.Writer
+	if format == "json" || format == "" { // "" for backwards compat
```
mtrmac, line 216: We don’t need the `""` case.

cmd/skopeo/inspect.go
```diff
@@ -187,10 +201,31 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
...
+		out = formats.StdoutTemplate{Output: data, Template: inspectFormat(format)}
+	}
+	return out.Out()
+}
+
+func inspectFormat(row string) string {
+	r := strings.NewReplacer(
+		"{{.Id}}", formats.IDString,
+		".Src", ".Source",
+		".Dst", ".Destination",
```
mtrmac, line 228: What does all of this do? It doesn’t match any of the fields actually present in the data.

cmd/skopeo/inspect.go
```diff
@@ -115,14 +118,21 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
...
+			err        error
+		)
 		if err := retry.RetryIfNecessary(ctx, func() error {
 			configBlob, err = img.ConfigBlob(ctx)
 			return err
 		}, opts.retryOpts); err != nil {
 			return errors.Wrapf(err, "Error reading configuration blob")
 		}
-		_, err = stdout.Write(configBlob)
+		if len(opts.format) == 0 {
```
mtrmac, line 131: If this affects `--config --raw`, it should affect a non-config `--raw` as well.

cmd/skopeo/inspect.go
```diff
@@ -187,10 +201,31 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
...
+}
+
+func outputFormat(data interface{}, format string) error {
+	var out formats.Writer
+	if format == "json" || format == "" { // "" for backwards compat
+		out = formats.JSONStruct{Output: data}
+	} else {
+		out = formats.StdoutTemplate{Output: data, Template: inspectFormat(format)}
+	}
+	return out.Out()
```
mtrmac, line 221: What does the `formats.` package do? AFAICT the only non-trivial effect is the `humanNewLine` call. Is that useful enough to add this dependency, complicating mass rebases? (Alternatively, moving the subpackage to c/common would mean no new dependencies are introduced.)

cmd/skopeo/inspect.go
mtrmac, line 216: It’s a bit surprising to have a `json` option, when the output is JSON already by default, and for that option to mean “indent the output nicely”.

Not impossible, so if there is a strong precedent, I guess… still, a bit disappointing.

docs/skopeo-inspect.1.md
```diff
@@ -4,38 +4,57 @@
 skopeo\-inspect - Return low-level information about _image-name_ in a registry
```
mtrmac, line 4: (Note to self: I haven’t reviewed this.)

mtrmac, line 4: Also completions/bash needs updating.

cmd/skopeo/inspect.go
rhatdan: We don't explain it in any of our tools, and yet people seem to understand it.

cmd/skopeo/inspect.go
rhatdan: Since we don't have backwards compatibility I am removing this check altogether.

rhatdan: I am removing, this is just stolen code from Buildah, which had this previously.

cmd/skopeo/inspect.go
rhatdan: There is an example in the man page.

cmd/skopeo/inspect.go
rhatdan: Removed

cmd/skopeo/inspect.go
rhatdan: As far as I can see the --raw option seems to have no effect currently if the --config flag is not specified.  Is it supposed to?

With the current code these commands have the same output.
```
/bin/skopeo  inspect --raw=true --raw docker://docker.io/golang 
/bin/skopeo  inspect --raw=false --raw docker://docker.io/golang 
```

cmd/skopeo/inspect.go
rhatdan: https://github.com/containers/common/pull/320

completions/bash/skopeo
```diff
@@ -77,6 +77,7 @@ _skopeo_inspect() {
      --cert-dir
      --retry-times
      --registry-token
+     --format
```
TomSweeneyRedHat, line 80: nit
could you move this up a few lines to alphabetize?

docs/skopeo-inspect.1.md
```diff
@@ -4,38 +4,57 @@
...
+
+**--cert-dir** _path_
+
+Use certificates at _path_ (\*.crt, \*.cert, \*.key) to connect to the registry.
 
-  _image-name_ name of image to retrieve information about
+**--config**
 
-  **--config** output configuration in OCI format, default is to format in JSON
+Output configuration in OCI format, default is to format in JSON
```
TomSweeneyRedHat, line 28:
```suggestion
Output configuration in OCI format, default is to format in JSON format.
```

docs/skopeo-inspect.1.md
```diff
@@ -4,38 +4,57 @@
...
-  **--tls-verify** _bool-value_ Require HTTPS and verify certificates when talking to container registries (defaults to true).
+**--registry-token** _Bearer token_
 
-  **--no-creds** _bool-value_ Access the registry anonymously.
+Registry token for accessing the registry.
 
-  **--registry-token** _Bearer token_ for accessing the registry.
+**--retry-times**
+
+The number of times to retry, retry wait time will be exponentially increased based on the number of failed attempts.
```
TomSweeneyRedHat, line 53:
```suggestion
The number of times to retry; retry wait time will be exponentially increased based on the number of failed attempts.
```

cmd/skopeo/inspect.go
mtrmac: To be specific, at least please put the JSON value in quotes (`or "json"`) to make it clear that it is a literal to be used, not a incorrectly-lowercased reference to a generic concept.

mtrmac: … actually, the `json` value was completely removed? So, this description needs to be updated but it’s easy now.

cmd/skopeo/inspect.go
```diff
@@ -22,6 +23,7 @@ import (
 type inspectOptions struct {
 	global    *globalOptions
 	image     *imageOptions
+	format    string
 	retryOpts *retry.RetryOptions
```
mtrmac, line 27: (Absolutely non-blocking nit: move the `format` option one line down, below the “use shared options” sets.)

cmd/skopeo/inspect.go
mtrmac: > With the current code these commands have the same output.

That’s not what I see:
```console
% bin/skopeo --override-os linux inspect --raw=true docker://docker.io/golang  
{"manifests":[{"digest":"sha256:097…

% bin/skopeo --override-os linux inspect --raw=false docker://docker.io/golang 
{
    "Name": "docker.io/library/golang",
…
```

cmd/skopeo/inspect.go
```diff
@@ -90,7 +93,7 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
 
 	defer func() {
 		if err := src.Close(); err != nil {
-			retErr = errors.Wrapf(retErr, fmt.Sprintf("(could not close image: %v) ", err))
+			retErr = errors.Wrapf(retErr, "could not close image")
```
mtrmac, line 96: This loses the `err` value.

cmd/skopeo/inspect.go
```diff
@@ -135,9 +145,13 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
 		}, opts.retryOpts); err != nil {
 			return errors.Wrapf(err, "Error reading OCI-formatted configuration data")
 		}
-		err = json.NewEncoder(stdout).Encode(config)
+		if len(opts.format) == 0 {
+			err = json.NewEncoder(stdout).Encode(config)
+		} else {
+			err = outputFormat(config, opts.format)
```
mtrmac, line 151: Hum, output is supposed to go the `stdout` parameter, not to `os.Stdout` (to allow writing in-process unit tests), but the `formats` package does not support that.

docs/skopeo-inspect.1.md
mtrmac: I mean, the _format_ is not the point, both are JSON; option causes the _value_ to be printed to be the result of converting the config to OCI standard, instead using “the usual inspect outcome” which mixes input/tag/config/… data in a single object.

Especially now that `--format` can cause arbitrary subsets of the data to be formatted as arbitrary text, with or without `--config`, it’s even more important to talk about values (that are either printed in JSON or input to the template).

docs/skopeo-inspect.1.md
```diff
@@ -4,38 +4,57 @@
...
+**--no-creds**
 
-  **--creds** _username[:password]_ for accessing the registry.
+Access the registry anonymously.
 
-  **--cert-dir** _path_ Use certificates at _path_ (\*.crt, \*.cert, \*.key) to connect to the registry.
+**--raw**
 
-  **--retry-times**  the number of times to retry, retry wait time will be exponentially increased based on the number of failed attempts.
+Output raw manifest, default is to format in JSON
```
mtrmac, line 45: This is incorrect for `--config --raw`.

go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	github.com/stretchr/testify v1.6.1
 	github.com/syndtr/gocapability v0.0.0-20180916011248-d98352740cb2
 	go4.org v0.0.0-20190218023631-ce4c26f7be8e // indirect
-	golang.org/x/text v0.3.3 // indirect
 	gopkg.in/yaml.v2 v2.3.0
 )
+
+replace github.com/containers/common => github.com/rhatdan/common v0.10.1-0.20201009102218-77bb0d1ee370
```
mtrmac, line 28: Note to self: This needs to be replaced before merging, to ensure we keep up with future fixes to c/common.

docs/skopeo-inspect.1.md
mtrmac: Reviewed now. (Would have been a bit easier if the sorting of options were a separate commit.)

completions/bash/skopeo
rhatdan: Fixed

go.mod
rhatdan: This no longer exists.

cmd/skopeo/inspect.go
rhatdan: Fixed

cmd/skopeo/inspect.go
rhatdan: Not sure what you mean here?

cmd/skopeo/inspect.go
rhatdan: Yes you are right.  My mistake.

cmd/skopeo/inspect.go
rhatdan: Removed json.

cmd/skopeo/inspect.go
mtrmac:
```go
type inspectOptions struct {
	global    *globalOptions
	image     *imageOptions
	retryOpts *retry.RetryOptions
	format    string
```

cmd/skopeo/inspect.go
```diff
@@ -135,9 +148,18 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
 		}, opts.retryOpts); err != nil {
 			return errors.Wrapf(err, "Error reading OCI-formatted configuration data")
 		}
-		err = json.NewEncoder(stdout).Encode(config)
+		if report.IsJSON(opts.format) || opts.format == "" {
+			out, err := json.MarshalIndent(config, "", "    ")
+			if err == nil {
+				fmt.Fprintf(stdout, "%s\n", string(out))
+			}
```
QiWang19, line 155: This error will not be reported if not nil?

rhatdan: Fixed

cmd/skopeo/inspect.go
rhatdan: Ok, but why?

rhatdan: BTW I did it, but I am curious why you want this moved?

rhatdan:
@containers/podman-maintainers PTAL
@QiWang19 @vrothberg @mtrmac PTAL
This should be ready to go in, and then we can cut a release.

TomSweeneyRedHat:
needs a rebase @rhatdan

completions/bash/skopeo
TomSweeneyRedHat: You said fixed, but this review isn't reflecting that.

TomSweeneyRedHat:
One missed nit, otherwise LGTM.

QiWang19 (commented):
LGTM

cmd/skopeo/inspect.go
```diff
@@ -182,15 +205,35 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
...
+			fmt.Fprintf(stdout, "%s\n", string(out))
+		}
+		return err
+	}
+	row := "{{range . }}" + report.NormalizeFormat(opts.format) + "{{end}}"
+	data = append(data, outputData)
+	return printTmpl(row, data)
+}
+
+func inspectNormalize(row string) string {
```
mtrmac, line 225: Does anything call this?

cmd/skopeo/inspect.go
```diff
@@ -135,9 +148,19 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
 		}, opts.retryOpts); err != nil {
 			return errors.Wrapf(err, "Error reading OCI-formatted configuration data")
 		}
-		err = json.NewEncoder(stdout).Encode(config)
+		if report.IsJSON(opts.format) || opts.format == "" {
```
mtrmac, line 151: The two formatting blocks are the same; is there a reason not to share?

cmd/skopeo/inspect.go
```diff
@@ -135,9 +148,19 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
...
+			var out []byte
+			out, err = json.MarshalIndent(config, "", "    ")
+			if err == nil {
+				fmt.Fprintf(stdout, "%s\n", string(out))
+			}
+		} else {
+			row := "{{range . }}" + report.NormalizeFormat(opts.format) + "{{end}}"
```
mtrmac, line 158: Why is `NormalizeFormat` desirable? The man page says “Go template”, nothing about treating `\t` specially, nor about the `table ` special case.

What does the `{{ range . }}…{{end}}` do that we need the `data` variable instead of just using an `interface{}` for the struct directly? At least https://golang.org/pkg/text/template/#example_Template suggests that should work.

(And if it is important, shouldn’t the `report` package do that?)

cmd/skopeo/inspect.go
mtrmac, line 151: The `IsJSON` case is, AFAICS, undocumented — and especially for the `{{json}}` case, which looks very much like a valid Go template, surprising. Why is it useful?

cmd/skopeo/inspect.go
mtrmac: The `*Options` values are conceptually different (sets of options shared between commands) from the other command-specific options, so it seems more intuitive for me to group the like together.

Also, that allows the declaration of the struct and its preparation for use in `inspectCmd` to be roughly in the same order, making it easier to verify that the two correspond.

I did say it is an “absolutely non-blocking nit”.

cmd/skopeo/inspect.go
```diff
@@ -182,15 +205,35 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
...
+}
+
+func printTmpl(row string, data []interface{}) error {
+	t, err := template.New("skopeo inspect").Parse(row)
 	if err != nil {
 		return err
 	}
-	fmt.Fprintf(stdout, "%s\n", string(out))
-	return nil
+	w := tabwriter.NewWriter(os.Stdout, 8, 2, 2, ' ', 0)
```
mtrmac, line 237: This should use the `stdout` parameter to `.run`.

mtrmac, line 237: The flexible tab behavior of `tabwriter` is undocumented; I’d expect “Go template” to mean just that, with no modifications.

mtrmac (commented):
Late review — I can work on the adjustments as necessary.

@rhatdan about half is a set of questions about how the templates/outputs are modified beyond just “Go template”. I don’t feel strongly at all about having those modifications, but if we do have them, it really should be documented somehow. Is there an established name for these things?

PR #1080, opened 2020-10-17
[DNM] Test ocicrypt upgrade

lumjjb:
Testing downstream users of ocicrypt library regression before release.

Signed-off-by: Brandon Lum <lumjjb@gmail.com>


rhatdan:
@lumjjb What is going on with this one?

lumjjb:
Hi @rhatdan we were resolving some build issues with the skopeo/buildah cross builder. One of the components needed to be special case based on whether cgo was being used or not. We have resolved this, so am closing this - and will close the other ones as well!

PR #1091, opened 2020-11-02
Use osusergo build tag for static build

saschagrunert:
We now use the osusergo build tag to not use the glibc functions which
occur in the warnings but them from golang the os/user package.


rhatdan:
LGTM
@vrothberg @mtrmac @QiWang19 PTAL

vrothberg (commented):
LGTM

I restarted the test. I've a similar flake in c/image PRs so this might be something we have to fix. Let's wait for the result.

saschagrunert:
> I restarted the test. I've a similar flake in c/image PRs so this might be something we have to fix. Let's wait for the result.

Thank you! :pray:

rhatdan:
Failure is unrelated to this PR>

PR #1094, opened 2020-11-04
Travis: bump go to 1.15.x

vrothberg:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ matrix:
   -  os: osx
 
 go:
-  - 1.13.x
+  - 1.15.x
 
 notifications:
   email: false
--- a/integration/copy_test.go
+++ b/integration/copy_test.go
@@ -1114,10 +1114,12 @@ func (s *CopySuite) TestCopyVerifyingMirroredSignatures(c *check.C) {
 	// Create a signature for mirroring-primary:primary-signed without pushing there. This should be easier than using standalone-sign.
 	signingDir := filepath.Join(topDir, "signing-temp")
 	assertSkopeoSucceeds(c, "", "copy", "--src-tls-verify=false", regPrefix+"primary:unsigned", "dir:"+signingDir)
-	c.Logf("%s", combinedOutputOfCommand(c, "ls", "-laR", signingDir))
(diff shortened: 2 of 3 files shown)


rhatdan:
Looks like we need the openssl package installed on the linux box?

vrothberg:
@mtrmac FYI ... let's see if can get this green again

vrothberg:
@rhatdan we're green

integration/copy_test.go
```diff
@@ -1114,10 +1114,12 @@ func (s *CopySuite) TestCopyVerifyingMirroredSignatures(c *check.C) {
 	// Create a signature for mirroring-primary:primary-signed without pushing there. This should be easier than using standalone-sign.
 	signingDir := filepath.Join(topDir, "signing-temp")
 	assertSkopeoSucceeds(c, "", "copy", "--src-tls-verify=false", regPrefix+"primary:unsigned", "dir:"+signingDir)
-	c.Logf("%s", combinedOutputOfCommand(c, "ls", "-laR", signingDir))
+	// Unknown error in Travis: https://github.com/containers/skopeo/issues/1093
+	//	c.Logf("%s", combinedOutputOfCommand(c, "ls", "-laR", signingDir))
```
mtrmac, line 1118: These are “just” diagnostic logs, but I’m not too happy to just ignore a weird failure like this. Let’s see what https://github.com/containers/skopeo/pull/1096 does.

.travis.yml
```diff
@@ -9,7 +9,7 @@ matrix:
   -  os: osx
 
 go:
-  - 1.13.x
+  - 1.15.x
```
mtrmac, line 12: (Note that this AFAICS only affects the macOS builds; Linux builds are driven by `Dockerfile` == what `golang-bin` package is installed by `fedora:latest`. But that does end up being 1.15 currently.)

vrothberg:
@mtrmac can we merge this PR to unblock the queue? We can continue investigating the permission error.

mtrmac:
Let’s.

In #1096 I ended up with
```
cannot statx '/': Operation not permitted
```
and I have absolutely no idea what’s going on or how to debug that.

I’m worried that we will run into actually needing that kind of data sooner or later, but with no clear path to finding the root cause, it makes no sense to block.

Issue #1093, opened 2020-11-04
CI failure


vrothberg:
The below failure happens on Skopeo master for a while and is also blocking c/image.

I'm looking into it but help is appreciated.

@rhatdan @mtrmac @QiWang19 @edsantiago 

```
FAIL: copy_test.go:1070: CopySuite.TestCopyVerifyingMirroredSignatures

Running skopeo copy --dest-tls-verify=false docker://busybox docker://localhost:5006/myns/mirroring-primary:unsigned

Running skopeo --policy /tmp/policy.json172747137 --registries.d /tmp/mirrored-signatures545258538/registries.d --registries-conf fixtures/registries.conf copy --src-tls-verify=false docker://localhost:5006/myns/mirroring-primary:unsigned dir:/tmp/mirrored-signatures545258538/unused-dest

Running skopeo --registries.d /tmp/mirrored-signatures545258538/registries.d copy --src-tls-verify=false --dest-tls-verify=false --sign-by personal@example.com docker://localhost:5006/myns/mirroring-primary:unsigned docker://localhost:5006/myns/mirroring-primary:direct

Running skopeo --policy /tmp/policy.json172747137 --registries.d /tmp/mirrored-signatures545258538/registries.d --registries-conf fixtures/registries.conf copy --src-tls-verify=false docker://localhost:5006/myns/mirroring-primary:direct dir:/tmp/mirrored-signatures545258538/unused-dest

Running skopeo --registries.d /tmp/mirrored-signatures545258538/registries.d copy --src-tls-verify=false --dest-tls-verify=false --sign-by personal@example.com docker://localhost:5006/myns/mirroring-primary:unsigned docker://localhost:5006/myns/mirroring-mirror:mirror-signed

Running skopeo --policy /tmp/policy.json172747137 --registries.d /tmp/mirrored-signatures545258538/registries.d --registries-conf fixtures/registries.conf copy --src-tls-verify=false docker://localhost:5006/myns/mirroring-mirror:mirror-signed dir:/tmp/mirrored-signatures545258538/unused-dest

Running skopeo --registries-conf fixtures/registries.conf copy --src-tls-verify=false docker://localhost:5006/myns/mirroring-primary:mirror-signed dir:/tmp/mirrored-signatures545258538/unused-dest

[...]

PR #1096, opened 2020-11-04
Log full temp directory contents

mtrmac:
… to try to resolve what #1093 is about.


mtrmac:
```console
... stat: cannot statx '/tmp/mirrored-signatures739221836': Operation not permitted
```
what the…

mtrmac:
… and
```
cannot statx '/': Operation not permitted
```
I give up.

PR #1099, opened 2020-11-06
Avoid overriding LDFLAGS in Makefile

mbargull:
Similar to https://github.com/containers/buildah/pull/2752:

#### What this PR does / why we need it:

`LDFLAGS` is a common/standard variable to hold linker flags for C/C++ compilers.
This PR uses a different name (`SKOPEO_LDFLAGS`) for the `-ldflags` passed on to Go so that we avoid overriding `LDFLAGS` set in the environment.

@mtrmac in https://github.com/containers/skopeo/pull/1099#pullrequestreview-525574335 :
> For the record, a missing part of the motivation is that CGo uses LDFLAGS when referring to C libraries, so this breaks overriding e.g. paths with `-L`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,7 @@ CONTAINER_RUN := $(CONTAINER_CMD) "$(IMAGE)"
 GIT_COMMIT := $(shell git rev-parse HEAD 2> /dev/null || true)
 
 EXTRA_LDFLAGS ?=
-LDFLAGS := -ldflags '-X main.gitCommit=${GIT_COMMIT} $(EXTRA_LDFLAGS)'
+SKOPEO_LDFLAGS := -ldflags '-X main.gitCommit=${GIT_COMMIT} $(EXTRA_LDFLAGS)'
 
 MANPAGES_MD = $(wildcard docs/*.md)
 MANPAGES ?= $(MANPAGES_MD:%.md=%)
@@ -127,9 +127,9 @@ static:
 # Build w/o using containers
 .PHONY: bin/skopeo
 bin/skopeo:
-	$(GPGME_ENV) $(GO) build $(MOD_VENDOR) ${GO_DYN_FLAGS} ${LDFLAGS} -gcflags "$(GOGCFLAGS)" -tags "$(BUILDTAGS)" -o $@ ./cmd/skopeo
+	$(GPGME_ENV) $(GO) build $(MOD_VENDOR) ${GO_DYN_FLAGS} ${SKOPEO_LDFLAGS} -gcflags "$(GOGCFLAGS)" -tags "$(BUILDTAGS)" -o $@ ./cmd/skopeo
 bin/skopeo.%:
-	GOOS=$(word 2,$(subst ., ,$@)) GOARCH=$(word 3,$(subst ., ,$@)) $(GO) build $(MOD_VENDOR) ${LDFLAGS} -tags "containers_image_openpgp $(BUILDTAGS)" -o $@ ./cmd/skopeo
+	GOOS=$(word 2,$(subst ., ,$@)) GOARCH=$(word 3,$(subst ., ,$@)) $(GO) build $(MOD_VENDOR) ${SKOPEO_LDFLAGS} -tags "containers_image_openpgp $(BUILDTAGS)" -o $@ ./cmd/skopeo
 local-cross: bin/skopeo.darwin.amd64 bin/skopeo.linux.arm bin/skopeo.linux.arm64 bin/skopeo.windows.386.exe bin/skopeo.windows.amd64.exe
 
 build-container:


rhatdan:
LGTM

rhatdan:
@containers/podman-maintainers PTAL

TomSweeneyRedHat (commented):
LGTM

QiWang19 (commented):
LGTM

mtrmac (commented):
For the record, a missing part of the motivation is that CGo uses LDFLAGS when referring to C libraries, so this breaks overriding e.g. paths with `-L`.

---

Grepping the source, does `default.nix` need updating? I have no idea what that thing even is.

mbargull:
> For the record, a missing part of the motivation is that CGo uses LDFLAGS when referring to C libraries, so this breaks overriding e.g. paths with `-L`.

Indeed. I should've added such an explanatory sentence here too, but forgot to. Thanks.

> Grepping the source, does `default.nix` need updating? I have no idea what that thing even is.

Ah, good catch! I could make a followup PR for that -- but that line looks non-functional anyway: it does `export LDFLAGS=...` but the `Makefile` has `LDFLAGS :=...` and `make` isn't invoked with `LDFLAGS=...` or `-e` (plus, even it were to override the `Makefile`'s `LDFLAGS`, it would then miss the `-ldflags` for `EXTRA_LDFLAGS`).
Let's just ask the author: @hswong3i, what is the intent of the `export LDFLAGS` line in the `nix/default.nix` file?

mbargull:
`buildah` and `podman` have the same putatively non-functional `export LDFLAGS=...` line.

hswong3i:
> Ah, good catch! I could make a followup PR for that -- but that line looks non-functional anyway: it does `export LDFLAGS=...` but the `Makefile` has `LDFLAGS :=...` and `make` isn't invoked with `LDFLAGS=...` or `-e` (plus, even it were to override the `Makefile`'s `LDFLAGS`, it would then miss the `-ldflags` for `EXTRA_LDFLAGS`).
> Let's just ask the author: @hswong3i, what is the intent of the `export LDFLAGS` line in the `nix/default.nix` file?

For your reference:
  - https://github.com/containers/crun/blob/master/nix/default.nix
  - https://github.com/containers/conmon/blob/master/nix/default.nix
  - https://github.com/containers/skopeo/blob/master/nix/default.nix
  - https://github.com/containers/buildah/blob/master/nix/default.nix
  - https://github.com/containers/podman/blob/master/nix/default.nix
  - https://github.com/cri-o/cri-o/blob/master/nix/default.nix

Here I share as much as similar skeleton across repos, so when running `nix build -f nix/` locally they could share as much of complied package result as possible, therefore reduce overall compile duration.

Hopefully we could also synchronize similar changes on `CFLAGS`/`LDFLAGS`/`EXTRA_LDFLAGS`/`BUILDTAGS`/etc across above repos: a simple stupid implementation could give us more intention for maintenance, daily ;-)

mbargull:
From that I (kind of) gather that the `LDFLAGS` in there is indeed meant to be just the "normal C" `LDFLAGS` -- which is good, in general and because it means this PR was beneficial for your intentions too.
The thing that threw me off a bit is the `-w` flag in there -- that one is rather uncommon to see (in `LDFLAGS` and overall you rather encounter `-W...` flags in the wild).

mbargull:
CI failure seems unrelated (maybe connection issue?).
https://travis-ci.org/github/containers/skopeo/jobs/741944968#L2178
```
----------------------------------------------------------------------

FAIL: copy_test.go:186: CopySuite.TestCopyWithManifestListStorage

Running skopeo copy docker://estesp/busybox:latest containers-storage:[vfs@/tmp/copy-manifest-list-storage442766416/root+/tmp/copy-manifest-list-storage442766416/runroot]test

Running skopeo copy docker://estesp/busybox:latest dir:/tmp/copy-manifest-list-storage-dir703840111

copy_test.go:198:

    assertSkopeoSucceeds(c, "", "copy", "docker://estesp/busybox:latest", "dir:"+dir1)

utils.go:62:

... value *exec.ExitError = &exec.ExitError{ProcessState:(*os.ProcessState)(0xc0000b6840), Stderr:[]uint8(nil)} ("exit status 1")

... Getting image source signatures

Copying blob sha256:57c14dd66db0390dbf6da578421c077f6de8e88edd0815b4caa94607ba5f4c09

time="2020-11-07T01:17:04Z" level=fatal msg="Error reading config blob sha256:3a093384ac306cbac30b67f1585e12b30ab1a899374dabc3170b9bca246f1444: Get \"https://registry-1.docker.io/v2/estesp/busybox/blobs/sha256:3a093384ac306cbac30b67f1585e12b30ab1a899374dabc3170b9bca246f1444\": EOF"

----------------------------------------------------------------------
```

rhatdan:
@mtrmac @mbargull @hswong3i So this is ready to merge?

mbargull:
@rhatdan, yes, good to go! (The `LDFLAGS` usage in `nix/default.nix` is untouched/unimpeded by this PR.)

PR #1100, opened 2020-11-09
WIP added implementation for scoped-without-registry support on sync

### LINKED ISSUE #1072: Sync: Add option to keep full name on target when using YAML
Hello,

* I know that this was already discussed in #898 and especially in #854.
* However, maybe I can add an usecase:
   * We currently use another program to sync images from a 8 registries to our internal one (docker.example.com).
   * To ease the use, all our internal users (and build systems) have configured docker.example.com as `registry-mirrors`.
   * So when they run `docker run --rm -it alpine`, they automatically pull this from docker.example.com/library/alpine and same goes for `docker run --rm jenkins/jnlp-slave`.
   *  We did not find any collisions when using the "group" as path.
   * Now I would like to keep all synced images from e.g. registry.hub.docker.com in one YAML file (currently we sync 185 repositories from there).
   * Another feature which only seems to be available in the YAML file is to define `images-by-tag-regex`. 
   * Without the option to keep the full name, I would need to split this into 185 files and commands.
* So I really would like to see something like `--scoped-without-registry` as outlined in #854.

c0ffee:
Adds support to sync 2 registries with all namespaces but without adding the registry as prefix

Fixes #1072 based on the comments of #854 

Name of the commandline argument can be discussed

Help is appreciated especially for writing meaningful tests!

:heavy_check_mark: Code
:x: Tests
:x: Documentation


rhatdan:
@c0ffee Needs a rebase.
@mtrmac @vrothberg PTAL

cmd/skopeo/sync.go
```diff
@@ -548,6 +550,13 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) error {
 			case docker.Transport:
 				// docker -> dir or docker -> docker
 				destSuffix = ref.DockerReference().String()
+				if opts.scopedWithoutRegistry {
+					removeRegistry := func(s string) string {
```
vrothberg, line 554: No need for the anonymous function.

vrothberg (commented):
Thanks! Did you try @mtrmac's suggestion in https://github.com/containers/skopeo/issues/854#issuecomment-605065731? Not sure a new flag is needed. Can you elaborate why you prefer a flag over https://github.com/containers/skopeo/issues/854#issuecomment-605065731?

mtrmac:
Yeah, I think it’s been a mistake to hard-code any naming policy in sync as a default behavior, and adding more and more options of that kind won’t make the basic mistake go away.

AFAICS to this day you can’t copy $registry1/ns1/repo to $registry2/ns2/some-other-repo-name if I’m not mistaken.
- on the CLI, a fix for that would automatically give users an _easy-to-understand_ way to build any other repo hierarchy as well, wouldn’t it?
- The YAML format is not trivially extensible to allow this (there is no way to put the “destination” in a `$registry`.`images-by-tag` for example… it _still_ should happen, or maybe deprecate the YAML format and just let users write a shell script, I’m still not sure what makes the YAML format necessary.

c0ffee:
we use this basically to synchronize some of our on-prem host registries to a merged remote one and some images use several namespaces which makes it uncomfortable

from my perspective the behavior of scope is not what you would expect and this is why all these issues exists, so the best way would be to fix this, but this is not possible right?

i actually like the yaml format, one nice way to use it is to check it in a repo and use it as a kind of promoting way which images should be available on the external mirror(controlled access on the repo and CI chain is using it as input)

before i put now more work into this? what would be the preferred way to continue?

mtrmac:
For CLI, an option (sadly it has to be an option) to do no mapping and just let the user specify a destination repo, per the previous comment.

For YAML, I’m not immediately sure. It would be possible to keep the existing format and add a repo map (source repo → destination repo) at the same level as the current `images`/`images-by-tag-regex`, but that’s pretty clumsy.

koweblomke:
I understand the concerns of not littering the CLI with all kind of options. But in this case couldn't it be said that the current way is the default way and when you want the namespace of the source repository to be respected is actually an option?

So instead of --scoped-without-registry add on option --keep-namespace

I understand this doesn't make the mistake go away, but it makes the tool much more useful.

PG2000:
@vrothberg Can you explain what the right direction would be here?
For me the use case makes also sense and i would continue to work on it if thats ok for @c0ffee

vrothberg:
> For CLI, an option (sadly it has to be an option) to do no mapping and just let the user specify a destination repo, per the previous comment.
> 
> For YAML, I’m not immediately sure. It would be possible to keep the existing format and add a repo map (source repo → destination repo) at the same level as the current `images`/`images-by-tag-regex`, but that’s pretty clumsy.

@PG2000, that looks like a good starting point to me.

davidseve:
This feature is highly important to us. We aim to copy images to production repositories, and the addition of the registry as a prefix complicates this process significantly. We've tested the changes in this PR, and it works well for us. We look forward to its integration.

PR #1048, opened 2020-09-19
install: make commanbds copy-pasteable

glensc:
make the commands copy-pasteble: double click on a line and copy

PR #1103, opened 2020-11-12
install: make commands copy-pasteable

rhatdan:
Based on Elan Ruusamäe patch.

### DIFF
--- a/install.md
+++ b/install.md
@@ -6,13 +6,13 @@
 ### Fedora
 
 ```sh
-$ sudo dnf -y install skopeo
+sudo dnf -y install skopeo
 ```
 
 ### RHEL/CentOS ≥ 8 and CentOS Stream
 
 ```sh
-$ sudo dnf -y install skopeo
+sudo dnf -y install skopeo
 ```
 
 Newer Skopeo releases may be available on the repositories provided by the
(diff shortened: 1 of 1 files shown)


rhatdan:
Replaces: https://github.com/containers/skopeo/pull/1048

TomSweeneyRedHat:
I think this is one of these "PoTAYto" vs "PoTAHto" arguments.  I fall on the other side of the fence for this change.  Regardless:
LGTM

rhatdan:
The problem, is that we are split on both sides of the fence, and that is not a comfortable place to be.

rhatdan:
@podman-maintainers PTAL

rhatdan:
@vrothberg @mtrmac PTAL

vrothberg (commented):
LGTM

glensc:
Yay 🎉 for the merge!

I hope it takes so much effort and confirmation when the next round decides to do the exact opposite!

ps: @rhatdan you can preserve authorship with `git commit --author "$email" [--amend]`, or `git cherry-pick $commit`

rhatdan:
@glensc Hopefully you are right.  We have many more eyes on Podman, but need to focus on all containers repos.

PR #1119, opened 2020-11-28
Update to macOS 10.14

mtrmac:
<s>We are updating it because Homebrew warns about the old version, so updating it only afterwards does not help.</s>

Homebrew:
> Warning: You are using macOS 10.13.
> We (and Apple) do not provide support for this old version.
> You will encounter build failures with some formulae.
    
So, update to the 10.14 major version, fully-updated.
    
Also remove the Xcode update attempt, it was added before Homebrew was warning about the Xcode version, but updating only after running Homebrew does not help, and anyway it does not complain anymore after the update.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,6 +57,7 @@ jobs:
       <<: *local-build
       name: local build for osx
       os: osx
+      osx_image: xcode11.3
       install:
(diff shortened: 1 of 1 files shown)


mtrmac:
This seems to help with the testing failures, e.g. in #1118.

vrothberg (commented):
LGTM

PR #1124, opened 2020-12-03
020-copy.bats: check that we set the manifest type correctly

nalind:
When copying to an OCI layout destination, forcing zstd compression, check that the manifest correctly describes the type of the layer blob.

### DIFF
--- a/systemtest/020-copy.bats
+++ b/systemtest/020-copy.bats
@@ -45,7 +45,7 @@ function setup() {
 }
 
 # Compression zstd
-@test "copy: oci, round trip, zstd" {
+@test "copy: oci, zstd" {
     local remote_image=docker://quay.io/libpod/busybox:latest
 
     local dir=$TESTDIR/dir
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
Tests are still angry @nalind

nalind:
That's intentional - the new bits of the test add a case that https://github.com/containers/image/pull/1089 is expected to fix.

mtrmac:
@nalind do you want to update this now, or only after a c/image release? (Or maybe there should be a release straight away?)

nalind:
I know you've got some things that you want to merge soonish, and this PR can wait until there's a tagged release.  I was manually running this test scenario before pushing updates to https://github.com/containers/image/pull/1089, so I don't expect any problems here.

A new tag will get picked up by various dependabot instances when it runs over the coming weekend, so my preference would be to have a tagged release before then.  OpenShift also prefers to use a tag rather than a commit ID, for example [in OKD](https://github.com/openshift/okd/issues/402#issuecomment-762369650), but I can try to `go mod replace` to a commit ID if you're not planning to tag a release soon.

mtrmac (approved):
LGTM. Thanks!

TomSweeneyRedHat:
LGTM

PR #1125, opened 2020-12-04
Skip updating Homebrew on macOS tests

mtrmac:
Currently, the Homebrew commands take about 15 out of 17-18 minutes of the macOS test run.

So, drop updating to latest versions of everything, and let's see what happens.


mtrmac:
`brew install gpgme` still takes over 10 minutes, so this is not a big enough improvement IMHO to be worth using outdated dependencies.

PR #1140, opened 2020-12-20
Spelling

jsoref:
This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

The misspellings have been reported at https://github.com/jsoref/skopeo/commit/f3fde540f2211a3256f067d907d3671ceb53e49f#commitcomment-45289138

The action reports that the changes in this PR would make it happy: https://github.com/jsoref/skopeo/commit/2f22e380de9e9525b95973c304a0c21b981e29c8

Note: this PR does not include the action. If you're interested in running a spell check on every PR and push, that can be offered separately.

### DIFF
--- a/cmd/skopeo/manifest_test.go
+++ b/cmd/skopeo/manifest_test.go
@@ -17,8 +17,8 @@ func TestManifestDigest(t *testing.T) {
 	}
 
 	// Error reading manifest
-	out, err := runSkopeo("manifest-digest", "/this/doesnt/exist")
-	assertTestFailed(t, out, err, "/this/doesnt/exist")
+	out, err := runSkopeo("manifest-digest", "/this/does/not/exist")
+	assertTestFailed(t, out, err, "/this/does/not/exist")
 
 	// Error computing manifest
 	out, err = runSkopeo("manifest-digest", "fixtures/v2s1-invalid-signatures.manifest.json")
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -142,7 +142,7 @@ When new PRs for [containers/image](https://github.com/containers/image) break `
 - as soon as possible after that, in the skopeo PR, restore the `containers/image` line in `vendor.conf` to use `containers/image:master`
 - run `make vendor`
 - update the skopeo PR with the result, drop the “DO NOT MERGE” marking
-- after tests complete succcesfully again, merge the skopeo PR
+- after tests complete successfully again, merge the skopeo PR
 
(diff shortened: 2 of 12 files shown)


rhatdan:
Squash your PRs

rhatdan:
LGTM

cmd/skopeo/manifest_test.go
```diff
@@ -17,8 +17,8 @@ func TestManifestDigest(t *testing.T) {
 	}
 
 	// Error reading manifest
-	out, err := runSkopeo("manifest-digest", "/this/doesnt/exist")
-	assertTestFailed(t, out, err, "/this/doesnt/exist")
+	out, err := runSkopeo("manifest-digest", "/this/does/not/exist")
+	assertTestFailed(t, out, err, "/this/does/not/exist")
```
mtrmac, line 21: (I must question the value of spell-checking paths and identifiers, with things like `/etc/passwd` and `creat(3)` a long-standing practice, but as long as it results in only PRs and not bots bringing extra work, it’s very nice to have.)

mtrmac (approved):
LGTM. Thanks!

cmd/skopeo/manifest_test.go
jsoref: (This is just an explanation of my logic, not an argument.)

It's much easier for me to apply changes globally than it is to try to add lots of filters against paths. -- In this case, it's easier for me to say "When writing a sentence, `doesnt` is wrong, therefore, don't use it anywhere" than "Sometimes `doesnt` is ok, and sometimes it isn't" -- it's a much harder programmatic rule to write. And if I'm not using a program, it means I have to spend more of my time reviewing and saying "oh, yeah, I can/should skip this instance". And if I revisit in 2 years (not uncommon), I'll spend that time again.

Also there will be comments that reference paths and people will get confused. Yes, `creat`, `passwd`, `Referer` and `umount` exist.
But, one of my recent PRs included a fix because someone wrote docs that talked about `unmount` expecting that to be a thing, where in fact the command for the specific tool is `umount` -- consistency is important -- people will reference docs from code and vice-versa, and they will search for things, and if it isn't spelled the way they're expecting, they may totally miss it (or lose considerable time trying to find it).

mtrmac: Oh I’m absolutely not _complaining_, more in awe of your dedication.

rhatdan:
LGTM
Thanks @jsoref

PR #1146, opened 2021-01-05
integration test: sync k8s.gcr.io/pause instead of docker.io/alpine

nalind:
Do repository synchronization tests using k8s.gcr.io/pause instead of docker.io/alpine.  The k8s.gcr.io/pause repository includes multiple tags, at least one of which is a single image ("1.0"), and at least one of which is a manifest list ("3.2", "3.3"), so I don't think we sacrifice coverage by switching to it.

### DIFF
--- a/integration/sync_test.go
+++ b/integration/sync_test.go
@@ -11,8 +11,24 @@ import (
 	"strings"
 
 	"github.com/containers/image/v5/docker"
+	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/types"
 	"github.com/go-check/check"
+)
+
+const (
+	// A repository with a path with multiple components in it which
+	// contains multiple tags, preferably with some tags pointing to
+	// manifest lists, and with some tags that don't.
+	pullableRepo = "k8s.gcr.io/coredns/coredns"
+	// A tagged image in the repository that we can inspect and copy.
+	pullableTaggedImage = "k8s.gcr.io/coredns/coredns:v1.6.6"
+	// A tagged manifest list in the repository that we can inspect and copy.
+	pullableTaggedManifestList = "k8s.gcr.io/coredns/coredns:v1.8.0"
+	// A repository containing multiple tags, some of which are for
+	// manifest lists, and which includes a "latest" tag.  We specify the
+	// name here without a tag.
+	pullableRepoWithLatestTag = "k8s.gcr.io/pause"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

mtrmac (commented):
Good idea, we definitely don’t need to sync 56 tags to test this.

The failure in `TestYamlUntagged` is real, though — the first `yamlConfig` is incorrect (the top level still says `docker.io` and the keys in `images` are apparently relative to the top-level registry name).

nalind:
Oof, thanks for catching that, I hadn't had a chance to yet.

integration/sync_test.go
```diff
@@ -202,8 +202,9 @@ func (s *SyncSuite) TestDocker2DirUntagged(c *check.C) {
 	defer os.RemoveAll(tmpDir)
 
 	// FIXME: It would be nice to use one of the local Docker registries instead of needing an Internet connection.
-	image := "alpine"
-	imageRef, err := docker.ParseReference(fmt.Sprintf("//%s", image))
+	registry := "k8s.gcr.io"
+	image := "pause"
+	imageRef, err := docker.ParseReference(fmt.Sprintf("//%s/%s", registry, image))
```
mtrmac, line 207: This now results in `… sync --src docker pause`, not what was intended.

integration/sync_test.go
```diff
@@ -226,8 +227,9 @@ func (s *SyncSuite) TestYamlUntagged(c *check.C) {
 	defer os.RemoveAll(tmpDir)
 	dir1 := path.Join(tmpDir, "dir1")
 
-	image := "alpine"
-	imageRef, err := docker.ParseReference(fmt.Sprintf("//%s", image))
+	registry := "k8s.gcr.io"
+	image := "pause"
+	imageRef, err := docker.ParseReference(fmt.Sprintf("//%s/%s", registry, image))
```
mtrmac, line 232: An unrelated and absolutely non-blocking note: The second use of `imageRef` should probably refer to `v2DockerRegistryURL+"/"+imagePath`, otherwise the `localTags` checks are meaningless.

mtrmac, line 232: (I’d be a tiny bit happier with avoiding this bit of string formatting, i.e. keep `image := "k8s.gcr.io/pause"`, and use `reference.Domain(imageRef.DockerReference()), imagePath` for creating the YAML. That also makes the two YAML files generated completely consistently.

But this works just fine, feel free to leave as is.)

integration/sync_test.go
nalind: Thanks for catching that.

integration/sync_test.go
mtrmac: Filed #1150 , mostly as a reminder to myself to fix that.

TomSweeneyRedHat:
LGTM

nalind:
Switched some of the "copying manifest lists" tests to use the fedora-minimal image instead of estesp's busybox image, for the same reason.

nalind:
Rebased.

TomSweeneyRedHat:
LGTM
and happy green test buttons.

mtrmac:
@nalind Thanks!

PR #1151, opened 2021-01-07
WIP: Experiment with Cirrus

mtrmac:
- Migrate the macOS task ~as is, without modifying (TO DO: or even renaming) the script)
- Smoke-test the out-of-container parts of local build

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,18 @@
+
+macOS_task:
+  macos_instance:
+    image: catalina-xcode-11.3.1
+  brew_script: brew install gpgme go
+  test_script: hack/travis_osx.sh
+
+# NOTE: This does not run most tests (yet)
+build_task:
+  container:
+    image: golang:1.15
+  modules_cache:
+    fingerprint_script: cat go.sum
+    folder: $GOPATH/pkg/mod
+  vendor_script:
+    - make vendor
+    - ./hack/tree_status.sh
+    - make local-cross


cevich:
Hint: There's a way to have Cirrus-CI run a testing script inside a container image which it manages.  Cirrus will cache the image after the first build, and only rebuild it if the `Dockerfile` changes or if the image gets purged due to inactivity.

Docs: https://cirrus-ci.org/guide/docker-builder-vm/#dockerfile-as-a-ci-environment

cevich:
@mtrmac I spent quite a bit of time looking at the output and test code...with little headway :disappointed: 

Do we get any more helpful output of the failure if you remove (even temporarily) the `| grep -v '^github\.com/containers/skopeo/\(integration\|vendor/.*\)$'` on the `test-unit-local` target?

mtrmac:
@cevich If you are talking about the “key not found” failure on macOS, I haven’t looked at that yet at all. My first guess is something in the environment or permissions that causes `gpg` to ignore the files in `cmd/skopeo/fixtures`; testing that separately without involving the `gpgme` library might be easier than getting and understanding logs from that library.

cevich:
ya, `gpgme` (and gpg) isn't always straight-forward to diagnose.  I'm not aware of anything in `gpg` that would cause it to ignore files by default.  Unless otherwise instructed, permissions there should be the same as what was used to clone the repo.   The reference to something called `/dev/full` makes me think it's something in the environment as well.  Perhaps the version of something changed WRT what was being done before in Travis?

rhatdan:
@cevich could you take this PR over and finish getting skopeo under cirrus?

cevich:
@rhatdan I could but...I think the best way forward is to get more people involved, not just change ownership.  Unless you had some other reason in mind?

@mtrmac and I are both "stuck" in debugging test-internals to figure out what needs fixing in the environment.  I am completely unfamiliar with both the code-base's tests, and the VM environment, so am of little use other than pulling high-level strings :disappointed: 

@ashley-cui has experience working in MacOS for podman, maybe some of her knowledge could be of use here?

Who else knows the skopeo integration tests well, and could help debug them?

ashley-cui:
I'll take a look and try to figure out why the Mac stuff is failing :sweat_smile:

cevich:
@ashley-cui any news? Were you able to reproduce the failure?

ashley-cui:
@cevich I'm so sorry I got way too wrapped up in secrets work, I definitely will take a look this afternoon and update here on my findings

ashley-cui:
@cevich Okay, took a little look at it. It's just that the test here: https://github.com/containers/skopeo/blob/84232cf3069a1b2f94792a28c9c980cea68ac356/cmd/skopeo/signing_test.go#L76
uses `/dev/full` to test an output error. Unfortunately, `/dev/full` which doesn't exist on a Mac. So if we either skip that test on a Mac or use something else to test a bad output, everything should work on the Mac.

cevich:
@ashley-cui Hmmm, [what if we did like this](https://www.thedroidsonroids.com/blog/dev-full-osx), and made a symlink to `/dev/full`  like they suggest, before running the tests?

Not having a Mac at hand, it's hard for me to try any of this out.  Does the script in the article work as-written for you?

ashley-cui:
@cevich tested the script and it looks to work as-written :)

mtrmac:
> @ashley-cui Hmmm, [what if we did like this](https://www.thedroidsonroids.com/blog/dev-full-osx), and made a symlink to `/dev/full` like they suggest, before running the tests?

That test only exercises an “`io.WriteFile` failed” error path, without caring all that much _how_ it fails. That path can be quite easily reached by providing an existing but non-writable (0400 perms) file, or a `/path/to/a/file/in/a/nonexistent/directory`. Especially the latter one should be very portable without needing any OS-specific infrastructure.

----

The `key "1D8230F6CDB6A06716E414C1DB72F2188BB46CC8" not found` failure might be more tricky; maybe https://github.com/containers/skopeo/pull/1151#issuecomment-762321813 , see if/how `GNUPGHOME=… gpg --sign …` fails. (But I still haven’t tried looking into it myself; it works on my personal Mac fine, and it works fine in Travis, this is something to do with the Cirrus CI environment.)

cevich:
> this is something to do with the Cirrus CI environment.

Suggestion: in travis and cirrus, print all the environment variables (`printenv`).  Then compare. Maybe it's something simple like Cirrus not setting `$HOME` or something.

cevich:
@lsm5 I think the next step is easy: Clone and push your own copy of this PR (then close this one).  Then there are two suggestions above, [one for fixing the broken test, another hinting at some gpg/environment problem.](https://github.com/containers/skopeo/pull/1151#issuecomment-767107627)

lsm5:
> @lsm5 I think the next step is easy: Clone and push your own copy of this PR (then close this one). Then there are two suggestions above, [one for fixing the broken test, another hinting at some gpg/environment problem.](https://github.com/containers/skopeo/pull/1151#issuecomment-767107627)

will do, thanks! :+1:

lsm5:
Closing in favor of: https://github.com/containers/skopeo/pull/1213

PR #1156, opened 2021-01-09
V1.2.1

TomSweeneyRedHat:
As the title says.

I'll be creating a v1.3.0 PR shortly.  We should only merge one of the two based on @rhatdan and @vrothberg's thoughts.

PR #1157, opened 2021-01-09
V1.3.0

TomSweeneyRedHat:
As the title says.  Only this or #1156 should be merged at this time.


TomSweeneyRedHat:
Killing as it looks like we're going to v1.2.1 at the moment.

Issue #1109, opened 2020-11-16
[feature request] Set a unique user-agent for skopeo


jonjohnsonjr:
I did some quick grepping around and don't see any skopeo-specific user-agent being set, which makes debugging things from the registry side a bit challenging.

vrothberg:
Thanks for the request, @jonjohnsonjr.

@rhatdan @mtrmac WDYT?

mtrmac:
Alternatively, should c/image have a default User-Agent value? Podman and Buildah set a value, but it’s unclear that the registry cares about the difference.

(OTOH I vaguely remember that at one time we have traced back a behavior difference between Podman and Skopeo down to User-Agent, and I can’t remember which one worked; so it might not be safe enough to default to an User-Agent value — or maybe that case was sufficiently obscure that we shouldn’t need to care; I can’t remember the details and I can’t search for them today.)

jonjohnsonjr:
> Alternatively, should c/image have a default User-Agent value?

I'd be fine with that as well, especially if some version information is included. Mostly I'm looking for a way to reproduce issues and determine how many users are affected by said issues.

Some kind of containers/image UA by default like `containers-image/5.8.0` (or whatever string) would work just as well.

skopeo could supply its own to produce `skopeo/1.2.0 containers-image/5.8.0` (not sure if this conforms to the grammar), which would save a little bit of time correlating versions across projects, but honestly I'd be fine with just `skopeo/1.2.0` as well!

> I can’t remember the details and I can’t search for them today.

https://github.com/containers/podman/issues/1792 might be what you're thinking of? Looks like it's fixed. That kind of user-agent sniffing is super brittle and I'd argue any registries doing this are broken :)

mtrmac:
I was actually thinking of https://github.com/containers/podman/issues/4254#issuecomment-662000140 , which is at least per the issue ~unresolved , though the other one is also relevant (as is, to an extent, https://github.com/containers/podman/issues/3788 “Our setup uses a nginx reverse proxy that only forwards requests to https://repo:443/... if the User-Agent string matches the pattern docker,” — but we really can’t support _arbitrary_ user-agent-restrictive configurations, so that one is moot).

---

Anyway, yes, I think this should happen, and the syntax you suggest does seem to be valid per RFC 7231 section 5.5.3 . c/image will probably need a new `DockerRegistryPartialUserAgent` or something like that, rather than changing the semantics of the existing field.

jonjohnsonjr:
[...]

PR #1159, opened 2021-01-12
Set User-Agent to skopeo/$VERSION

jonjohnsonjr:
Prior to this, the User-Agent used by containers/image would default to
the default User-Agent for golang, which makes it difficult to
distinguish skopeo from any other golang binaries in registry logs.

Fixes https://github.com/containers/skopeo/issues/1109

If you think this requires a test, I'm happy to add something, but `go test` fails to build for me because I'm missing certain header files, and I figured I'd give this a shot before diving into that 😉 

Signed-off-by: Jon Johnson <jonjohnson@google.com>

### DIFF
--- a/cmd/skopeo/main_test.go
+++ b/cmd/skopeo/main_test.go
@@ -23,7 +23,10 @@ func TestGlobalOptionsNewSystemContext(t *testing.T) {
 	// Default state
 	opts, _ := fakeGlobalOptions(t, []string{})
 	res := opts.newSystemContext()
-	assert.Equal(t, &types.SystemContext{}, res)
+	assert.Equal(t, &types.SystemContext{
+		// User-Agent is set by default.
+		DockerRegistryUserAgent: defaultUserAgent,
+	}, res)
 	// Set everything to non-default values.
 	opts, _ = fakeGlobalOptions(t, []string{
 		"--registries.d", "/srv/registries.d",
@@ -43,5 +46,6 @@ func TestGlobalOptionsNewSystemContext(t *testing.T) {
 		BigFilesTemporaryDir:        "/srv",
 		SystemRegistriesConfPath:    "/srv/registries.conf",
 		DockerInsecureSkipTLSVerify: types.OptionalBoolTrue,
+		DockerRegistryUserAgent:     defaultUserAgent,
 	}, res)
 }
--- a/cmd/skopeo/main.go
+++ b/cmd/skopeo/main.go
@@ -17,6 +17,8 @@ import (
 // and will be populated by the Makefile
 var gitCommit = ""
 
+var defaultUserAgent = "skopeo/" + version.Version
+
 type globalOptions struct {
 	debug              bool          // Enable debug output
 	tlsVerify          optionalBool  // Require HTTPS and verify certificates (for docker: and docker-daemon:)
@@ -143,6 +145,7 @@ func (opts *globalOptions) newSystemContext() *types.SystemContext {
 		VariantChoice:            opts.overrideVariant,
 		SystemRegistriesConfPath: opts.registriesConfPath,
 		BigFilesTemporaryDir:     opts.tmpDir,
+		DockerRegistryUserAgent:  defaultUserAgent,
 	}
 	// DEPRECATED: We support this for backward compatibility, but override it if a per-image flag is provided.
 	if opts.tlsVerify.present {
--- a/cmd/skopeo/utils_test.go
+++ b/cmd/skopeo/utils_test.go
@@ -37,7 +37,9 @@ func TestImageOptionsNewSystemContext(t *testing.T) {
 	opts := fakeImageOptions(t, "dest-", []string{}, []string{})
 	res, err := opts.newSystemContext()
 	require.NoError(t, err)
-	assert.Equal(t, &types.SystemContext{}, res)
+	assert.Equal(t, &types.SystemContext{
+		DockerRegistryUserAgent: defaultUserAgent,
+	}, res)
 
 	// Set everything to non-default values.
 	opts = fakeImageOptions(t, "dest-", []string{
@@ -72,6 +74,7 @@ func TestImageOptionsNewSystemContext(t *testing.T) {
 		DockerDaemonCertPath:              "/srv/cert-dir",
 		DockerDaemonHost:                  "daemon-host.example.com",
 		DockerDaemonInsecureSkipTLSVerify: true,
+		DockerRegistryUserAgent:           defaultUserAgent,
 		BigFilesTemporaryDir:              "/srv",
 	}, res)
 
@@ -129,7 +132,9 @@ func TestImageDestOptionsNewSystemContext(t *testing.T) {
 	opts := fakeImageDestOptions(t, "dest-", []string{}, []string{})
 	res, err := opts.newSystemContext()
 	require.NoError(t, err)
-	assert.Equal(t, &types.SystemContext{}, res)
+	assert.Equal(t, &types.SystemContext{
+		DockerRegistryUserAgent: defaultUserAgent,
+	}, res)
 
 	oldXRD, hasXRD := os.LookupEnv("REGISTRY_AUTH_FILE")
 	defer func() {
@@ -149,7 +154,10 @@ func TestImageDestOptionsNewSystemContext(t *testing.T) {
 	})
 	res, err = opts.newSystemContext()
 	require.NoError(t, err)
-	assert.Equal(t, &types.SystemContext{AuthFilePath: authFile}, res)
+	assert.Equal(t, &types.SystemContext{
+		AuthFilePath:            authFile,
+		DockerRegistryUserAgent: defaultUserAgent,
+	}, res)
 
 	// Set everything to non-default values.
 	opts = fakeImageDestOptions(t, "dest-", []string{
@@ -184,6 +192,7 @@ func TestImageDestOptionsNewSystemContext(t *testing.T) {
 		DockerDaemonCertPath:              "/srv/cert-dir",
 		DockerDaemonHost:                  "daemon-host.example.com",
 		DockerDaemonInsecureSkipTLSVerify: true,
+		DockerRegistryUserAgent:           defaultUserAgent,
 		DirForceCompress:                  true,
 		BigFilesTemporaryDir:              "/srv",
 	}, res)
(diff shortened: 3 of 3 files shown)


jonjohnsonjr:
Hah, of course the test would fail.

Looking at `cmd/skopeo/main.go`, it seems like everything in `SystemContext` is overrideable via flags. Should I do the same for the user agent?

mtrmac (commented):
Thanks!

- Most importantly, do we need this? With https://github.com/containers/image/pull/1115 , registries probably care more about the client implementation than Skopeo specifically. #1109 talked about having both, that would be even better, but just c/image might well be enough.
- The unit test failures are real.
- The unit test failures also show that the implementation is unnecessarily using `%s`.

mtrmac:
> If you think this requires a test, I'm happy to add something

Other than those that already exists and need adjusting, an actual test that this really works is up to you, if you think this is important enough to notice if it ever got broken –  and the cost/benefit ratio of adding such a test (a custom HTTP server?) seems pretty dubious to me.

jonjohnsonjr:
> Most importantly, do we need this?

I would prefer it, but I'm open to being convinced otherwise. Given that podman and buildah have distinct user-agents, I feel that skopeo deserves its own as well :)

This would at least allow me to distinguish between "skopeo is doing this" and "some tools that uses containers/image" is doing this. In terms of supportability, it's much easier for me to ask a customer "are you using skopeo for anything?" vs "are you using something that leverages the containers/image library?", as the former can usually be answered by grepping source code.

mtrmac:
> Looking at `cmd/skopeo/main.go`, it seems like everything in `SystemContext` is overrideable via flags. Should I do the same for the user agent?

No; it’s rather the other way around, options were added to `SystemContext` to allow implementing those flags.

Making the User-Agent variable would somewhat defeat the point of having it — we don’t typically add options (and take on the extra maintenance cost) without anticipating that they will be used. (It is possible that the User-Agent value chance could break some users, but those users can, in the short-term, revert to an older version and long-term we would like the servers to adjust instead of documenting a work-around that every user has to manually apply. And if we _had_ to have an user override for this long-term, a file like `registries.conf` would make more sense than a CLI flag.)

mtrmac:
> In terms of supportability, it's much easier for me to ask a customer "are you using skopeo for anything?" vs "are you using something that leverages the containers/image library?", as the former can usually be answered by grepping source code.

That’s a good point. How about the `DockerRegistryPartialUserAgent` option discussed in #1109 , then?

jonjohnsonjr:
> (It is possible that the User-Agent value chance could break some users, but those users can, in the short-term, revert to an older version and long-term we would like the servers to adjust instead of documenting a work-around that every user has to manually apply. And if we had to have an user override for this long-term, a file like registries.conf would make more sense than a CLI flag.)

Whoops, pushed this right as your comment came in -- let me revert and just fix the `%s`.

jonjohnsonjr:
> That’s a good point. How about the `DockerRegistryPartialUserAgent` option discussed in #1109 , then?

I thought about this a little more, and I'm not sure that it's worth the overhead of having both of these in the API. Having just containers/image default to _something_ gives me ~95% of the benefit if a consumer doesn't bother to set a user-agent. Client that _do_ care enough to set a user-agent would likely care enough to set it to something useful...

I do think that being able to segment clients by library _is_ useful for other reasons (I've done something similar in other libraries), but that feels like a bigger change (inventing semantics and adding to the API surface vs just setting some defaults). I'm happy to leave the issue open and have this PR be a partial fix?

mtrmac:
> > That’s a good point. How about the `DockerRegistryPartialUserAgent` option discussed in #1109 , then?
> 
> I thought about this a little more, and I'm not sure that it's worth the overhead of having both of these in the API.
> …
> Edit: It might also make sense to just have containers/image expose some `func (string) string` in `version` that enriches the useragent? Then we don't have to have two options...

I kind of think about this the other way: Having two very simple clearly-scoped options that contain a `string` and don’t affect anything else is not a big deal (and without trying to write tests the implementation is about 5 lines that will probably never change). We can always add a few more options like this (`DockerRegistryPretendToBeDockerInUserAgent` or whatever ridiculous thing might be needed in the future) without breaking any past callers, on a strictly opt-in basis for future library users, and affecting only an extremely small part of the c/image codebase.

A `func (string) string` API would not allow us to remove the existing `DockerRegistryUserAgent` field (we really don’t want to bump the major version of c/image again if at all possible), so we would still have 2 fields to maintain, and keeping the callback API compatible with past/future callers could be much more difficult. (What if c/image versioning changes? What if we eventually realize that we need _both_ c/image and a Go http stack version in there? Would we change the format of the string submitted to that function, or just not include some input and modify what the callback returns? Either way could break someone.)

cmd/skopeo/main_test.go
```diff
@@ -23,7 +24,11 @@ func TestGlobalOptionsNewSystemContext(t *testing.T) {
 	// Default state
 	opts, _ := fakeGlobalOptions(t, []string{})
 	res := opts.newSystemContext()
-	assert.Equal(t, &types.SystemContext{}, res)
+	// User-Agent is set by default.
+	defaultContext := &types.SystemContext{
```
mtrmac, line 28: (Non-blocking: It would be consistent with the other test to inline this into the `assert.Equal`. This works fine, of course.)

jonjohnsonjr:
I'm not sure that I understand what you're referring to as a callback. Maybe my edit wasn't super clear. I'm imagining something like this:

```go
ctx := &types.SystemContext{
  DockerRegistryUserAgent: cversion.UserAgent("skopeo/"+version.Version),
}
```


```go
// containers/image/version/version.go
func UserAgent(base string) string {
  ua := "containers-image/"+Version
  if base == "" {
    return ua
  }
  return fmt.Sprintf("%s %s", base, ua)
}
```

docker/docker does something similar in https://github.com/moby/moby/blob/master/pkg/useragent/useragent.go

And https://github.com/moby/moby/blob/46cdcd206c56172b95ba5c77b827a722dab426c5/dockerversion/useragent.go#L18 + https://github.com/moby/moby/blob/46cdcd206c56172b95ba5c77b827a722dab426c5/api/server/server.go#L132

(Which callers could also use, if they really wanted to pretend to be docker...)

It just seems simpler to have an option for "set the user-agent to this string" and then some accessory helpers for producing than string than to have a bunch of options for different string-assembling behavior.

> (It is possible that the User-Agent value chance could break some users, but those users can, in the short-term, revert to an older version and long-term we would like the servers to adjust instead of documenting a work-around that every user has to manually apply. And if we had to have an user override for this long-term, a file like registries.conf would make more sense than a CLI flag.)

Re-reading this, I _think_ I might understand what you're getting at -- you're thinking of this from the perspective of twiddling a string in a config file vs me thinking of this from the perspective of the golang API?

cmd/skopeo/main_test.go
jonjohnsonjr: Done.

rhatdan:
@mtrmac @vrothberg PTAL

Is this one ready to go in?

vrothberg (commented):

Let's wait for a final head nod from @mtrmac

jonjohnsonjr:
I'm happy to rebase and squash, not sure what y'all would prefer.

jonjohnsonjr:
@mtrmac any strong opinions either way?

vrothberg (commented):
LGTM, @rhatdan PTAL

Could you do a final rebase? Then we can merge.

mtrmac:
> I'm not sure that I understand what you're referring to as a callback. Maybe my edit wasn't super clear. I'm imagining something like this:
> 
> ```go
> ctx := &types.SystemContext{
>   DockerRegistryUserAgent: cversion.UserAgent("skopeo/"+version.Version),
> }
> ```

Ah.

> It just seems simpler to have an option for "set the user-agent to this string" and then some accessory helpers for producing that string than to have a bunch of options for different string-assembling behavior.

People, especially under time pressure, can be incredibly inventive about not using / bypassing optional helpers — and I don’t think we really want a whole general infrastructure for parsing/formatting/combining strings.

mtrmac:
> Is this one ready to go in?

I’d still like to have _both_ c/image and Skopeo in the User-Agent header, but I guess that can wait for the user, if any, that needs that…

jonjohnsonjr:
> Could you do a final rebase? Then we can merge.

Done.

> People, especially under time pressure, can be incredibly inventive about not using / bypassing optional helpers — and I don’t think we really want a whole general infrastructure for parsing/formatting/combining strings.

Definitely agree -- which is why I think good defaults are really important. This PR is actually a pretty good example. Go doesn't force people to set a user agent, so most people don't think about it.

> I’d still like to have both c/image and Skopeo in the User-Agent header

I'd be happy to change this default to:

```go
var defaultUserAgent = fmt.Sprintf("skopeo/%s containers/%s (github.com/containers/image)", version.Version, cversion.Version)
```

I could also expose github.com/containers/image/v5/docker.defaultUserAgent so we can just import the second half? I'm always in favor of a little bit of copying to avoid adding the dependency and having to bump versions, but I'd leave that up to you.

The exact string isn't super important to me, I just want to be able to pinpoint "this is skopeo". My other PR is mostly sufficient for that, given that podman and buildah set their own UA, but this PR is still necessary to distinguish skopeo from anything else using containers/image.

> but I guess that can wait for the user, if any, that needs that…

I'm happy if this merges as-is, but I'd also like you to be happy with it.

Let me know whatever you'd prefer.

mtrmac:
> I could also expose github.com/containers/image/v5/docker.defaultUserAgent so we can just import the second half?

An advantage of the `DockerRegistryPartialUserAgent` approach (along with soft-deprecating `DockerRegistryUserAgent`) would be that the caller _does not have to care_. Whatever process the caller has for choosing a user agent can continue to be unmodified, with no imports, no calls, no conventions, no opportunities for typos in the caller-irrelevant parts; just pass the value that the caller _does care about_ along.

mtrmac (approved):
LGTM. Thanks!

thomasmckay:
Is the user-agent configurable? Quay uses skopeo for to implement "repo mirror" and there was an ask to set user-agent.

vrothberg:
It's not configurable in Skopeo but in the containers/image library.  Are you looking for something similar to `curl -A`?  I am supportive of this idea; I see Skopeo as a curl-like utility for container images.

Interested in opening a PR for that? I imagine a top-level flag making it available to all (sub)commands.

thomasmckay:
@vrothberg sorry, golang is outside my skill set so no PR from me

mtrmac:
> Is the user-agent configurable? Quay uses skopeo for to implement "repo mirror" and there was an ask to set user-agent.

Set user-agent for what purpose? User-agent is not an authentication mechanism; AFAIK it’s primarily to help interoperability (allow bug workarounds and the like) — and in that sense, all Skopeo clients (and really all c/image clients) with a particular version should be, I think, treated the same.

PR #1164, opened 2021-01-15
Bump to v1.2.2

rhatdan:


mtrmac:
@rhatdan Is there something specific you want included in the release? AFAICS the c/common update is just a renaming cleanup. Is that to align versions with other components, or for some other reason?

(Also, there should be a post-release bump to 1.2.3-dev.)

rhatdan:
Yes for the realignment, no other reason.

mtrmac:
OK, so please add a post-release bump, and we’ll include this on both `master` and `release-1.2` branches.

TomSweeneyRedHat:
LGTM

rhatdan:
This is not needed so closing.

Issue #787, opened 2019-12-12
Help for overriding paths during build


glensc:
in brew packaging, the paths have gone wrong at least twice now:

- https://github.com/Homebrew/homebrew-core/pull/47766
- https://github.com/Homebrew/homebrew-core/pull/45834

Is there anything from skopeo side that can be done to improve this?

For example support for building to use standard env variables or build from makefile and make arguments.

Or alias the build time configurable into namespace that does not change with each release for example to main module? config module?

```
        "-X main.systemRegistriesDirPath=#{etc/"containers/registries.d"}",
        "-X main.unixTempDirForBigFiles=/var/tmp",
        "-X main.systemDefaultPolicyPath=#{etc/"containers/policy.json"}",
        "-X main.systemRegistriesConfPath=#{etc/"containers/registries.conf"}",
```

also, perhaps `skopeo --version --verbose` (or `skypeo --config`, or `skopeo config`)could output configured paths? so could use assert_match in brew formula test stage?

TomSweeneyRedHat:
@lsm5 any tips?

mtrmac:
Thanks for your report.

https://github.com/Homebrew/homebrew-core/pull/47766/files does indeed look painful — likely enough to happen again, _and_ it fails silently.

In principle I think the skopeo `Makefile` should be handling that, e.g. making sure the destinations used to install files (`CONTAINERSSYSCONFIGDIR` and the like) and the paths hard-coded into the binary match. (It may impact integration tests, though.)

Figuring out the c/image major version component automatically within the Makefile should be possible as well. Testing that the overrides are applied and not ignored would, I suppose, happen through the integration tests “impacted” above.

rhatdan:
@glensc @lsm5 Could you work together to open a PR to fix this issue.
I opened a PR to fix Makefile handling of PREFIX and DESTDIR.  https://github.com/containers/skopeo/pull/1168
What else needs to change?  Seems like we could fix this one rather quickly.

mateuszkwiatkowski:
This is also makes porting to FreeBSD harder - FreeBSD ports install files with `PREFIX = /usr/local` and skopeo can't find configuration without manual intervention or patching.

rhatdan:
@mateuszkwiatkowski Please open a PR to fix issues that you see.  My PR was merged back in January, what other issues do you want fixed?

mtrmac:
I don’t think that adding _more_ hard-coded paths to c/image is going to make relocation any easier.

[...]

PR #1168, opened 2021-01-19
Fix Makefile to handle PREFIX correctly

rhatdan:
PREFIX should default to /usr/local according to GNU coding standards.

DESTDIR should just be used for install

This will allow us to specify

make PREFIX=/usr DESTDIR=/tmp/build/ install

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,9 @@
 .PHONY: all binary build-container docs docs-in-container build-local clean install install-binary install-completions shell test-integration .install.vndr vendor vendor-in-container
 
 export GOPROXY=https://proxy.golang.org
+PREFIX ?= /usr/local
 
 ifeq ($(shell uname),Darwin)
-PREFIX ?= ${DESTDIR}/usr/local
 DARWIN_BUILD_TAG=
 # On macOS, (brew install gpgme) installs it within /usr/local, but /usr/local/include is not in the default search path.
 # Rather than hard-code this directory, use gpgme-config. Sadly that must be done at the top-level user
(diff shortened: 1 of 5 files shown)


rhatdan:
Replaces: https://github.com/containers/skopeo/pull/1165

mtrmac (commented):
- This does not replace #1165 , which also cares about GPG compile flags
- #787 would be easier to fix if things like `REGISTRIESDDIR` did _not_ contain `DESTDIR`, so that the variables can be used to substitute the hard-coded paths; and `DESTDIR` were only handled by the `install`* targets.

hack/travis_osx.sh
```diff
@@ -13,5 +13,5 @@ GO111MODULE=off go get -u github.com/cpuguy83/go-md2man golang.org/x/lint/golint
 
 cd ${_containers}/skopeo
 make validate-local test-unit-local bin/skopeo
-sudo make install
+sudo make PREFIX=/usr install
```
mtrmac, line 16: This one seems incorrect (and breaks tests).

mtrmac (commented):
Looking at the other fix-ups, I’m not at all sure changing the Linux default is worth the effort (and the breakage which would probably affect many other users as well). OTOH it _does_ improve consistency with standards, so I don’t have a strong opinion.

vrothberg (commented):
LGTM

mateuszkwiatkowski:
Hello,
What's blocking this PR? I'd like to update my PR (https://github.com/containers/skopeo/pull/1165) so that it could be merged too.

Makefile
```diff
@@ -149,23 +147,23 @@ clean:
 	rm -rf bin docs/*.1
 
 install: install-binary install-docs install-completions
-	install -d -m 755 ${SIGSTOREDIR}
-	install -d -m 755 ${CONTAINERSSYSCONFIGDIR}
-	install -m 644 default-policy.json ${CONTAINERSSYSCONFIGDIR}/policy.json
-	install -d -m 755 ${REGISTRIESDDIR}
-	install -m 644 default.yaml ${REGISTRIESDDIR}/default.yaml
+	install -d -m 755 ${DESTDIR}/${SIGSTOREDIR}
+	install -d -m 755 ${DESTDIR}/${CONTAINERSSYSCONFIGDIR}
```
mtrmac, line 151: `CONTAINERSSYSCONFIGDIR` is now undefined, see the macOS failure.

rhatdan:
Since this PR has now passed test, merging.

PR #1169, opened 2021-01-20
Migrate tests from docker.io

edsantiago:
Switch to using images from quay.io/libpod instead, where
we're not (yet) subject to rate limiting.

Disable (skip) one test that I have no idea how to migrate,
because it needs only-one-arch images from docker.io/$arch.

### DIFF
--- a/systemtest/010-inspect.bats
+++ b/systemtest/010-inspect.bats
@@ -65,59 +65,47 @@ END_EXPECT
 }
 
 @test "inspect: env" {
-    remote_image=docker://docker.io/fedora:latest
+    remote_image=docker://quay.io/libpod/fedora:31
     run_skopeo inspect $remote_image
     inspect_remote=$output
 
     # Simple check on 'inspect' output with environment variables.
     #    1) Get remote image values of environment variables (the value of 'Env')
     #    2) Confirm substring in check_array and the value of 'Env' match.
-    check_array=(PATH=.* )
+    check_array=(FGC=f31 DISTTAG=f31container)
     remote=$(jq '.Env[]' <<<"$inspect_remote")
     for substr in ${check_array[@]}; do
         expect_output --from="$remote" --substring "$substr"
     done
 }
 
+# Tests https://github.com/containers/skopeo/pull/708
 @test "inspect: image manifest list w/ diff platform" {
-    # When --raw is provided, can inspect show the raw manifest list, w/o
-    # requiring any particular platform to be present
-    # To test whether container image can be inspected successfully w/o
-    # platform dependency.
-    #    1) Get current platform arch
-    #    2) Inspect container image is different from current platform arch
-    #    3) Compare output w/ expected result
+    # This image's manifest is for an os + arch that is... um, unlikely
+    # to support skopeo in the foreseeable future. Or past. The image
+    # is created by the make-noarch-manifest script in this directory.
+    img=docker://quay.io/libpod/notmyarch:20210121
 
-    # Here we see a revolting workaround for a podman incompatibility
-    # change: in April 2020, podman info completely changed format
-    # of the keys. What worked until then now throws an error. We
-    # need to work with both old and new podman.
-    arch=$(podman info --format '{{.host.arch}}' || true)
-    if [[ -z "$arch" ]]; then
-        arch=$(podman info --format '{{.Host.Arch}}')
-    fi
+    # Get our host arch (what we're running on). This assumes that skopeo
+    # arch matches podman; it also assumes running podman >= April 2020
+    # (prior to that, the format keys were lower-case).
+    arch=$(podman info --format '{{.Host.Arch}}')
 
-    case $arch in
-        "amd64")
-            diff_arch_list="s390x ppc64le"
-            ;;
-        "s390x")
-            diff_arch_list="amd64 ppc64le"
-            ;;
-        "ppc64le")
-            diff_arch_list="amd64 s390x"
-            ;;
-        "*")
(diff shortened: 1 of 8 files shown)


edsantiago:
@chuanchang @mtrmac could you look at the skipped test and suggest a way to reenable it? I don't understand the point of the test, whether it's to check multiple images, or if it's OK to just check one single image with multiple arches. TIA.

systemtest/010-inspect.bats
```diff
@@ -65,21 +65,27 @@ END_EXPECT
...
     done
 }
 
 @test "inspect: image manifest list w/ diff platform" {
+    # But how? What is the point of this test? Can we just remove the
+    # entire test and do a simple inspect of quay.io/libpod/busybox:latest?
+    # That has a ton of arches:
+    #  inspect... | jq -r '.manifests | .[] | .platform.architecture' | sort -u
+    skip "FIXME FIXME FIXME THIS TEST NEEDS TO BE REWRITTEN"
+
```
TomSweeneyRedHat, line 88: @edsantiago if you have not yet, could you make an issue against Skopeo so we don't forget to circle back to this?

TomSweeneyRedHat:
LGTM
I'm fine with getting this in now to get the RHEL gating tests back up and then circling back to fix up the one test.

edsantiago:
Oh - I was hoping to get this solved within this PR, before merging. But yes, if this merges with the skip, I will file an issue.

edsantiago:
Update, just to show my thinking: `--override-arch` doesn't help. Even with it, skopeo dumps the entire arch list:
```bin
$ bin/skopeo --override-arch=<whatever> inspect --raw docker://quay.io/libpod/busybox:latest |jq .
...long long list
```
where `whatever` can be a real arch (`s390x`) or gibberish (`sdfsdf`).

I suppose we could create a number of `quay.io/libpod/foo_<arch>` images, but it seems wasteful if not needed. I still don't understand the purpose of this test, or, to put it another way, which earlier version of skopeo would have failed this test, and what would that failure look like?

edsantiago:
Looks like integration tests need a lot of dockerpocalypse fixes. That's more than I can take on in this PR.

edsantiago:
No response. Executive decision: completely rewrite the multiarch manifest test, and make my best effort to describe what it's testing (which I still don't fully understand)

mtrmac:
> @chuanchang @mtrmac could you look at the skipped test and suggest a way to reenable it? I don't understand the point of the test, whether it's to check multiple images, or if it's OK to just check one single image with multiple arches. TIA.

I default to thinking that the description is accurate. This came as #746 , probably to test #708 : a failure would be that `skopeo --inspect `**`--raw`** on a manifest list that does not contain the current architecture or OS (most commonly inspecting of a Linux-only image on macOS, but in this case inspecting e.g. an x86+ARM image (no PPC variant) on PPC would fail the same way a non-`--raw` inspect fails, e.g.
```
% bin/skopeo inspect docker://quay.io/libpod/busybox:latest                           
FATA[0002] Error parsing manifest for image: Error choosing image instance: no image found in manifest list for architecture amd64, variant "", OS darwin
```

The test current test is, I guess, elegant in that it works without `--override-arch`, but that also makes it complex and with external dependencies. I’d suggest finding a multi-arch image with a limited set of platforms, and testing that `skopeo --override-arch platform_not_on_the_list inspect --raw $image` works. (Actually a literal `platform_not_on_the_list` + any existing multi-arch image would be a valid, although unrealistic, test of the code path.)

mtrmac:
> Looks like integration tests need a lot of dockerpocalypse fixes. That's more than I can take on in this PR.

They have been reduced somewhat recently. There is OTOH some definite value in testing that the c/image code works with the Docker Hub; we rely on the integration tests for that. Of course, at some point it just makes sense to say that if they make it too hard to test against, we won’t bother.

edsantiago:
> probably to test #708 : a failure would be that skopeo --inspect --raw on a manifest list that does not contain the current architecture or OS

@mtrmac that was exactly what I needed, thank you! Can you see if my new approach looks reasonable?

systemtest/010-inspect.bats
```diff
@@ -65,59 +65,47 @@ END_EXPECT
...
-
-    for arch in $diff_arch_list; do
-        remote_image=docker://docker.io/$arch/golang
-        run_skopeo inspect --tls-verify=false --raw $remote_image
-        remote_arch=$(jq -r '.manifests[0]["platform"]["architecture"]' <<< "$output")
-        expect_output --from="$remote_arch" "$arch" "platform arch of $remote_image"
-    done
+    # This image's manifest is for an os + arch that is... um, unlikely
+    # to support skopeo in the foreseeable future. Or past. The image
+    # is created by the make-noarch-manifest script in this directory.
```
mtrmac, line 86: Oh, that’s clever. And keeping build script as well is perfect.

mtrmac (approved):
LGTM. Thanks!

systemtest/010-inspect.bats
edsantiago: >could you make an issue against Skopeo so we don't forget to circle back to this?

@TomSweeneyRedHat final followup: thanks to valuable feedback from @mtrmac , the test is reenabled in a form that relies only on quay, and should pass on all RHEL builds that I know of. 

One important caveat is that the test will fail if we switch to running CI on AmigaOS. I choose to defer that problem until such date.

TomSweeneyRedHat:
@edsantiago I concur on the fix the other problem later if it becomes a problem.  More importantly, TYVM for all of your work on the gating test this week.

@jnovy FYI

PR #1189, opened 2021-02-10
Enable 'OptimizeDestinationImageAlreadyExists' feature

gjenkins8:
Enable `OptimizeDestinationImageAlreadyExists` feature, developed in https://github.com/containers/image/pull/1041


cmd/skopeo/sync.go
```diff
@@ -531,6 +531,8 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) error {
 		imageListSelection = copy.CopyAllImages
 	}
 
+	optimizeDestinationImageAlreadyExists := true
```
gjenkins8, line 534: Do we want to unilaterally enable this? Or some sort of conditional enablement/disablement via command line flag or environment variable? In case of some sort of unintended behaviour regression?

```
    optimizeDestinationImageAlreadyExists := true
    if len(os.Getenv("SKOPEO_EXPERIMENTAL_DISABLE_OPTIMIZE_DESTINATION_IMAGE_ALREADY_EXISTS")) != 0 {
        optimizeDestinationImageAlreadyExists = false
    }
```

vrothberg: I think we should enable it by default. For `sync` I cannot think of case where it could fail but I have been proven wrong in the past and think we should play it safe: adding a boolean flag to the CLI sounds like a good way forward to me.

Now comes the hard part, naming :^) `--overwrite-destination`, `--destination-exists`?

rhatdan: I hate multi word options.  Just do with --exists and explain it, default to true, and then users can disable it if it causes issues

vrothberg: In that case, I would prefer `overwrite` and let it default to false.

vrothberg: Would that work for you, @rhatdan?

mtrmac: Random drive-by comment: Do we really _need_ an option, just so that we have a workaround for a bug we are not anticipating but afraid of? Once the code stabilizes, we will never recommend use of that option again, will we?

vrothberg: Sounds good to me :+1:

rhatdan: I am really in favor of no option. if that is possible.

gjenkins8: > Once the code stabilizes, we will never recommend use of that option again, will we?

If the behaviour of `OptimizeDestinationImageAlreadyExists` is correct, then it should be functionality transparent. So, I wouldn't expect there to be functional utility in being able to disable (ie. no cli flag desired)

For an 'initial' sync, where destination images don't/are not expected to already exist, attempts to fetch the manifest's for non-existent destination images will be unnecessary overhead. There is perhaps an argument for a cli flag for performance reasons. But in this scenario, this 'initial' sync is probably going to be "slow" anyway (due to content upload). And so, I wouldn't expect the spurious manifest fetches to be significant. So it still doesn't seem desirable to add a cli flag.

gjenkins8: > Do we really need an option, just so that we have a workaround for a bug we are not anticipating but afraid of?

This was the main reason I was thinking of. I would normally look an environment variable for this situation (like in my code snippet). As normally a cli-flag I would expect to remain as part of a permanent interface. Or at least, removing that flag in the future, will cause an breakage for users in such situations like scripting (as minor as that might be in this case).

That said, if we are comfortable with the test coverage and release process. No 'feature flag' disablement is necessary. I think my concern comes from the number of different situations required to be accounted for. During the implementation's code-review. That I'm not familiar with, but I am sure @mtrmac and others understand well.

gjenkins8: Summary:
* Do we need conditional enablement? For feature and/or behavior regression?
* If so, do we want a cli flag (or environment variable)?

I am certainly willing to follow the preferences of maintainers here (of course)

vrothberg: >     * Do we need conditional enablement? For feature and/or behavior regression?
> 
>     * If so, do we want a cli flag (or environment variable)?

I think it should always be and no CLI flag. @mtrmac, @rhatdan and you convinced me :)

gjenkins8: Nice. I will mark the PR ready for review. Thanks!

vrothberg (commented):
> cmd/skopeo/sync.go:564:3: unknown field 'OptimizeDestinationImageAlreadyExists' in struct literal of type copy.Options

You also need to do a `go get github.com/containers/image/v5@master` to fetch the latest c/image code.

vrothberg:
Looks like one last `make vendor` should help make CI happy.

```
GO111MODULE=on go mod vendor

GO111MODULE=on go mod verify

all modules verified

tree is dirty, please commit all changes and sync the vendor.conf

 M go.sum

The command "make vendor && ./hack/tree_status.sh && make local-cross && make check" exited with 1.
```

vrothberg (commented):
LGTM, thank you

CEikermann:
Do you have already scheduled a release which contains this merge request?

PR #1211, opened 2021-03-01
Add local integration and system test targets

cevich:
These tests need to operate as part of the c/image repository CI to
verify downstream-usage.  That environment is already inside the
container built from the Dockerfile (here).  Support this use-case by
adding 'local' targets which bypass the container build.  Also,
simplify the "in-container" check to more specifically verify the exact
container image it's operating under.

Ref: https://github.com/containers/image/pull/1140

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -173,7 +173,11 @@ check: validate test-unit test-integration test-system
 
 # The tests can run out of entropy and block in containers, so replace /dev/random.
 test-integration: build-container
-	$(CONTAINER_RUN) bash -c 'rm -f /dev/random; ln -sf /dev/urandom /dev/random; SKOPEO_CONTAINER_TESTS=1 BUILDTAGS="$(BUILDTAGS)" hack/make.sh test-integration'
+	$(CONTAINER_RUN) bash -c 'rm -f /dev/random; ln -sf /dev/urandom /dev/random; SKOPEO_CONTAINER_TESTS=1 BUILDTAGS="$(BUILDTAGS)" $(MAKE) test-integration-local'
+
+# Intended for CI, shortcut 'build-container' since already running inside container.
+test-integration-local:
+	hack/make.sh test-integration
 
 # complicated set of options needed to run podman-in-podman
 test-system: build-container
@@ -181,10 +185,14 @@ test-system: build-container
 	$(CONTAINER_CMD) --privileged \
 	    -v $$DTEMP:/var/lib/containers:Z -v /run/systemd/journal/socket:/run/systemd/journal/socket \
             "$(IMAGE)" \
-            bash -c 'BUILDTAGS="$(BUILDTAGS)" hack/make.sh test-system'; \
+            bash -c 'BUILDTAGS="$(BUILDTAGS)" $(MAKE) test-system-local'; \
 	rc=$$?; \
 	$(RM) -rf $$DTEMP; \
 	exit $$rc
+
+# Intended for CI, shortcut 'build-container' since already running inside container.
+test-system-local:
+	hack/make.sh test-system
 
 test-unit: build-container
 	# Just call (make test unit-local) here instead of worrying about environment differences
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,6 +46,7 @@ RUN set -x \
 
 ENV GOPATH /usr/share/gocode:/go
 ENV PATH $GOPATH/bin:/usr/share/gocode/bin:$PATH
+ENV container_magic 85531765-346b-4316-bdb8-358e4cca9e5d
 RUN go version
 WORKDIR /go/src/github.com/containers/skopeo
 COPY . /go/src/github.com/containers/skopeo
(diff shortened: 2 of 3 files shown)


mtrmac (commented):
A few drive-by comments:
- ACK in principle, this is clearly useful.
- Please make the in-container targets invoke the new `-local` ones (e.g. have `test-integration` run `make $params test-integration-local` instead of `hack/make.sh integration`, to reduce duplication). That might invite a few more questions about whether things like the `/dev/random` or `SKOPEO_CONTAINER_TESTS` parts belong in the `-local` rule or the run-container rule. (I _haven’t_ looked into these two things in particular at all!)
- WRT the new comments:
  - I’m not very sure which “comments in hack/make.sh” they refer to. The general environment setup documented at the top?
  - Maybe document the `-local` convention somewhere globally? [Or, clearly out of scope, even automate it via some kind of pattern rule? Probably impractical.]
  - Should the new targets have comments with explicit warnings to the effect of “unless you are implementing a CI system and understand how destructive the tests are, never invoke these Make targets”?

cevich:
> Please make the in-container targets invoke the new -local ones

Sure thing, good idea.

> WRT the new comments

Okay, I can put more thought/time into better documenting/commenting these.  Another possibility is I could just drop the `Makefile` changes entirely and instead run the hack script directly from `.cirrus.yml` instead.  Somehow that seems less developer-friendly though.

mtrmac:
WRT the comments, just a quick “this is for experts, don’t use if if you don’t know you are one” is the most important part to me, to decrease risk of accidents.

Given that, it would be perfectly fair to not bother with the details, or even drop the other comments instead of fine-tuning them :) After all, we have hardly have any documentation for `make test-integration` and `make test-system`.

cevich:
> even drop the other comments

I agree, this is what I was working on just before you commented :smile:

mtrmac:
(BTW of `.cirrus.yml`, `Makefile` and `hack/make.sh`, my very weak preference would be to get rid of `hack/make.sh`; it’s a vestigial part of a partial re-use of a much larger build system, probably obscuring more than helping. But clearly whoever is doing the work gets to make the decisions, and I’m not even here. This comment does not exist :) )

Makefile
```diff
@@ -174,19 +174,27 @@ check: validate test-unit test-integration test-system
...
+	hack/make.sh test-integration
 
 # complicated set of options needed to run podman-in-podman
 test-system: build-container
 	DTEMP=$(shell mktemp -d --tmpdir=/var/tmp podman-tmp.XXXXXX); \
 	$(CONTAINER_CMD) --privileged \
 	    -v $$DTEMP:/var/lib/containers:Z -v /run/systemd/journal/socket:/run/systemd/journal/socket \
             "$(IMAGE)" \
-            bash -c 'BUILDTAGS="$(BUILDTAGS)" hack/make.sh test-system'; \
+            bash -c 'BUILDTAGS="$(BUILDTAGS)" hack/make.sh test-system-local'; \
```
mtrmac, line 189:
```suggestion
            bash -c 'make test-system-local BUILDTAGS="$(BUILDTAGS)"'; \
```
? (No idea if it actually works.)

Makefile
```diff
@@ -195,7 +203,7 @@ validate: build-container
 	$(CONTAINER_RUN) hack/make.sh validate-git-marks validate-gofmt validate-lint validate-vet
 
 # This target is only intended for development, e.g. executing it from an IDE. Use (make test) for CI or pre-release testing.
-test-all-local: validate-local test-unit-local
+test-all-local: validate-local test-unit-local test-integration-local test-system-local
```
mtrmac, line 206: Starting years-old OpenShift instances, several registry servers, and the other things `test-integration` and `test-system` do, seem rather undesirable to do directly on the developer’s workstation.

(Up to you if you think the target should be renamed.)

Makefile
cevich: Yeah, I ran into this. Sending it through `make` appears to work, trying that...

Makefile
cevich: Oh okay, I reads the "fort CI testing" part of the comment and got excited :grin:  I'll put it back.

cevich:
@mtrmac I'm also thinking I should try and rig something up to test the changes here, under c/image's `.cirrus.yml` script. My gut tells me this PR is overly simplistic.

cevich:
Confirmed: These changes do allow the c/image CI to at least run the integration tests.  The system-tests currently fail due to a difficult-to-fix podman-in-docker conundrum.  I propose we fix that separately to allow the double travis->cirrus migrations to move forward.

rhatdan:
Works for me.

cevich:
Oof, we're also going to need a workaround for the [silly `toomanyrequests: You have reached your pull rate limit.` thing](https://travis-ci.org/github/containers/skopeo/jobs/761176748) :confounded:

cevich:
@mtrmac PTAL, this one is ready to go.

vrothberg (commented):
LGTM, thank you

Issue #1240, opened 2021-04-09
skopeo login may need write permissions to /run/containers


jonstelly:
I'm trying to run Skopeo on Ubuntu 20.04 in WSL2.  I followed [this guide](https://www.answertopia.com/ubuntu/working-with-containers-on-ubuntu/) for installing.  `skopeo version 1.2.2`

## Initial Issue - docker:// prefix in README.md
Following the instructions from the skopeo README.md at the root of this repository I couldn't connect to any docker registries.  I have a private registry and an azure acr registry.  Both fail with the same error

```
> az acr credential show -n <my-acr>
{
  "passwords": [
    {
      "name": "password",
      "value": "..."
    },
    {
      "name": "password2",
      "value": "..."
    }
  ],
  "username": "<my-acr>"
}
> skopeo login docker://<my-acr>.azurecr.io --username <my-acr>
Password: ...
FATA[0033] error authenticating creds for "docker:": error pinging docker registry docker:: Get "https://docker/v2/": dial tcp: lookup docker on 172.20.208.1:53: no such host
```

I tried running this in an ubuntu docker image (using the 20.10/rolling image) and got a similar result:

```
> docker run -it --rm ubuntu:rolling
> apt update && apt install skopeo
> skopeo login --username <my-arc> docker://<my-arc>.azurecr.io
Password:
FATA[0028] error authenticating creds for "docker:": error pinging docker registry docker:: Get "https://docker/v2/": dial tcp: lookup docker on 192.168.65.5:53: read udp 172.17.0.5:59199->192.168.65.5:53: i/o timeout
```

I then noticed that the [login docs](https://github.com/containers/skopeo/blob/master/docs/skopeo-login.1.md) show a different format than what's in the [README.md at the root of this repository](https://github.com/containers/skopeo/blob/master/README.md) (no `docker://` prefix), so maybe this is just a documentation error?

## Next Issue - /run/containers missing in WSL?
If I omit the `docker://` scheme/prefix then I now get a different error: `FATA[0001] mkdir /run/containers: permission denied`.  So if I run with `sudo skopeo login ...` it works, same with `skopeo copy ...` but I'm a bit confused because the skopeo README says `skopeo does not require the user to be running as root to do most of its operations.`

Is there something I'm doing wrong or missing?  It would be nice to not have to runt his as root, trying to test it out for my CI/CD build agents.

vrothberg:
Regarding the login error:
[...]

PR #1241, opened 2021-04-12
Fix for login / logout registry argument

jonstelly:
login and logout don't seem to support or require the `docker://` scheme prefix.

Fixes part of #1240


vrothberg (approved):
LGTM

Thank you!  The CI flake relates to the Docker Hub rate limit, so we're good to merge

PR #1213, opened 2021-03-02
[WIP - DO NOT MERGE] switch to cirrus

lsm5:
Continuing from pr#1151

Signed-off-by: Lokesh Mandvekar <lsm5@fedoraproject.org>

PR #1249, opened 2021-04-15
Travis -> Cirrus: validate, vendor, and test

cevich:
Partial replacement for #1213 (does ***not*** migrate the problematic OSX-cross build)

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -12,9 +12,13 @@ env:
     GOBIN: "${GOPATH}/bin"
     GOCACHE: "${GOPATH}/cache"
     GOSRC: &gosrc "/var/tmp/go/src/github.com/containers/skopeo"
+    # Required for consistency with containers/image CI
+    SKOPEO_PATH: *gosrc
     CIRRUS_WORKING_DIR: *gosrc
     # The default is 'sh' if unspecified
     CIRRUS_SHELL: "/bin/bash"
+    # Save a little typing (path relative to $CIRRUS_WORKING_DIR)
+    SCRIPT_BASE: "./contrib/cirrus"
 
     ####
     #### Cache-image names to test with (double-quotes around names are critical)
@@ -60,17 +64,26 @@ validate_task:
     script: make validate-local
 
 
-vendor_task:
-    only_if: $CIRRUS_PR != ''
-    container: *build_container
-    script: make vendor && ./hack/tree_status.sh
+cross_task:
+    # TODO: hack/travis_osx.sh doesn't work under cirrus-ci
+    #       for some unknown reason.  Remove the next line when
+    #       fixed (remember to remove from .travis.yml also).
+    allow_failures: $CI == $CI
+    macos_instance:
+        image: catalina-xcode-11.3.1
+    brew_script: brew install gpgme go go-md2man
+    test_script: hack/travis_osx.sh
 
 
-test_task:
-    alias: test
+#####
+##### NOTE: This task is subtantially duplicated in the containers/image
+##### repository's `.cirrus.yml`.  Changes made here should be fully merged
+##### prior to being manually duplicated and maintained in containers/image.
+#####
+test_skopeo_task:
+    alias: test_skopeo
     depends_on:
         - validate
-        - vendor
     gce_instance:
         image_project: libpod-218412
         zone: "us-central1-f"
@@ -87,35 +100,20 @@ test_task:
         - name: "Skopeo Test w/ opengpg"
           env:
               BUILDTAGS: 'btrfs_noversion libdm_no_deferred_remove containers_image_openpgp'
-    env:
-        # TODO: Remove test technical-debt requiring integration tests to run
-        #       inside a container and with --privileged.  This is hiding bugs!
-        PODMANMAKE: >-
-            podman run -it --rm --privileged
-                -e GOPATH=$GOPATH
-                -v $GOPATH:$GOPATH:Z
-                -w $GOSRC
-                $SKOPEO_CI_CONTAINER_FQIN
-                    make
-    setup_script: |
-        # This is required as part of the standard Fedora VM setup
-        growpart /dev/sda 1
-        resize2fs /dev/sda1
-
-        # VM's come with the distro. skopeo pre-installed
-        dnf erase -y skopeo
-
-        # Ensure we start with a clean-slate
-        podman system reset --force
-    # Note: We _HAVE_ to spam BUILDTAGS in this way due to how Cirrus-CI
-    # mangles quoted strings when performing in-line 'env' substitution.
-    main_script: |
-        set -x
-        $PODMANMAKE bin/skopeo BUILDTAGS="$BUILDTAGS"
-        $PODMANMAKE test-unit-local BUILDTAGS="$BUILDTAGS"
-        $PODMANMAKE test-integration-local BUILDTAGS="$BUILDTAGS"
-        # Do not run inside podman, uses containers internally.
-        make test-system-local BUILDTAGS="$BUILDTAGS"
+    setup_script: >-
+        "${GOSRC}/${SCRIPT_BASE}/runner.sh" setup
+    vendor_script: >-
+        "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" vendor
+    build_script: >-
+        "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" build
+    validate_script: >-
+        "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" validate
+    unit_script: >-
+        "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" unit
+    integration_script: >-
+        "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" integration
+    system_script: >
+        "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" system
 
 
 # This task is critical.  It updates the "last-used by" timestamp stored
@@ -153,8 +151,8 @@ success_task:
     # N/B: ALL tasks must be listed here, minus their '_task' suffix.
     depends_on:
         - validate
-        - vendor
-        - test
+        - cross
+        - test_skopeo
         - meta
     container: *smallcontainer
     env:
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -11,7 +11,7 @@ env:
     GOPATH: &gopath "/var/tmp/go"
     GOBIN: "${GOPATH}/bin"
     GOCACHE: "${GOPATH}/cache"
-    GOSRC: &gosrc "/var/tmp/go/src/github.com/containers/podman"
+    GOSRC: &gosrc "/var/tmp/go/src/github.com/containers/skopeo"
     CIRRUS_WORKING_DIR: *gosrc
     # The default is 'sh' if unspecified
     CIRRUS_SHELL: "/bin/bash"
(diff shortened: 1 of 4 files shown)


cevich:
Oof, well :hankey: ! Things were running along fine,[ then ploP!](https://cirrus-ci.com/task/4643706780254208?logs=main#L200-L196)

```
# $ podman --cgroup-manager=cgroupfs run -d --name reg -v /tmp/skopeo_bats.DNuIGc/auth:/auth:Z -p 5000:5000 quay.io/libpod/registry:2
# time="2021-04-15T15:00:48-05:00" level=error msg="User-selected graph driver \"vfs\" overwritten by graph driver \"overlay\" from database - delete libpod local files to resolve"
# 9760db076413233f2a16402c70188f52191256762d6d188ecc25174c85a93517
# /var/tmp/go/src/github.com/containers/skopeo/systemtest/helpers.bash: connect: No route to host
# /var/tmp/go/src/github.com/containers/skopeo/systemtest/helpers.bash: line 355: /dev/tcp/127.0.0.1/5000: No route to host
```

I don't remember hitting that when implementing the same test for c/image :confused:

cevich:
Indeed, [same test in c/image, same images (I think) 2 days ago passed](https://cirrus-ci.com/task/5288613098291200?logs=main#L256).

Harrumph!

cevich:
@vrothberg have you seen this before?  Any clue about the cause?

vrothberg:
> @vrothberg have you seen this before? Any clue about the cause?

This could suggest that podman already ran before with VFS.  A `podman system reset` may do the trick.

cevich:
> This could suggest that podman already ran before with VFS.

Right, the integration tests ran prior to this and I find

```golang
integration/copy_test.go:       storage = fmt.Sprintf("[vfs@%s/root+%s/runroot]", storage, storage)
```

However while the unit and integration tests run ***inside*** a `--privileged` container, they don't volume-mount anything other than `$GOPATH`.

Also, other than not replacing the c/image sources, the failing Cirrus task is exactly the same as what's used there - where we don't have this problem.  Suggesting it's some kind of rare race...that I happened to hit twice in a row.

Puzzling :confused:

Trying again with a `system reset` as suggested...

cevich:
...hrmmm.  Okay, it passed.  Let's go with that then, and I'll add the same change to c/image CI.

cevich:
Marked this as draft so I can work up a script to prevent lots of duplication w/ c/image automation.

cevich:
Okay this is ready.  Once it merges, I can finish up it's twin: https://github.com/containers/image/pull/1202

cevich:
**Note to me:** After this merges, need to adjust the branch-protection required status checks to include 'Total Success'.

cevich:
(Rebased and force-pushed)

rhatdan:
LGTM
@vrothberg PTAL

vrothberg (commented):
@cevich, can you do another rebase?

The changes LGTM

cevich:
Rebased

vrothberg:
Great work, @cevich !

.travis.yml
```diff
@@ -67,14 +67,6 @@ jobs:
...
       script:
       - hack/travis_osx.sh
-    - stage: local-build
-      <<: *local-build
-      name: local build for linux
-      os: linux
-      services:
-      - docker
-      script: 
-      - make vendor && ./hack/tree_status.sh && make local-cross && make check
```
mtrmac, line 77: Is there anything providing the equivalent of `make vendor && ./hack/tree_status.sh` (i.e. ensure that we are not committing any patches, or missing any updates, to the `vendor` subdirectory?

mtrmac, line 77: Was `local-cross` (e.g. testing that this can be built even on Windows) dropped intentionally?

.cirrus.yml
```diff
@@ -45,6 +52,70 @@ timeout_in: 30m
...
+          env:
+              BUILDTAGS: 'btrfs_noversion libdm_no_deferred_remove containers_image_openpgp'
+    setup_script: >-
+        "${GOSRC}/${SCRIPT_BASE}/runner.sh" setup
+    vendor_script: >-
+        "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" vendor
+    build_script: >-
+        "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" build
+    validate_script: >-
+        "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" validate
```
mtrmac, line 110: So `validate-local` is run both from its own `validate_task` and here. Is that intentional? (It might well be worth it for consistency with c/image .)

contrib/cirrus/runner.sh
```diff
@@ -0,0 +1,91 @@
...
+_run_build() {
+    podmanmake bin/skopeo BUILDTAGS="$BUILDTAGS"
+}
+
+_run_validate() {
+    podmanmake validate-local BUILDTAGS="$BUILDTAGS"
+}
+
+_run_unit() {
+    podmanmake test-integration-local BUILDTAGS="$BUILDTAGS"
```
mtrmac, line 65:
```suggestion
    podmanmake test-unit-local BUILDTAGS="$BUILDTAGS"
```

contrib/cirrus/runner.sh
```diff
@@ -0,0 +1,91 @@
...
+}
```
mtrmac, line 66: _In principle_ these might not need to run in a container, if the top level has all the relevant Go tools in the right versions. Doing it in a container doesn’t hurt, of course.

(`test-integration` more or less needs a container because it starts several registry servers, and `test-system` starts registry servers _in another container_ IIRC.

mtrmac (commented):
@cevich I’m (very) slowly catching up, and I have some questions about the intent/trade-offs.

I’ll of course prepare PRs if necessary, at this point I’m primarily trying to learn how this works.

cevich:
> I’m (very) slowly catching up, and I have some questions about the intent/trade-offs.

oh this is great, I'm glad you're reviewing this.  I'm just a wing-nut pushing buttons, so please don't be afraid to call me out.

.travis.yml
cevich: Yes (I just checked).  This should be [caught by the `validate` task](https://github.com/containers/skopeo/blob/main/.cirrus.yml#L64) (unless I'm missing something).

cevich: Hmmm...this might be an "oops".  Should it be added to the `cross` task (running on a mac), `validate` or a new task (so it can run in parallel)?  I'm not sure how long it takes and/or if it should block anything on failure. Please advise.

.cirrus.yml
cevich: Yes this was intentional (for the reason you identified).  Happy to adjust things if you see a need, but IIRC this particular operation doesn't take very long.

contrib/cirrus/runner.sh
cevich: :hankey: 

Good catch :+1:

contrib/cirrus/runner.sh
cevich: > Doing it in a container doesn’t hurt, of course.

Actually it does.  Unless it's been fixed, there was at least one test that absolutely refused to run correctly outside of a `--privileged` container.  This was missed (for possibly years) precisely *because* the tests only ever ran in a container.

My (strong) preference is that tests run 100% on the host, and then only secondarily run inside a container (if specifically required for the test-subject).  Compromises were made here (and a big-fat 'TODO' added) for the sake of getting off of travis faster.

cevich: I'll open a PR to drop the container stuffs, let's see what breaks (it's been quite a while since I've tried).

cevich: Ref: https://github.com/containers/skopeo/pull/1334

.travis.yml
mtrmac: `validate-local` AFAICS does not do this; there isn’t a single reference to `hack/tree_status.sh`, and I can’t find any equivalent in the repo.

(The exclusion from `validate-local` is somewhat intentional or at least useful: It’s useful to run `make validate-local` during development _on uncommitted code_, and that would trigger `tree_status.sh`.

The larger picture is that _Skopeo_ PRs, we should never need to run `make vendor` _on top_; the PR should already contain an update to the `vendor` subdirectory. So `make vendor && ./hack/tree_status.sh` is a sanity check that the PR is self-contained and no further `make vendor` is necessary. OTOH if c/image tests replace the c/image implementation vendored into Skopeo, we do need to run `make vendor` _somewhere_, and _not_ fail c/image tests on uncommitted `vendor/containers/image`.)

So maybe we want to:
- Do this “`vendor` is fully committed” check in `validate_task`
- I don’t really care where the code lives, a `make validate-vendor-is-committed` in the `Makefile` (along with comments about why it is not included in `validate-local`), keeping `hack/tree_status.sh`, would probably be most consistent with the rest; alternatively all in `Makefile` and getting rid of the separate shell script, or all only in `.cirrus.yml` and nothing in `Makefile`, would be fine with me as well.

mtrmac: It’s a set of 5 compiles, probably not _too_ expensive; with `go build -a` to remove some of the caching effects, on my local system, 91s of wall clock time, 362s of CPU time, but I didn’t measure this in Cirrus.

I think this should overall block `success_task`, but probably nothing else — if a Windows build fails, that’s good to know but not directly a reason not to run the Linux integration tests; both might fail for different reasons. _Typically_ the most interesting failures happen either early in `validate` (missing sign-offs, bad formatting) or in integration tests, so it would be nice for users to see such failures reasonably fast.

Conceptually it’s a part of the “non-Linux build” checks, i.e. probably `cross_task`. OTOH it is architecture-independent, and makes sense to run this on our cheapest machines, and perhaps to avoid the overhead of extra environment setup / pulls / go module fetches and all that, so adding it to either `validate_task` or `test_skopeo_task` to maximize parallelism and minimize total cost might be worthwhile.

TMTOWTDI:
- Rename `validate_task` to `cheaper_tests_task`, with a `validate_script` and `cross_script`.
- Repurpose `cross_task` to run `local-cross`, and move the existing macOS tests into a `mac_task`.

Many other things would work. Which one do you prefer?

contrib/cirrus/runner.sh
mtrmac: > Actually it does. Unless it's been fixed, there was at least one test that absolutely refused to run correctly outside of a `--privileged` container.

That should happen for `test-integration`, but not `test-unit`. (That typo identified elsewhere might well have played a part.)

The integration tests don’t care about a _container_ as such, it’s just that they are somewhat invasive (run multiple servers, and the like). If a host opts in by `SKOPEO_CONTAINER_TESTS=1` / `container_magic`, running them directly on the host is fine with me (as long as the host’s environment is kept similar to `Dockerfile`, with low maintenance effort).

contrib/cirrus/runner.sh
mtrmac: Filed #1339 for this, while we discuss/experiment with the rest.

.travis.yml
cevich: > So maybe we want to:

That sounds good.  IIRC, c/image has a similar 'validate' task...crap, it seems that doesn't have a `make vendor` +  `hack/tree_status.sh` either :disappointed: 

Yes we should keep `hack/tree_status.sh` for consistency.

I just looked...podman does it in a shell script (not Makefile), buildah does it directly in `.cirrus.yml`. 

Okay, I opened a PR: https://github.com/containers/skopeo/pull/1340

mtrmac: c/image doesn’t need a `tree_status.sh` because it doesn’t maintain the `vendor` subdirectory as part of the repo. (Vendoring dependencies doesn’t work well for libraries that have vendored types as part of the API.)

cevich: > makes sense to run this on our cheapest machines,

Oh! Okay, well that would actually be the `cross_task` - a Mac VM that we pay $0 for :grin: 

> maximize parallelism and minimize total cost might be worthwhile.

Ease of debugging is important too.  This (and c/image) traffic is so low, I'm not afraid to throw another VM at a 5-minute test.  That will make it parallel and easy for anybody to debug.

Opened PR: https://github.com/containers/skopeo/pull/1341

contrib/cirrus/runner.sh
cevich: >  host’s environment is kept similar to Dockerfile, with low maintenance effort

Yep, we're on the same page here.  This is what I was playing with in https://github.com/containers/skopeo/pull/1334 using a script to do the shared setup stuff, then calling the script from both the Dockerfile and Cirrus-CI.  The logistics and setup seem to be working, but the tests are breaking (differently from what I remember before).  In any case, we can continue that design/discussion over in that PR.

.travis.yml
cevich: Ya, I noticed that too (after I actually went looking for it).  Okay, we should be all caught up now then (I think).  I's dotted and T's crossed :smile: Let me know if you spot any other blunders I made.

mtrmac: > Okay, we should be all caught up now then (I think). I's dotted and T's crossed 😄 

Yup. Thanks again!

Issue #1262, opened 2021-04-23
Using OCI format for skopeo sync for multi-architecture images


Freakin:
I have a need to replicate a set of public images to our internal registry, and `skopeo sync` looked like it was working great for this.

However, when I attempted to use multi-arch images with `sync`, my registry (Harbor) only uses the tag on the most recently pushed image and the previously pushed/tagged images have their tags cleared.

When I use 
`skopeo copy docker://docker.io/nginx:stable-perl docker://harbor.my-domain.com/my-project/nginx:stable-perl --format oci --all`
It works perfectly and uploads the OCI manifest to Harbor with all the images included and properly tagged.

But it appears that `skopeo sync` does not have the format parameter.  Is there any plan to implement manifest format for `sync`?

rhatdan:
Interested in opening a PR to add it?

Freakin:
PR submitted @rhatdan

mtrmac:
Per https://github.com/containers/skopeo/pull/1263#issuecomment-827048749 , this seems to work as expected (with `--all`). Please reopen if that’s not the case.

PR #1263, opened 2021-04-23
Added format parameter to sync command

Freakin:
Signed-off-by: Brendan Aye <brendan.aye@t-mobile.com>

### DIFF
--- a/cmd/skopeo/sync.go
+++ b/cmd/skopeo/sync.go
@@ -31,12 +31,13 @@ type syncOptions struct {
 	srcImage          *imageOptions     // Source image options
 	destImage         *imageDestOptions // Destination image options
 	retryOpts         *retry.RetryOptions
-	removeSignatures  bool   // Do not copy signatures from the source image
-	signByFingerprint string // Sign the image using a GPG key with the specified fingerprint
-	source            string // Source repository name
-	destination       string // Destination registry name
-	scoped            bool   // When true, namespace copied images at destination using the source repository name
-	all               bool   // Copy all of the images if an image in the source is a list
+	removeSignatures  bool           // Do not copy signatures from the source image
+	signByFingerprint string         // Sign the image using a GPG key with the specified fingerprint
+	format            optionalString // Force conversion of the image to a specified format
+	source            string         // Source repository name
+	destination       string         // Destination registry name
+	scoped            bool           // When true, namespace copied images at destination using the source repository name
+	all               bool           // Copy all of the images if an image in the source is a list
 }
 
 // repoDescriptor contains information of a single repository used as a sync source.
@@ -95,6 +96,7 @@ See skopeo-sync(1) for details.
 	flags := cmd.Flags()
 	flags.BoolVar(&opts.removeSignatures, "remove-signatures", false, "Do not copy signatures from SOURCE images")
 	flags.StringVar(&opts.signByFingerprint, "sign-by", "", "Sign the image using a GPG key with the specified `FINGERPRINT`")
+	flags.VarP(newOptionalStringValue(&opts.format), "format", "f", `MANIFEST TYPE (oci, v2s1, or v2s2) to use when syncing image(s) to a destination (default is manifest type of source)`)
 	flags.StringVarP(&opts.source, "src", "s", "", "SOURCE transport type")
 	flags.StringVarP(&opts.destination, "dest", "d", "", "DESTINATION transport type")
 	flags.BoolVar(&opts.scoped, "scoped", false, "Images at DESTINATION are prefix using the full source image path as scope")
@@ -536,6 +538,14 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) error {
 		return err
 	}
 
+	var manifestType string
+	if opts.format.present {
+		manifestType, err = parseManifestFormat(opts.format.value)
+		if err != nil {
+			return err
+		}
+	}
+
 	ctx, cancel := opts.global.commandTimeoutContext()
 	defer cancel()
 
@@ -562,6 +572,7 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) error {
 		DestinationCtx:                        destinationCtx,
 		ImageListSelection:                    imageListSelection,
 		OptimizeDestinationImageAlreadyExists: true,
+		ForceManifestMIMEType:                 manifestType,
 	}
 
 	for _, srcRepo := range srcRepoList {
--- a/cmd/skopeo/utils.go
+++ b/cmd/skopeo/utils.go
@@ -2,14 +2,17 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"os"
 	"strings"
 
 	"github.com/containers/common/pkg/retry"
+	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/pkg/compression"
 	"github.com/containers/image/v5/transports/alltransports"
 	"github.com/containers/image/v5/types"
+	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -244,6 +247,21 @@ func parseImageSource(ctx context.Context, opts *imageOptions, name string) (typ
 		return nil, err
 	}
 	return ref.NewImageSource(ctx, sys)
+}
+
+// parseManifestFormat parses format parameter for copy and sync command.
+// It returns string value to use as manifest MIME type
+func parseManifestFormat(manifestFormat string) (string, error) {
+	switch manifestFormat {
+	case "oci":
+		return imgspecv1.MediaTypeImageManifest, nil
+	case "v2s1":
+		return manifest.DockerV2Schema1SignedMediaType, nil
+	case "v2s2":
+		return manifest.DockerV2Schema2MediaType, nil
+	default:
+		return "", fmt.Errorf("unknown format %q. Choose one of the supported formats: 'oci', 'v2s1', or 'v2s2'", manifestFormat)
+	}
 }
 
 // usageTemplate returns the usage template for skopeo commands
--- a/cmd/skopeo/utils_test.go
+++ b/cmd/skopeo/utils_test.go
@@ -4,7 +4,9 @@ import (
 	"os"
 	"testing"
 
+	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/types"
+	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/spf13/cobra"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -201,6 +203,38 @@ func TestImageDestOptionsNewSystemContext(t *testing.T) {
 	opts = fakeImageDestOptions(t, "dest-", []string{}, []string{"--dest-creds", ""})
 	_, err = opts.newSystemContext()
 	assert.Error(t, err)
+}
+
+func TestParseManifestFormat(t *testing.T) {
+	for _, testCase := range []struct {
+		formatParam          string
+		expectedManifestType string
+		expectErr            bool
+	}{
+		{"oci",
+			imgspecv1.MediaTypeImageManifest,
+			false},
+		{"v2s1",
+			manifest.DockerV2Schema1SignedMediaType,
+			false},
+		{"v2s2",
+			manifest.DockerV2Schema2MediaType,
+			false},
+		{"",
+			"",
+			true},
+		{"badValue",
+			"",
+			true},
+	} {
+		manifestType, err := parseManifestFormat(testCase.formatParam)
+		if testCase.expectErr {
+			require.Error(t, err)
+		} else {
+			require.NoError(t, err)
+		}
+		assert.Equal(t, manifestType, testCase.expectedManifestType)
+	}
 }
 
 // since there is a shared authfile image option and a non-shared (prefixed) one, make sure the override logic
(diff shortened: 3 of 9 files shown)


Freakin:
PR to fix #1262

cmd/skopeo/sync.go
```diff
@@ -562,6 +580,7 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) error {
 		DestinationCtx:                        destinationCtx,
 		ImageListSelection:                    imageListSelection,
 		OptimizeDestinationImageAlreadyExists: true,
+		ForceManifestMIMEType:                 manifestType,
```
TomSweeneyRedHat, line 583: any man page changes?

vrothberg (commented):
Thanks!

As @TomSweeneyRedHat mentioned, could you add the new option to the `skopeo-sync` man page?  Also, a test or two in `./integration` would be good.  The formatting code paths are well exercised but I want to make sure that we're not silently regressing in the future.

mtrmac (commented):
\*shrug\* I don’t care strongly about having this option in `sync`, as such.

But looking at the original report

> However, when I attempted to use multi-arch images with `sync`, my registry (Harbor) only uses the tag on the most recently pushed image and the previously pushed/tagged images have their tags cleared.

Do you mean that all tags in the same repo are removed on a multi-arch push? Or that somehow before there were multiple images with the same tag, and those are now cleared? Or are you just saying that after copying image1 to repo:tag and image2 to repo:tag, image1 no longer has the repo:tag label? Something else? Either way, that report is rather concerning.

Is it possible that the difference is something else, maybe using `--all`? Or, unlikely, were you trying to create a multi-arch image by repeatedly copying the various single-arch images to the same tag?

If that described behavior happens (and it’s not just a tag move on a copy of new image to the same tag), it’s very surprising and rather concerning; AFAICS there’s nothing `skopeo sync` specific to that, and it’s not clear how changing the image format should matter. Having `skopeo copy` untag unrelated images seems like a very serious unwanted side effect that should be understood in detail before adding workarounds like this and papering over the situation.

cmd/skopeo/sync.go
mtrmac: Also `completions/bash/skopeo` (which seems to be missing most of `sync` options).

docs/skopeo-sync.1.md
```diff
@@ -54,6 +54,8 @@ Path of the authentication file for the destination registry. Uses path given by
 
 **--dest** _transport_ Destination transport.
 
+**--format, -f** _manifest-type_ Manifest type (oci, v2s1, or v2s2) to use when saving image to directory using the 'dir:' transport (default is manifest type of source)
```
mtrmac, line 57: The `dir:` transport restriction is not true.

cmd/skopeo/sync.go
```diff
@@ -536,6 +540,20 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) error {
...
+		case "oci":
+			manifestType = imgspecv1.MediaTypeImageManifest
+		case "v2s1":
+			manifestType = manifest.DockerV2Schema1SignedMediaType
+		case "v2s2":
+			manifestType = manifest.DockerV2Schema2MediaType
+		default:
+			return fmt.Errorf("unknown format %q. Choose one of the supported formats: 'oci', 'v2s1', or 'v2s2'", opts.format.value)
+		}
+	}
```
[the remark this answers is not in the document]
Freakin: > Do you mean that all tags in the same repo are removed on a multi-arch push? Or that somehow before there were multiple images with the same tag, and those are now cleared? Or are you just saying that after copying image1 to repo:tag and image2 to repo:tag, image1 no longer has the repo:tag label? Something else? Either way, that report is rather concerning.

When copying image1 to `repo:tag` and then copying image2 to `repo:tag`, the `tag` is removed from image1

I think this is likely a limitation of Harbor rather than any issue with Skopeo.  When I attempt to push a multi-arch image using `-all` parameter in `skopeo sync` it fails on Harbor < 2.x.  If multiple architectures for the same `image:tag` are pushed sequentially, the above-mentioned behavior is observed.

With Harbor 2.x and their support for OCI manifests, it appears to correctly infer the MIME type and uses an OCI manifest by default to support multi-arch.  


> Please share this in utils.go instead of copy&pasting from the copy subcommand.

Just to clarify, you're suggesting something like a `parseManifestFormat` function in `utils.go` to be used for `copy` and `sync` commands?

mtrmac:
> When copying image1 to `repo:tag` and then copying image2 to `repo:tag`, the `tag` is removed from image1

Thanks. That looks entirely expected in the container registry model: If a client later requests to download `repo:tag`, it will expect to get image2 and never image1; so the `repo:tag` is no longer pointing to image1, i.e. that it was “removed” from image1 in a sense.

So that is unrelated to the need for this option?

---

> When I attempt to push a multi-arch image using `-all` parameter in `skopeo sync` it fails on Harbor < 2.x.

A schema2 image fails by default but it does work with a manual `--format ? How does that fail? The code is supposed to _automatically_ try various formats. If there is a bug in that code, we should fix it if possible, rather than force users to manually use an option.

> With Harbor 2.x and their support for OCI manifests, it appears to correctly infer the MIME type and uses an OCI manifest by default to support multi-arch.

Or are you saying that all multi-arch pushes fail on Harbor < 2.x, and they automatically work without the `--format` option on ≥ 2.x? So the option is not _necessary_ to make sync work?

If the multi-arch code should work by default but doesn’t fall back correctly, please share a `skopeo --debug sync` output (maybe in the issue, not in this PR).

If this option is not required for the sync to work, it might still make sense to provide it, I just want to make sure we are not ignoring a bug that should be fixed.

cmd/skopeo/sync.go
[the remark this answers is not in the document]
mtrmac: > Just to clarify, you're suggesting something like a `parseManifestFormat` function in `utils.go` to be used for `copy` and `sync` commands?

Yes, please. (Alternatively, it could make sense to share much more of the CLI handling for setting up `copy.Options` between `copy` and `sync`, similarly to `imageOptions` providing `imageOptions.newSystemContext`, but that’s way out of scope of this PR.)

Freakin:
> > When copying image1 to `repo:tag` and then copying image2 to `repo:tag`, the `tag` is removed from image1
> 
> Thanks. That looks entirely expected in the container registry model: If a client later requests to download `repo:tag`, it will expect to get image2 and never image1; so the `repo:tag` is no longer pointing to image1, i.e. that it was “removed” from image1 in a sense.
> 
> So that is unrelated to the need for this option?
> 
> > When I attempt to push a multi-arch image using `-all` parameter in `skopeo sync` it fails on Harbor < 2.x.
> 
> A schema2 image fails by default but it does work with a manual `--format` ? How does that fail? The code is supposed to _automatically_ try various formats. If there is a bug in that code, we should fix it if possible, rather than force users to manually use an option.
> 
> > With Harbor 2.x and their support for OCI manifests, it appears to correctly infer the MIME type and uses an OCI manifest by default to support multi-arch.
> 
> Or are you saying that all multi-arch pushes fail on Harbor < 2.x, and they automatically work without the `--format` option on ≥ 2.x? So the option is not _necessary_ to make sync work?
> 
> If the multi-arch code should work by default but doesn’t fall back correctly, please share a `skopeo --debug sync` output (maybe in the issue, not in this PR).
> 
> If this option is not required for the sync to work, it might still make sense to provide it, I just want to make sure we are not ignoring a bug that should be fixed.

Yeah looking like I'm just an idiot.  Not sure what problem I was having initially with Harbor 2.x, but doing a normal `sync --all` from docker hub to Harbor 2.x works fine, automatically determines OCI manifest format, and uploads the properly tagged multi-arch image without any issues.

Harbor 1.x does not support multi-arch image manifests and fails as expected with `Manifest.list is not supported` response

I'm happy to finish up this PR or it can be closed if you don't see a reason to add the param to `sync`.  Doesn't look like I need it for my purposes as originally thought.

mtrmac:
Thanks very much for the update.

> I'm happy to finish up this PR or it can be closed if you don't see a reason to add the param to `sync`.

Up to you, I don’t feel strongly about this at all. As the above shows, sometimes the discussion about the option can be more helpful than the existence of the option :)

OTOH it’s certainly attractive for `copy` and `sync` to behave very similarly and to provide substantially the same tunables.

rhatdan:
I like the idea of the option.

cmd/skopeo/sync.go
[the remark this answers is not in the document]
Freakin: Moved parsing to `utils.go` and added a test for it

cmd/skopeo/sync.go
Freakin: @mtrmac would adding completions for `sync` (which don't exist today) be out of scope for this PR?

Freakin:
Added basic integration test for `sync --format`

Let me know if anything else is needed for this PR

cmd/skopeo/sync.go
rhatdan: No you should add them.

cmd/skopeo/utils.go
```diff
@@ -2,17 +2,21 @@ package main
...
 
 	"github.com/containers/common/pkg/retry"
+	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/pkg/compression"
 	"github.com/containers/image/v5/transports/alltransports"
 	"github.com/containers/image/v5/types"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+
```
rhatdan, line 18: Remove empty line.

cmd/skopeo/utils_test.go
```diff
@@ -4,10 +4,13 @@ import (
 	"os"
 	"testing"
 
+	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/types"
 	"github.com/spf13/cobra"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
```
rhatdan, line 12: Remove blank line.

integration/sync_test.go
```diff
@@ -12,8 +12,11 @@ import (
 
 	"github.com/containers/image/v5/docker"
 	"github.com/containers/image/v5/docker/reference"
+	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/types"
 	"github.com/go-check/check"
+
```
rhatdan, line 18: Remove blank line.

rhatdan (approved):
Nice job, a few nits and missing completions.

cmd/skopeo/sync.go
mtrmac: Please do add a completion for the new `--format` option.

I don’t think it would be fair to block this PR on adding full completion support for the pre-existing options, though we certainly wouldn’t reject that work :)

mtrmac:
> Let me know if anything else is needed for this PR

Completions (discussed elsewhere) and an update to `docs/skopeo-sync.1.md`, I think. Code looks good.

Freakin:
> Completions (discussed elsewhere) and an update to `docs/skopeo-sync.1.md`, I think. Code looks good.

Added all completions for sync.  

What update is needed for `docs/skopeo-sync.1.md`?  I added the format parameter previously and corrected the language regarding the `dir:` restrictions based on previous feedback.

Freakin:
Looks like `validate` is failing on `gofmt` due to wanting to sort imports.  Sorted imports don't look like are being used in the current HEAD of master.

Should I add a commit with sorted imports?

mtrmac:
> What update is needed for `docs/skopeo-sync.1.md`? I added the format parameter previously and corrected the language regarding the `dir:` restrictions based on previous feedback.

I’m sorry, I must have mistakenly viewed only one of the commits or something.

cmd/skopeo/sync.go
```diff
@@ -95,6 +96,7 @@ See skopeo-sync(1) for details.
 	flags := cmd.Flags()
 	flags.BoolVar(&opts.removeSignatures, "remove-signatures", false, "Do not copy signatures from SOURCE images")
 	flags.StringVar(&opts.signByFingerprint, "sign-by", "", "Sign the image using a GPG key with the specified `FINGERPRINT`")
+	flags.VarP(newOptionalStringValue(&opts.format), "format", "f", `MANIFEST TYPE (oci, v2s1, or v2s2) to use when saving image to directory using the 'dir:' transport (default is manifest type of source)`)
```
mtrmac, line 99:
```suggestion
	flags.VarP(newOptionalStringValue(&opts.format), "format", "f", `MANIFEST TYPE (oci, v2s1, or v2s2) to use when saving image (default is manifest type of source)`)
```

or maybe without the (default is…) part because the code conversions automatically, and explaining that in `--help` would be too verbose.

Same in the man page.

mtrmac:
> Looks like `validate` is failing on `gofmt` due to wanting to sort imports. Sorted imports don't look like are being used in the current HEAD of master.

The failure says
> `Please reformat the above files using "gofmt -s -w" and commit the result.`

and that should be all that is required. (The imports are sorted by package name, not purely line-by-line, and only within the empty-line-delineated sections. So removing an empty line can change the order.)

> Should I add a commit with sorted imports?

It might be easier to squash all the commits into one, including the changes made by the `gofmt` step.

Freakin:
> > Looks like `validate` is failing on `gofmt` due to wanting to sort imports. Sorted imports don't look like are being used in the current HEAD of master.
> 
> The failure says
> 
> > `Please reformat the above files using "gofmt -s -w" and commit the result.`
> 
> and that should be all that is required. (The imports are sorted by package name, not purely line-by-line, and only within the empty-line-delineated sections. So removing an empty line can change the order.)
> 
> > Should I add a commit with sorted imports?
> 
> It might be easier to squash all the commits into one, including the changes made by the `gofmt` step.

Fixed docs and help language and squashed commits

rhatdan:
LGTM

TomSweeneyRedHat:
LGTM
@vrothberg ?

vrothberg (commented):

PR #1285, opened 2021-05-14
Add `go get` install method

stevegt:
Probably should also remove the GOPATH instructions, but for now this just adds a `go get`.


install.md
```diff
@@ -159,7 +159,13 @@ brew install gpgme
 sudo zypper install libgpgme-devel device-mapper-devel libbtrfs-devel glib2-devel
 ```
 
-Make sure to clone this repository in your `GOPATH` - otherwise compilation fails.
+For newer versions of Go with modules support, after installing the above prereqs, you can build and install Skopeo using a simple `go get`:
```
rhatdan, line 162: Could you define newer?  IE >= golang 1.14, 1.15, 1.16?

stevegt: It's not version-specific; it depends on environment variable content as well as the version of Go.  I'll include a URL.

install.md
```diff
@@ -159,7 +159,18 @@ brew install gpgme
 sudo zypper install libgpgme-devel device-mapper-devel libbtrfs-devel glib2-devel
 ```
 
-Make sure to clone this repository in your `GOPATH` - otherwise compilation fails.
+For newer versions of Go configured with modules support, after installing the
+above prereqs, you can build and install Skopeo using a simple `go
+get`.  (For more information about module support and related environment
+configuration in Go 1.16, see [New module changes in Go 1.16](https://blog.golang.org/go116-module-changes).
```
mtrmac, line 165: That link explicitly recommends _against_ `go get` for installation, preferring `go install`.

install.md
```diff
@@ -159,7 +159,18 @@ brew install gpgme
 sudo zypper install libgpgme-devel device-mapper-devel libbtrfs-devel glib2-devel
 ```
 
-Make sure to clone this repository in your `GOPATH` - otherwise compilation fails.
+For newer versions of Go configured with modules support, after installing the
+above prereqs, you can build and install Skopeo using a simple `go
+get`.  (For more information about module support and related environment
+configuration in Go 1.16, see [New module changes in Go 1.16](https://blog.golang.org/go116-module-changes).  
+For earlier Go versions, see [Go Modules in 2019](https://blog.golang.org/modules2019) 
+and [Using Go modules](https://blog.golang.org/using-go-modules).)
```
mtrmac, line 167: It’s unclear to me what I’m supposed to be looking for in these links to long documents..

mtrmac (commented):
Overall, I’m not at all sure this is worth it.

We don’t want to recommend N different ways to do a similar thing; the less variation, the easier debugging and collaboration.

Just using `go get` drops various features of the existing `Makefile`, like including the git commit in for `skopeo --version`, automatically using the right c/storage flags, correctly linking with GPGME. (If users know about these downsides and want to use `go {get,install}` directly, fine, but to me those should be intentional choices by people who have read the `Makefile` and perhaps monitor it for changes, not the _default_ approach.)

And the following instructions for building documentation / system-wide installation don’t work without a user-visible checkout.

So I’m leaning towards not merging this. What do others think?

stevegt:
What jumped out at me, caused me to use `go get` instead, and caused me to write this PR, was the reference to GOPATH -- it's not likely that newer Go users will even have it set, particularly after 1.17.

I'd be fine with not merging this PR, but then someone ought to tackle getting rid of that apparent dependence on GOPATH and making sure the Makefile still works regardless of how GOPATH is set or if it's not set at all.

rhatdan:
I think making the Makefile without GOPATH makes total sense.  If not set, then just set it to $HOME/go, and you get 90% there.

rhatdan:
BTW, Does it work without GOPATH?

unset=GOPATH make
CGO_CFLAGS="" CGO_LDFLAGS="-L/usr/lib64 -lgpgme" GO111MODULE=on go build -mod=vendor "-buildmode=pie" -ldflags '-X main.gitCommit=5af5f8a0e7520a93c736a03bf58c26ea006f9ec6 ' -gcflags "" -tags "  " -o bin/skopeo ./cmd/skopeo

Would 
```
git diff
diff --git a/Makefile b/Makefile
index 6bacd12e..94998e2d 100644
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,7 @@ GO ?= go
 GOBIN := $(shell $(GO) env GOBIN)
 GOOS ?= $(shell go env GOOS)
 GOARCH ?= $(shell go env GOARCH)
+GOPATH ?= $(shell go env GOPATH)
 
 ifeq ($(GOBIN),)
 GOBIN := $(GOPATH)/bin
```
Be enough

mtrmac:
Right now checking out into $GOPATH works every time (and, to an extent, steers users who don’t know what $GOPATH is towards distribution packages or containers, though that was not an intentional plot).

It would absolutely make sense to simplify to just say “check it out somewhere and run `make`” _after_ we drop support for those old Go versions (it seems we are testing on Go 1.16, and at least
```
vendor/gopkg.in/yaml.v2/go.mod:go 1.15
```
). And the `Makefile` is already using `-mod=vendor` ~everywhere, commit 8eba94f271ba180a86886dea6e4f8f392f862051 suggests we have needed this since since 1.13 already.

I haven’t now looked how old versions of Go actually allow building Skopeo, and where is the relevant point from which on `GOPATH` is no longer relevant. We might easily be well past that point already, but it does require research and/or testing.

mtrmac:
> I think making the Makefile without GOPATH makes total sense. If not set, then just set it to $HOME/go, and you get 90% there.

It can be rather surprising, to the point of suggesting an attack, if `make` in a random checkout starts creating unannounced directories in the home directory.

rhatdan:
I think the upstream Makefile can safely assume golang >= 1.15 and be done with this.  If people want to install on older versions then they need to setup their environment correctly.

lsm5:
@mtrmac @rhatdan can we close this?

mtrmac:
Thanks for the reminder. I do think that we don‘t want to recommend the `go get` installation mechanism.

WRT the outside-of-GOPATH setups, that’s an interesting discussion but let’s do that in a separate issue/PR, if there is anyone who has the time and interest to pursue it (which won’t be me any time soon, I’m afraid).

Issue #1255, opened 2021-04-21
v1.2.3 docker image?


jess-belliveau:
Hello, I'm wondering when the docker image for the tagged release 1.2.3 will be available?
https://quay.io/repository/skopeo/stable?tab=tags

It seems latest was built/pushed recently but still reports 1.2.2
```
➜ docker run quay.io/skopeo/stable:latest -v
skopeo version 1.2.2
```

vrothberg:
@TomSweeneyRedHat PTAL

samdoran:
It would also be nice to have tags for v1.2.1 and v1.2.2. The latest tag is currently v1.2.0 and I would like to use a newer version of `skopeo` without having to use `latest`.

samdoran:
It actually looks like the v1.2.0 image actually has `skopeo` 1.2.2, though the images do have different hashes.

```
> docker image inspect quay.io/containers/skopeo:v1.2.0 -f '{{join .RepoDigests ", "}}'
quay.io/containers/skopeo@sha256:f13b21b67066d451881f65675ca8b0b55dc69cb1637a0e8a03a4fa5106d7781e

> docker image inspect quay.io/containers/skopeo:latest -f '{{join .RepoDigests ", "}}'
quay.io/containers/skopeo@sha256:0ec6e3bc99478f74c5f2f6b094d65f42580d52276246f91f9bf2ef1df335491f

> docker run --rm quay.io/containers/skopeo:v1.2.0 -v
skopeo version 1.2.2

> docker run --rm quay.io/containers/skopeo:latest -v
skopeo version 1.2.2
```

lsm5:
@TomSweeneyRedHat where are the build scripts or whatever for the image located? Wondering if this can be part of release tagging / CI job?

TomSweeneyRedHat:
Sorry for  the delay on this, I've been looking for a good hour to sit down and work through this, but haven't yet.  Still on the Radar. 
The problem  in the past has been that the Skopeo kits hadn't been promoted to Stable on bodhi, I'm not sure if that's still the case here.  

The images these days are created via the CI and the Dockerfiles that create them can be found [here](https://github.com/containers/skopeo/tree/master/contrib/skopeoimage)

lsm5:
> The problem in the past has been that the Skopeo kits hadn't been promoted to Stable on bodhi, I'm not sure if that's still the case here.

[...]

PR #1288, opened 2021-05-17
Multi-arch github-action workflow unification

cevich:
This is a port from the podman and buildah repository workflows.
It's purpose is to build and push multi-arch images containing the
latest upstream, testing, and stable versions of skopeo.  It fully
replaces the last remaining use of Travis in this repo, for
substantially the same purpose.

In a future commit, I intend to de-duplicate this workflow from
podman and buildah, such that all three share a common set of details.
Until then, any changes will need to be manually duplicated across
all three repos.

Fixes: #1255


cevich:
***Note:*** I have a working prototype of a more unified workflow (re: commit message).  However, development on it is slow and non-optimized.  Rather than force the community to live with broken Travis automation, I thought we'd just duplicate this workflow yet again, but hopefully temporarily.

rhatdan:
LGTM

rhatdan:
@vrothberg PTAL
@cevich Are we building skopeo for the mac and windows?

.github/workflow/multi-arch-build.yaml
```diff
@@ -0,0 +1,208 @@
...
+            # workaround vim syntax-highlight bug: '
+            ALLTAGS=$(skopeo list-tags \
+                      docker://$CONTAINERS_QUAY_REGISTRY/skopeo | \
+                      jq -r '.Tags[]')
+
+            # New image? Push quay.io/containers/skopeo:vX.X.X and :latest
+            if ! fgrep -qx "$VERSION" <<<"$ALLTAGS"; then
+              FQIN="$CONTAINERS_QUAY_REGISTRY/skopeo:$VERSION,$CONTAINERS_QUAY_REGISTRY/skopeo:latest"
+            else # Not a new version-tagged image, but contents may be updated
+              FQIN="$CONTAINERS_QUAY_REGISTRY/skopeo:latest"
```
TomSweeneyRedHat, line 129: I'm confused by this, for the upstream and testing versions, we should only be pushing a `:latest` tag.  We don't have to worry about adding a `:v1.2.0` or whatever version tag.

.github/workflow/multi-arch-build.yaml
```diff
@@ -0,0 +1,208 @@
...
+                      jq -r '.Tags[]')
+
+            # New image? Push quay.io/skopeo/stable:vX.X.X and :latest
+            if ! fgrep -qx "$VERSION" <<<"$ALLTAGS"; then
+              # Assume version-tag is also the most up to date (i.e. "latest")
+              FQIN="$SKOPEO_QUAY_REGISTRY/stable:$VERSION,$SKOPEO_QUAY_REGISTRY/stable:latest"
+            else # Not a new version-tagged image
+              # Assume other contents changed, so this is the "new" latest.
+              FQIN="$SKOPEO_QUAY_REGISTRY/stable:latest"
+            fi
```
TomSweeneyRedHat, line 96: When  a new version of Skopeo is created on GitHub, I think this will create a skopeo/stable:{newversion}`, but then the `:latest` tagged image will still have the old version until the next run of this script?  If that's right, then we might want to try to figure out later how to update latest at the same time.

.github/workflow/multi-arch-build.yaml
```diff
@@ -0,0 +1,208 @@
...
+            echo "::error::Unknown matrix item '${{ matrix.source }}'"
+            exit 1
+          fi
+          echo "::warning::Pushing $FQIN"
+          echo "::set-output name=fqin::${FQIN}"
+          echo '::set-output name=push::true'
+
+      # This is substantially the same as the above step, except the
+      # $CONTAINERS_QUAY_REGISTRY is used and the "testing"
+      # flavor is never pushed.
```
TomSweeneyRedHat, line 113: Why isn't the testing flavor pushed?  A `testing:latest` should be pushed everytime this script runs me thinks.

.github/workflow/multi-arch-build.yaml
```diff
@@ -0,0 +1,208 @@
...
+          org.opencontainers.image.source=https://github.com/${{ github.repository }}.git
+          org.opencontainers.image.revision=${{ github.sha }}
+          org.opencontainers.image.created=$(date -u --iso-8601=seconds)
+          DELIMITER
+          EOF
+
+      # Separate steps to login and push for skopeo and containers quay
+      # repositories are required, because 2 sets of credentials are used and `docker
+      # login` as well as `skopeo login` do not support having 2 different
+      # credential sets for 1 registry.
```
TomSweeneyRedHat, line 158: Really?  When I login to my account on quay.io using skopeo_login, I can push to either the skopeo or containres repos.  For that matter, a couple of others too.

contrib/skopeoimage/README.md
```diff
@@ -6,22 +6,35 @@
...
+The container images are:
+
+  * `quay.io/containers/skopeo:<version>` and `quay.io/skopeo/stable:<version>` -
+    These images are built when a new Skopeo version becomes available in
+    Fedora.  These images are intended to be unchanging and stable, they will
+    never be updated by automation once they've been pushed.  For build details,
+    please [see the configuration file](stable/Dockerfile).
+  * `quay.io/containers/skopeo:latest` and `quay.io/skopeo/stable:latest` -
+    Built daily using the same Dockerfile as above.  The skopeo version
+    will remain the "latest" available in Fedora, however the other image
```
TomSweeneyRedHat, line 28:
```suggestion
    will remain the "latest" available in Fedora, however the image
```

.github/workflow/multi-arch-build.yaml
vrothberg: In that case you're using the same credentials for both pushes but it is correct that auth files allow for storing only 1 credential set for a given registry.

vrothberg (commented):
LGTM from my side.  Good to merge once Tom's comments are addressed.

.github/workflow/multi-arch-build.yaml
cevich: This is not the case because `if ! fgrep -qx "$VERSION" <<<"$ALLTAGS"` (new version detected) then we push both tags with `FQIN="$SKOPEO_QUAY_REGISTRY/stable:$VERSION,$SKOPEO_QUAY_REGISTRY/stable:latest"`.  If there isn't a new version, we only push `:latest`.

.github/workflow/multi-arch-build.yaml
cevich: One is for `quay.io/skopeo`, just not `quay.io/containers` because "this is how it was before" :grin: I can change this behavior but...remember this config. is a copy-paste of what we already do in podman and buildah.  Should I change those also?

cevich: Clarification: `quay.io/containers/skopeo` has tags for `latest` and `$VERSION` only.  This is different from `quay.io/skopeo` where we have an **image** called `testing`. So if we want to push testing to containers, I would need to add something like a `:testing` tag.

.github/workflow/multi-arch-build.yaml
cevich: This entire block excludes `testing` with `if: matrix.source != 'testing'`...But, I think your confusion is justified as it seems this logic will overwrite `latest` from "stable" and again with "upstream".  I'll need to fix this here and in buildah/podman.

.github/workflow/multi-arch-build.yaml
cevich: Yep, and this is enforced even more for "robot" accounts, as they're forced to a single repository namespace.

contrib/skopeoimage/README.md
cevich: Drat, I'll need to update this in podman and buildah as well.

.github/workflow/multi-arch-build.yaml
cevich: Follow-up fixes:
* https://github.com/containers/podman/pull/10414
* https://github.com/containers/buildah/pull/3247

cevich:
Fixed-up and rebased.

.github/workflow/multi-arch-build.yaml
TomSweeneyRedHat: Thanks, misread the code here then.

.github/workflow/multi-arch-build.yaml
TomSweeneyRedHat: sorry for the confusion.  I do NOT want a `:testing` label for any of our images.   I wasn't seeing that the `quay.io/skopeo/testing:latest` image was getting pushed as part of this processing.  Did I miss that?

.github/workflow/multi-arch-build.yaml
```diff
@@ -0,0 +1,212 @@
...
+            if ! fgrep -qx "$VERSION" <<<"$ALLTAGS"; then
+              # Assume version-tag is also the most up to date (i.e. "latest")
+              FQIN="$SKOPEO_QUAY_REGISTRY/stable:$VERSION,$SKOPEO_QUAY_REGISTRY/stable:latest"
+            else # Not a new version-tagged image
+              # Assume other contents changed, so this is the "new" latest.
+              FQIN="$SKOPEO_QUAY_REGISTRY/stable:latest"
+            fi
+          elif [[ "${{ matrix.source }}" == 'testing' ]]; then
+            # Assume some contents changed, always push latest testing.
+            FQIN="$SKOPEO_QUAY_REGISTRY/testing:latest"
```
TomSweeneyRedHat, line 99: and here's the testing processing that I was looking for last night and could note see.  DOH!

.github/workflow/multi-arch-build.yaml
TomSweeneyRedHat: found it, this is good, ty!

.github/workflow/multi-arch-build.yaml
```diff
@@ -0,0 +1,212 @@
...
+          else
+            echo "::error::Unknown matrix item '${{ matrix.source }}'"
+            exit 1
+          fi
+          echo "::warning::Pushing $FQIN"
+          echo "::set-output name=fqin::${FQIN}"
+          echo '::set-output name=push::true'
+
+      # This is substantially the same as the above step, except the
+      # $CONTAINERS_QUAY_REGISTRY is used and the "testing"
```
TomSweeneyRedHat, line 112: upstream too no?  At least to `quay.io/containers/upstream`, only to `quay.io/skopeo/upstream`

TomSweeneyRedHat:
@cevich I've one more testing question, might be a late night brain fart.  If you think so, perhaps we can chat for a min or three tomorrow?

.github/workflow/multi-arch-build.yaml
cevich: Yes, exclusively "upstream" and "stable".  This is selected by `if: matrix.source != 'testing'`

cevich: Okay, this should be fixed now.  Thanks Tom.

cevich:
Fixed as discussed, rebased, and force-pushed.

TomSweeneyRedHat:
And of course the ci barfed without leaving a trace behind.  I've hit restart.

TomSweeneyRedHat:
LGTM
assuming happy tests.  However, logic has been changed pretty substantially since the first pass, so another review from @edsantiago , @rhatdan or @vrothberg would be appreciated.

rhatdan:
LGTM

PR #1298, opened 2021-05-26
Respect $PREFIX for config files paths

mateuszkwiatkowski:
Handling $PREFIX for binaries and documentation was fixed in
https://github.com/containers/skopeo/pull/1168.
This commit puts configuration files to $PREFIX/etc/containers.

Fixes https://github.com/containers/skopeo/issues/787

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,10 +12,10 @@ GPGME_ENV := CGO_CFLAGS="$(shell gpgme-config --cflags 2>/dev/null)" CGO_LDFLAGS
 
 # Normally empty, DESTDIR can be used to relocate the entire install-tree
 DESTDIR ?=
-CONTAINERSCONFDIR ?= ${DESTDIR}/etc/containers
+PREFIX ?= ${DESTDIR}/usr/local
+CONTAINERSCONFDIR ?= ${PREFIX}/etc/containers
 REGISTRIESDDIR ?= ${CONTAINERSCONFDIR}/registries.d
 SIGSTOREDIR ?= ${DESTDIR}/var/lib/containers/sigstore
-PREFIX ?= ${DESTDIR}/usr/local
 BINDIR ?= ${PREFIX}/bin
 MANDIR ?= ${PREFIX}/share/man
 BASHCOMPLETIONSDIR ?= ${PREFIX}/share/bash-completion/completions
@@ -69,7 +69,13 @@ CONTAINER_RUN := $(CONTAINER_CMD) "$(IMAGE)"
 GIT_COMMIT := $(shell git rev-parse HEAD 2> /dev/null || true)
 
 EXTRA_LDFLAGS ?=
-SKOPEO_LDFLAGS := -ldflags '-X main.gitCommit=${GIT_COMMIT} $(EXTRA_LDFLAGS)'
+
+SKOPEO_LDFLAGS := -ldflags '-X main.gitCommit=${GIT_COMMIT} \
+		-X github.com/containers/image/v5/docker.systemRegistriesDirPath=${REGISTRIESDDIR} \
+		-X github.com/containers/image/v5/signature.systemDefaultPolicyPath=${CONTAINERSCONFDIR}/policy.json \
+		-X github.com/containers/image/v5/pkg/sysregistriesv2.systemRegistriesConfPath=${CONTAINERSCONFDIR}/registries.conf \
+		-X github.com/containers/image/v5/internal/tmpdir.unixTempDirForBigFiles=/var/tmp \
+		$(EXTRA_LDFLAGS)'
 
 MANPAGES_MD = $(wildcard docs/*.md)
 MANPAGES ?= $(MANPAGES_MD:%.md=%)


vrothberg (commented):
Thanks!

I am a bit worried about overriding the paths unconditionally.  If we run into the hypothetical situation of changing defaults in c/image, Skopeo would continue using the defaults from its Makefile.

@mtrmac WDYT?

mateuszkwiatkowski:
@vrothberg I agree this is not ideal solution and it already bitten homebrew maintainers as described in https://github.com/containers/skopeo/issues/787. The good thing is it's now consistent with other paths, the bad is it has to reach that deep to vendored package. I'm open to other solutions of this problem but unfortunately I'm not proficient in Go build system so help would be needed. :-)

mtrmac:
> I am a bit worried about overriding the paths unconditionally. If we run into the hypothetical situation of changing defaults in c/image, Skopeo would continue using the defaults from its Makefile.

Well, that’s not hypothetical at all:
> `const builtinRegistriesConfPath = "/etc/containers/registries.conf"`
vs. this PR which would use `/usr/local/etc/containers` by default.


OTOH if Skopeo is already installing default files in the paths decided in its own `Makefile`, it only makes sense for the Skopeo binary to default to paths that contains those files.

mtrmac:
> I am a bit worried about overriding the paths unconditionally. If we run into the hypothetical situation of changing defaults in c/image, Skopeo would continue using the defaults from its Makefile.

Well, that’s not really hypothetical at all:
> `const builtinRegistriesConfPath = "/etc/containers/registries.conf"`

vs. this PR which would use `/usr/local/etc/containers` by default.

There are three aspects to this:
- Is it a good idea to default to `/usr/local/etc/`? I don’t know; https://www.gnu.org/prep/standards/standards.html#Directory-Variables and https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s09.html#specificOptions14 suggests so, but then the logic should probably be more complex, both the FHS link and https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.70/autoconf.html#Site-Defaults say that `/usr/etc` is never used.
- To the extent Skopeo ships some of the default configuration, being inconsistent with c/image defaults breaks Podman/Buildah’s use; so we do need to be consistent.
- Even if we made that consistent by changing c/image to default to `/usr/local/etc`, can we just change the path used by default on a Skopeo version update, without fallback/compatibility? Here I’d say _definitely not_.

---

#787 is a very valid request; that does not, AFAICS, require changing any file locations. The `-X …` arguments can be added without moving to `/usr/local/etc`, so that’s what I think this PR should do, and we can discuss the move to `/usr/local/etc` (and compatibility impact), separately. (Users can override the `make` variables in any case.)

---

WRT unconditionally overriding — if Skopeo is already installing default files in the paths decided in its own `Makefile`, we need to keep Skopeo’s `Makefile` and c/image in sync already, both for Skopeo to work and for Buildah/Podman to work. That’s not a _new_ concern, in that sense.

It would still be very nice to add an automated tests for this, of some kind, maybe very vaguely (probably doesn’t work at all):
```make
# use this everywhere
POLICY_JSON_PATH := ${CONTAINERSCONFDIR}/policy.json

test-path-consistency:
    somehow_with_empty_environment_and_no_overrides $(MAKE) test-path-consistency-internal

# or maybe just a script in hack/* that gets passed POLICY_JSON_PATH and all the others?
test-path-consistency-internal: 
    # Actually parse as Go, or run some kind of a helper program??
    # This is not a correct Go parser, but works well enough to highlight the need to keep consistency.
    grep -q 'const builtinDefaultPolicyPath = "$(POLICY_JSON_PATH)"' vendor/github.com/containers/image/v5/signature/*.go
```

… if someone had time to make that actually work. (An architecture-astronaut approach would be to maintain the mapping of package+variable name vs. make variable name only once, and apply it both for the `-X` flags and this check, but that seems way more trouble than is warranted.)

(I don’t think this automated test needs to be be a blocker for this PR.)

rhatdan:
What do we want to do with this PR?

mtrmac:
- Definitely drop the parts that change the default path from `/etc/containers` to `/usr/local/etc/containers`
- Maybe the automated test as suggested above. I don’t think it’s a blocker but it would certainly give us peace of mind.

lsm5:
ping @mateuszkwiatkowski

rhatdan:
Most of the time when we install we set PREFIX=/usr, which would mean the config files would end up in /usr/etc/containers, which is clearly broken.

@mateuszkwiatkowski Unless we can work around that problem, I say we close this PR.

lsm5:
Closing until we hear further from @mateuszkwiatkowski or anyone willing to take this PR

mateuszkwiatkowski:
I'm sorry, I didn't look into project involved with skopeo since then. I understand your concerns and currently don't have resources to work on this further. I see that work continue at https://github.com/containers/skopeo/pull/1507.

PR #1303, opened 2021-06-01
Fix GitHub workflows + Add [CI:DOCS] mode

cevich:
Github checks the `workflows` sub-directory for github actions workflow
files.  Since it was called `workflow`, nothing was running.  Fix this
by renaming the directory.


edsantiago:
Added string matches the one in podman. New task map: 
cirrus-map-pr1303

LGTM

rhatdan:
LGTM

PR #1308, opened 2021-06-08
Fix multi-arch build version check

cevich:
This was copy-pasted from buildah and podman, unfortunately the
Dockerfile entrypoint is different for skopeo. Fix it.

Ref: https://github.com/containers/skopeo/runs/2771691404?check_suite_focus=true

### DIFF
--- a/.github/workflows/multi-arch-build.yaml
+++ b/.github/workflows/multi-arch-build.yaml
@@ -72,10 +72,9 @@ jobs:
         if: matrix.source == 'stable'
         id: sniff_test
         run: |
-          VERSION_OUTPUT="$(docker run localhost:5000/skopeo/${{ matrix.source }} \
-                            skopeo --storage-driver=vfs version)"
+          VERSION_OUTPUT="$(docker run localhost:5000/skopeo/${{ matrix.source }} --version)"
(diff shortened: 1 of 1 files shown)


cevich:
@TomSweeneyRedHat another option here (possibly more desirable), is to remove the `ENTRYPOINT` here (so it matches `buildah` and `podman`).

rhatdan:
Don't remove the entrypoint, at this point users are probably relying on it.

We went back and forth on this when we were first creating these images.

cevich:
okay.

cevich:
This is ready to go then (all builds are currently failing).

cevich:
Woopsie-doodle...okay, *now* it's ready to go :grin:

cevich:
Pressed re-run on failed tests (they looked like networking flakes)

cevich:
(rebased and force-pushed)

TomSweeneyRedHat:
LGTM
and happy green test buttons.

rhatdan:
LGTM

cevich:
[Noted automatic run failed](https://github.com/containers/skopeo/runs/2837054486) due to a permissions problem on quay.io.  I fixed that and re-ran the job.

PR #1317, opened 2021-06-16
Cirrus: New VM Images w/ podman 3.2.1

cevich:
Ref: https://github.com/containers/automation_images/pull/75


mtrmac (commented):
I’m new to this; the linked PR is still open, is that OK?

Feel free to merge if so.

cevich:
> I’m new to this; the linked PR is still open, is that OK?

Yep.  The `automation_images` PR workflow is what produces the images.  So merging that doesn't actually do much, except maybe influence future PRs (and image builds).  I normally wait for the images to pass muster in the target-repos. before closing the `automation_images` PR (in case there's a problem found).

In this case, there [actually is a problem](https://github.com/containers/common/issues/631), though it (clearly) doesn't affect Skopeo testing (likely because runc isn't used).

PR #1319, opened 2021-06-17
Fix documentation of the --format option of skopeo copy and skopeo sync

### LINKED ISSUE #1265: skopeo-copy docs incorrectly state that `--format` only applies to `dir:`
https://github.com/containers/skopeo/blob/master/docs/skopeo-copy.1.md currently says:

> --format, -f manifest-type Manifest type (oci, v2s1, or v2s2) to use when saving image to directory using the 'dir:' transport (default is manifest type of source)

However, this also works with e.g. `docker:` targets:

```bash
skopeo copy --format=v2s2 oci-archive:<ref> docker://<ref>
```

The above results in a Docker-format image being pushed, vs. an OCI image without the `--format` option.

mtrmac:
It affects all transports; and without `--format`, we try several manifest formats.

Fixes #1265 .


rhatdan:
LGTM
@vrothberg @TomSweeneyRedHat PTAL

rhatdan:
@lsm5 PTAL

vrothberg (approved):
LGTM

PR #1328, opened 2021-06-23
Fix automation re: master->main rename

### LINKED ISSUE #1282: Please rename the master branch to the main branch.

cevich:
Ref: https://github.com/containers/image/pull/1272 and https://github.com/containers/skopeo/issues/1282

Fixes #1282

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,6 +21,7 @@ RUN dnf -y update && dnf install -y make git golang golang-github-cpuguy83-md2ma
 # both. This allows integration-cli tests to cover push/pull with both schema1
 # and schema2 manifests.
 RUN set -x \
+	&& export GO111MODULE=off \
 	&& REGISTRY_COMMIT_SCHEMA1=ec87e9b6971d831f0eff752ddb54fb64693e51cd \
 	&& REGISTRY_COMMIT=47a064d4195a9b56133891bbb13620c3ac83a827 \
 	&& export GOPATH="$(mktemp -d)" \
@@ -34,6 +35,7 @@ RUN set -x \
 	&& rm -rf "$GOPATH"
 
 RUN set -x \
+	&& export GO111MODULE=off \
 	&& export GOPATH=$(mktemp -d) \
 	&& git clone --depth 1 -b v1.5.0-alpha.3 git://github.com/openshift/origin "$GOPATH/src/github.com/openshift/origin" \
 	# The sed edits out a "go < 1.5" check which works incorrectly with go ≥ 1.10. \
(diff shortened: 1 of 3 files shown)


cevich:
Oof, [the Dockerfile build](https://quay.io/repository/skopeo/ci/build/7d2d1016-77fd-438e-948d-f8e11281c5d9) is currently broken as well, gunna needa fix that too...

cevich:
...okay, local testing shows the Dockerfile now builds successfully.  I added a notification to the auto-builder (on quay.io) for this image: It should now post future build-failure notices to our podman-monitor list.

mtrmac:
> Oof, the Dockerfile build is currently broken as well, gunna needa fix that too...

To be explicit, those are the `GO111MODULE` changes, which are otherwise unrelated?

---

The `cross` task is failing for what seems to be a Brew server failure, I have triggered a re-run.

Dockerfile
```diff
@@ -34,6 +35,7 @@ RUN set -x \
 	&& rm -rf "$GOPATH"
 
 RUN set -x \
+    && export GO111MODULE=off \
```
mtrmac, line 38: A nit: Could you indent the new lines the same way as the rest, please?

cevich: Oh woops, my mistake, yep I got it...

cevich:
@mtrmac PTAL, I fixed the whitespace and bumped up the timeout a bit to see if that helps the cross task.

mtrmac (approved):
Thanks! LGTM.

(I’m not sure the timeout is desirable, this just seems to be a Brew infrastructure issue where we instead of pre-built binaries burn CPU time re-building locally. But even if it gets CI unblocked for a week and then we can revert, that’s valuable for that week.)

mtrmac:
@cevich Looking at https://github.com/Homebrew/brew/issues/11119 or https://github.com/Homebrew/brew/pull/11366 , the `brew` failure seems not to be a temporary situation (although https://status.bintray.com does currently report a scheduled maintenance).

Is the `brew` version in the macOS environment new enough, or could it be updated? If not, a `brew update` before package installation might well pay off in time.

cevich:
> But even if it gets CI unblocked for a week and then we can revert

I'm good either way.  The same key-name can be added to individual tasks for more specific/specialized timeout values, I only stuck in a global one out of sheer laziness.

TomSweeneyRedHat:
LGTM

mtrmac:
I have filed a completely uninformed #1329 with changes from this PR + `brew update`, just to see what happens.

cevich:
> To be explicit, those are the GO111MODULE changes, which are otherwise unrelated?

Yes, mostly unrelated.  Though auto-builder on quay was specifically targeting the master branch.  When I fixed that, I noticed the image build has been failing for months, so, yeah, :disappointed: Anyway, I also added some monitoring, so we'll now get e-mails to the podman-monitor list any time it breaks in the future.

rhatdan:
LGTM

PR #1329, opened 2021-06-23
Start macOS tests with (brew update)

mtrmac:


.cirrus.yml
```diff
@@ -70,6 +70,7 @@ cross_task:
         image: catalina-xcode
     setup_script: |
         export PATH=$GOPATH/bin:$PATH
+        brew update
```
cevich, line 73: Oh, okay, this is new to me.  I'm fine if this works, we can take this over my PR, no problem.

cevich (approved):

.cirrus.yml
mtrmac: (If this indeed is the cause of the problem,) ideally the OS image should be refreshed to contain a new version, so that we don’t have to do this at runtime. But I can only find https://cirrus-ci.org/guide/macOS/#list-of-available-images , nothing about the time when they were built or if they are rebuilt regularly.

mtrmac:
So the “setup” step of the “cross” task now runs in 9:21, which is an improvement — but the Brew self-update (+ update of its environment, like Ruby) is still quite a few minutes which would be nice to avoid.

cevich:
> still quite a few minutes which would be nice to avoid.

Ya, we've got very little control over that environment :disappointed: It's provided at no cost to us by Cirrus-CI, for which I'm very grateful.

Even getting past the $$$ hurdles of using our own MacStadium cloud setup, I'm afraid we'd just be thrown into months of debugging "Because Apple" problems.  Shoot, I even have a hard-time justifying time to add arm64 support via AWS, and that wouldn't involve twerking a weird bastardized-BSD environment...all standard Linux stuff.

mtrmac:
Rebased; see https://github.com/containers/skopeo/pull/1328#issuecomment-867039337 for the rationale.

rhatdan:
LGTM

PR #1330, opened 2021-06-23
[CI:DOCS] Workaround quay.io image build failure

cevich:
Ref: [quay.io build failure](https://quay.io/repository/skopeo/ci/build/3992488d-ab5b-430a-856e-10a382c4b7cf)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM fedora
+FROM registry.fedoraproject.org/fedora:latest
 
 RUN dnf -y update && dnf install -y make git golang golang-github-cpuguy83-md2man \
 	# storage deps


Dockerfile
```diff
@@ -1,4 +1,8 @@
-FROM fedora
+# Normally it's illadvised to use a FQDN, but this image is auto-built
```
mtrmac, line 1: Personally curious: Is it? We typically warn against short names (acknowledging that they are convenient.)

Dockerfile
```diff
@@ -1,4 +1,8 @@
...
+# on quay.io which has no setting to specify a non-docker default registry
+# or build-args.
+ARG REGISTRY=registry.fedoraproject.org/
+FROM ${REGISTRY}fedora:latest
```
mtrmac, line 5: Absolutely non-blocking: Do we expect anyone to override the variable? Just hard-coding `registry.fedoraproject.org/fedora:latest` might be good enough.

rhatdan:
LGTM

rhatdan:
@mtrmac @vrothberg PTAL

mtrmac (commented):
I can’t see the failure log, but this looks clearly fine.

Please merge as necessary.

Dockerfile
cevich: It's what I've been told/read as a best-practice.  It's more portable to use the short-name in the `Dockerfile` and rely on the build-host to configure the appropriate registry/registries within it's context.  However in practice, esp. with these CI-related Dockerfiles and repository auto-builders (like quay and docker hub), it's often easier to just jam the FQIN in and forget about it.

Dockerfile
cevich: TBH I didn't/don't know who/what uses this Dockerfile, so I made the more conservative choice.  I have no problem with "good enough" if that's what it is :grin:

rhatdan:
I believe make uses this Dockerfile to build an image to run tests with.

cevich:
> I believe make uses this Dockerfile to build an image to run tests with.

Correct.  Though rather than complicate things further, I think we can get away with hard-coding a non-docker-hub FQIN for now.  If an issue is raised, I'm happy to go the build-arg route.  Otherwise, I think @mtrmac is right, the vast majority of users and robots will never care.

PR #1252, opened 2021-04-19
Travis -> Cirrus: MacOS Cross test

cevich:
Fixes #1213

***Depends on:*** #1249

Also cleanup `Makefile` WRT `${DESTDIR}` definition and use to make it's
purpose more clear.

PR #1331, opened 2021-06-23
Reintroduce the GNU semantics of DESTDIR

mtrmac:
This partially reverts commit a81cd747341e04783e7e3a94a76c4444c3bd5cd8 so that most path variables like `PREFIX` and `BINDIR` refer to paths on the installed system, and `DESTDIR` is used only in `make install`, following the philosophy of the GNU coding standards for path variables. (But not precisely the variable names, which are lowercase in the standard, nor the principle that even `SYSCONFDIR` should be under `$PREFIX`.)

Keep the use of `?=` instead of `=` because it somewhat better expresses the idea that the values can be overridden.

Use `${DESTDIR}${BINDIR}` instead of `${DESTDIR}/${BINDIR}` etc, so that a plain `make install` does not use paths like `//usr/bin/...` ; strictly speaking they are IIRC reserved by POSIX, and more importantly it just looks untidy :)

Cc: @cevich This originally appeared in #1252, and I might be missing something that expects the other semantics.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,12 +10,12 @@ export GOPROXY=https://proxy.golang.org
 # (and the user will probably find out because the cgo compilation will fail).
 GPGME_ENV := CGO_CFLAGS="$(shell gpgme-config --cflags 2>/dev/null)" CGO_LDFLAGS="$(shell gpgme-config --libs 2>/dev/null)"
 
-# Normally empty, DESTDIR can be used to relocate the entire install-tree
+# The following variables very roughly follow https://www.gnu.org/prep/standards/standards.html#Makefile-Conventions .
 DESTDIR ?=
-CONTAINERSCONFDIR ?= ${DESTDIR}/etc/containers
+PREFIX ?= /usr/local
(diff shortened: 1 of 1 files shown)


Foxboron:
Thank you for doing the revert! I was sadly a bit low on my todo list :disappointed:

rhatdan:
LGTM
@vrothberg PTAL

vrothberg (commented):
LGTM but I want a final head nod from @cevich

TomSweeneyRedHat:
LGTM, but a test isn't happy and I too would like a head nod from @ceveich

cevich:
> I want a final head nod from @cevich

Well now...looks like we got a head-nod from Cirrus, so it must be okay :grin:

cevich:
Just to be transparent: I've designed nearly all our automation (everywhere) to ignore details inside Makefiles.  We strive to use unadulterated `make` commands when convenient. Otherwise the scripts simply call the needed shell commands directly (bypassing make).  That's why "tests passing" is good enough for me.

PR #1334, opened 2021-06-24
Cirrus: Run checks directly on the host

cevich:
In order to meet achievable deadlines converting from Travis to Cirrus
CI, one significant artifact was carried forward (instead of fixing):

Depending on a `--privileged` container to execute all/most automated
checks/tests.

Prior attempts to remove this aspect resulted in several test failures.
Fixing the problems was viewed as more time-consuming than simply
preserving this runtime environment.

Time has passed, and the code has since moved on.  This commit removes
the legacy need to execute CI operations in a `--privileged`
container, instead running them directly on the host.  At the same time,
the necessary test binaries are obtained from the same container used
for development/local testing purposes.  This ensures the two experiences
are virtually always identical.


cevich:
ahh right [most of the tests are skipped when not running in a container](https://cirrus-ci.com/task/6314613785493504), there was something I needed to tweak to get around that...

cevich:
Wait...what?!?!?! It passed??????

@mtrmac PTAL (it's still a prototype).

cevich:
oh! wait, I made a typo, and we probably want developers to be able to build a container for testing in...

Dockerfile
```diff
@@ -1,56 +1,26 @@
...
-	&& (cd "$GOPATH/src/github.com/docker/distribution" && git checkout -q "$REGISTRY_COMMIT") \
-	&& GOPATH="$GOPATH/src/github.com/docker/distribution/Godeps/_workspace:$GOPATH" \
-		go build -o /usr/local/bin/registry-v2 github.com/docker/distribution/cmd/registry \
-	&& (cd "$GOPATH/src/github.com/docker/distribution" && git checkout -q "$REGISTRY_COMMIT_SCHEMA1") \
-	&& GOPATH="$GOPATH/src/github.com/docker/distribution/Godeps/_workspace:$GOPATH" \
-		go build -o /usr/local/bin/registry-v2-schema1 github.com/docker/distribution/cmd/registry \
-	&& rm -rf "$GOPATH"
+ENV GOPATH="/var/tmp/go"
+ENV GOSRC="$GOPATH/src/github.com/containers/skopeo"
+ENV PATH="$GOPATH/bin:$GOSRC/bin:/usr/share/gocode/bin:/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin:/bin:/sbin"
```
mtrmac, line 5: (Is there a benefit to hard-coding the various default instead of just prepending to distribution-specified $PATH ?)

Dockerfile
```diff
@@ -1,56 +1,26 @@
...
-RUN dnf -y update && dnf install -y make git golang golang-github-cpuguy83-md2man \
-	# storage deps
-	btrfs-progs-devel \
-	device-mapper-devel \
-	# gpgme bindings deps
-	libassuan-devel gpgme-devel \
-	gnupg \
-	# htpasswd for system tests
-	httpd-tools \
-	# OpenShift deps
```
mtrmac, line 12: Any chance to preserve these comments?

Dockerfile
```diff
@@ -1,56 +1,26 @@
...
+		bsdtar socat ethtool device-mapper iptables \
+		tree findutils nmap-ncat e2fsprogs xfsprogs lsof \
+		docker iproute bats jq podman runc golint \
+		openssl \
+	&& dnf -y upgrade \
+	&& dnf erase -y skopeo \
+	&& dnf clean all
 
-#ENTRYPOINT ["hack/dind"]
+COPY . $GOSRC
```
mtrmac, line 21: Having this `COPY` last allows more sharing on interactive debugging of the tests. Maybe start with a `COPY hack`, or even more specific, and end with a `COPY .`?

Dockerfile
```diff
@@ -1,56 +1,26 @@
...
-RUN set -x \
-	&& export GO111MODULE=off \
-	&& export GOPATH=$(mktemp -d) \
-	&& git clone --depth 1 -b v1.5.0-alpha.3 git://github.com/openshift/origin "$GOPATH/src/github.com/openshift/origin" \
-	# The sed edits out a "go < 1.5" check which works incorrectly with go ≥ 1.10. \
-	&& sed -i -e 's/\[\[ "\${go_version\[2]}" < "go1.5" ]]/false/' "$GOPATH/src/github.com/openshift/origin/hack/common.sh" \
-	&& (cd "$GOPATH/src/github.com/openshift/origin" && make clean build && make all WHAT=cmd/dockerregistry) \
-	&& cp -a "$GOPATH/src/github.com/openshift/origin/_output/local/bin/linux"/*/* /usr/local/bin \
-	&& cp "$GOPATH/src/github.com/openshift/origin/images/dockerregistry/config.yml" /atomic-registry-config.yml \
-	&& rm -rf "$GOPATH" \
```
mtrmac, line 46: The previous versions removed the source checkouts (and unnecessary build products) from the container; IIRC that fairly significantly reduced the size.

hack/make.sh
```diff
@@ -23,22 +23,9 @@ export SKOPEO_PKG='github.com/containers/skopeo'
 export SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 export MAKEDIR="$SCRIPTDIR/make"
 
-# We're a nice, sexy, little shell script, and people might try to run us;
-# but really, they shouldn't. We want to be in a container!
```
mtrmac, line 27: This is still somewhat a concern; we don’t want someone inattentive to run `make test-integration-local` on the main development machine.

mtrmac (commented):
The integration tests fail on OpenShift startup… Assuming that’s not transient (I have now triggered a re-run), debugging that doesn’t look like much fun.

Another sort-of-concern is that while running a custom container inside a custom VM is rather wasteful, and running CI directly on a VM is certainly nice, for local development it is convenient that `make test-integration` (and `make test-system`?) can be run in a container, so it is useful for CI to enforce that those two continue to work inside a container.

So, would running vendor+build+validate+unit directly on the VM, and integration+system in a container be (even possible and) worth the effort?

cevich:
> for local development it is convenient

Yes, we need to do both, the challenge is keeping the environments and experience as similar as possible.

> Would running vendor+build+validate+unit directly on the VM, and integration+system in a container be (even possible and) worth the effort?

There's no reason we can't/shouldn't do both in parallel (on separate VMs).

Dockerfile
cevich: My understanding is that you can't reference the container's `$PATH`  from inside an `ENV` instruction.  It will simply be blank (I need to confirm this).  Normally `$PATH` is setup by pam, and/or `/etc/profile*` + the `$HOME` config/script files.  So assuming there's even a shell executable inside the container, it would have to be executed to first obtain the value, then somehow pass that back into the build context.  So I think it _has_ to be hard-coded if we want to change it in the Dockerfile.  Checking...

cevich: ...okay, I stand corrected.  It looks like `podman` build either hard-codes in a default value, or is somehow obtaining it from inside the container.

mtrmac: OK, if the previous version didn’t work, that’s a pretty strong reason.

Dockerfile
cevich: Actually, I'm not a big fan of maintaining a package list inside the Dockerfile (it often becomes unwieldy).  Let me investigate an alternative that will also allow keeping these comments.

Dockerfile
cevich: Yes, I noticed that compile/setup time, I don't like it either.  I think probably the best way to handle this for both developers and CI, is to separately build a base-image.  Then use that here to add in only the finishing touches, like overwriting $GOSRC in the container.  That way we gain all the benefits of environment-consistency, it will be reasonably up-to-date (i.e. built daily), and fast for CI + Developer use.

The disadvantages are: A daily out-of-date base image, means both CI and Developers will always be pulling it down - annoying, but quite likely faster than building + compiling each time.  The other disadvantage is it will absolutely be a 'rolling' environment tightly-coupled to work on the *main* branch only.  It will be useless/dangerous trying to use it on any release branches - but maybe this is a separate discussion.

cevich: In case it sparks an idea:  What we do in podman and buildah is rely on container images built at the same time as VM images.  They 99.999% match the VM setup, and **NEVER** use a `latest` tag (i.e. only `c123456789` strings matching the VM images).  This means the CI and developer experience on both *main* and all release branches remains consistent - it only ever changes when someone deliberately changes the ID.

The down-side here is the container images are VERY large (almost 1gb) with almost zero room for reduction.  On the other hand, they change much more slowly/deliberately compared to a rolling `latest` image.  IMHO this makes development work easier, because you don't have *two* moving targets (the environment and the WIP code).

The other down-side is, it's desirable to keep `.cirrus.yml` as the single source-of-truth for the image ID (and tag).  That complicates operations outside of CI, since there's an extra step in obtaining the *current* ID (e.g. parsing YAML to JSON and filtering the ID with `jq`).  It can be done, it's just an extra step.

Dockerfile
cevich: Right.  This is an important attribute for container images, I'll update the script to do something similar.

hack/make.sh
cevich: This may be a problem we can simply ignore, developers/users must assume some amount of risk when running `make` and tinkering with development-project scripts.  It can be tricky (in scripts) to distinguish between "some random development environment" and a VM.  It's slightly easier in a container, but likely an unnecessary complexity.

Is this behavior you strongly want to preserve?   Assuming so, I could add a check similar to before (a canary env. var.) for containers, or alternately confirm `$CI` is set to `true` (guaranteed by Cirrus and most other CI systems).  Worth?

Dockerfile
cevich: Nope, it probably did work.  I was wrong, so I put it back to the old way.

cevich:
The openshift thingie seems to be failing to start due to SSl/TLS crap:

`transport: http2Client.notifyError got notified that the client transport was broken remote error: tls: bad certificate.`

But I also see a fair number of what looks like errors binding to port 8443 of the local IP...I wonder if maybe something in the GCP networking setup is (somehow) blocking this?   I would think the system/kernel smart enough to recognize that as the local address and not actually send anything out.

Dockerfile
mtrmac: Nice.

Dockerfile
mtrmac: > I think probably the best way to handle this for both developers and CI, is to separately build a base-image.

That’s very attractive for _most_ cases. IIRC the OpenShift compilation is by far the most resource-intensive part of Skopeo builds, and letting users just download a binary would be convenient.

Looking at the history of `Dockerfile`, it only changes a few times a year, so having a _named_ image pulled down instead of relying on local build cache feels like an improvement. I don’t have a numbers on the (world-wide) relative cost of pulling a gigabyte of data vs. a local OpenShift build, but it seems reasonable to guess that pulling the gigabyte image would not be _prohibitively_ worse for anyone.

_Occasionally_, some changes to the `Dockerfile` happen in bursts of development, and at that time it should be possible to work with a locally-built container; so there’s a downside in having to maintain, or at least enable, a rarely-used code path. (Also, there would need to be a way to trigger a fresh image/VM build after such a change happens.)

WRT using rolling `latest` vs. non-rolling explicitly tagged images/VMs, I don’t have a strong opinion. It sounds important to have some, preferably automated, or at least easy, way to run tests on old stable branches, but you have much better idea what the trade-offs/limitations in this area (I vaguely remember something about expiring images?) are.

hack/make.sh
mtrmac: I feel somewhat strongly about having some safeguard, primarily so that `make` or `make check` (something users are a bit likely to just type without reading much documentation) is not disruptive; alternatively we could make sure the Makefile and documentation is (over time maintained to be) consistent and safe, and it seems significantly easier to me to code a safeguard :)

I don’t really care very much at all _how_ that safeguard works exactly. (The whole `hack/make.sh` is inherited/stripped-down from a very different project, probably with way too much abstraction for how little it does; the in-container check is here because it was inherited in that file. Other than that check, maybe we don’t really need that file much.)

I don’t think it’s necessary to protect against directly running `hack/*`; such users should read the scripts first.

---

A manual `make check` (run on a non-CI Linux system) should be possible, so that integration/system tests can be developed locally, and should create a container; so a Cirrus-_only_ solution is not ideal. Checking for ($in_container or $CI) would work, I guess with a comment about where $CI comes from.

---

(Historical note: The old design that the disruptive checks require a caller to set SKOPEO_CONTAINER_TESTS, and that was typically set by `make check` for code running inside a container; Travis ran all tests inside a container. A CI could just run `make test-integration-local SKOPEO_CONTAINER_TESTS=1` on a test VM directly. The is a bit misleading, it didn’t anticipate dedicated VMs, it’s just an opt-in to disruptive behavior.)

---

There are now two mechanisms: `SKOPEO_CONTAINER_TESTS` and `container_magic`, AFAICS both roughly comparably safe; we probably need only one, I don’t care which one.

mtrmac:
> [The openshift thingie seems to be failing to start due to SSl/TLS crap:](https://cirrus-ci.com/task/5463315578093568?logs=integration#L259)
> 
> `transport: http2Client.notifyError got notified that the client transport was broken remote error: tls: bad certificate.`

Yeah. I don’t know what that’s about :/ It’s quite possible that a networking difference changes the detected primary IP address inconsistently in some parts of the code vs. others, or something like that. 

> But I also see a fair number of what looks like errors binding to port 8443 of the local IP...

Binding or connecting? Connecting retries are expected, that’s just https://github.com/containers/skopeo/blob/71e7a5839ef6bed3e280742d916e073b465f6d81/integration/utils.go#L117 waiting for the server to start up.

Binding to `0.0.0.0:8443` seems to succeed in those logs; also the “remote error” strings at least hint in the direction that a “remote” connection was actually established. Do you mean that it fails from time to time?

> I would think the system/kernel smart enough to recognize that as the local address and not actually send anything out.

Yes — OTOH OpenShift might be trying to be “useful” on the local network; we just run an `openshift start master` (in this old version that didn’t use dozens of separate containers with no configuration at all), and I don’t really know much at all what that does.

Dockerfile
cevich: > pulling the gigabyte image would not be prohibitively worse for anyone.

I know anybody with a poor internet link would absolutely hate it and curse the universe. In fact, random networking glitches have caused Cirrus-CI to timeout running jobs (maybe 10 times in the last few months) due to the image size.  I also don't know how in-sync quay.io's content-distribution proxies are (or if they even exist).  So it could be quite slow for anybody network-far from the source.  Give it a try: `podman pull quay.io/libpod/fedora_podman:c4687311653830656` (chosen at random).

> Also, there would need to be a way to trigger a fresh image/VM build after such a change happens

If I setup a dedicated base-image build for this on `c/automation_images`, needing to hand-build is actually a very easy and relatively un-complicated `make foobar` type command (on that repo).  It's the VM images which are complex.  This will keep the size down as well.

> idea what the trade-offs/limitations in this area

I have an idea of how best to keep automation stable/reliable.  It's the local-developer workflow that I'm much less savvy about.  The two often have conflicting priorities of "latest" vs "stable".  It's possible to solve for both, but adds risk of sometimes CI != Development (in very non-obvious ways).

> I vaguely remember something about expiring images

That's a big problem on the VM side, and a small problem for container images.  I'm not sure of the exact details, but I am sure that quay.io will delete tagged images when they're not pulled for some interval.  However, it sounds to me like I need to go ahead and setup cirrus-cron jobs on skopeo release branches anyway (to preserve VM images).  Doing that will also cause the container images to be pulled, and (in theory) block their removal.

cevich: To be clear: What I'm thinking of is setting up `c/automation_images` to build a base image which is consumed by the Dockerfile (here).  It will produce a `latest` image daily, and a tagged image every time updated VM images are called for.  I'll add in a `ARG tag=latest`, but have CI override that at runtime with `tag=${IMAGE_SUFFIX}`.

Secondarily, I'll setup cirrus-cron jobs to run on skopeo release branches.  Going-forward (as new release branches are made) this will block removal of both VM and the tagged container base image.  However the "old" `latest` image will obviously be lost so development work on release branches will need a way to override the `tag` arg via a `make` option.

Clear as mud? :grin:

hack/make.sh
cevich: Thanks for the explanation and for your opinion.  It sounds like we can use SKOPEO_CONTAINER_TESTS, default that to 0 (disabled) in the Makefile unless `CI` is also set.  Conveniently that also provides a handy place to document the potential dangers of `SKOPEO_CONTAINER_TESTS=1`.

As for `hack/make` and friends, my preference is to leave them mostly alone for now.  Even if inefficient/overly complex, they do function as good detail-encapsulation for both humans and automation.  So I think it's perfectly safe to leave and overhaul of them to a future PR, after all the "big-picture" stuff is functioning.

cevich:
> quite possible that a networking difference changes the detected primary IP address inconsistently

Oh, good thought.  I can check this manually by comparing the logs from a job run from master.

> Binding to 0.0.0.0:8443 seems to succeed in those logs;

Interesting, I didn't notice that.  

> OTOH OpenShift might be trying to be “useful” on the local network

Hmmm, so the VMs definitely have TWO IP addresses.  One is internal-network only, the other is external.  IIRC, the logs mentioned the 10.x.x.x (internal) network.  That _should_ work, but maybe not.  We're using a very vanilla/default GCP networking setup, it's possible the default is to block inter-VM communications.

In any case, I think this is going to require some hands-on work.

Dockerfile
mtrmac: > > pulling the gigabyte image would not be prohibitively worse for anyone.
> 
> I know anybody with a poor internet link would absolutely hate it and curse the universe.

Sure, but (assuming the manually-tagged rarely-changing images) the alternative is not magic that pulls down 0 bytes, but all the `git clone`s and RPM updates necessary to build the container image locally, with many more round-trips and other latency than a straight pull of a gigabyte of data.

mtrmac: > To be clear: What I'm thinking of is setting up `c/automation_images` to build a base image which is consumed by the Dockerfile (here). It will produce a `latest` image daily, and a tagged image every time updated VM images are called for. I'll add in a `ARG tag=latest`, but have CI override that at runtime with `tag=${IMAGE_SUFFIX}`.

Sounds reasonable. Maybe have the `Makefile` for local builds also use the same tag, so that local users (if any) don’t keep re-downloading every day?

[Anyway is all of this predicated on fixing the OpenShift failure we are seeing?]

cevich: > not magic that pulls down 0 bytes

Of course not.  The 1g image is generic, and matches the VM images.  I think I can do a custom/special-purpose base-image just for Skopeo that will be smaller, maybe half that size.  So twice as fast to download, at the cost of some extra tinkering in c/image (which I have no problem with doing).

cevich: > local builds also use the same tag

This would be ideal, as we could completely avoid the whole ever-changing 'latest' image.  The cost is in extra complexity needed to obtain the `$IMAGE_SUFFIX` value out of `.cirrus.yml`, but this is doable.  In fact, I have a small container-image which can do it in a generic way.

cevich:
> this old version that didn’t use dozens of separate containers

@mtrmac QQ: Is the purpose of this to verify Skopeo can talk to the OpenShift registry?  Is there a (testing) reason why skopeo needs to remain compatible with an unsupported OpenShift?

Reason I ask is...might it be easier to get a newer/supported registry to test against rather than fight/fix old one?

mtrmac:
> QQ: Is the purpose of this to verify Skopeo can talk to the OpenShift registry?

Yes,

> Is there a (testing) reason why skopeo needs to remain compatible with an unsupported OpenShift?

Somewhat.

We need the registry for testing two things:
- The X-Registry-Supports-Signatures API, a registry-native way of storing signatures, independent of OpenShift’s k8s API. That one is, AFAIK, supported by the OpenShift integrated registry in current versions as well, is _much_ easier to use than external sigstores, so it’s something we fairly strongly need to keep functional, i.e. keep testing. (It’s fairly reasonable to argue that testing against _old_ OpenShift will over time be less representative than testing against currently-supported versions, so if testing against the current full-cluster monsters isn’t feasible, so we could just give up and test against completely fake mock servers instead, and get similar coverage as we do now. It’s arguable but it’s not obviously the right call, and anyway replacing the current tests with the fake-mock-server tests would be a somewhat sizable effort.)
- The preceding OpenShift-k8s-specific signature API, using the `atomic:` transport. That is still shipped but fairly strongly deprecated (it was created before the X-R-S-S API, so we need it either not to break old `atomic:` users, or to allow reading/writing signatures for OpenShift versions that support the `atomic:` API but not X-R-S-S), and less and less relevant over time. OpenShift has already changed its API so that `atomic:` does not work against recent servers, we keep it only to access the old ones. At some point, I don’t know how to decide when, we could just cut our losses and stop testing `atomic:`, or even remove it altogether.

In an ideal world I don’t expect to happen any time soon, we would teach GitHub.com/distribution/distribution the X-R-S-S API, and test against that — a simple registry server is much smaller and easier to run. But nobody has worked on such a server, to my knowledge; and even in that case we would probably want to ensure interoperability with the OpenShift implementation of X-R-S-S. (… and in a _truly ideal_ world, we would have testing coverage of the most of the major registry products, but that’s another problem space entirely.)

cevich:
Okay, thanks for the insights.

Yeah, mocking it would be a major undertaking and additional maintenance, much better to use the real-deal.  Too bad it's such a behemoth.

There are also now registries on both github and gitlab, in addition to quay and in the google cloud.  I assume testing those wouldn't add much value, but please correct me if I'm wrong.

cevich:
Okay, I setup an POC [container build over in the c/automation_images repo.](https://github.com/containers/automation_images/pull/83).  Next I will work to incorporate it's use into both CI and the Makefile.

Makefile
```diff
@@ -136,9 +135,6 @@ bin/skopeo.%:
 	GOOS=$(word 2,$(subst ., ,$@)) GOARCH=$(word 3,$(subst ., ,$@)) $(GO) build $(MOD_VENDOR) ${SKOPEO_LDFLAGS} -tags "containers_image_openpgp $(BUILDTAGS)" -o $@ ./cmd/skopeo
 local-cross: bin/skopeo.darwin.amd64 bin/skopeo.linux.arm bin/skopeo.linux.arm64 bin/skopeo.windows.386.exe bin/skopeo.windows.amd64.exe
 
-build-container:
```
mtrmac, line 139: 🎉

Makefile
```diff
@@ -105,10 +106,8 @@ help:
 	@echo " * 'shell' - Run the built image and attach to a shell"
 	@echo " * 'clean' - Clean artifacts"
 
-# Build a container image (skopeobuild) that has everything we need to build.
-# Then do the build and the output (skopeo) should appear in current dir
+# Do the build and the output (skopeo) should appear in current dir
 binary: cmd/skopeo
-	${CONTAINER_RUNTIME} build ${BUILD_ARGS} -f Dockerfile.build -t skopeobuildimage .
```
mtrmac, line 111: Note the difference between `Dockerfile.build` and `Dockerfile`; this would break the following line.

(I don’t think any of our build processes use this target, so it might not be worth spending much time on changing it. OTOH some users do use this, at least from time to time, it seems…)

cevich: I ran it locally and it worked.  All the packages from `Dockerfile.build` wound up in the image I built (mostly by accident).  I've made that explicit now - [Ref. packages.txt file in my other PR.](https://github.com/containers/automation_images/pull/83/files)

mtrmac: I’m not saying we need two different containers (I haven’t checked); my point really is that this line used to build `skopeobuildimage` and the following line continues to run `skopeobuildimage`. So, [with the code the way I see it in this PR, which might not be relevant], it can only work if you have run the target previously and have a local image tagged `skopeobuildimage`, AFAICS.

cevich: Oh...I thought I fixed that...

cevich: ...ya, it's fixed.  Everything now references `$(IMAGE)` which I need to figure out how to look up from `.cirrus.yml` to keep things consistent...

cevich:
@mtrmac  I've taught `make` how to lookup and use the container image shared with CI.  This involves some compromises (pulling & running two container images), though it's a one-time cost (only pulled if absent) and the images are relatively small.  The big benefit here is, CI and local development will always remain in-sync (with an option to override for trying out a newer image).  Please clone this PR and try out the Makefile so far (`make docs-in-container`, `validate`, `shell`, `test-unit`, `test-integration`, etc).  I know some of the tests are still broken, but I'd like your blessing on the "look and feel" aspect.  

Meanwhile...let me see about fixing some of these docker-rate-limit problems, they're starting to drive me crazy :angry:

Makefile
cevich: ...okay, this has been solved by re-using the get_ci_vm container image.  It contains a script that can properly resolve env. var. values from inside `.cirrus.yml`.

cevich:
@mtrmac I'm pretty sure the etcd certificate error is coming from bad IP-address detection (as you suspect).  A long while ago, I was playing with etcd and seem to recall there is a special/magic config option (or env. var.) you can set that makes it relax it's verification when running in single-node "testing" mode.

```
Log line: 2021-07-26 10:48:21.219416 I | etcdserver/api/v3rpc: transport: http2Client.notifyError got notified that the client transport was broken remote error: tls: bad certificate.
```

Ring any bells?

cevich:
@mtrmac I'm looking through `integration/openshift.go` and the `go test` output, trying to reconcile what's happening between executing inside a container and directly on the host.  There are certainly some major networking differences, notably in `/etc/hosts`, `/etc/resolv.conf`, and of course outside the container there's the podman CNI network (bridge) in addition to the normal system interfaces.  Any number of these things could be breaking client certificate validation :confused: 

Is there some kind of special/extra debugging I can enable that will show the complete commands being executed?

Also a long-short:  Do you have any ideas on how we can get more details out of `etcd` regarding the client certificate validation, and why it's failing?  (this info. may not be available)

cevich:
Update: I tried testing the certificates/keys themselves both in a container and on a VM using openssl's `s_client` and `s_server`.  In both cases the certificates appear to work fine (w/o any validation settings enabled).  So to move forward here...we really need to know exactly why etcd is rejecting the client certificate when running on a VM.

contrib/cirrus/runner.sh
```diff
@@ -22,62 +22,83 @@ OS_RELEASE_ID="$(source /etc/os-release; echo $ID)"
...
+    errmsg="Somehow $SKOPEO_CI_CONTAINER_FQIN is not based on $OS_REL_VER."
+    if fgrep -qx "ID=$OS_RELEASE_ID" $mnt/etc/os-release; then
+        if ! fgrep -qx "VERSION_ID=$OS_RELEASE_VER" $mnt/etc/os-release; then
+            die "$errmsg"
+        fi
+    else
+        die "$errmsg"
+    fi
+    msg "Copying test binaries from $SKOPEO_CI_CONTAINER_FQIN /usr/local/bin/"
+    podman pull --quiet $SKOPEO_CI_CONTAINER_FQIN
```
mtrmac, line 85: This was already done a bit above; does this do anything extra?

hack/get_fqin.sh
```diff
@@ -0,0 +1,29 @@
...
+# the same script we need here.
+$CONTAINER_RUNTIME run --rm \
+    --security-opt label=disable \
+    -v $REPO_DIRPATH:/src:ro \
+    --entrypoint=/usr/share/automation/bin/cirrus-ci_env.py \
+    quay.io/libpod/get_ci_vm:latest \
+    --envs="Skopeo Test" /src/.cirrus.yml | \
+    egrep -m1 '^SKOPEO_CI_CONTAINER_FQIN' | \
+    awk -F "=" -e '{print $2}' | \
+    tr -d \'\"
```
mtrmac, line 29: /me resists temptation to play code golf :)

hack/get_fqin.sh
```diff
@@ -0,0 +1,29 @@
...
+# values.  More specifically, it allows obtaining the FQIN for the CI/Dev.
+# container from a single source (.cirrus.yml).
+
+set -e
+
+SCRIPT_FILEPATH=$(realpath "${BASH_SOURCE[0]}")
+SCRIPT_DIRPATH=$(dirname "$SCRIPT_FILEPATH")
+REPO_DIRPATH=$(realpath "$SCRIPT_DIRPATH/../")
+
+if [[ -n $(type -P docker) ]]; then CONTAINER_RUNTIME=docker; fi
```
mtrmac, line 16: Please prefer Podman here as well. (I have both binaries installed, and the build tried using both.)

Makefile
```diff
@@ -54,17 +51,28 @@ ifeq ($(GOOS), linux)
...
 # if this session isn't interactive, then we don't want to allocate a
 # TTY, which would fail, but if it is interactive, we do want to attach
 # so that the user can send e.g. ^C through.
 INTERACTIVE := $(shell [ -t 0 ] && echo 1 || echo 0)
 ifeq ($(INTERACTIVE), 1)
 	CONTAINER_CMD += -t
 endif
-CONTAINER_RUN := $(CONTAINER_CMD) "$(IMAGE)"
+GOSRC = /src/github.com/containers/skopeo
+CONTAINER_RUN ?= $(CONTAINER_CMD) --security-opt label=disable --privileged -v $(CURDIR):$(GOSRC) -w $(GOSRC) -e SKOPEO_CONTAINER_TESTS=$(SKOPEO_CONTAINER_TESTS) $(IMAGE)
```
mtrmac, line 75: All but one user sets `SKOPEO_CONTAINER_TESTS=1`; maybe `CONTAINER_RUN` could do that already, and the users could be simplified.

contrib/cirrus/runner.sh
```diff
@@ -22,62 +22,83 @@ OS_RELEASE_ID="$(source /etc/os-release; echo $ID)"
...
-        -e GOPATH=$GOPATH \
-        -v $GOPATH:$GOPATH:Z \
-        -w $SKOPEO_PATH \
-        $SKOPEO_CI_CONTAINER_FQIN \
-            make "$@"
-}
-
 _run_setup() {
+    local mnt
+    local errmsg
```
mtrmac, line 32: These variables aren’t set here AFAICS (but exist in `_run_integration`).

mtrmac (commented):
Just reading the implementation to see how it works; I haven’t tried to diagnose the OpenShift failure yet, I’m afraid.

For the things I rely on (local builds, in-container tests), this works perfectly. Thanks!

.cirrus.yml
```diff
@@ -41,8 +41,8 @@ env:
     UBUNTU_CONTAINER_FQIN: "quay.io/libpod/ubuntu_podman:${IMAGE_SUFFIX}"
     PRIOR_UBUNTU_CONTAINER_FQIN: "quay.io/libpod/prior-ubuntu_podman:${IMAGE_SUFFIX}"
 
-    # Equivilent to image produced by 'make build-container'
-    SKOPEO_CI_CONTAINER_FQIN: "quay.io/skopeo/ci:${DEST_BRANCH}"
+    # Equivilent to image produced by 'make build-container'. Auto-built by quay.
```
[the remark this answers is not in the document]
cevich: Woops, good catch, I'll fix that.

Makefile
```diff
@@ -29,11 +29,8 @@ ifeq ($(GOBIN),)
 GOBIN := $(GOPATH)/bin
 endif
 
-# Required for integration-tests to detect they are running inside a specific
-# container image.  Env. var defined in image, make does not automatically
-# pass to children unless explicitly exported
-export container_magic
-CONTAINER_RUNTIME := $(shell command -v podman 2> /dev/null || echo docker)
+# Multiple scripts are sensitive to this value, make sure it's exported/available
+export CONTAINER_RUNTIME ?= $(if $(shell type -P podman),podman,docker)
```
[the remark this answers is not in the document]
cevich: Oh, this is new to me, I will change it out of sake for consistency....hmmm, the man page on F33 doesn't show a `-v` option.  It does show a `-s` option for "search", can we use that?

cevich:
> I’m afraid.

You and me both! :scream: 

However, I think if we can hack on it together, between you knowledge of openshift and my ample supply of bubble-gum and duct-tape, we can figure it out :grin:

Makefile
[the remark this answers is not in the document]
cevich: okay, got it.  `-v` is a documented POSIX compatibility option.  Let's go with that.

Makefile
```diff
@@ -54,17 +51,28 @@ ifeq ($(GOOS), linux)
...
+CONTAINER_CMD ?= ${CONTAINER_RUNTIME} run --rm -i -e TESTFLAGS="$(TESTFLAGS)" -e CI=$(CI)
 # if this session isn't interactive, then we don't want to allocate a
 # TTY, which would fail, but if it is interactive, we do want to attach
 # so that the user can send e.g. ^C through.
 INTERACTIVE := $(shell [ -t 0 ] && echo 1 || echo 0)
 ifeq ($(INTERACTIVE), 1)
 	CONTAINER_CMD += -t
 endif
-CONTAINER_RUN := $(CONTAINER_CMD) "$(IMAGE)"
+GOSRC = /src/github.com/containers/skopeo
```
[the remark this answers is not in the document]
cevich: This is all still very highly-experimental, so yes, anything you feel makes it clearer or more understandable is good by me.  I'll change it.

Makefile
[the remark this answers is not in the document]
cevich: Oh I jammed that in there while debugging something...I can't remember.  I'll yank it back out.

Makefile
```diff
@@ -79,6 +87,9 @@ LIBDM_BUILD_TAG = $(shell hack/libdm_tag.sh)
 LOCAL_BUILD_TAGS = $(BTRFS_BUILD_TAG) $(LIBDM_BUILD_TAG)
 BUILDTAGS += $(LOCAL_BUILD_TAGS)
 
+# When running inside the CI/Dev container, all tags are known.
+CONTAINER_BUILDTAGS ?= btrfs_noversion libdm_no_deferred_remove
```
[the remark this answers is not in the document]
cevich: > Is this just to avoid running the hack/ autodetection scripts inside the container’s make?

Yes, I tried doing that and got it working, but wasn't happy that it would take up a few extra seconds of precious developer time.  The other thing I found was the scripts were ultimately resulting in an empty `BUILDTAGS` (maybe that's correct behavior?).

In any case, the contents of the container are entirely known and under our control, so I thought sticking in a default made sense.

> it might be a a bit surprising

Yeah, that's a good point, we don't want that :frowning: 

Maybe my fear of causing confusion between host and container BUILDTAGS is unfounded?

I'm open to other ideas/suggestions to make this easier, if you have any.

contrib/cirrus/runner.sh
cevich: Nope, must be leftover. I'll yank it.

hack/get_fqin.sh
cevich: Yep, np.  I'll also switch it to use `command -v` as well.

Makefile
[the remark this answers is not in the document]
cevich: :hankey: I remembered why I did this.  If you run `make test-integration` rootless, on F33 I get this error:

```
FAIL: copy_test.go:205: CopySuite.TestCopyWithManifestListStorageMultiple

Running skopeo --override-arch amd64 copy docker://registry.fedoraproject.org/fedora-minimal containers-storage:[vfs@/tmp/copy-manifest-list-storage-multiple800499411/root+/tmp/copy-manifest-list-storage-multiple800499411/runroot]test
copy_test.go:216:
    assertSkopeoSucceeds(c, "", "--override-arch", "amd64", "copy", knownListImage, "containers-storage:"+storage+"test")
utils.go:67:
... value *exec.ExitError = &exec.ExitError{ProcessState:(*os.ProcessState)(0xc0006b33c8), Stderr:[]uint8(nil)} ("exit status 1")
... time="2021-07-27T14:33:33Z" level=warning msg="error reading allowed ID mappings: error reading subuid mappings for user \"root\" and subgid mappings for group \"root\": No subuid ranges found for user \"root\" in /etc/subuid"
time="2021-07-27T14:33:33Z" level=warning msg="Found no UID ranges set aside for user \"root\" in /etc/subuid."
time="2021-07-27T14:33:33Z" level=warning msg="Found no GID ranges set aside for user \"root\" in /etc/subgid."
Copying blob sha256:033c7516884e47498394d3d1e2a10cc583802710fb877c357b3529b1dff65e02
time="2021-07-27T14:33:35Z" level=error msg="Error while applying layer: ApplyLayer exit status 1 stdout:  stderr: potentially insufficient UIDs or GIDs available in user namespace (requested 0:35 for /usr/libexec/utempter): Check /etc/subuid and /etc/subgid: lchown /usr/libexec/utempter: invalid argument"
time="2021-07-27T14:33:35Z" level=fatal msg="writing blob: adding layer with blob \"sha256:033c7516884e47498394d3d1e2a10cc583802710fb877c357b3529b1dff65e02\": ApplyLayer exit status 1 stdout:  stderr: potentially insufficient UIDs or GIDs available in user namespace (requested 0:35 for /usr/libexec/utempter): Check /etc/subuid and /etc/subgid: lchown /usr/libexec/utempter: invalid argument"
time="2021-07-27T14:33:35Z" level=error msg="exit status 1"
```

Any idea of a more elegant way to get past that?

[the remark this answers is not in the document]
mtrmac: OK, that’s way outside my expertise. @nalind is it reasonably possible to run a rootless unprivileged container, and use c/storage inside? ^^^

Makefile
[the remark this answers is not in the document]
mtrmac: > The other thing I found was the scripts were ultimately resulting in an empty BUILDTAGS (maybe that's correct behavior?).

I think so, the tags disable missing components or work around missing API.  But then `CONTAINER_BUILDTAGS` should be empty, to match the outcome, shouldn’t it?

---

Isn’t the problem here is that we mix user-input BUILDTAGS with the autodetected values with BUILDTAGS += LOCAL_TAGS? Without that, we could just forward the user choice directly into the container, and let both the outside and inside environment choose LOCAL_TAGS as appropriate (e.g. no autodetection in a container). But fixing that feels like a good topic for a separate PR, not something this one should deal with…

---

A reasonably cheap hack might be to skip the LOCAL_TAGS auto detection if SKOPEO_CONTAINER_TESTS == 1; that way we know we are in a known environment, we avoid running the slow shell script, all _extra_ user overrides via BUILDTAGS still work. The downside is that out-of-container LOCAL_TAGS still leak inside.

hack/get_fqin.sh
mtrmac: (`command -v` doesn’t matter here because of the `#!/usr/bin/env bash`. But it doesn’t hurt either.)

Makefile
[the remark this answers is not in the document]
cevich: Right, since when `make` is re-executed inside the container, it should do the right thing w/ running the scripts (and they're quick).  Okay, so I think this is a non-problem and It's safe for me to "un-fix" it :grinning:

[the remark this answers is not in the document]
cevich: I've stripped it out, and also removed all the passing-in of `BUILDTAGS` when **not** running in a container (so there is no leaking).  We/Me/Someone can open a separate PR to add in something like 'EXTRA_BUILDTAGS' to give developers some control, when running the container-based targets.

contrib/cirrus/runner.sh
cevich: Yep, I moved 'em.  You'll see, it's fixed in my next push (waiting for images to build first).

Makefile
cevich: okay, fixed.  Thanks.

Makefile
```diff
@@ -171,40 +172,42 @@ install-completions:
 	install -m 755 -d ${DESTDIR}${BASHCOMPLETIONSDIR}
 	install -m 644 completions/bash/skopeo ${DESTDIR}${BASHCOMPLETIONSDIR}/skopeo
 
-shell: build-container
-	$(CONTAINER_RUN) bash
+shell:
+	$(CONTAINER_RUN) env bash
```
cevich, line 176: woops, no need for `env` here.  I've stripped it out in the next version.

Makefile
```diff
@@ -171,40 +174,42 @@ install-completions:
...
 	install -m 644 completions/bash/skopeo ${DESTDIR}${BASHCOMPLETIONSDIR}/skopeo
 
-shell: build-container
-	$(CONTAINER_RUN) bash
+shell:
+	$(CONTAINER_RUN) env SKOPEO_CONTAINER_TESTS=1 bash
 
 check: validate test-unit test-integration test-system
 
-# The tests can run out of entropy and block in containers, so replace /dev/random.
```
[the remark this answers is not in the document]
cevich: I don't think it's necessary, at least in Cirrus-CI.  A few years ago I made contact with someone who (supposedly) convinced google to add in the option to turn on virtual RNG device in QEMU.  Additionally, we also have the "Make sure" solution of running the `rngd` service.  Otherwise yes, we had entropy problems before...to the extent that even the "login" getty service couldn't start :laughing:

In any case, short answer is "no", lack of entropy shouldn't be causing problems anymore.

Makefile
```diff
@@ -171,40 +174,42 @@ install-completions:
...
 
 # complicated set of options needed to run podman-in-podman
-test-system: build-container
+test-system:
 	DTEMP=$(shell mktemp -d --tmpdir=/var/tmp podman-tmp.XXXXXX); \
 	$(CONTAINER_CMD) --privileged \
-	    -v $$DTEMP:/var/lib/containers:Z -v /run/systemd/journal/socket:/run/systemd/journal/socket \
-            "$(IMAGE)" \
-            bash -c 'BUILDTAGS="$(BUILDTAGS)" $(MAKE) test-system-local'; \
+		-e SKOPEO_CONTAINER_TESTS=1 \
```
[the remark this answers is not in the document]
cevich: Ya, I just wanted to avoid the warning.

cevich:
Progress :clap:

[We're successfully starting up OpenShift now and talking to the registry.  ](https://cirrus-ci.com/task/4888452610129920?logs=integration#L41).  Next problem is the one I reproduced manually: Timeout in the next OpenShift startup + registry test (the traceback isn't very helpful :face_with_head_bandage:)

I've got to sign-off for the day, so we'll tackle that one tomorrow.

P.S. Thanks a **TON** for all your help so far.

Makefile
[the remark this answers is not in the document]
mtrmac: Great, thanks for the information (and having a written record here).

mtrmac:
> Timeout in the next OpenShift startup + registry test (the traceback isn't very helpful 🤕)

Try just bumping it to 25 minutes? IIRC I’ve had that locally for some time, though it was not necessary in Travis — and it’s also not necessary for the current implementation. So it’s not very likely to be the cause, or even necessarily the right fix (fast tests are better than slow tests, quite possibly we are doing something stupid), but there’s a chance it’s a half a bit of relevant information.

cevich:
> So it’s not very likely to be the cause, or even necessarily the right fix

No it's definitely not the right fix...remember these tests all pass fine when run under a `--privileged` container.  The one that's timing out is another registry-test, the one where it endlessly loops waiting to bind & connect to port 5000 then 5001.  Maybe https://github.com/containers/skopeo/blob/main/integration/check_test.go#L33 ?

What's the correct way to tell make to **only** run one test.  I tried the other day with `-test.run "SkopeoSuite.TestLoginLogout"` but then it ran no tests :disappointed:

cevich:
@mtrmac Since it never shows the failing test's name, just the unhelpful go-check timeout traceback, I ran all the tests one at a time.  The test that get's stuck in a loop is `CopySuite.TestCopyAllWithManifestList`:

```
...cut...
Output /usr/local/bin/oc oc create -f - stdout: Closed
Output registry-5000 stdout: waiting
Output registry-5000 stderr: waiting
Waiting for registry to start
Checking for port 5000...
Sleeping for port 5000
Output registry-5000 stderr: got 147,<nil>: time="2021-07-28T12:42:43-05:00" level=fatal msg="Unable to open configuration file: open /atomic-registry-config.yml: no such file or directory"
Output registry-5000 stderr: waiting
Output registry-5000 stderr: got 0,&errors.errorString{s:"EOF"}:
Output registry-5000 stderr: Closed
Output registry-5000 stdout: got 0,&errors.errorString{s:"EOF"}:
Output registry-5000 stdout: Closed
Sleeping for port 5000 done, will retry
Checking for port 5000...
Sleeping for port 5000
Sleeping for port 5000 done, will retry
Checking for port 5000...
Sleeping for port 5000
...repeats...
```

mtrmac:
> What's the correct way to tell make to **only** run one test. I tried the other day with `-test.run "SkopeoSuite.TestLoginLogout"` but then it ran no tests 😞

`openshift_shell_test.go` uses (also to enable logging as it goes): `-v -check.v -check.vv -check.f='CopySuite.TestRunShell'`

(We are only using that “check” testing mechanism because that’s what we started with. One of these days with copious spare time we probably should rip it out…)

mtrmac:
> Output registry-5000 stderr: got 147,<nil>: time="2021-07-28T12:42:43-05:00" level=fatal msg="Unable to open configuration file: open /atomic-registry-config.yml: no such file or directory"

Thanks. That looks like a root cause for the failure, and that file was created by `Dockerfile`; I think the part of `runner.sh` that copies files from `/usr/local/bin` should copy this one as well.

(I don’t care act all about continuing to place it in the root directory; maybe `/usr/local/etc/skopeo-tests` or something similarly easy to identify/search for would work better.)

Makefile
mtrmac: (I still see `SKOPEO_CONTAINER_TESTS=$(SKOPEO_CONTAINER_TESTS)`. If it’s in the next version you will push eventually, that’s perfectly fine, no need to push it right now.)

mtrmac:
> > So it’s not very likely to be the cause, or even necessarily the right fix
> 
> No it's definitely not the right fix...remember these tests all pass fine when run under a `--privileged` container. The one that's timing out is another registry-test, the one where it endlessly loops waiting to bind & connect to port 5000 then 5001. Maybe https://github.com/containers/skopeo/blob/main/integration/check_test.go#L33 ?

(Not that one, that one does fail eventually.)

Let’s fix the infinite hangs as well: #1390 .

cevich:
> That looks like a root cause for the failure, and that file was created by Dockerfile

Oh fantastic, and likely something I missed.  That should be an easy fix...

cevich:
Okay let's try this now (just pushed).  Note: I moved and combined the `TESTFLAGS` definition into the Makefile where it will be easier to maintain.

mtrmac:
> ... error: Missing or incomplete configuration info.  Please login or point to an existing, complete config file:

Huh. Is it possible that the home directory does not exist, or is not writeable? The `out` variable at https://github.com/containers/skopeo/blob/8060e41dce9009b9ec7e3cd3fe888b44085bba3b/integration/openshift.go#L219 might be useful, in case it reported successful login + inability to record that.

cevich:
> Is it possible that the home directory does not exist, or is not writeable?

We're running as root, so I doubt that simple case is the problem, but could it be an SELinux issue maybe?

I'm hitting all kinds of problems in podman land with these same VM images, due to policy updates: https://github.com/containers/podman/issues/10522

Specifically, any volume mounts with `:ro` or those missing a `:Z` or `:z` are causing all kinds of problems with registry containers.

I'm having a really (**REALLY**) hard time parsing all the openshift/registry output mixed in with the test output.  Is there any way we could send the openshift stuff into a file instead of printing it all?  It's really easy for me to tell Cirrus to attach a file after the task runs (pass or fail).

cevich:
confirmed: `$HOME` is set to `/root` and the permissions seem correct (`0755`), and SELinux type `admin_home_t:s0`

Makefile
cevich: Oh no, I had to un-fix it.  I did it wrong and there was one case where it was passing 0 instead of 1.  TBH, I'm a bit overloaded right now, so let's leave this as a FIXME (I marked it un-resolved).

mtrmac:
> Huh. Is it possible that the home directory does not exist, or is not writeable?

Close enough of a guess, it seems.

> confirmed: `$HOME` is set to `/root` and the permissions seem correct (`0755`), and SELinux type `admin_home_t:s0`

That’s not _quite_ what I see. The home directory, per `getpwnam(3)`/`/etc/passwd`, is indeed `/root`, but `$HOME` is either unset or empty.

See https://github.com/containers/skopeo/pull/1403 , which includes various debugging printouts, and where the integration tests pass when run with `HOME=/root make …`.

mtrmac:
> I'm having a really (**REALLY**) hard time parsing all the openshift/registry output mixed in with the test output. Is there any way we could send the openshift stuff into a file instead of printing it all? It's really easy for me to tell Cirrus to attach a file after the task runs (pass or fail).

I’m probably thinking about this incorrectly — I think those log streams have identifiable prefixes, so it seems possible to filter them out after-the-fact, using tools if necessary. So far it seemed convenient enough to me to have everything in the same log stream, usually in the natural order (“starting registry” … registry start-up logs … “waiting for registry start-up finished”); separating some of those logs out into external files would almost every time require looking up one or more of those files (presumably the main log would contain paths?), which seems to be more work to me. Is it the visual noise on each individual line? Too many useless log lines? The fact that everything is mixed in a single stream, regardless of what the individual lines look like? Something entirely different? Some combination of factors?

Admittedly:
- It is _definitely_ confusing that when start-up of the first cluster fails, we don’t tear it down and then the other test setup fails with a “port is already in use”. That’s should be fixable with a little effort.
- The `http2Client.notifyError got notified that the client transport was broken remote error: tls: bad certificate` logs certainly spam the log, and IIRC weren’t there when writing the tests originally. But if the tests pass anyway, I’m not _too_ tempted to dig deep into this old version of software to understand what that’s about; it is likely to be non-trivial effort and the outcome of that investigation is  not very probable to be reusable to any modern software.

cevich:
> So far it seemed convenient enough to me to have everything in the same log stream

That's fine too, it's just hard on my eyes.  But if it makes sense to you and you can make sense of it, I'm fine just being the lever-puller.

> Some combination of factors?

From the perspective of someone who doesn't see it much / ever, it's hard to visually distinguish between test harness output and subject (skopeo) output and tooling (openshift, registry, etc.) output.

> IIRC weren’t there when writing the tests originally

The horrendous amount of redundant messages from openshift/etcd ("blah blah tls cert invalid blah blah") are also really confusing.  But yes, this can be a problem for another day and PR.  I too am concerned as I don't remember seeing them before, maybe they were simply not shown before (somehow)?

> the integration tests pass when run with HOME=/root

Oh wait, this just struck me.  When the tests run, it's **not** a login shell, but in a `get_ci_vm.sh` environment, it is.  This may be exactly the problem, and I remember having it in tests on the other repos.  Let me copy-paste the fix...

Makefile
cevich: Oh, I got it now.  We can just set `SKOPEO_CONTAINER_TESTS=1` in `$CONTAINER_CMD` and remove it in all the other targets.  When not set (for whatever reason) it will be set to `1` only when `$CI == "true"` (nearly universal standard).

mtrmac:
> > So far it seemed convenient enough to me to have everything in the same log stream
> 
> That's fine too, it's just hard on my eyes. But if it makes sense to you and you can make sense of it, I'm fine just being the lever-puller.
> 
> > Some combination of factors?
> 
> From the perspective of someone who doesn't see it much / ever, it's hard to visually distinguish between test harness output and subject (skopeo) output and tooling (openshift, registry, etc.) output.

I think there’s basically two cases of the _asynchronous_ output: `Log line:` comes from OpenShift, and `Output registry-$port {stdout,stderr}` comes from registry processes. It should be easy enough to make the two more similar, or use a more prominent log prefix, or something like that.

Some other synchronous commands log output. Skopeo itself typically doesn’t log output, unless the test fails and the output is included in an assertion message.

Probably none of that is 100% consistent, as the code grew over time.

---

I can’t really remember any more, and of course past problems are not _that_ much of an indication of future problems, but it seems reasonable that those noisy `Output /usr/local/bin/oc oc create -f - stderr: waiting` lines were added for a reason, and losing the relative timing of those vs. other output might be bad. OTOH it’s also quite possible that those log lines were added only while developing the log catching implementation and were never relevant for timing; my commit message doesn’t say.

I don’t feel about the output format strongly… I guess PRs are welcome?

---

> The horrendous amount of redundant messages from openshift/etcd ("blah blah tls cert invalid blah blah") are also really confusing. But yes, this can be a problem for another day and PR. I too am concerned as I don't remember seeing them before, maybe they were simply not shown before (somehow)?

My first guess is that this has something to do with recent Go only looking for host names in subjectAltName instead of CN. But that’s really just a guess with no supporting evidence at all.

Makefile
mtrmac: Yup!

Makefile
```diff
@@ -54,17 +52,42 @@ ifeq ($(GOOS), linux)
...
+CI ?=
+
+# This env. var. is interpreted by some tests as a permission to
+# modify local configuration files and services.
+export SKOPEO_CONTAINER_TESTS ?= $(if $(CI),1,0)
+
+# This is a compromise, we either use a container for this or require
+# the local user to have a compatible python3 development environment.
+# Define it as a "resolve on use" variable to avoid calling out when possible
+export SKOPEO_CIDEV_CONTAINER_FQIN ?= $(shell hack/get_fqin.sh)
```
mtrmac, line 80: It turns out this doesn’t work, the script is run even if not used (which I noticed because macOS doesn’t have `realpath(1)`, sigh).

Removing `export` seems to help; and because we only use `SKOPEO_CIDEV_CONTAINER_FQIN` to start a container, not to do anything inside the container, it seems to me that removing `export` shouldn’t cause any extra script invocations.

cevich: Hrmmm...looking at the script I think not having the export should be okay?  I was trying to cover the case where the developer wants to override the value.  That won't be available unless it's explicitly exported (or I'm mistaken in understanding the docs).  Either way, I can remove it for now.

cevich: Also, I don't think it will matter about `realpath` since the mac won't be able to run the container anyway.

BTW, I'm not happy about needing to run a container to lookup the name of the image.  Unf. the script I have that properly interprets the `.cirrus.yml` values is in python.  We can do away with the "run a container" part, but will then need to jump through hoops (or document) to ensure the python 3 interpreter + python3-yaml module are both available :disappointed: 

Okay, I have the export removed in the next version.

cevich:
Huh...well [it seems (somehow) the tests passed.](https://cirrus-ci.com/build/5326981723062272)  I didn't notice anything funny in the output, but I'm a bit skeptical.  Let me push this small `Makefile` fixup and let's see if it passes again...

Makefile
mtrmac: > I was trying to cover the case where the developer wants to override the value. That won't be available unless it's explicitly exported (or I'm mistaken in understanding the docs).

I couldn’t immediately find anything in `info make` to the contrary, and
```make
with_default = with_default_value

target:
	echo $(no_default)
	echo $(with_default)
```
allows overriding using `make -s no_default=a with_default=b` without an export.

mtrmac: > Also, I don't think it will matter about `realpath` since the mac won't be able to run the container anyway.

Yes.

> BTW, I'm not happy about needing to run a container to lookup the name of the image.

\*shrug\* for users that want to run builds/tests in a container, i.e. prefer not to change the current user environment, running the lookup in another container seems likely to be acceptable, whereas imposing requirements on the current user environment might be unwanted; so using a container makes sense to me.

cevich: Ya, that was my line of thought as well.  Thanks for checking out the Make behavior.  I remember reading the specifics on the official GNU docs page.   But the actual behavior is what matters most.

cevich:
[Well, okay then it passed again.](https://cirrus-ci.com/build/4858500447731712)  Removing the draft status, let's get on with final review before something breaks :grin:

contrib/cirrus/runner.sh
```diff
@@ -23,61 +36,101 @@ OS_RELEASE_VER="$(source /etc/os-release; echo $VERSION_ID | tr -d '.')"
...
+    # within the CI/Dev container image, but we want to run
+    # things directly on the host VM.  Fortunately they're all
+    # located in the container under /usr/local/bin
+    msg "Accessing contents of $SKOPEO_CIDEV_CONTAINER_FQIN"
+    podman pull --quiet $SKOPEO_CIDEV_CONTAINER_FQIN
+    mnt=$(podman mount $(podman create $SKOPEO_CIDEV_CONTAINER_FQIN))
+
+    # This should never ever ever happen, so double-check to make sure.
+    errmsg="Somehow $SKOPEO_CIDEV_CONTAINER_FQIN is not based on $OS_REL_VER."
+    if fgrep -qx "ID=$OS_RELEASE_ID" $mnt/etc/os-release; then
```
mtrmac, line 78: Is this somehow different from
```sh
if !fgrep … \
       || !fgrep ; then
    die "Somehow…"
fi
```
? If there is a trick to what this doing, please document it so that unaware people like me don’t break it in the future.

Makefile
```diff
@@ -171,40 +186,42 @@ install-completions:
...
-test-integration: build-container
-	$(CONTAINER_RUN) bash -c 'rm -f /dev/random; ln -sf /dev/urandom /dev/random; SKOPEO_CONTAINER_TESTS=1 BUILDTAGS="$(BUILDTAGS)" $(MAKE) test-integration-local'
+test-integration:
+	$(CONTAINER_RUN) $(MAKE) test-integration-local
+
 
-# Intended for CI, shortcut 'build-container' since already running inside container.
-test-integration-local:
+# Intended for CI, assumed to be running in quay.io/libpod/skopeo_cidev container.
+test-integration-local: bin/skopeo
```
mtrmac, line 199: `hack/make/test-integration` invokes `make bin/skopeo` already — I guess that command in `hack/make/test-integration` can be dropped in favor of adding this dependency, here we have more visible control over the build parameters.

Moving the dependency from shell here is unrelated and non-blocking; I’d prefer not to have _both_ and the potential confusion about how it is expected to work, though.

Makefile
```diff
@@ -171,40 +186,42 @@ install-completions:
...
+			$(MAKE) test-system-local; \
 	rc=$$?; \
-	$(RM) -rf $$DTEMP; \
+	-$(RM) -rf $$DTEMP; \
 	exit $$rc
 
-# Intended for CI, shortcut 'build-container' since already running inside container.
-test-system-local:
+# Intended for CI, assumed to already be running in quay.io/libpod/skopeo_cidev container.
+test-system-local: bin/skopeo
```
mtrmac, line 216: `hack/make/test-system` invokes `make bin/skopeo` already (see `integration` above).

mtrmac (commented):
Thanks for all the work!

I’m really only worried about the `/etc/os-release` check question; the rest are non-blocking nits.

.cirrus.yml
```diff
@@ -57,8 +57,8 @@ validate_task:
     # under Cirrus-CI, due to challenges obtaining the starting commit ID.
     # Only do validation for PRs.
     only_if: $CIRRUS_PR != ''
-    container:  &build_container
-        image: "${SKOPEO_CI_CONTAINER_FQIN}"
+    container:
+        image: '${SKOPEO_CIDEV_CONTAINER_FQIN}'
```
mtrmac, line 61:
```
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
	LANGUAGE = (unset),
	LC_ALL = (unset),
	LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
```
That might be worth fixing in the image.

Makefile
```diff
@@ -54,17 +52,42 @@ ifeq ($(GOOS), linux)
...
+# You can increase test outputerbosity with the option '-test.vv'.
+# You can select certain tests to run, with `-test.run <regex>` for example:
+#
+#     make test-unit TESTFLAGS='-test.run ^TestBuild$'
+#
+# For integration test, we use [gocheck](https://labix.org/gocheck).
+# You can increase test outputerbosity with the option '-check.vv'.
+# You can limit test selection with `-check.f <regex>`, for example:
+#
+#     make test-integration TESTFLAGS='-check.f DockerSuite.TestBuild*'
```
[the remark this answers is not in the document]
cevich: well you gave me the copy-paste, so might as well use it :grinning:

contrib/cirrus/runner.sh
cevich: No tricks, I just thought the logic made more sense that way.  But NP doing it the way that makes sense to you...

Makefile
cevich: > here we have more visible control over the build parameters.

Yes that was my thought too...did I forget to remove it from the script?

I did, woops.   Okay it's gone now.

Makefile
cevich: Yep, my bad.  I got it.

contrib/cirrus/runner.sh
```diff
@@ -23,61 +36,101 @@ OS_RELEASE_VER="$(source /etc/os-release; echo $VERSION_ID | tr -d '.')"
...
 
 _run_unit() {
-    podmanmake test-unit-local BUILDTAGS="$BUILDTAGS"
+    make test-unit-local BUILDTAGS="$BUILDTAGS"
 }
 
 _run_integration() {
-    podmanmake test-integration-local BUILDTAGS="$BUILDTAGS"
+    # Ensure we start with a clean-slate
+    podman system reset --force
```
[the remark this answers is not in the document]
cevich: I'd prefer to keep it since we are doing some container stuffs during _run_setup().  We do the same thing for system tests too, but there I think it's even more necessary.

.cirrus.yml
cevich: Oof, that means re-rolling them all again.  I opened https://github.com/containers/automation_images/issues/85 for this.

cevich:
@mtrmac thanks for all your help on this too, couldn't have done it w/o you.  What a doozy of a PR! :sweat_smile:

contrib/cirrus/runner.sh
mtrmac: Mostly the `fgrep` + `!fgrep` combination caught my eye — the negation difference between two similarly-looking conditions seemed likely to be intentional.

contrib/cirrus/runner.sh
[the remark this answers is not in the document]
mtrmac: `_run_setup` is a good point.

mtrmac (approved):
LGTM. Thanks again!

mtrmac:
FWIW https://cirrus-ci.com/build/6216764259303424 lists some “Issues found while parsing configuration” — most look like stylistic preferences rather than something that needs fixing. I have no idea about the severity (or it might have been a well-considered long-standing decision), I just happened to randomly notice that field (well, the whole page, actually) for the first time today.

cevich:
>some “Issues found while parsing configuration” — most look like stylistic preferences rather than something that needs fixing.

Ugh, I know it does that everywhere now.  For some reason they really don't like how we use `alias` + `name` in jobs, but we literally have to because of a bug they (supposedly) can't fix.  It's safe to ignore for now.

PR #1346, opened 2021-06-29
[release-1.3] Bump to Skopeo v1.3.1, c/common v0.38.11, c/storage v1.31.3 

TomSweeneyRedHat:
Bumping to Skopeo v1.3.1, c/common to v0.38.11, and c/storage to v1.31.3 in preparation for RHEL 8.4.0.2


mtrmac (commented):
Looks good, but `validate` fails.

@vrothberg PTAL as well.

TomSweeneyRedHat:
The validate seems to be failing on the second commit.  We need to get this in today, I think I'll do a new PR with just the first commit and then do a follow up PR to bump the version to dev.

PR #1332, opened 2021-06-24
Add --retry-times to markdown docs

rittneje:
Fixes #1322.

PR #1347, opened 2021-06-29
man page checker - part 1 of 2

edsantiago:
Add new script, hack/man-page-checker, copied from podman. Run it
in 'make validate-local' target.

This is NOT the checker requested in #1332 (verify that flags
listed in 'skopeo foo --help' are documented in man pages and
vice-versa). This is a much simpler script that merely looks
for very basic typos or discrepancies between skopeo.1.md
and skopeo-foo.1.md.

The next part (cross-checking flags) is in progress but will
require a huge number of changes to the man pages. I'm submitting
this now because it's easy to review.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -209,8 +209,10 @@ validate: build-container
 # This target is only intended for development, e.g. executing it from an IDE. Use (make test) for CI or pre-release testing.
 test-all-local: validate-local test-unit-local
 
+.PHONY: validate-local
 validate-local:
 	hack/make.sh validate-git-marks validate-gofmt validate-lint validate-vet
+	hack/man-page-checker
 
 test-unit-local:
 	$(GPGME_ENV) $(GO) test $(MOD_VENDOR) -tags "$(BUILDTAGS)" $$($(GO) list $(MOD_VENDOR) -tags "$(BUILDTAGS)" -e ./... | grep -v '^github\.com/containers/skopeo/\(integration\|vendor/.*\)$$')
--- a/cmd/skopeo/login.go
+++ b/cmd/skopeo/login.go
@@ -21,7 +21,7 @@ func loginCmd(global *globalOptions) *cobra.Command {
 		global: global,
 	}
 	cmd := &cobra.Command{
-		Use:     "login",
+		Use:     "login [command options] REGISTRY",
 		Short:   "Login to a container registry",
 		Long:    "Login to a container registry on a specified server.",
 		RunE:    commandAction(opts.run),
--- a/docs/skopeo-login.1.md
+++ b/docs/skopeo-login.1.md
@@ -4,7 +4,7 @@
 skopeo\-login - Login to a container registry.
 
 ## SYNOPSIS
-**skopeo login** [*options*] *registry*
+**skopeo login** [*options*] _registry_
 
 ## DESCRIPTION
 **skopeo login** logs into a specified registry server with the correct username
--- a/docs/skopeo-standalone-sign.1.md
+++ b/docs/skopeo-standalone-sign.1.md
@@ -4,7 +4,7 @@
 skopeo\-standalone-sign - Debugging tool - Publish and sign an image in one step.
 
 ## SYNOPSIS
-**skopeo standalone-sign** _manifest docker-reference key-fingerprint_ **--output**|**-o** _signature_
+**skopeo standalone-sign** [*options*] _manifest_ _docker-reference_ _key-fingerprint_
 
 ## DESCRIPTION
 This is primarily a debugging tool, or useful for special cases,
@@ -31,4 +31,3 @@ skopeo(1), skopeo-copy(1), containers-signature(5)
 ## AUTHORS
 
 Antonio Murdaca <runcom@redhat.com>, Miloslav Trmac <mitr@redhat.com>, Jhon Honce <jhonce@redhat.com>
-
--- a/docs/skopeo.1.md
+++ b/docs/skopeo.1.md
@@ -78,15 +78,15 @@ See [containers-transports(5)](https://github.com/containers/image/blob/master/d
 | Command                                   | Description                                                                    |
 | ----------------------------------------- | ------------------------------------------------------------------------------ |
 | [skopeo-copy(1)](skopeo-copy.1.md)        | Copy an image (manifest, filesystem layers, signatures) from one location to another. |
-| [skopeo-delete(1)](skopeo-delete.1.md)    | Mark image-name for deletion.                                                  |
-| [skopeo-inspect(1)](skopeo-inspect.1.md)  | Return low-level information about image-name in a registry.                   |
-| [skopeo-list-tags(1)](skopeo-list-tags.1.md)  | List the tags for the given transport/repository.                           |
+| [skopeo-delete(1)](skopeo-delete.1.md)    | Mark the _image-name_ for later deletion by the registry's garbage collector.  |
+| [skopeo-inspect(1)](skopeo-inspect.1.md)  | Return low-level information about _image-name_ in a registry.                 |
+| [skopeo-list-tags(1)](skopeo-list-tags.1.md)  | Return a list of tags for the transport-specific image repository.         |
 | [skopeo-login(1)](skopeo-login.1.md)  | Login to a container registry. |
 | [skopeo-logout(1)](skopeo-logout.1.md)  | Logout of a container registry. |
-| [skopeo-manifest-digest(1)](skopeo-manifest-digest.1.md)    | Compute a manifest digest of manifest-file and write it to standard output.|
-| [skopeo-standalone-sign(1)](skopeo-standalone-sign.1.md)    | Sign an image.                                               |
-| [skopeo-standalone-verify(1)](skopeo-standalone-verify.1.md)| Verify an image.                                             |
-| [skopeo-sync(1)](skopeo-sync.1.md)| Copy images from one or more repositories to a user specified destination.             |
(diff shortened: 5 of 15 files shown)


docs/skopeo-login.1.md
```diff
@@ -4,7 +4,7 @@
 skopeo\-login - Login to a container registry.
 
 ## SYNOPSIS
-**skopeo login** [*options*] *registry*
+**skopeo login** [*options*] _registry_
```
edsantiago, line 7: The skopeo convention is to use underscores, not asterisks.

cmd/skopeo/login.go
```diff
@@ -21,7 +21,7 @@ func loginCmd(global *globalOptions) *cobra.Command {
 		global: global,
 	}
 	cmd := &cobra.Command{
-		Use:     "login",
+		Use:     "login [command options] REGISTRY",
```
edsantiago, line 24: The skopeo convention is to emit `[command options]` instead of just `[options]`

edsantiago (commented):
Tests finally green; the hardest part was dealing with the Macintosh version of sed

TomSweeneyRedHat:
LGTM
thanks for getting this underway @edsantiago

Makefile
```diff
@@ -209,8 +209,10 @@ validate: build-container
 # This target is only intended for development, e.g. executing it from an IDE. Use (make test) for CI or pre-release testing.
 test-all-local: validate-local test-unit-local
 
+.PHONY: validate-local
```
mtrmac, line 212: OK. I guess failing to update the `.PHONY` at the top is good enough evidence that adding it directly above the target is more likely to cause future rule additions to notice. :)

Makefile
```diff
@@ -209,8 +209,10 @@ validate: build-container
...
 validate-local:
 	hack/make.sh validate-git-marks validate-gofmt validate-lint validate-vet
+	hack/man-page-checker
```
mtrmac, line 215: The `validate` target above should invoke `make validate-local` instead of open-coding it so that it also triggers this check.

rhatdan:
Great job @edsantiago 
LGTM

hack/man-page-checker
```diff
@@ -0,0 +1,153 @@
...
+#
+# Make sure the SYNOPSIS line in skopeo-foo.1.md reads '**skopeo foo** ...'
+for md in *.1.md;do
+    synopsis=$(egrep -A1 '^#* SYNOPSIS' $md|tail -1)
+
+    # Command name must be bracketed by double asterisks; options and
+    # arguments are bracketed by single ones.
+    #   E.g. '**skopeo copy** [*options*] _..._'
+    # Get the command name, and confirm that it matches the md file name.
+    cmd=$(echo "$synopsis" | sed -E -e 's/(.*)\*\*.*/\1/' | tr -d \*)
```
mtrmac, line 124: Maybe
```suggestion
    cmd=$(echo "$synopsis" | sed -E -e 's/(.+)\*\*.*/\1/' | tr -d \*)
```
or
```suggestion
    cmd=$(echo "$synopsis" | sed -E -e 's/^\*\*(.+)\*\*.*/\1/' | tr -d \*)
```
?

I ultimately had to test this to understand how this works, and whether this matches the _initial_ `**`. The current code is fine because regexes prefer longest matches, but that feels borderline too subtle.

hack/man-page-checker
```diff
@@ -0,0 +1,153 @@
...
+    fi
+done
+
+# Helper function: compares man page synopsis vs --help usage message
+function compare_usage() {
+    local cmd="$1"
+    local from_man="$2"
+
+    # Sometimes in CI we run before skopeo gets built.
+    test -x ../bin/skopeo || return
```
mtrmac, line 74: Non-blocking: This works but feels a bit fragile.

Maybe split this into two checks, one run in `validate-local`, and one in a new `validate-built-binary` that depends on `bin/skopeo`, and run the latter one only in `test_skopeo_task`?

OTOH if you don’t think it’s worth the effort, it’s absolutely fine to leave it as it is.

hack/man-page-checker
```diff
@@ -0,0 +1,153 @@
+#!/usr/bin/env bash
+#
+# man-page-checker - validate and cross-reference man page names
+#
+# This is the script that cross-checks BETWEEN MAN PAGES. It is not the
+# script that cross-checks that each option in skopeo foo --help is listed
+# in skopeo-foo.1.md and vice-versa; that one is xref-helpmsgs-manpages.
+#
+# IMPORTANT NOTE: this script runs on Macs, on which sed is an ancient
+# non-gnu version. To make sed work on all platforms, we invoke with '-E'.
```
mtrmac, line 10: (I don’t expect a high risk of the inputs or outputs varying across architectures; so if you wanted to make this a separate `make` target that isn’t run on the “build+unit-test on macOS” Cirrus task, that would be perfectly fine with me.

OTOH now that the work has been done, there might be no particular reason to do _even more_ work to drop macOS…)

hack/man-page-checker
```diff
@@ -0,0 +1,153 @@
...
+
+    # Run 'cmd --help', grab the line immediately after 'Usage:'
+    local help_output=$(../bin/$cmd --help)
+    local from_help=$(echo "$help_output" | grep -A1 '^Usage:' | tail -1)
+
+    # strip off command name from both
+    from_man=$(sed -E -e "s/\*\*$cmd\*\*[[:space:]]*//" <<<"$from_man")
+    from_help=$(sed -E -e "s/^[[:space:]]*$cmd[[:space:]]*//" <<<"$from_help")
+
+    # man page lists 'foo [*options*]', help msg shows 'foo [flags]'.
```
mtrmac, line 84:
```suggestion
    # man page lists 'foo [*options*]', help msg shows 'foo [command options]'.
```

hack/man-page-checker
```diff
@@ -0,0 +1,153 @@
...
+    elif expr "$from_help" : "\[command options\]" >/dev/null; then
+        echo "WARNING: $cmd: --help shows [command options], man page does not show [*options*]"
+        rc=1
+    fi
+
+    # Strip off options and flags; start comparing arguments
+    from_man=$(sed  -E -e 's/^\[\*options\*\][[:space:]]*//' <<<"$from_man")
+    from_help=$(sed -E -e 's/^\[command options\][[:space:]]*//'      <<<"$from_help")
+
+    # Args in man page are '*foo*', in --help are 'FOO'. Convert all to
```
mtrmac, line 102:
```suggestion
    # Args in man page are '_foo_', in --help are 'FOO'. Convert all to
```

docs/skopeo-standalone-sign.1.md
```diff
@@ -4,7 +4,7 @@
 skopeo\-standalone-sign - Debugging tool - Publish and sign an image in one step.
 
 ## SYNOPSIS
-**skopeo standalone-sign** _manifest docker-reference key-fingerprint_ **--output**|**-o** _signature_
+**skopeo standalone-sign** [*options*] _manifest_ _docker-reference_ _key-fingerprint_
```
mtrmac, line 7: It’s somewhat attractive to change the `--help` output instead (the`--output` option is mandatory), OTOH it would not be worth making the script any more complex.

docs/skopeo.1.md
```diff
@@ -78,15 +78,15 @@ See [containers-transports(5)](https://github.com/containers/image/blob/master/d
 | Command                                   | Description                                                                    |
 | ----------------------------------------- | ------------------------------------------------------------------------------ |
 | [skopeo-copy(1)](skopeo-copy.1.md)        | Copy an image (manifest, filesystem layers, signatures) from one location to another. |
-| [skopeo-delete(1)](skopeo-delete.1.md)    | Mark image-name for deletion.                                                  |
-| [skopeo-inspect(1)](skopeo-inspect.1.md)  | Return low-level information about image-name in a registry.                   |
-| [skopeo-list-tags(1)](skopeo-list-tags.1.md)  | List the tags for the given transport/repository.                           |
+| [skopeo-delete(1)](skopeo-delete.1.md)    | Mark the _image-name_ for later deletion by the registry's garbage collector.  |
+| [skopeo-inspect(1)](skopeo-inspect.1.md)  | Return low-level information about _image-name_ in a registry.                 |
+| [skopeo-list-tags(1)](skopeo-list-tags.1.md)  | Return a list of tags for the transport-specific image repository.         |
```
mtrmac, line 83: “Return” is not quite what commands do on stdout, maybe `List tags in the transport-specific image repository`?

mtrmac:
WRT the “The Skopeo convention is” notes, it’s pretty likely those conventions happened without much deliberate thought, so if changing things helped for consistency, sure.

Makefile
edsantiago: I don't feel comfortable making so many changes to this Makefile - there's a lot I don't understand, especially wrt `$(CONTAINER_RUN)`, and `hack/make.sh`, and `contrib/cirrus/runner.sh`, and what-gets-run-where in Cirrus and in devel environments. I was hoping to do the smallest possible change, for ease of future maintainability; and it sounds like I didn't get it right. I will take a fresh look tomorrow.

edsantiago:
I've really, really missed your reviews @mtrmac. You caught a lot of subtleties I missed, and a number of issues that I deliberately ignored because I didn't dare make bigger changes. I will address them all tomorrow morning. Thank you.

mtrmac:
#1351 to put _some_ effort to my words, but that’s just a stopgap for today.

Makefile
mtrmac: To be fair if we are not running `make validate` in CI, probably no-one else is running it anywhere else :)

PR #1350, opened 2021-06-29
Man page validation: part 2 of 2

edsantiago:
This is the script that runs 'skopeo COMMAND --help' and
cross-checks that all the option flags are documented
in man pages, and vice-versa (all options listed in man
pages appear in COMMAND's --help message).

Copied from podman, with changes for skopeo-land (removing
the rst checks, and conforming to skopeo conventions).

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -65,6 +65,15 @@ validate_task:
         make validate-local
         make vendor && hack/tree_status.sh
 
+doccheck_task:
+    only_if: $CIRRUS_PR != ''
+    depends_on:
+      - validate
+    container:  *build_container
+    script: |
+      "${GOSRC}/${SCRIPT_BASE}/runner.sh" setup
+      "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" build
+      "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" doccheck
 
 osx_task:
     only_if: &not_docs $CIRRUS_CHANGE_TITLE !=~ '.*CI:DOCS.*'
@@ -185,6 +194,7 @@ success_task:
     # N/B: ALL tasks must be listed here, minus their '_task' suffix.
     depends_on:
         - validate
+        - doccheck
         - osx
         - cross
         - test_skopeo
--- a/Makefile
+++ b/Makefile
@@ -207,12 +207,17 @@ validate: build-container
 	$(CONTAINER_RUN) make validate-local
 
 # This target is only intended for development, e.g. executing it from an IDE. Use (make test) for CI or pre-release testing.
-test-all-local: validate-local test-unit-local
+test-all-local: validate-local validate-docs test-unit-local
 
 .PHONY: validate-local
 validate-local:
 	hack/make.sh validate-git-marks validate-gofmt validate-lint validate-vet
+
+# This invokes bin/skopeo, hence cannot be run as part of validate-local
+.PHONY: validate-docs
+validate-docs:
 	hack/man-page-checker
+	hack/xref-helpmsgs-manpages
 
 test-unit-local:
 	$(GPGME_ENV) $(GO) test $(MOD_VENDOR) -tags "$(BUILDTAGS)" $$($(GO) list $(MOD_VENDOR) -tags "$(BUILDTAGS)" -e ./... | grep -v '^github\.com/containers/skopeo/\(integration\|vendor/.*\)$$')
--- a/contrib/cirrus/runner.sh
+++ b/contrib/cirrus/runner.sh
@@ -65,6 +65,10 @@ _run_validate() {
     podmanmake validate-local BUILDTAGS="$BUILDTAGS"
 }
 
+_run_doccheck() {
+    podmanmake validate-docs BUILDTAGS="$BUILDTAGS"
+}
+
 _run_unit() {
     podmanmake test-unit-local BUILDTAGS="$BUILDTAGS"
 }
--- a/docs/skopeo-copy.1.md
+++ b/docs/skopeo-copy.1.md
@@ -20,7 +20,11 @@ automatically inherit any parts of the source name.
 
 ## OPTIONS
 
-**--all**
+**--additional-tag**=_strings_
+
+Additional tags (supports docker-archive).
+
+**--all**, **-a**
 
 If _source-image_ refers to a list of images, instead of copying just the image which matches the current OS and
 architecture (subject to the use of the global --override-os, --override-arch and --override-variant options), attempt to copy all of
@@ -42,57 +46,119 @@ Path of the authentication file for the source registry. Uses path given by `--a
 
 Path of the authentication file for the destination registry. Uses path given by `--authfile`, if not provided.
 
+**--dest-shared-blob-dir** _directory_
+
+Directory to use to share blobs across OCI repositories.
+
 **--digestfile** _path_
 
 After copying the image, write the digest of the resulting image to the file.
 
-**--format, -f** _manifest-type_ MANIFEST TYPE (oci, v2s1, or v2s2) to use in the destination (default is manifest type of source, with fallbacks)
+**--encrypt-layer** _ints_
 
-**--quiet, -q** suppress output information when copying images
+*Experimental* the 0-indexed layer indices, with support for negative indexing (e.g. 0 is the first layer, -1 is the last layer)
 
-**--remove-signatures** do not copy signatures, if any, from _source-image_. Necessary when copying a signed image to a destination which does not support signatures.
+**--format**, **-f** _manifest-type_
 
-**--sign-by=**_key-id_ add a signature using that key ID for an image name corresponding to _destination-image_
+MANIFEST TYPE (oci, v2s1, or v2s2) to use in the destination (default is manifest type of source, with fallbacks)
 
-**--encryption-key** _protocol:keyfile_ specifies the encryption protocol, which can be JWE (RFC7516), PGP (RFC4880), and PKCS7 (RFC2315) and the key material required for image encryption. For instance, jwe:/path/to/key.pem or pgp:admin@example.com or pkcs7:/path/to/x509-file.
+**--help**, **-h**
 
-**--decryption-key** _key[:passphrase]_ to be used for decryption of images. Key can point to keys and/or certificates. Decryption will be tried with all keys. If the key is protected by a passphrase, it is required to be passed in the argument and omitted otherwise.
+Print usage statement
 
-**--src-creds** _username[:password]_ for accessing the source registry.
+**--quiet**, **-q**
 
-**--dest-compress** _bool-value_ Compress tarball image layers when saving to directory using the 'dir' transport. (default is same compression type as source).
+Suppress output information when copying images.
 
-**--dest-oci-accept-uncompressed-layers** _bool-value_ Allow uncompressed image layers when saving to an OCI image using the 'oci' transport. (default is to compress things that aren't compressed).
+**--remove-signatures**
 
-**--dest-creds** _username[:password]_ for accessing the destination registry.
+Do not copy signatures, if any, from _source-image_. Necessary when copying a signed image to a destination which does not support signatures.
 
-**--src-cert-dir** _path_ Use certificates at _path_ (*.crt, *.cert, *.key) to connect to the source registry or daemon.
+**--sign-by**=_key-id_
 
-**--src-no-creds** _bool-value_ Access the registry anonymously.
+Add a signature using that key ID for an image name corresponding to _destination-image_
 
-**--src-tls-verify** _bool-value_ Require HTTPS and verify certificates when talking to container source registry or daemon (defaults to true).
+**--src-shared-blob-dir** _directory_
 
-**--dest-cert-dir** _path_ Use certificates at _path_ (*.crt, *.cert, *.key) to connect to the destination registry or daemon.
+Directory to use to share blobs across OCI repositories.
 
-**--dest-no-creds** _bool-value_  Access the registry anonymously.
+**--encryption-key** _protocol:keyfile_
 
-**--dest-tls-verify** _bool-value_ Require HTTPS and verify certificates when talking to container destination registry or daemon (defaults to true).
+Specifies the encryption protocol, which can be JWE (RFC7516), PGP (RFC4880), and PKCS7 (RFC2315) and the key material required for image encryption. For instance, jwe:/path/to/key.pem or pgp:admin@example.com or pkcs7:/path/to/x509-file.
 
-**--src-daemon-host** _host_ Copy from docker daemon at _host_. If _host_ starts with `tcp://`, HTTPS is enabled by default. To use plain HTTP, use the form `http://` (default is `unix:///var/run/docker.sock`).
+**--decryption-key** _key[:passphrase]_
 
-**--dest-daemon-host** _host_ Copy to docker daemon at _host_. If _host_ starts with `tcp://`, HTTPS is enabled by default. To use plain HTTP, use the form `http://` (default is `unix:///var/run/docker.sock`).
+Key to be used for decryption of images. Key can point to keys and/or certificates. Decryption will be tried with all keys. If the key is protected by a passphrase, it is required to be passed in the argument and omitted otherwise.
+
+**--src-creds** _username[:password]_
+
+Credentials for accessing the source registry.
+
+**--dest-compress** _bool-value_
+
+Compress tarball image layers when saving to directory using the 'dir' transport. (default is same compression type as source).
+
+**--dest-oci-accept-uncompressed-layers** _bool-value_
+
+Allow uncompressed image layers when saving to an OCI image using the 'oci' transport. (default is to compress things that aren't compressed).
+
+**--dest-creds** _username[:password]_
+
+Credentials for accessing the destination registry.
+
+**--src-cert-dir** _path_
+
+Use certificates at _path_ (*.crt, *.cert, *.key) to connect to the source registry or daemon.
+
+**--src-no-creds** _bool-value_
+
+Access the registry anonymously.
+
+**--src-tls-verify** _bool-value_
+
+Require HTTPS and verify certificates when talking to container source registry or daemon (defaults to true).
+
+**--dest-cert-dir** _path_
+
+Use certificates at _path_ (*.crt, *.cert, *.key) to connect to the destination registry or daemon.
+
+**--dest-no-creds** _bool-value_
+
+Access the registry anonymously.
+
+**--dest-tls-verify** _bool-value_
+
+Require HTTPS and verify certificates when talking to container destination registry or daemon (defaults to true).
+
+**--src-daemon-host** _host_
+
+Copy from docker daemon at _host_. If _host_ starts with `tcp://`, HTTPS is enabled by default. To use plain HTTP, use the form `http://` (default is `unix:///var/run/docker.sock`).
+
+**--dest-daemon-host** _host_
+
+Copy to docker daemon at _host_. If _host_ starts with `tcp://`, HTTPS is enabled by default. To use plain HTTP, use the form `http://` (default is `unix:///var/run/docker.sock`).
 
 Existing signatures, if any, are preserved as well.
 
-**--dest-compress-format** _format_ Specifies the compression format to use.  Supported values are: `gzip` and `zstd`.
+**--dest-compress-format** _format_
 
-**--dest-compress-level** _format_ Specifies the compression level to use.  The value is specific to the compression algorithm used, e.g. for zstd the accepted values are in the range 1-20 (inclusive), while for gzip it is 1-9 (inclusive).
+Specifies the compression format to use.  Supported values are: `gzip` and `zstd`.
 
-**--src-registry-token** _Bearer token_ for accessing the source registry.
+**--dest-compress-level** _format_
 
-**--dest-registry-token** _Bearer token_ for accessing the destination registry.
+Specifies the compression level to use.  The value is specific to the compression algorithm used, e.g. for zstd the accepted values are in the range 1-20 (inclusive), while for gzip it is 1-9 (inclusive).
 
-**--retry-times**  the number of times to retry, retry wait time will be exponentially increased based on the number of failed attempts.
+**--src-registry-token** _token_
+
+Bearer token for accessing the source registry.
+
+**--dest-registry-token** _token_
+
+Bearer token for accessing the destination registry.
+
+**--retry-times**
+
+The number of times to retry. Retry wait time will be exponentially increased based on the number of failed attempts.
 
 ## EXAMPLES
 
--- a/docs/skopeo-inspect.1.md
+++ b/docs/skopeo-inspect.1.md
@@ -31,10 +31,18 @@ Output configuration in OCI format, default is to format in JSON format.
 
 Username and password for accessing the registry.
 
+**--daemon-host** _host_
+
+Use docker daemon host at _host_ (`docker-daemon:` transport only)
+
 **--format**, **-f**=*format*
 
 Format the output using the given Go template.
 The keys of the returned JSON can be used as the values for the --format flag (see examples below).
+
+**--help**, **-h**
+
+Print usage statement
 
 **--no-creds**
 
@@ -53,9 +61,9 @@ Registry token for accessing the registry.
 
 The number of times to retry; retry wait time will be exponentially increased based on the number of failed attempts.
 
-**--tls-verify**
+**--shared-blob-dir** _directory_
 
-Require HTTPS and verify certificates when talking to container registries (defaults to true).
+Directory to use to share blobs across OCI repositories.
 
 ## EXAMPLES
 
--- a/hack/xref-helpmsgs-manpages
+++ b/hack/xref-helpmsgs-manpages
@@ -1,0 +1,277 @@
+#!/usr/bin/perl
+#
+# xref-helpmsgs-manpages - cross-reference --help options against man pages
+#
+package LibPod::CI::XrefHelpmsgsManpages;
+
+use v5.14;
+use utf8;
+
+use strict;
+use warnings;
+
+(our $ME = $0) =~ s|.*/||;
+our $VERSION = '0.1';
+
+# For debugging, show data structures using DumpTree($var)
+#use Data::TreeDumper; $Data::TreeDumper::Displayaddress = 0;
+
+# unbuffer output
+$| = 1;
+
+###############################################################################
+# BEGIN user-customizable section
+
+# Path to skopeo executable
+my $Default_Skopeo = './bin/skopeo';
+my $SKOPEO = $ENV{SKOPEO} || $Default_Skopeo;
+
+# Path to all doc files (markdown)
+my $Docs_Path = 'docs';
+
+# Global error count
+my $Errs = 0;
+
+# END   user-customizable section
+###############################################################################
+
+###############################################################################
+# BEGIN boilerplate args checking, usage messages
+
+sub usage {
+    print  <<"END_USAGE";
+Usage: $ME [OPTIONS]
+
+$ME recursively runs 'skopeo --help' against
+all subcommands; and recursively reads skopeo-*.1.md files
+in $Docs_Path, then cross-references that each --help
+option is listed in the appropriate man page and vice-versa.
+
+$ME invokes '\$SKOPEO' (default: $Default_Skopeo).
+
+Exit status is zero if no inconsistencies found, one otherwise
+
+OPTIONS:
+
+  -v, --verbose  show verbose progress indicators
+  -n, --dry-run  make no actual changes
+
+  --help         display this message
+  --version      display program name and version
+END_USAGE
+
+    exit;
+}
+
+# Command-line options.  Note that this operates directly on @ARGV !
+our $debug   = 0;
+our $verbose = 0;
+sub handle_opts {
+    use Getopt::Long;
+    GetOptions(
+        'debug!'     => \$debug,
+        'verbose|v'  => \$verbose,
+
+        help         => \&usage,
+        version      => sub { print "$ME version $VERSION\n"; exit 0 },
+    ) or die "Try `$ME --help' for help\n";
+}
+
+# END   boilerplate args checking, usage messages
+###############################################################################
+
+############################## CODE BEGINS HERE ###############################
+
+# The term is "modulino".
+__PACKAGE__->main()                                     unless caller();
+
+# Main code.
+sub main {
+    # Note that we operate directly on @ARGV, not on function parameters.
+    # This is deliberate: it's because Getopt::Long only operates on @ARGV
+    # and there's no clean way to make it use @_.
+    handle_opts();                      # will set package globals
+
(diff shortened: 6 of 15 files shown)


edsantiago:
I think this is going to fail, because I bet Macs don't have perl, but I'll figure that all out tomorrow when I incorporate @mtrmac's review feedback from #1347. Right now I just want to see how far we get. (It passes on my home system).

mtrmac:
FWIW
```
% perl --version

This is perl 5, version 30, subversion 2 (v5.30.2) … 
```
on macOS 11.4, but it’s deprecated (Google quickly finds https://forums.macrumors.com/threads/are-the-scripting-languages-still-here.2300956/ ).

But with that I’m done for today.

Makefile
```diff
@@ -212,7 +212,12 @@ test-all-local: validate-local test-unit-local
 .PHONY: validate-local
 validate-local:
 	hack/make.sh validate-git-marks validate-gofmt validate-lint validate-vet
+
+# This invokes bin/skopeo, hence cannot be run as part of validate-local
+.PHONY: validate-docs
+validate-docs:
```
mtrmac, line 218: Maybe `validate-docs-local` for consistency, and have `test-all-local` depend on it.

(I wouldn’t bother adding the `$CONTAINER_RUN` non-`local` variant.)

contrib/cirrus/runner.sh
```diff
@@ -62,7 +62,7 @@ _run_cross() {
 }
 
 _run_validate() {
-    podmanmake validate-local BUILDTAGS="$BUILDTAGS"
+    podmanmake validate-local validate-docs BUILDTAGS="$BUILDTAGS"
```
mtrmac, line 65: This works but I’m worried that it will be confusing.

Maybe name this `…/runner.sh validate-and-validate-docs`, something that doesn’t directly correspond to a `Makefile` rule.

mtrmac (commented):
(Just looking at the build system integration for now, for ideas to use in #1351 .)

Makefile
mtrmac: … on second though, I don’t feel strongly about the `-local` suffix at all. Having a convention to always type extra is a hassle, and Chris is strongly against tests that _must_ run in a container.

This might be a good time to move towards changing the convention.

So just the `test-all-local` dependency would be nice.

contrib/cirrus/runner.sh
```diff
@@ -65,6 +65,10 @@ _run_validate() {
     podmanmake validate-local BUILDTAGS="$BUILDTAGS"
 }
 
+_run_doccheck() {
```
edsantiago, line 68: Will this do? New runner.sh target, invoked in CI, with a name unambiguously different from the Makefile target.

mtrmac: Works for me.

RFC: @cevich  I don’t think c/image tests _need_ to run this, but there might well be something I’m missing about keeping the two test systems consistent.

hack/xref-helpmsgs-manpages
```diff
@@ -0,0 +1,334 @@
...
+            # Special case: these are actually global options
+            next if $k =~ /^--(cni-config-dir|runtime)$/ && $man =~ /-build/;
+
+            # Special case: weirdness with Cobra and global/local options
+            next if $k eq '--namespace' && $man =~ /-ps/;
+
+            next if "@subcommand" eq 'system' && $k eq 'service';
+
+            # Special case: skopeo completion is a hidden command
+            next if $k eq 'completion';
```
mtrmac, line 174: This looks (mostly?) Podman-specific.

(I have no idea how to sanely maintain the script across repositories and prevent drift, and this script is already very useful even with these exceptions, so it well might be worth keeping it as is to decrease the maintenance overhead.)

hack/xref-helpmsgs-manpages
```diff
@@ -0,0 +1,334 @@
...
+        elsif ($section eq 'commands') {
+            # In skopeo.1.md
+            if ($line =~ /^\|\s*\[skopeo-(\S+?)\(\d\)\]/) {
+                # $1 will be changed by recursion _*BEFORE*_ left-hand assignment
+                my $subcmd = $1;
+                $man{$subcmd} = skopeo_man("skopeo-$1");
+            }
+
+            # In skopeo-<subcommand>.1.md
+            elsif ($line =~ /^\|\s+(\S+)\s+\|\s+\[\S+\]\((\S+)\.1\.md\)/) {
```
mtrmac, line 278: (I don’t think this happens in Skopeo — but see the other comment about maintainability, that might be a very good reason to keep it as is.)

hack/xref-helpmsgs-manpages
```diff
@@ -0,0 +1,334 @@
...
+                $previous_subcmd = $subcmd;
+                $man{$subcmd} = skopeo_man($2);
+            }
+        }
+
+        # Options should always be of the form '**-f**' or '**\-\-flag**',
+        # possibly separated by comma-space.
+        elsif ($section eq 'flags') {
+            # e.g. 'skopeo run --ip6', documented in man page, but nonexistent
+            if ($line =~ /^not\s+implemented/i) {
```
mtrmac, line 294: Skopeo doesn’t have `--ip6`, and I think this doesn’t currently match anything.  Again see elsewhere, defining a convention in advance might well be the ideal approach.

hack/xref-helpmsgs-manpages
```diff
@@ -0,0 +1,334 @@
...
+                # Keep track of them, in case we see 'Not implemented' below
+                push @most_recent_flags, $1;
+            }
+        }
+    }
+    close $fh;
+
+    # Special case: the 'image trust' man page tries hard to cover both set
+    # and show, which means it ends up not being machine-readable.
+    if ($command eq 'skopeo-image-trust') {
```
mtrmac, line 319: podman-image-trust I guess.

docs/skopeo-copy.1.md
```diff
@@ -42,57 +46,119 @@ Path of the authentication file for the source registry. Uses path given by `--a
...
+Directory to use to share blobs across OCI repositories.
+
 **--digestfile** _path_
 
 After copying the image, write the digest of the resulting image to the file.
 
-**--format, -f** _manifest-type_ MANIFEST TYPE (oci, v2s1, or v2s2) to use in the destination (default is manifest type of source, with fallbacks)
+**--encrypt-layer** _ints_
+
+*(Experimental)*: key with the encryption protocol to use needed to encrypt the image (e.g. jwe:/path/to/key.pem)
```
mtrmac, line 59:
```suggestion
*Experimental* the 0-indexed layer indices, with support for negative indexing (e.g. 0 is the first layer, -1 is the last layer)
```
(I haven’t reviewed the details of the man pages any further for now.)

docs/skopeo-inspect.1.md
```diff
@@ -53,9 +61,9 @@ Registry token for accessing the registry.
 
 The number of times to retry; retry wait time will be exponentially increased based on the number of failed attempts.
 
-**--tls-verify**
```
mtrmac, line 56: Hum, so this exposes a mess (in several subcommands).

The previous CLI framework had a strong distinction between global options and local options, and required the order `skopeo $global-options $command $local-options`. In that context, `--tls-verify` originally was a global option, but became a local per-command option (for _some_ commands, it’s local precisely because `skopeo copy` etc. have separate `--src-tls-verify` and `--dest-tls-verify`), and we have left around global `--tls-verify` for compatibility, but hidden and with a warning.

Cobra has basically no distinction between global/local options; the `tis-verify` option name ends up sharing (some) state. So now we have the global option (in `createApp`) still marked as hidden, so that it doesn’t show up in help in `skopeo copy`, but that hidden bit is shared, and we don’t list it in help of `skopeo inspect` and others.

IIRC I have looked into this some time ago and there wasn’t a clean way to fix this, the `tis-verify` option is either hidden everywhere, or not. It might be worth another look… or, sadly, we might need an exception in the checker.

mtrmac (commented):
Thanks! This is very useful.

I couldn’t resist trying to see if I can understand at least a bit of Perl. It’s entirely up to you how you prefer to deal with the various special cases in a way that is maintainable.

This explicitly exposes the problems with Skopeo’s `--tls-verify` implementation.

mtrmac:
FWIW I have filed https://github.com/containers/skopeo/issues/1354 for the “/dev/tcp/127.0.0.1/5000: No route to host” failures. These come up about daily now, and a re-start almost(?) always succeeds.

(So far, I’m ashamed to admit, I’ve just been pressing Re-run because I don’t know where to start with “no route to 127.0.0.1”.)

hack/xref-helpmsgs-manpages
edsantiago: I'm sorry! This is part of the WIP: I intend to strip this and other such sections before removing the WIP.

As you identified, maintaining across repositories is a no-hope situation.

hack/xref-helpmsgs-manpages
edsantiago: Yes, I did a simple brainless `s/podman/skopeo/` just to get the script to run. This, and the above, and a few other instances need scrutiny which I meant to apply today but didn't.

docs/skopeo-copy.1.md
edsantiago: Thank you, this is an important part of the feedback I'm looking for. I just copy-pasted the `--help` output.

hack/xref-helpmsgs-manpages
mtrmac: There’s no rush, I’m sorry if I made that impression.

docs/skopeo-inspect.1.md
edsantiago: >--tls-verify originally was a global option, but became a local per-command option

Ahhh, thank you, that is important context that helps me understand. I think the important factor here is that `--help` and man pages should match. Right now `--tls-verify` does not appear in any `skopeo help` output, neither globally nor locally, even in `login` which appears to override with a local `tls-verify` option. If we add an exception in the code, that would mean we allow `--tls-verify` to appear in some man pages without appearing in `--help`. That makes me really uncomfortable. I think this needs deeper thought...

docs/skopeo-copy.1.md
mtrmac: Copy&pasting `--help` is absolutely fair, this PR should not be blocked on agonizing over details of the full man page set.

I noted the above because the “0-indexed layer indices” text above is what I see in the `skopeo copy --help` output, so I was wondering if it suggests a bug in some automation used to generate the new text.

docs/skopeo-inspect.1.md
mtrmac: A possible path would be to not mark `--tls-verify` as hidden in Cobra, but “just” filter it out put the main `skopeo --help` output. I don’t know exactly what this would take but from a quick look it seems possible in principle (`SetHelpFunc`).

That’s extremely ugly but it is _an_ option, if we can’t find anything better.

rhatdan: Can't we just mark it hidden for individual commands instead of globally?

mtrmac: AFAIK no, because
- The “hidden” bit is shared; so marking the “local” variant of `inspect --tls-verify` hidden would still mark the global one as hidden
- The flag data is basically data-only; we build all subcommand structures in advance, and AFAIK don’t get a change to _change_ what the flags are after Cobra recognizes a subcommand (and for maintainability I would prefer to steer away from doing so anyway)

edsantiago:
OK, I think this is ready for review. I've reworked the CI invocation so it doesn't run on macs; incorporated @mtrmac's feedback from #1347 (the few bits he didn't fix in #1351); and removed unnecessary bits from the xref script.

contrib/cirrus/runner.sh
cevich: Yes, keeping the `skopeo` testing task YAML 100% the same between the two repos is important since it simplifies the burden of maintaining the necessary duplication.  But I'm fine with dropping the extra validate call in that task here (and in c/image).  I was slightly afraid the extra validation might be necessary to confirm golang is still happy (in c/image), after grafting c/skopeo code in.  But it sounds like you feel this is unnecessary, and I'm fine with removing it.  That way the docs testing only happens in skopeo CI, keeping things simple.

Makefile
```diff
@@ -207,12 +207,17 @@ validate: build-container
...
-test-all-local: validate-local test-unit-local
+test-all-local: validate-local validate-docs test-unit-local
 
 .PHONY: validate-local
 validate-local:
 	hack/make.sh validate-git-marks validate-gofmt validate-lint validate-vet
+
+# This invokes bin/skopeo, hence cannot be run as part of validate-local
+.PHONY: validate-docs
+validate-docs:
```
mtrmac, line 218: Non-blocking? Maybe wrap this an “if not on Linux, succeed” check, so that `test-all-local` is usable for macOS users. (Or wrap only the `man-page-checker` script if that’s the only one that fails; I should test that.)

hack/xref-helpmsgs-manpages
cevich: > I have no idea how to sanely maintain the script across repositories and prevent drift

> As you identified, maintaining across repositories is a no-hope situation.

FWIW: The way this is done with other shared scripts is a bit complex, but functional and reliable:  Stick the script into c/automation (along with it's unit tests) so it installs directly into the VMs via c/automation_images.  Then call the script from the task.  Stability is maintained by tagging "releases" of c/automation, and [pinning that version into the VM build process](https://github.com/containers/automation_images/blob/main/lib.sh#L27).

I will admit it's a lot of work up-front, but I think it does pay off long-term by not having weird problems happening on release branches (where VM images don't ever change).

mtrmac (commented):
Thanks! A quick impression:
- Overall the integration makes sense to me. I’d like @cevich ’s review to confirm this won’t break the c/image sharing, I am not caught up with how that works at all yet.
- I didn’t review the actual man page updates… yet?
- My main concern is the `--tls-verify` flag handling. I’m tempted to try to fix it before applying this (and removing the pre-existing documentation of that flag, only to re-add it again, hopefully soon). OTOH that might be an unreasonable obstacle.

.cirrus.yml
```diff
@@ -142,6 +142,8 @@ test_skopeo_task:
         "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" build
     validate_script: >-
         "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" validate
+    doccheck_script: >-
+        "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" doccheck
```
cevich, line 146: Mentioned in a reply to a review comment (above), but might as well put it here.

I would prefer this be included in the 'validate' task higher up.  I *think* we can get away with dropping `validate_script` here (it's redundant).  This will also simplify the necessary copy/paste of this task over into c/image's `.cirrus.yml`.

Hopefully that makes sense.

cevich: Alternatively, I'm also okay with having this be in a dedicated/standalone task.

hack/xref-helpmsgs-manpages
edsantiago: If the question is "can we share a version of this script between podman, buildah, and skopeo", the answer is "not in any sane manner". Each repo has different conventions and special cases. Maintaining this in a share repo would mean that every edit would need to be tested in podman/buildah/skopeo ... and we know that will not happen.

History has shown that these scripts are fine as they are, and do not get new-functionality edits, only new-special-case-handling edits (see git log in podman and buildah). As much as I hate duplication, I think it's the most reasonable approach in this case.

.cirrus.yml
mtrmac: I agree `validate` in `test_skopeo_task` seems unnecessary (it is also not very related to the c/image tests, `validate` mostly targets the non-`vendor` parts of the codebase).

Moving the man page checks into `validate` might not be a good fit because the two checks depend on a built `bin/skopeo` binary, and at that point a separate `validate` “quick check” that is a prerequisite for other CI tasks loses some of its benefits. (That’s assuming that `validate_task` is intended to be fast and small, which is a guess, not a documented design.)

edsantiago: Running in `validate` was my first approach; @mtrmac pushed against it as being confusing. Also, running in `validate` might make it run on Macs, which we 100% do not want.

hack/xref-helpmsgs-manpages
mtrmac: It seems vaguely possible to share the core of the logic, and then either have a shared script that has the special cases conditioned on the program under check (like the `podman-image-trust` case that came up earlier, with that name), or just have the special cases maintained directly in the individual Podman/Skopeo/… repos.

OTOH now that the core logic works, it seems quite unlikely to need substantial changes, so the payoff of sharing _that_ part might well be absolutely not worth the effort to share it.

Makefile
edsantiago: I'm totally fine with blocking this until a solution is found for `--tls-verify`. 

Re: changes to man pages -- I'm a little uncomfortable with the options not being sorted, but that's how they are now and I'm not willing to change that in this PR (it would become unreviewable). If someone agrees with me that options should be sorted, I'll happily submit a precursor PR next week to fix that, then rebase this PR once that merges.

.cirrus.yml
mtrmac: The thing I was worried about being confusing in this PR is `runner.sh validate` running both `validate-local` and `validate-docs`.

I think @cevich is suggesting moving this into `validate_task`, which does not involve `runner.sh` at all.

(AFAICS neither `validate_task` nor `test_skopeo_task` currently runs on Macs, though the current `validate_task`, before the dependency on recent `sed`, could in principle be moved to that platform.)

docs/skopeo-inspect.1.md
mtrmac: > A possible path would be to not mark `--tls-verify` as hidden in Cobra, but “just” filter it out of the main `skopeo --help` output. I don’t know exactly what this would take but from a quick look it seems possible in principle (`SetHelpFunc`).
> 
> That’s extremely ugly but it is _an_ option, if we can’t find anything better.

That might not actually work either; I can’t get `--tls-verify` to show up in `skopeo inspect --help`.

https://github.com/containers/skopeo/pull/1358 to show what it takes, and maybe someone can figure out another way to get this done. (That PR is already pretty close to unacceptably ugly, even with tests to verify this keeps working.)

mtrmac: Alternatively, _if_ Cobra really doesn’t care about option positioning, we could just drop the global option and have existing commands add a local `--tls-verify`, hidden or not and warning or not, depending on circumstances (and we wouldn’t add it to future commands that shouldn’t have it.) That might be quite clean, after all.

edsantiago: It's more subtle than that: Cobra does care, but only in one direction: it can accept a global option on the right-hand side, but cannot accept a local (right-hand) option on the left (before the subcommand). Moving `--tls-verify` to a local per-subcommand option would (hypothetically) break legacy tools that run `skopeo --tls-verify=...`. I assume that's the reason for keeping it global and deprecated.

mtrmac: Yes, that’s the reason. So much for that idea.

edsantiago: Here's a wild idea:
```diff
diff --git a/cmd/skopeo/main.go b/cmd/skopeo/main.go
index 670260e0..4f959932 100644
--- a/cmd/skopeo/main.go
+++ b/cmd/skopeo/main.go
@@ -55,7 +55,7 @@ func createApp() (*cobra.Command, *globalOptions) {
        var dummyVersion bool
        rootCommand.Flags().BoolVarP(&dummyVersion, "version", "v", false, "Version for Skopeo")
        rootCommand.PersistentFlags().BoolVar(&opts.debug, "debug", false, "enable debug output")
-       flag := optionalBoolFlag(rootCommand.PersistentFlags(), &opts.tlsVerify, "tls-verify", "Require HTTPS and verify certificates when accessing the registry")
+       flag := optionalBoolFlag(rootCommand.PersistentFlags(), &opts.tlsVerify, "tls-verifyx", "Require HTTPS and verify certificates when accessing the registry")
        flag.Hidden = true
        rootCommand.PersistentFlags().StringVar(&opts.policyPath, "policy", "", "Path to a trust policy file")
        rootCommand.PersistentFlags().BoolVar(&opts.insecurePolicy, "insecure-policy", false, "run the tool without any policy check")
```
(that is: append 'x' to 'tls-verify'). In my testing, that works to the extent that:
* skopeo --help does not list the option
* skopeo login --help **does** list the option (because it explicitly declares it, I don't know why)
* skopeo --tls-verify=false inspect, delete, list-tags seem to recognize `--tls-verify` as a shortcut for `--tls-verifyx`, and emit the expected warnings (I did not confirm if the option actually works as intended). And
* skopeo --tls-verify=falce copy, sync, standalone*, those don't work at all. I don't know why.

Anhow, seems worth playing around with?

mtrmac: I’ll also ask Cobra maintainers, maybe it’s something that could be enabled with a reasonable code change…

hack/xref-helpmsgs-manpages
cevich: Yes I agree. It only makes sense if the core logic is substantial, and important enough to have it's own unit-tests.  I'm not familiar with this code at all, so I trust Ed's judgement.  My intention was simply to point out the possibility in case it made sense.

.cirrus.yml
cevich: It sounds to me like this should get it's own task then.  It could probably run in a container, but I'm okay if a VM is needed (note: this comes with a 1-minute spin-up time hit).  Something like (not tested):

```yaml
validate_task:
    only_if: $CIRRUS_PR != ''
    depends_on:
        - validate
    container:  *build_container
    script: |
        "${GOSRC}/${SCRIPT_BASE}/runner.sh" setup
        "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" build
        "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" doccheck
```

**note:** The `setup` step is probably only needed for the (accident prevention) `dnf erase skopeo`.  The default VM setup has 1-2gig of disk space available, which should be fine for this task.  I don't have actual knowledge of this though, so maybe `setup` is needed.

Makefile
mtrmac: WRT sorting, I don’t really have a preference. It would make sense to me to group the options by topic in principle, but that’s hard to maintain/enforce over PR reviews with little context, so even a man page that starts with good topic grouping probably fairly soon results in an unordered mess of options — and at that point a machine-enforced alphabetical ordering ends up being clearly better.

.cirrus.yml
mtrmac: If `validate_task` can run in a container, this one looks like it should run in a container similarly easily (same dependencies on Go to build the binary, + shell/perl to run the tests.)

(Does `runner.sh setup` work from inside a container? I’d expect the `growpart` to fail. Anyway the existing `validate_task` could be a starting point.)

cevich: Right, in that case just `s/.*\/runner.sh" setup/dnf erase -y skopeo/` and it should be good to go.  Unless there are new perl requirements, that particular container image should already have what's needed.

cevich: "new" meaning, something beyond what Podman and Buildah need for the same check.

lsm5:
rebase needed

edsantiago:
@lsm5 not really, this is just the stupid github bug that complains if PR isn't on the very tip of main. The warning can (and should) be ignored.

lsm5:
> @lsm5 not really, this is just the stupid github bug that complains if PR isn't on the very tip of main. The warning can (and should) be ignored.

ah i remember now. my bad.

.cirrus.yml
edsantiago: I'm completely lost here. IIUC you want a new `something_task` in `.cirrus.yml`? If so, I've just pushed a new change that (I think) does that. I'm not really the best at cirrus yaml, so I may have completely gotten it wrong.

edsantiago:
New task map: 
cirrus-map-man_page_checker_2

edsantiago:
It failed. Moving on to other things, this isn't something I can debug. If someone would like to take over this PR, I'm fine with it. 
```
"${GOSRC}/${SCRIPT_BASE}/runner.sh" setup
WARNING: It does not appear that containers/automation was installed.
         Functionality of most of /var/tmp/go/src/github.com/containers/skopeo/./contrib/cirrus/runner.sh will be negatively
         impacted.
/var/tmp/go/src/github.com/containers/skopeo/./contrib/cirrus/runner.sh: line 87: req_env_vars: command not found

Exit status: 127
```

mtrmac:
I don’t know that much about the CI setup, but comparing with `validate_task`: just replacing the three `script:` entries with a simple `make validate-docs`, with no `runner.sh` use, might work.

docs/skopeo-inspect.1.md
mtrmac: > (that is: append 'x' to 'tls-verify'). In my testing, that works to the extent that:

…
> * skopeo --tls-verify=false inspect, delete, list-tags seem to recognize `--tls-verify` as a shortcut for `--tls-verifyx`, and emit the expected warnings (I did not confirm if the option actually works as intended). 

I don’t see any warnings in this case (right now, with Cobra 1.2.1 pflag 1.0.5).

Also AFAICS prefixes aren’t matched, as a rule: `skopeo --hel`, `skopeo --override-o inspect` just fail for me.

> * skopeo --tls-verify=falce copy, sync, standalone*, those don't work at all. I don't know why.

That’s consistent with no prefix matching: After the global `--tls-verify` is renamed, there isn’t any `--tls-verify` option for these subcommands, so it seems expected to me that this would fail.

---

In my experiments, the warning actually doesn’t work as expected. It does trigger with `skopeo copy` (when there isn’t a local option, the global option is found and triggers a warning), but both `skopeo inspect --tls=verify=false` and `skopeo --tls-verify=false inspect` edit the local option, and no warning is triggered. (I think this used to work, I’ll look into when it broke.)

So the duplication doesn’t buy us that much; we could just as well have only the global option, and then selectively warn on it based on subcommand choice. (OTOH that doesn’t do what we _truly_ want, which is for the option to only show up in `--help` of some subcommands.)

---

> It's more subtle than that: Cobra does care, but only in one direction: it can accept a global option on the right-hand side, but cannot accept a local (right-hand) option on the left (before the subcommand).

It’s even _more_ subtle than that: it _sometimes_ works in the local → global direction as well:
```console
% bin/skopeo --src-creds=foo copy a b
# does not complain about --src-creds being unknown
```
and
```console
% bin/skopeo --unknown=a copy a b
FATA[0000] unknown flag: --unknown
```
vs.
```console
% bin/skopeo --unknown copy a b
FATA[0000] unknown command "a" for "skopeo"             
```

When interpreting the command line (without `TraverseChildren`), Cobra first looks for the (optionally nested) subcommand name by stripping flags from the command line; after it finds the subcommand, it merges all relevant flags into one `FlagSet` (somehow, perhaps with priorities), strips the recognized (nested) subcommand name from the command line, and then uses the merged `FlagSet` on the flags-only command line.

The subtlety is that dividing the command line into subcommands/flags requires the parser, in some cases, to know which flags require an argument, and the code happens to assume that unknown flags do require an argument; so `--unknown copy a b` is interpreted as `--unknown=copy a b`, which explains the error message. And _this_ part does not use merged flags, i.e. a local option is not interpreted correctly when put to the left of the subcommand name (but it may be accepted if the parser ends up identifying the subcommand).

mtrmac: There’s even more subtleties to this, see https://github.com/spf13/cobra/issues/1449 .

But I think we might actually get the outcome we want, it seems: don’t use `PersistentFlags`, but a separate local flag for the root and for (all current) subcommands; and set `TraverseChildren`. If there isn’t an inherited persistent flag, all the ambiguity goes away.

Test program:
```go
package main

import (
	"bytes"
	"fmt"
	"strings"

	"github.com/spf13/cobra"
)

type options struct {
	deprecatedFlag bool
	sameNameFlag   bool
	srcFlag        bool
}

func dummyRun(_ *cobra.Command, _ []string) {}

func createApp() (*cobra.Command, *options) {
	opts := options{}

	rootCommand := &cobra.Command{Use: "main", Long: "Top-level main help"}
	rootCommand.Flags().BoolVar(&opts.deprecatedFlag, "flag", false, "deprecated - use per-subcommand flags")
	if err := rootCommand.Flags().MarkHidden("flag"); err != nil {
		panic(`unable to mark "flag" as hidden`)
	}
	rootCommand.PersistentFlags().BoolVar(&opts.sameNameFlag, "global", false, "foo")
	rootCommand.TraverseChildren = true

	sameNameCmd := &cobra.Command{Use: "same-name", Long: "same-name help", Run: dummyRun}
	sameNameCmd.Flags().BoolVar(&opts.sameNameFlag, "flag", false, "this is not deprecated")
	rootCommand.AddCommand(sameNameCmd)

	otherNameCmd := &cobra.Command{Use: "other-name", Long: "other-name help", Run: dummyRun}
	otherNameCmd.Flags().BoolVar(&opts.deprecatedFlag, "flag", false, "deprecated - use src-flag")
	if err := otherNameCmd.Flags().MarkHidden("flag"); err != nil {
		panic(`unable to mark "flag" as hidden`)
	}
	otherNameCmd.Flags().BoolVar(&opts.srcFlag, "src-flag", false, "this is not deprecated either")
	rootCommand.AddCommand(otherNameCmd)
	return rootCommand, &opts
}

func main() {
	for _, t := range []struct {
		args     []string
		expected options
	}{
		{[]string{"same-name"}, options{}}, // No flags
		{[]string{"same-name", "--flag"}, options{sameNameFlag: true}},

		{[]string{"--flag", "same-name"}, options{deprecatedFlag: true}},

		{[]string{"--flag", "other-name"}, options{deprecatedFlag: true}},
		{[]string{"other-name", "--flag"}, options{deprecatedFlag: true}},
		{[]string{"other-name", "--src-flag"}, options{srcFlag: true}},
	} {
		app, opts := createApp()
		app.SetArgs(t.args)
		if err := app.Execute(); err != nil {
			panic(fmt.Sprintf("Execute failed for %#v", t.args))
		}
		if *opts != t.expected {
			panic(fmt.Sprintf("Unexpected parsing results for %#v: %#v vs. %#v", t.args, opts, t.expected))
		}
	}

	for _, t := range []struct {
		args    []string
		visible bool
	}{
		{[]string{"--help"}, false},
		{[]string{"same-name", "--help"}, true}, // Would like to have true here
		{[]string{"other-name", "--help"}, false},
	} {
		app, _ := createApp()
		stdout := bytes.Buffer{}
		app.SetOut(&stdout)
		app.SetArgs(t.args)
		if err := app.Execute(); err != nil {
			panic(fmt.Sprintf("Execute failed for %#v", t.args))
		}
		help := stdout.String()
		fmt.Printf("===\n%s\n===\n", help)
		visible := strings.Contains(help, "--flag")
		if visible != t.visible {
			panic(fmt.Sprintf("Unexpected visibility for %#v: %v vs. %v", t.args, visible, t.visible))
		}
	}
}
```

cevich:
> If someone would like to take over this PR, I'm fine with it.

oh, I know how to fix that.  I can take this over, np.

PR #1351, opened 2021-06-29
Misc man-page-checker RFCs

mtrmac:
@edsantiago A few proposed follow-ups on #1347 .

One non-trivial issue is that the macOS `sed` does not support `\U` in replacements. That doesn’t matter in CI because the CI’s macOS build target runs `validate-local` _before_ building `bin/skopeo`, so the `--help` consistency checks are skipped, but it’s a hassle for local development where a `bin/skopeo` is almost always around:
```console
% hack/man-page-checker 
skopeo copy:              man='Usource-image Udestination-image' help='SOURCE-IMAGE DESTINATION-IMAGE'
skopeo delete:            man='Uimage-name' help='IMAGE-NAME'
…
```

For now this PR just carelessly compares case-insensitively, to have _something_ to fall back on if doing something better turned out to be too difficult.

### DIFF
--- a/hack/man-page-checker
+++ b/hack/man-page-checker
@@ -99,6 +99,9 @@ function compare_usage() {
     from_man=$(sed  -E -e 's/^\[\*options\*\][[:space:]]*//' <<<"$from_man")
     from_help=$(sed -E -e 's/^\[command options\][[:space:]]*//'      <<<"$from_help")
 
+    # Constant strings in man page are '**foo**', in --help are 'foo'.
+    from_man=$(sed -E -e 's/\*\*([^*]+)\*\*/\1/g' <<<"$from_man")
+
     # Args in man page are '*foo*', in --help are 'FOO'. Convert all to
     # UPCASE simply because it stands out better to the eye.
     from_man=$(sed -E -e 's/_([a-z-]+)_/\U\1/g' <<<"$from_man")
@@ -121,7 +124,7 @@ for md in *.1.md;do
     # arguments are bracketed by single ones.
     #   E.g. '**skopeo copy** [*options*] _..._'
     # Get the command name, and confirm that it matches the md file name.
-    cmd=$(echo "$synopsis" | sed -E -e 's/(.*)\*\*.*/\1/' | tr -d \*)
+    cmd=$(echo "$synopsis" | sed -E -e 's/^\*\*([^*]+)\*\*.*/\1/' | tr -d \*)
     # Use sed, not tr, so we only replace the first dash: we want
     # skopeo-list-tags -> "skopeo list-tags", not "skopeo list tags"
     md_nodash=$(basename "$md" .1.md | sed -e 's/-/ /')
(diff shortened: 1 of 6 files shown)


hack/man-page-checker
```diff
@@ -99,9 +99,16 @@ function compare_usage() {
...
+    # Constant strings in man page are '**foo**', in --help are 'foo'.
+    from_man=$(sed -E -e 's/\*\*([^*]+)\*\*/\1/g' <<<"$from_man")
+
+    # Args in man page are '*foo*', in --help are 'FOO'. Selectively
+    # converting to upper-case is too hard, so just convert everything
+    # to upper case below to compare case-insensitively.
+    from_man=$(sed -E -e 's/_([a-z-]+)_/\1/g' <<<"$from_man")
+
+    from_man=$(tr '[:lower:]' '[:upper:]' <<<"$from_man")
+    from_help=$(tr '[:lower:]' '[:upper:]' <<<"$from_help")
```
edsantiago, line 111: Real quick note about this: I deliberately wanted `from_help` to remain as-is, because the convention is for `--help` args to be UPPER CASE. If they are lower case, I wanted to flag that. (But it's late in the day, and I may not be fully understanding your purpose here)

mtrmac: This is almost certainly too broad - the `--src` / `--dest` literals from `skopeo sync` should not be modified.

I just wanted have _something_ so that the other commits could be at least partially verified, and I can’t think of any simple replacement for `\U`.

mtrmac: Alternatively, it might be worth it to just run this on Linux only, and let macOS users deal with CI-only failures that don’t show up locally.

edsantiago: The more I think about it, the better that sounds. I think it's a waste of our time to work around Mac bugs, especially when some of us don't even have access to Macs. And I can't imagine a scenario in which man-page validation would catch something on a Mac that wouldn't be caught on Linux.

edsantiago (approved):
Changes LGTM, but I'm more inclined toward finding a way to run this check only on Linux -- I just don't know how to do that.

mtrmac:
(Note to self: `Makefile` used to use `ifeq ($(shell uname),Darwin)` fairly recently. The Linux value is `Linux`.)

hack/man-page-checker
mtrmac: I have dropped the macOS workaround for now, leaving `make validate-local` failing on macOS, so that this doesn’t conflict with the work in #1350.

rhatdan:
LGTM
@edsantiago PTAL

edsantiago:
LGTM but I don't have /lgtm or merge privs here

mtrmac:
This was, so far, marked as a draft anyway…

I have dropped the macOS bits, #1350 is on a path to resolving that by moving the documentation checks out of `validate-local`, and then we can skip running the scripts on macOS.

PR #1353, opened 2021-06-29
[release-1.3] Bump to Skopeo v1.3.2-dev

TomSweeneyRedHat:
Bump the release branch version to the next numbered dev.

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -1,4 +1,4 @@
 package version
 
 // Version is the version of the build.
-const Version = "1.3.1"
+const Version = "1.3.2-dev"


rhatdan:
I did not think we did this on release branched only on the main branch?

mtrmac:
Ah, it’s possible that this was an intentional decision — I can’t remember such a rule, and the history is not consistent.

(We have no written policy AFAICS, at least not in the repo.)

rhatdan:
I guess this makes sense to prevent people from accidently releasing a -dev branch
LGTM

PR #1358, opened 2021-07-01
BROKEN try to filter out --tls-verify from skopeo --help

mtrmac:
Apparently that part works, but (skopeo inspect --tls-verify) doesn't show up...


mtrmac:
Dropping the global `--tls-verify` option does cause the option to show up in `skopeo inspect --help`. But that’s not what we want…

PR #1363, opened 2021-07-07
disable `completion` command

vrothberg:
Disable the implicit `completion` command that cobra 1.2.x added.

### DIFF
--- a/cmd/skopeo/main.go
+++ b/cmd/skopeo/main.go
@@ -102,6 +102,11 @@ func main() {
 		return
 	}
 	rootCmd, _ := createApp()
+	// Currently, skopeo uses manually written completions.  Cobra allows
+	// for auto-generating completions for various shells.  Podman is
+	// already making us of that.  If Skopeo decides to follow, please
+	// remove the line below (and hide the `completion` command).
+	rootCmd.CompletionOptions.DisableDefaultCmd = true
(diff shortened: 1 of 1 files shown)


cmd/skopeo/main.go
```diff
@@ -102,6 +102,8 @@ func main() {
 		return
 	}
 	rootCmd, _ := createApp()
+	// Disable the implicit `completion` command in cobra.
```
mtrmac, line 105: (I’m not sure this comment helps much. Pointing at the existing `completions/bash/skopeo` script would at least document the rationale.)

mtrmac (commented):
ACK, this makes sense short-term. Wouldn’t it be nice to use the native multi-shell support instead of manually (not) maintaining a bash-only completion script, eventually?

[Either way it would be esthetically nice to hide the completion command from `--help`, especially now that Ed is working on a script to enforce man page / `--help` correspondence; and I can’t see any way to do that right now, other than disabling it completely.]

cmd/skopeo/main.go
vrothberg: Better?

mtrmac: Sure. Thanks!

mtrmac (approved):

PR #1364, opened 2021-07-07
CONTRIBUTING: update vendoring instructions

moio:
:wave: skopeo maintainers!

This updates the contribution instructions to use go modules vendoring instead of `vndr`.

I hope I interpreted the intention correctly as I stumbled into this while trying to open my first PR (trying to do that according to documentation!).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -117,29 +117,31 @@ commit automatically with `git commit -s`.
 
 ### Dependencies management
 
-Make sure [`vndr`](https://github.com/LK4D4/vndr) is installed.
+Dependencies are managed via [standard go modules](https://golang.org/ref/mod).
 
 In order to add a new dependency to this project:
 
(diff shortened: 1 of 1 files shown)


mtrmac (commented):
Thanks!

Everything in this PR looks correct; since you are working in this area (and are looking at it with fresh eyes), would you care to change the instructions further to use the Go tools? I think that’s `go get`, `go get -u`, and more importantly `go mod edit -replace` and `go mod edit -dropreplace`; in particular, IIRC `go mod edit -replace …=…@latest` would automatically figure out the right date/commit number so that it doesn’t have to be computed manually.

moio:
Thanks @mtrmac, all good points. I hope the text is OK now!

rhatdan:
Could you squash your commits.

moio:
> Could you squash your commits.

Done

mtrmac (approved):
Thanks again!

PR #1367, opened 2021-07-09
Revert "integration tests: disable `ls` for logs"

mtrmac:
This reverts commit 65c5b0bf8d391ab19c86c0e7d337c457369ecba5.

We never diagnosed the cause for #1093; but now that we are on a different CI implementation, let’s just blindly try to see if it happens to work now.


vrothberg (commented):
LGTM, thank you!

PR #1368, opened 2021-07-09
Bump main version to v1.4.0-dev

lsm5:

@vrothberg @mtrmac @rhatdan PTAL. v1.3.1 is already released.

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -1,4 +1,4 @@
 package version
 
 // Version is the version of the build.
-const Version = "1.3.1-dev"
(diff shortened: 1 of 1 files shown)


vrothberg:
Since we have a `1.3` branch, I think that the next version but from the `main` branch should be 1.4.0.

@mtrmac WDYT?

mtrmac:
Yeah, let’s not make it that easy to release 1.3.x from two different branches.

mtrmac (approved):

PR #1369, opened 2021-07-09
Fix --tls-verify

mtrmac:
- Make the --help presence match deprecation status
- Warn on (skopeo --tls-verify command) again

See https://github.com/containers/skopeo/pull/1350#discussion_r667184279 for past discussion

<s>WIP, this needs unit tests and much more documentation.</s>

@edsantiago I think this works, but I’d very much appreciate a second pair of eyes.

### DIFF
--- a/cmd/skopeo/main.go
+++ b/cmd/skopeo/main.go
@@ -50,6 +50,12 @@ func createApp() (*cobra.Command, *globalOptions) {
 		// already making us of that.  If Skopeo decides to follow, please
 		// remove the line below (and hide the `completion` command).
 		CompletionOptions: cobra.CompletionOptions{DisableDefaultCmd: true},
+		// This is documented to parse "local" (non-PersistentFlags) flags of parent commands before
+		// running subcommands and handling their options. We don't really run into such cases,
+		// because all of our flags on rootCommand are in PersistentFlags, except for the deprecated --tls-verify;
+		// in that case we need TraverseChildren so that we can distinguish between
+		// (skopeo --tls-verify inspect) (causes a warning) and (skopeo inspect --tls-verify) (no warning).
+		TraverseChildren: true,
 	}
 	if gitCommit != "" {
 		rootCommand.Version = fmt.Sprintf("%s commit: %s", version.Version, gitCommit)
@@ -60,8 +66,6 @@ func createApp() (*cobra.Command, *globalOptions) {
 	var dummyVersion bool
 	rootCommand.Flags().BoolVarP(&dummyVersion, "version", "v", false, "Version for Skopeo")
 	rootCommand.PersistentFlags().BoolVar(&opts.debug, "debug", false, "enable debug output")
-	flag := optionalBoolFlag(rootCommand.PersistentFlags(), &opts.tlsVerify, "tls-verify", "Require HTTPS and verify certificates when accessing the registry")
-	flag.Hidden = true
 	rootCommand.PersistentFlags().StringVar(&opts.policyPath, "policy", "", "Path to a trust policy file")
 	rootCommand.PersistentFlags().BoolVar(&opts.insecurePolicy, "insecure-policy", false, "run the tool without any policy check")
 	rootCommand.PersistentFlags().StringVar(&opts.registriesDirPath, "registries.d", "", "use registry configuration files in `DIR` (e.g. for container signature storage)")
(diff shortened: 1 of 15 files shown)


cmd/skopeo/manifest.go
```diff
@@ -11,17 +11,23 @@ import (
...
 
 type manifestDigestOptions struct {
+	deprecatedTLSVerify *deprecatedTLSVerifyOption
 }
 
 func manifestDigestCmd() *cobra.Command {
-	var opts manifestDigestOptions
+	deprecatedTLSVerifyFlags, deprecatedTLSVerifyOpt := deprecatedTLSVerifyFlags()
+	opts := manifestDigestOptions{
+		deprecatedTLSVerify: deprecatedTLSVerifyOpt,
```
mtrmac, line 20: BTW it makes so little sense to accept `skopeo manifest-digest --tls-verify` and the like, that I’m very tempted to not add these compatibility-only flags and just break those users.

cmd/skopeo/main.go
```diff
@@ -50,6 +50,12 @@ func createApp() (*cobra.Command, *globalOptions) {
 		// already making us of that.  If Skopeo decides to follow, please
 		// remove the line below (and hide the `completion` command).
 		CompletionOptions: cobra.CompletionOptions{DisableDefaultCmd: true},
+		// This is documented to parse "local" (non-PersistentFlags) flags of parent commands before
+		// running subcommands and handling their options. We don't really run into such cases,
+		// because all of our flags on rootCommand are in PersistenFlags, except for the deprecated --tls-verify;
```
mtrmac, line 55:
```suggestion
		// because all of our flags on rootCommand are in PersistentFlags, except for the deprecated --tls-verify;
```

cmd/skopeo/manifest.go
edsantiago: I don't think it was ever documented in the `manifest` man page. Nor `copy` nor `sync`, where it also makes no sense. I see no reason to support it in any of those.

edsantiago:
I'm really sorry, I started to review this last week but it's way over my head. All I can do is compile and see what `--help` looks like, and run my man page xref on it. The latter works fine, it now suggests adding `--tls-verify` back to the man pages for `delete`, `inspect`, `list-tags`, and `login`. 

What I don't really like is the `--help` output. This isn't your fault, it's already broken in current skopeo, but:
```console
$ bin/skopeo inspect --help | grep tls
      --tls-verify                  require HTTPS and verify certificates when talking to the container registry or daemon (defaults to true) (default )
```
Note the duplicate 'default', one useful, one not. I don't know enough Cobra to know how to fix that, but mention it in case you do.

Anyhow, in principle, this LGTM. It has the desired effect of documenting `--tls-verify` in the `help` output where it belongs, and that makes me happy.

cmd/skopeo/manifest.go
mtrmac: I think we did have `skopeo copy --tls-verify=false` (meaning `--src-tls-verify=false && --dest-tls-verify=false`) for a time. That’s one of the reasons why the global option is deprecated, to nudge users towards only disabling TLS for one of the two registries instead of both.

But I’ll remove it from the non-registry users.

mtrmac:
@edsantiago Thanks, that’s exactly the kind of check I was looking for.

---

> Note the duplicate 'default', one useful, one not.

That is actually my fault, as the author of `optionalBoolFlag` and something I didn’t notice before. Let split that out to https://github.com/containers/skopeo/issues/1383 , it’s probably fixable but needs some investigation.

mtrmac:
This is now ready for review.
- `skopeo logout` newly accepts non-deprecated `--tls-verify`
- `skopeo manifest-digest` and the signing commands now don’t accept the `--tls-verify` subcommand option and fail (but still accept the global one with a warning)
- The visibility of `--tls-verify` in `--help`, and warning on use, should otherwise work as expected.

rhatdan:
LGTM
@vrothberg PTAL

vrothberg (commented):
Changes LGTM

But since we're on it, can we tackle the empty default value `(default )`?
>      --tls-verify        require HTTPS and verify certificates when accessing the registry (default )

mtrmac:
> But since we're on it, can we tackle the empty default value `(default )`?

That’s filed separately as https://github.com/containers/skopeo/issues/1383 ; it might be a simple change but it’s not obviously trivial. I can work on that now if you think it is a blocker.

vrothberg:
> > But since we're on it, can we tackle the empty default value `(default )`?
> 
> That’s filed separately as #1383 ; it might be a simple change but it’s not obviously trivial. I can work on that now if you think it is a blocker.

Thanks! Not a blocker at all, let's get it in.

edsantiago:
Belated LGTM :grin:

PR #1370, opened 2021-07-09
Set cobra.Command.CompletionOption already in createApp

mtrmac:
... because our unit tests use `createApp`, so the current `main()`-only edit is not visible to unit tests.

Cc: @vrothberg , I’m afraid I have missed this originally.


rhatdan:
LGTM

vrothberg (approved):
LGTM, thank you

PR #1374, opened 2021-07-12
Implement manpage checker in CI

cevich:
Taking over https://github.com/containers/skopeo/pull/1350

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -65,6 +65,15 @@ validate_task:
         make validate-local
         make vendor && hack/tree_status.sh
 
+doccheck_task:
+    only_if: $CIRRUS_PR != ''
+    depends_on:
+      - validate
+    container:  *build_container
+    script: |
+      "${GOSRC}/${SCRIPT_BASE}/runner.sh" setup
+      "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" build
+      "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" doccheck
 
 osx_task:
     only_if: &not_docs $CIRRUS_CHANGE_TITLE !=~ '.*CI:DOCS.*'
@@ -185,6 +194,7 @@ success_task:
     # N/B: ALL tasks must be listed here, minus their '_task' suffix.
     depends_on:
         - validate
+        - doccheck
         - osx
         - cross
         - test_skopeo
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -69,9 +69,17 @@ doccheck_task:
     only_if: $CIRRUS_PR != ''
     depends_on:
       - validate
-    container:  *build_container
+    container:
+        image: "${FEDORA_CONTAINER_FQIN}"
+        cpu: 4
+        memory: 8
+    env:
+        BUILDTAGS: &withopengpg 'btrfs_noversion libdm_no_deferred_remove containers_image_openpgp'
     script: |
-      "${GOSRC}/${SCRIPT_BASE}/runner.sh" setup
+      # TODO: Can't use 'runner.sh setup' inside container.  However,
+      # removing the pre-installed package is the only necessary step
+      # at the time of this comment.
+      dnf erase -y skopeo  # Guarantee non-interference
       "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" build
       "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" doccheck
 
@@ -110,7 +118,7 @@ cross_task:
         disk: 200
         image_name: ${FEDORA_CACHE_IMAGE_NAME}
     env:
-        BUILDTAGS: 'btrfs_noversion libdm_no_deferred_remove containers_image_openpgp'
+        BUILDTAGS: *withopengpg
     setup_script: >-
         "${GOSRC}/${SCRIPT_BASE}/runner.sh" setup
     cross_script: >-
@@ -142,15 +150,13 @@ test_skopeo_task:
               BUILDTAGS: 'btrfs_noversion libdm_no_deferred_remove'
         - name: "Skopeo Test w/ opengpg"
           env:
-              BUILDTAGS: 'btrfs_noversion libdm_no_deferred_remove containers_image_openpgp'
+              BUILDTAGS: *withopengpg
     setup_script: >-
         "${GOSRC}/${SCRIPT_BASE}/runner.sh" setup
     vendor_script: >-
         "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" vendor
     build_script: >-
         "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" build
-    validate_script: >-
-        "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" validate
     unit_script: >-
         "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" unit
     integration_script: >-
--- a/docs/skopeo-delete.1.md
+++ b/docs/skopeo-delete.1.md
@@ -19,24 +19,46 @@ $ docker exec -it registry /usr/bin/registry garbage-collect /etc/docker-distrib
 
 ```
 
+## OPTIONS
+
 **--authfile** _path_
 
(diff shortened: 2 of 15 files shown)


docs/skopeo-delete.1.md
```diff
@@ -19,24 +19,46 @@ $ docker exec -it registry /usr/bin/registry garbage-collect /etc/docker-distrib
...
+
+**--daemon-host** _host_
 
-**--creds** _username[:password]_ for accessing the registry.
+Use docker daemon host at _host_ (`docker-daemon:` transport only)
 
-**--cert-dir** _path_ Use certificates at _path_ (*.crt, *.cert, *.key) to connect to the registry.
+**--help**, **-h**
 
-**--tls-verify** _bool-value_ Require HTTPS and verify certificates when talking to container registries (defaults to true).
```
mtrmac, line 31: Just to make sure this isn’t lost, there’s been a fairly long discussion / research around the `--tls-verify` flag, https://github.com/containers/skopeo/pull/1350#discussion_r661757345 .

cevich:
*DO NOT MERGE*: Holding for https://github.com/containers/skopeo/pull/1375

docs/skopeo-delete.1.md
cevich: Oof, yeah, "long" is right.  Would you mind opening an issue about it? I feel that might be a better way to ensure the discussion isn't lost, since addressing it in this (or that #1350) PR seems onerous.

mtrmac: The thing is, this PR is mostly blocked on that discussion — without fixing `--help`, enforcing correspondence between `--help` and man pages would make the state of documentation worse.

But contrary to the bleak mood of most of that discussion, there might be a fairly clean way forward now: https://github.com/containers/skopeo/pull/1369 . It definitely needs an independent review, but looks promising.

cevich:
[Stuck on an osx failure:](https://cirrus-ci.com/task/5461578599366656)

```
export PATH=$GOPATH/bin:$PATH
brew update
remote: fatal: packfile /data/repositories/b/nw/b6/07/5c/123272362/network.git/objects/pack/pack-b33165f7e4627d75eb4db0f1f037c9cfe8722882.pack cannot be accessed        
fatal: protocol error: bad pack header
Error: Fetching /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core failed!
...cut...
You have 1 outdated formula and 2 outdated casks installed.
You can upgrade them with brew upgrade
or list them with brew outdated.
```

I'm hoping that's just some kind of flake (I already re-ran it once).  Is "update" and "upgrade" the same thing?

mtrmac:
`brew update` = upgrade Homebrew itself, download fresh metadata.

`brew upgrade` = given current metadata, upgrade any currently installed packages (but not Homebrew itself, which uses the above special mechanism)

cevich:
Hmmm, so it sounds like we should do an `update` then and `upgrade`?  Though I'm also concerned about the `remote: fatal: packfile` error, and wondering why isn't a non-zero exit thrown?  Maybe just a poorly written error message?

mtrmac:
> Hmmm, so it sounds like we should do an `update` then and `upgrade`

I think _ideally_ we should do neither: the images we deploy should be frequently refreshed with the latest Homebrew, and we would just do `brew install`. At least historically with Travis IIRC we’ve changed that with a good (and hopefully documented?) reason, I can’t recall looking into the Cirrus situation.

> Though I'm also concerned about the `remote: fatal: packfile` error, and wondering why isn't a non-zero exit thrown? Maybe just a poorly written error message?

No idea. As a very wild guess this just looks like an infrastructure issue, either at the Homebrew’s sponsor that provides hosting for downloads, or within the current Cirrus image, and especially if it’s the latter it might be one that isn’t encountered frequently enough to be handled quite correctly. Checking reports in http://github.com/homebrew/ might be informative, but it’s too late for me today.

cevich:
>  the images we deploy should be frequently refreshed with the latest Homebrew

For Mac, this is somewhat out of our control as the images are managed by Cirrus (I believe they're rebuilt/refreshed at some interval).  Same story for Windows.  The only viable alternatives for automated Mac testing involve either less control and added code duplication (github actions) or TONs more scripting and plumbing work (Cirrus + paid MacStadium account).  Basically...it would be better for humanity if Apple behaved more like Linux :grin: 

> At least historically with Travis IIRC we’ve changed that with a good (and hopefully documented?) reason, I can’t recall looking into the Cirrus situation.

Travis routinely was taking an hour to *start* running a 5-minute job, as well as severely limiting the total minutes-quota available for FOSS projects to use.  So it's not really a choice to leave for something else, we really needed to drop Travis to avoid severe impacts to development.

>No idea. As a very wild guess this just looks like an infrastructure issue

This was the case, I re-ran the job today and it passed.  Still, it's concerning that (the tool) doesn't throw a non-zero exit on a fatal error.  It's possible, thought I highly doubt Cirrus-CI is not setting +e for the shell (they're pretty careful about that stuff elsewhere).

cevich:
(rebase)

mtrmac:
I was absolutely not suggesting this as a reason to return to Travis; I just wanted to point at the https://github.com/containers/skopeo/blob/146af8cd59a995ec4483ceef25e7de4b81573a53/.travis.yml#L62-L66 comment block documenting why+what.

It’s was not immediately clear to me (OTOH I hadn’t looked into the history at all deeply when writing my previous comment) whether the Cirrus version was doing `brew update` for a reason … and it turns out there was a reason, documented in the commit message of https://github.com/containers/skopeo/pull/1329 , and I have nobody to blame but myself for this goose chase.

cevich:
>  I have nobody to blame but myself for this goose chase.

No worries, at least we can rely on git/github to keep a record of stuff of our failing memories :grin:

cevich:
I believe this is ready to go.  @edsantiago and @mtrmac PTAL.

cevich:
note-to-self: Need to copy-paste the `.cirrus.yml` changes to to the `test_skopeo` task over to `c/image`.

.cirrus.yml
```diff
@@ -133,15 +150,13 @@ test_skopeo_task:
...
-              BUILDTAGS: 'btrfs_noversion libdm_no_deferred_remove containers_image_openpgp'
+              BUILDTAGS: *withopengpg
     setup_script: >-
         "${GOSRC}/${SCRIPT_BASE}/runner.sh" setup
     vendor_script: >-
         "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" vendor
     build_script: >-
         "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" build
-    validate_script: >-
-        "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" validate
```
edsantiago, line 144: Could you also remove this then? https://github.com/containers/skopeo/blob/main/contrib/cirrus/runner.sh#L64-L66

mtrmac (commented):
The CI automation part looks plausible to me, but I only very minimally understand it and I’ll happily defer to experts.

What about the `--tls-verify` part, https://github.com/containers/skopeo/pull/1374#discussion_r668044012 ?

@edsantiago Could you see whether there are any obvious holes in https://github.com/containers/skopeo/pull/1369 that I might have missed, please? You probably understand the gotchas of Cobra better than I do.

(If that plan works, I’d be perfectly fine with merging this PR first, and taking on the work to actually update man pages in the other PR myself.)

cevich:
> but I only very minimally understand it

My overarching goal is for it to be at least somewhat approachable for everyone.  I'm not always around to fix/maintain things and am deliberately attempting to not be a bottleneck.  I realized that automation is somewhat of a niche specialization, but if I can help make it any clearer with comments or different names or links to docs or anything, please let me know.

edsantiago:
/lgtm

@mtrmac can you rebase #1369 on this, adding back the removed `--tls-verify` to manpages?

cevich:
> looks like I don't have /lgtm privs in this repo

There's no bot setup here.

mtrmac:
> > but I only very minimally understand it
> 
> My overarching goal is for it to be at least somewhat approachable for everyone. I'm not always around to fix/maintain things and am deliberately attempting to not be a bottleneck. I realized that automation is somewhat of a niche specialization, but if I can help make it any clearer with comments or different names or links to docs or anything, please let me know.

No worries; I feel reasonably confident in making targeted small changes, it’s primarily an “I don’t know what I don’t know” worry.

PR #1382, opened 2021-07-22
Fix using images from rate-limited docker hub

cevich:
The necessary images have been manually copied over to quay.  Code was
updated with centralized constants for the utilized images.  Tests then
all reference the constants (in case the image locations need to change
again).

### DIFF
--- a/integration/check_test.go
+++ b/integration/check_test.go
@@ -101,7 +101,7 @@ func (s *SkopeoSuite) TestCopyWithLocalAuth(c *check.C) {
 	assertSkopeoSucceeds(c, wanted, "login", "--tls-verify=false", "--username="+s.regV2WithAuth.username, "--password="+s.regV2WithAuth.password, s.regV2WithAuth.url)
 	// copy to private registry using local authentication
 	imageName := fmt.Sprintf("docker://%s/busybox:mine", s.regV2WithAuth.url)
-	assertSkopeoSucceeds(c, "", "copy", "--dest-tls-verify=false", "docker://docker.io/library/busybox:latest", imageName)
+	assertSkopeoSucceeds(c, "", "copy", "--dest-tls-verify=false", testFQIN+":latest", imageName)
 	// inspect from private registry
 	assertSkopeoSucceeds(c, "", "inspect", "--tls-verify=false", imageName)
 	// logout from the registry
--- a/integration/utils.go
+++ b/integration/utils.go
@@ -17,6 +17,10 @@ import (
 const skopeoBinary = "skopeo"
 const decompressDirsBinary = "./decompress-dirs.sh"
 
+const testFQIN = "docker://quay.io/libpod/busybox" // tag left off on purpose, some tests need to add a special one
+const testFQIN64 = "docker://quay.io/libpod/busybox:amd64"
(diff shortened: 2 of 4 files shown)


cevich:
@mtrmac PTAL.  I did the replacements somewhat brainlessly - I didn't scrutinize what each test was doing, mostly just guessed at what needed changing.  Since Ed already took care of the systemtests, with this PR skopeo can finally be independent from the damn Docker Hub.

integration/check_test.go
```diff
@@ -101,7 +101,7 @@ func (s *SkopeoSuite) TestCopyWithLocalAuth(c *check.C) {
 	assertSkopeoSucceeds(c, wanted, "login", "--tls-verify=false", "--username="+s.regV2WithAuth.username, "--password="+s.regV2WithAuth.password, s.regV2WithAuth.url)
 	// copy to private registry using local authentication
 	imageName := fmt.Sprintf("docker://%s/busybox:mine", s.regV2WithAuth.url)
-	assertSkopeoSucceeds(c, "", "copy", "--dest-tls-verify=false", "docker://docker.io/library/busybox:latest", imageName)
+	assertSkopeoSucceeds(c, "", "copy", "--dest-tls-verify=false", fmt.Sprintf("%s:latest", testFQIN), imageName)
```
mtrmac, line 104: (Absolutely non-blocking: Just `testFQIN+":latest"`, in these cases where `%s` is at the start, might be a tiny bit more readable. Also elsewhere.)

integration/utils.go
```diff
@@ -17,6 +17,10 @@ import (
 const skopeoBinary = "skopeo"
 const decompressDirsBinary = "./decompress-dirs.sh"
 
+const testFQIN = "docker://quay.io/libpod/busybox" // tag left off on purpose, some tests need to add a special one
+const testFQIN64 = "docker://quay.io/libpod/busybox:amd64"
+const testFQINml = "docker://quay.io/libpod/alpine_nginx:master" // multi-layer
```
mtrmac, line 22: This variable name is a bit too much of a puzzle.

`multiLayerTestFQIN`? `testFQINMultiLayer`?

mtrmac (commented):
Thanks for taking on this work. Just one nit really.

---

> I did the replacements somewhat brainlessly - I didn't scrutinize what each test was doing,

Yeah, that’s the better approach. Examining what each test was intended to do, is actually doing, eliminating duplication, and the like, leads to mad^H^H^H a much larger PR that would delay this important improvement.

integration/check_test.go
cevich: wait...I don't need to use `fmt.Sprintf()`?!?!?!

integration/utils.go
cevich: Good suggestion, okay willdo.

integration/check_test.go
cevich: Probably should have googled that first lol :tired_face:

cevich:
Okay, all fixed up, rebased, and `--force` pushed.  Fingers-crossed for good luck.

mtrmac (approved):
Thanks!

LGTM, assuming the tests pass.

Issue #1354, opened 2021-06-30
Intermittent system-tests failures


mtrmac:
e.g. https://github.com/containers/skopeo/pull/1350/checks?check_run_id=2953073576
```
not ok 12 copy: additional tag
# (from function `start_registry' in file systemtest/helpers.bash, line 359,
#  from function `setup' in test file systemtest/020-copy.bats, line 11)
#   `start_registry reg' failed
# $ podman --cgroup-manager=cgroupfs run -d --name reg -v /tmp/skopeo_bats.hJ9dA4/auth:/auth:Z -p 5000:5000 quay.io/libpod/registry:2
# c8ed7d9eae3602d4e8cc9e094dacd2526704b421192857296c7d190ecfc23987
# /var/tmp/go/src/github.com/containers/skopeo/systemtest/helpers.bash: connect: No route to host
# /var/tmp/go/src/github.com/containers/skopeo/systemtest/helpers.bash: line 355: /dev/tcp/127.0.0.1/5000: No route to host
# /var/tmp/go/src/github.com/containers/skopeo/systemtest/helpers.bash: connect: No route to host
# /var/tmp/go/src/github.com/containers/skopeo/systemtest/helpers.bash: line 355: /dev/tcp/127.0.0.1/5000: No route to host
# /var/tmp/go/src/github.com/containers/skopeo/systemtest/helpers.bash: connect: No route to host
# /var/tmp/go/src/github.com/containers/skopeo/systemtest/helpers.bash: line 355: /dev/tcp/127.0.0.1/5000: No route to host
# /var/tmp/go/src/github.com/containers/skopeo/systemtest/helpers.bash: connect: No route to host
# /var/tmp/go/src/github.com/containers/skopeo/systemtest/helpers.bash: line 355: /dev/tcp/127.0.0.1/5000: No route to host
# /var/tmp/go/src/github.com/containers/skopeo/systemtest/helpers.bash: connect: No route to host
# /var/tmp/go/src/github.com/containers/skopeo/systemtest/helpers.bash: line 355: /dev/tcp/127.0.0.1/5000: No route to host
# /var/tmp/go/src/github.com/containers/skopeo/systemtest/helpers.bash: connect: No route to host
# /var/tmp/go/src/github.com/containers/skopeo/systemtest/helpers.bash: line 355: /dev/tcp/127.0.0.1/5000: No route to host
# /var/tmp/go/src/github.com/containers/skopeo/systemtest/helpers.bash: connect: No route to host
[...]

PR #1413, opened 2021-08-11
systemtests: if registry times out, show container logs

edsantiago:
the 'signing' test is flaking; symptom is that we can never
connect to the port on the registry:

   https://api.cirrus-ci.com/v1/task/6208385738604544/logs/system.log

By all indications, the registry is up, i.e., the 'podman rm -f reg'
in teardown() succeeds, as shown by the 53c (CID) in the log. (It
bothers me that the FAIL message from die() does not appear in the
log, and I can't figure out why).

To try to diagnose this, run 'podman logs' on the registry upon
failure.

### DIFF
--- a/systemtest/helpers.bash
+++ b/systemtest/helpers.bash
@@ -359,6 +359,7 @@ start_registry() {
         timeout=$(expr $timeout - 1)
         sleep 1
     done
+    log_and_run $PODMAN logs $name
     die "Timed out waiting for registry container to respond on :$port"
 }
 


mtrmac (approved):
Sure.

I’m really wondering what “no route to host” `127.0.0.1` is supposed to mean, probably something deep in the container networking implementation I know nothing about.

mtrmac:
Thanks!

#1354 links some other discussion, for the record.

PR #1415, opened 2021-08-11
[release-1.4] Bump c/storage 1.34.0 and c/common 0.43.0

TomSweeneyRedHat:
As the titles says, bumping c/storage, c/image, and c/common in the next
vendor dance to get ready for RHEL 8.5

[NO TESTS NEEDED]


mtrmac (approved):
LGTM.

To confirm, we are not updating c/image on the 5.15 branch with a dependency update to c/storage, relying instead of the top-level tools to update their c/storage dependencies? Less work is always nice, I just want to make sure this is intended.

Do we want to tag a release immediately afterwards?

---

> [NO TESTS NEEDED]

I don’t see that this is a riskless change where testing is not necessary. Luckily that annotation does nothing here :)

TomSweeneyRedHat:
FWIW, the [NO TESTS NEEDED] just turns off the requirement of adding a test for the change in the PR, all existing tests are still run.

I'd not thought about c/image getting bumped too, I probably should look into that.

Yes, I'm planning on creating at Skopeo v1.4.1 once this gets merged in.

@mtrmac thanks!  As usual, an excellent review.

mtrmac:
> FWIW, the [NO TESTS NEEDED] just turns off the requirement of adding a test for the change in the PR, all existing tests are still run.

Ah, my mistake. Either way, that bot is not enabled on this repo (yet?), to my knowledge.

> I'd not thought about c/image getting bumped too, I probably should look into that.

Note that the dependency graph goes c/storage → c/image → c/common; so bumping image might imply one more c/common bump. (Without a c/common bump, we end up in the same situation of the final versions brought in by the top-level executables, except we would be bringing c/image instead of c/storage.)

I’m personally quite fine with not bumping c/image, but I don’t know what the conventions/trade-offs for vendor dance consistency are; I’ll completely defer to you, feel free to merge this or change it.

TomSweeneyRedHat:
Added vendoring for c/image v5/15.1 too

mtrmac (approved):
LGTM.

TomSweeneyRedHat:
And it looks like we'll need to bump both c/storage and c/image one more time, the dance continues.

PR #1420, opened 2021-08-17
[CI:DOCS] Add OWNERS file

rhatdan:

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,0 +1,17 @@
+approvers:
+  - mtrmac
+  - lsm5
+  - TomSweeneyRedHat
+  - rhatdan
+  - vrothberg
+reviewers:
+  - ashley-cui
+  - giuseppe
+  - containers/image-maintainers
+  - lsm5
+  - mtrmac
+  - QiWang19
+  - rhatdan
+  - runcom
+  - TomSweeneyRedHat
+  - vrothberg


vrothberg (approved):
LGTM

mtrmac (commented):
Does this do anything right now? Or is it a prerequisite for turning on a bot in the future?

TomSweeneyRedHat:
@mtrmac @jberkus is putting together some thoughts on Governance for the Containers organization.  The first step is to have an owners.md file in each of the containers/* projects.  I don't know if a bot will be involved or not for sure, I don't believe so.  It's primarily being used so the future Governance document, likely in c/common, can say "Look at the owners.md file for a project ...".  Not so much for this project, but for the smaller ones it would be handy to have such a file in case there's ever a CVE reported against it.

TomSweeneyRedHat:
I'm confused why a PR with only a new file would ever need a rebase?
LGTM

rhatdan:
This is for Governance,  We will be opening similar PRs for other REPOS.

mtrmac:
Thanks for the explanation.

---

> I'm confused why a PR with only a new file would ever need a rebase?

“Does it _need_ a rebase” is a different question, and GitHub indicates that one by “has conflicts that needs to be resolved”.

I think it’s just doing a graph shape check; that answers “did the submitter (get a chance to) test this on top of the latest version” — a zero-conflict PR could still be broken by other changes to the codebase; so some projects have a policy of asking the submitters to test against a recent version. Of course that doesn’t scale past some amount of changes.

Anyway, humans with judgement about such drift risk can use a big hammer to merge anyway, so let’s do that.

PR #1428, opened 2021-08-23
Update non-module dependencies

mtrmac:
Dependabot was apparently not picking these up (and several haven't had a release for a long time anyway).

Also move from `github.com/go-check/check` to its newly declared (and `go.mod`-enforced) name `gopkg.in/check.v1`.


vrothberg (commented):
LGTM

Issue #1383, opened 2021-07-23
Help of --tls-verify has `(default` twice


mtrmac:
`
      --tls-verify                  require HTTPS and verify certificates when talking to the container registry or daemon (defaults to true) (default )
`

It seems that this can right now be fixed by removing `optionalBoolValue.IsBoolFlag` (and the existing tests pass), but everything about this is brittle.

We should at least look into the history to understand why `IsBoolFlag` was required and isn’t required any more.

PR #1431, opened 2021-08-26
Remove the extra (defaults to true) help msg

rhatdan:
By default skopeo checks to see if the user actually uses one of the
--*tls-verify flags. Their initial value is ignored.  Setting the
initial value to false causes Cobra to not display the default value on
the screen when the user runs a `skopeo --help` command.

If the user does not specify a --*tls-verify option, it falls back to
using the value specified in the registries.conf file.

Fixes: #1383


rhatdan:
@mtrmac @vrothberg PTAL

lsm5 (commented):
LGTM, maybe needs a rebase or could be a github problem.

@rhatdan Nit RE: commit verification. It says `Unverified` against your commit probably because you haven't uploaded your GPG to github.

PR #1444, opened 2021-09-14
Update VM Images + Drop prior-ubuntu references

cevich:
These images contain a workaround for:
     https://github.com/containers/podman/issues/11123

Prior-Ubuntu support is being dropped everywhere.

Ref: https://github.com/containers/podman/issues/11070
     https://github.com/containers/automation_images/pull/88

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -26,20 +26,17 @@ env:
     FEDORA_NAME: "fedora-34"
     PRIOR_FEDORA_NAME: "fedora-33"
     UBUNTU_NAME: "ubuntu-2104"
-    PRIOR_UBUNTU_NAME: "ubuntu-2010"
 
     # Google-cloud VM Images
-    IMAGE_SUFFIX: "c6724387953967104"
+    IMAGE_SUFFIX: "c6431352024203264"
     FEDORA_CACHE_IMAGE_NAME: "fedora-${IMAGE_SUFFIX}"
     PRIOR_FEDORA_CACHE_IMAGE_NAME: "prior-fedora-${IMAGE_SUFFIX}"
     UBUNTU_CACHE_IMAGE_NAME: "ubuntu-${IMAGE_SUFFIX}"
-    PRIOR_UBUNTU_CACHE_IMAGE_NAME: "prior-ubuntu-${IMAGE_SUFFIX}"
 
     # Container FQIN's
     FEDORA_CONTAINER_FQIN: "quay.io/libpod/fedora_podman:${IMAGE_SUFFIX}"
     PRIOR_FEDORA_CONTAINER_FQIN: "quay.io/libpod/prior-fedora_podman:${IMAGE_SUFFIX}"
     UBUNTU_CONTAINER_FQIN: "quay.io/libpod/ubuntu_podman:${IMAGE_SUFFIX}"
-    PRIOR_UBUNTU_CONTAINER_FQIN: "quay.io/libpod/prior-ubuntu_podman:${IMAGE_SUFFIX}"
 
     # Built along with the standard PR-based workflow in c/automation_images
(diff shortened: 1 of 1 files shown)


cevich:
@mtrmac any idea [what this means](https://cirrus-ci.com/task/5632555316674560) and if/how I can fix it:

```
---> Making bundle: validate-vet
go build github.com/containers/storage/pkg/devicemapper: invalid flag in pkg-config --cflags: /usr/bin/sh:
make: *** [Makefile:231: validate-local] Error 1
```

mtrmac:
> go build github.com/containers/storage/pkg/devicemapper: invalid flag in pkg-config --cflags: /usr/bin/sh:

Two guesses:
- `pkg-config` missing (I see no reason to assume that)
- The `/bin/sh: warning: setlocale: LC_ALL: cannot change locale (en_US): No such file or directory` things that are all over the log leaking into `pkg-config` output. (I see no reason to assume that either, but it seems more likely to happen unintentionally, e.g. by `fedora:latest` updating to a new release.)

Maybe see what `pkg-config --cflags devmapper` outputs in that environment.

---

BTW the “cannot change locale” warnings worry me somewhat; the system can probably mostly work in such an environment, but it just screams “untested error handling paths” to me. Would it be possible to fix that? I don’t very much care whether it is by changing the locale choice or by adding the locale data, though it might turn out to matter in practice.

cevich:
> Maybe see what pkg-config --cflags devmapper outputs in that environment.

Thanks for the hint, that is something I can check...

> but it just screams “untested error handling paths” to me. Would it be possible to fix that?

Yeah I noticed that too, and it seems new.  I think the first step to a fix is understanding the cause...for which I have less than zero insight.  Do you have any idea where that error is coming from?  I cannot find references to `setlocate` anywhere under `hack/`.

cevich:
...Oh right, so `make validate` runs inside the container.  That's where (and maybe why?) we see the locale errors.  This is totally reproducible locally, it doesn't require CI:

```bash
$ make validate
podman run --rm -i -e TESTFLAGS="-v -check.v -test.timeout=15m" -e CI= -e SKOPEO_CONTAINER_TESTS=1 -t --security-opt label=disable -v /home/cevich/devel/skopeo:/src/github.com/containers/skopeo -w /src/github.com/containers/skopeo quay.io/libpod/skopeo_cidev:c6431352024203264 make validate-local
Trying to pull quay.io/libpod/skopeo_cidev:c6431352024203264...
Copying blob ecfb9899f4ce done
...cut...
/bin/sh: warning: setlocale: LC_ALL: cannot change locale (en_US): No such file or directory
...cut...
$ podman run -it --rm quay.io/libpod/skopeo_cidev:c6431352024203264
bash: warning: setlocale: LC_ALL: cannot change locale (en_US): No such file or directory
/usr/bin/sh: warning: setlocale: LC_ALL: cannot change locale (en_US): No such file or directory
/usr/bin/sh: warning: setlocale: LC_ALL: cannot change locale (en_US): No such file or directory
/usr/bin/sh: warning: setlocale: LC_ALL: cannot change locale (en_US): No such file or directory
[root@76cb0cc7dcaf /]# cat /etc/os-release
NAME=Fedora
VERSION="34 (Container Image)"
ID=fedora
VERSION_ID=34
VERSION_CODENAME=""
PLATFORM_ID="platform:f34"
...
[root@76cb0cc7dcaf /]# pkg-config --cflags devmapper

```

cevich:
Ahh ha! I found it.  I set this up in the `Containerfile` to make perl happy, apparently that broke everything else.  Working on a manual fix.

cevich:
...okay, I fixed this in a really hacky/not-recommended way.  I built a new image with a fix, then manually overwrote the tag and shoved it up into quay (after confirming the localle errors are gone).  Re-running the tests now.

rhatdan:
LGTM

PR #1445, opened 2021-09-15
Shuffle sync order

wzshiming:


rhatdan:
Please improve your PR description.

mtrmac:
Yeah, this feels like some kind of partial attempt to run the sync process in parallel, invoking separate `skopeo sync` processes that are unaware of each other.

I guess we can move `skopeo sync` in that direction if there is interest, but it seems much closer to a correct/well-controlled process to have a single Go process spawn off Goroutines, and assign work to _reliably_ avoid re-copying exactly the same repo. (Doing that in a _fully optimized_ way, only copying a particular layer once, even across images, does seem like a stretch with the current c/image/copy .) That would work better than unexpected randomness just to decrease, not eliminate, inefficiency when invoked by a shell script (which would everyone still have to write and maintain).

cmd/skopeo/sync.go
```diff
@@ -417,6 +418,15 @@ func imagesToCopyFromRegistry(registryName string, cfg registrySyncConfig, sourc
 // encountered.
 func imagesToCopy(source string, transport string, sourceCtx *types.SystemContext) ([]repoDescriptor, error) {
 	var descriptors []repoDescriptor
+	defer func() {
```
mtrmac, line 421: (I also don’t think using `defer` is helpful for understanding the code structure in this case, but that’s a side point when the whole idea currently seems problematic.)

wzshiming:
@mtrmac


> Yeah, this feels like some kind of partial attempt to run the sync process in parallel, invoking separate skopeo sync processes that are unaware of each other.

Yeah, I use `Github Action` to synchronize some images periodically, so **I think this patch is still necessary**
https://github.com/mirrorshub/cr/blob/main/.github/workflows/sync.yml


---

> I guess we can move skopeo sync in that direction if there is interest, but it seems much closer to a correct/well-controlled process to have a single Go process spawn off Goroutines, and assign work to reliably avoid re-copying exactly the same repo. 

I agree to use goroutines to perform synchronization in one process at the same time. But how to display the synchronization log is more friendly, I think it is worth discussing.  may I send a new PR to work on this 

---

> (Doing that in a fully optimized way, only copying a particular layer once, even across images, does seem like a stretch with the current c/image/copy .) That would work better than unexpected randomness just to decrease, not eliminate, inefficiency when invoked by a shell script (which would everyone still have to write and maintain).

https://github.com/containers/skopeo/blob/177443f47d2a0f2f0e5223cb3c8eecb1ad9b7e83/vendor/github.com/containers/image/v5/docker/docker_image_dest.go#L214-L238
There is already logic to avoid repeated uploading of layers

rhatdan:
@vrothberg has a PR or Issue about supporting parallell pulls at the same time, where only one process moves forward, where as the others wait.  But that has never moved up the priority list.

mtrmac:
> Yeah, I use `Github Action` to synchronize some images periodically, multiple actions are uploading the same blob, so **I think this patch is still necessary**

For what reason exactly? How does the existing code fail? Is it just that there are multiple concurrent attempts to upload the same blob, which works but is wasteful?

My naive reading of the above is that the GitHub action is running so frequently / there is such a large blob, that uploading a single blob takes longer than the frequency of the GitHub action; in that case, it seems reasonable to just decrease the frequency, because no matter what else happens, the sync has the long single-blob upload on a critical path, which seems to run into Amdahl’s law.

Or are there a large number of same-is blobs, or something like that?

---

Overall I’m not too enthusiastic about introducing randomness in `skopeo sync`. It makes things much harder to debug. _Maybe_, if the situation really is that special, it could be enabled by a command-line option, but I’d prefer to investigate other ways to achieve the high-level result first.

---

> @vrothberg has a PR or Issue about supporting parallell pulls at the same time, where only one process moves forward, where as the others wait.

Note that that involves _pulls_ to shared c/storage, which also implicitly provides a space for maintaining the logs.

Here, it’s so far unclear to me whether the goal is to have a _single_ sync complete faster (i.e. multiple parallel copies, and potentially locks/reuse within a _single process_), or to somehow support a rapid-fire sequence of syncs, perhaps even within containerized processes that have no shared state or other mechanisms that could be used to maintain a lock.

PR #1450, opened 2021-09-17
[release-1.3] Cirrus: Disable OSX task

cevich:
Release-branches infrequently change, but the OSX VM images constantly
change.  Disable this test to avoid encountering flakes.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -65,6 +65,7 @@ validate_task:
 
 
 cross_task:
+    only_if: 1 == 0
     macos_instance:
         image: catalina-xcode
     setup_script: |


lsm5:
@mtrmac @vrothberg do we need to update 1.3 branch at all ?

.cirrus.yml
```diff
@@ -65,6 +65,7 @@ validate_task:
 
 
 cross_task:
+    only_if: $CI != $CI
```
mtrmac, line 68: Is this intended to mean `false`, or is it something more subtle?

mtrmac:
> @mtrmac @vrothberg do we need to update 1.3 branch at all ?

That’s a good question. There’s probably a mapping of maintained RPM releases vs. upstream release branches somewhere, but I don’t know where that is.

.cirrus.yml
cevich: -sigh- no, I asked them about this a very long time ago.  Cirrus-CI cannot spell `false` :cry: 

If it's any clearer, we could do something like `only_if: 0 == 1`

cevich:
> There’s probably a mapping of maintained RPM releases

So long as those RPMs are outside of any support commitments, they will never need to be built again.  Since the solution to any problems is: upgrade to a newer version.  Unf. I'm also unaware of where/how/who uses this branch downstream :confused: 

Maybe @TomSweeneyRedHat or @lsm5 knows?  

Ref: In podman-CI-land, we essentially have three branch/CI states:
* Fully active and supported downstream packages and/or lots of code churn:  Fully active CI Testing
* Still supported packages, but almost zero / very low code churn: Partially active CI, check the code compiles and skip all actual testing.
* Fully dead/unsupported and zero code changes: Fully disabled/inactivated CI.  Nothing runs.

.cirrus.yml
mtrmac: \*shrug\* this works fine. I’m probably going to forget and keep asking anyway (it’s quite possible that I have asked about this sometime in the past, at least once, already).

It might be worth a comment, in principle, but this is a backport from `main`, so being consistent with the condition already there is at least as important, and this discussion already serves as a record.

rhatdan:
LGTM

PR #1449, opened 2021-09-17
Cirrus: Disable OSX task

cevich:
Release-branches infrequently change, but the OSX VM images constantly
change.  Disable this test to avoid encountering flakes.

Signed-off-by: Chris Evich <cevich@redhat.com>

PR #1446, opened 2021-09-16
release-1.4: Run (gofmt -s -w)

mtrmac:
This is a backport of #1427 for the release-1.4 branch.

Marked as draft to ensure there is consensus on this (vs. pinning an earlier Go version).

---

Go 1.17 introduces a much more reasonable build constraint format, and gofmt now fails without using it.

Sadly we still need the old format as well, to support <1.17 builds.

Signed-off-by: Miloslav Trmač <mitr@redhat.com>

PR #1454, opened 2021-09-20
[release-1.4] pin go

mtrmac:
1.17 has changed the expected `gofmt` format, and we don't want to follow such changes on the stable branch.

Then re-enable the `osx` task disabled in #1449.

An alternative to #1446 .

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -90,12 +90,13 @@ osx_task:
     macos_instance:
         image: catalina-xcode
     setup_script: |
-        export PATH=$GOPATH/bin:$PATH
+        # /usr/local/opt/go@1.16 will be populated by (brew install go@1.16) below
+        export PATH=$GOPATH/bin:/usr/local/opt/go@1.16/bin:$PATH
         brew update
-        brew install gpgme go go-md2man
+        brew install gpgme go@1.16 go-md2man
         go get -u golang.org/x/lint/golint
     test_script: |
-        export PATH=$GOPATH/bin:$PATH
+        export PATH=$GOPATH/bin:/usr/local/opt/go@1.16/bin:$PATH
         go version
         go env
         make validate-local test-unit-local bin/skopeo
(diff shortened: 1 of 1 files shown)


vrothberg (commented):
LGTM

mtrmac:
@cevich PTAL, compare with current state post-#1449 , and #1446.

I’ll defer entirely to your judgment as to which approach is the best; keeping the current state with `osx` disabled would be perfectly fine with me.

cevich:
> keeping the current state with osx disabled would be perfectly fine with me.

I'm leaning this way, only from the perspective that few-jobs == less fail chance (the osx job also flakes due to homebrew repo. and/or networking problems).  Just to confirm, am I correct believing these release branches see very little (if any) changes over time?

mtrmac:
> Just to confirm, am I correct believing these release branches see very little (if any) changes over time?

Historically that’s been very much the case. Right now the `release-1.4` activity feels a little higher than usual to me, but that’s been mostly build-system changes, not features/security fixes driving the purpose of the branch. So I do still expect to see very few, if any, changes, in the future.

cevich:
Oh I was going to say let's merge the other one, oh well, not a biggie.  We can revert-revert the osx disablement later if needed :grinning:

mtrmac:
@cevich So just to confirm, would you like the `release-1.4` branch to remain as is, and #1446 to be dropped? Or to revert this and return to just disabling the test? Something else, like make the test pass using one or both methods, but disable it anyway?

I don’t really have a preference, I’m just looking for a way to have this matter decided, for now at least.

cevich:
The PR was merged, so I became comfortable just ignoring it, now you've gone and ruined all my fun! :smile: 

Yes I think eventually (months/years) we'll probably end up disabling the OSX test again.  If in the meanwhile #1446 happens, it's fine / won't hurt anything, and may help avoid a flake or three.  I wouldn't suggest spending any more than 5-minutes on it though.

cevich:
Clarification: meaning apply #1446 and leave the test enabled for now.

mtrmac:
It’s 3 clicks to merge it, so I did that. Thanks!

cevich:
And thank you :smiley:

PR #1459, opened 2021-09-24
Add dry-run mode to skopeo-sync

### LINKED ISSUE #1175: sync with a yaml src should have a "dry-run" or validate mode
We keep a yaml file checked in for performing syncs, but it doesn't seem like there is a good way to either locally or in some automated way verify that nothing was fat-fingered. For example, if the source yaml looked something like:

```yaml
gcr.io:
  images:
    distroless/static-debian10:
      - latest
      - nonroot
```

It would be nice to be able to make sure that the relevant repository[1] exists, and in the case where there are explicit tags listed, it would be nice to make sure that they also exist. Today this is possible by just syncing to a local dir, but that requires going through and actually downloading everything. I was hoping for something a little lighter - possibly only skipping the actual image download.

Would something like this be possible? It might make it possible for us to wire up a CI check that could execute fast, and wouldn't need somewhere to dump tons of data (or need to download it in the first place). Basically anything that would help with improving confidence in editing the yaml file without needing to do the actual sync would be appreciated.

[1] I hope I'm not swapping words. Repository and registry get swapped all the time in my mind.

twexler:
I found this useful when testing `skopeo-sync`, instead of actually copying to the destination, `skopeo-sync` will output a message like 
```
time="2021-09-24T14:06:25-04:00" level=info msg="Would have copied image, but `--dry` is specified" from="docker://registry-1.docker.io/prom/prometheus:v2.9.0" to="docker://my-cool-docker-repo.local/registry-1.docker.io/prom/prometheus:v2.9.0"
```
Fixes #1175.


rhatdan:
@mtrmac @vrothberg PTAL

docs/skopeo-sync.1.md
```diff
@@ -50,6 +50,10 @@ Path of the authentication file for the source registry. Uses path given by `--a
 
 Path of the authentication file for the destination registry. Uses path given by `--authfile`, if not provided.
 
+**--dry**
+
+Run the sync without actually copying data to the destination
```
TomSweeneyRedHat, line 55:
```suggestion
Run the sync without actually copying data to the destination.
```

TomSweeneyRedHat:
@twexler thanks for bringing this forward.  I think this is indeed useful, but I think I'd prefer it if the code went a bit further and touched the registry and perhaps the destination to see if it was available and if the auths were OK.  A bigger effort for sure.

twexler:
@TomSweeneyRedHat Gotcha. I'll take a look at this again tomorrow, see if I can fit that into this PR with minimal headache...

vrothberg:
> @twexler thanks for bringing this forward. I think this is indeed useful, but I think I'd prefer it if the code went a bit further and touched the registry and perhaps the destination to see if it was available and if the auths were OK. A bigger effort for sure.

I am not sure that will work.  Let's assume we want to sync an image to `registry.com/repo/image:tag`.  The presence/credential checks would only work if `registry.com/repo/image:tag` is already present. I do not know of a reliable way to figure out if an image exists or not since most registries will return an auth error if an image doesn't exist:
```
~ $ podman pull quay.io/example312123/foo:latest
Trying to pull quay.io/example312123/foo:latest...
Error: initializing source docker://quay.io/example312123/foo:latest: reading manifest latest in quay.io/example312123/foo: unauthorized: access to the requested resource is not authorized
```

So I don't think that such checks should be performed in a dry run.

twexler:
@vrothberg You _could_ hit the [Version Check API](https://docs.docker.com/registry/spec/api/#api-version-check) on the registry to confirm it's a registry and potentially authn/z would play into that, but you're right: If that API isn't authenticated on the specific registry implementation, then we're only verifying that the defined repository is _probably_ a registry.

vrothberg (commented):
In the current form I find the logs too redundant:
```
INFO[0001] Would have copied image, but `--dry` is specified  from="docker://busybox:ubuntu" to="dir:/tmp/test/busybox:ubuntu"
INFO[0001] Copying image ref 173/179                     from="docker://busybox:ubuntu-12.04" to="dir:/tmp/test/busybox:ubuntu-12.04"
INFO[0001] Would have copied image, but `--dry` is specified  from="docker://busybox:ubuntu-12.04" to="dir:/tmp/test/busybox:ubuntu-12.04"
INFO[0001] Copying image ref 174/179                     from="docker://busybox:ubuntu-14.04" to="dir:/tmp/test/busybox:ubuntu-14.04"
INFO[0001] Would have copied image, but `--dry` is specified  from="docker://busybox:ubuntu-14.04" to="dir:/tmp/test/busybox:ubuntu-14.04"
INFO[0001] Copying image ref 175/179                     from="docker://busybox:uclibc" to="dir:/tmp/test/busybox:uclibc"
INFO[0001] Would have copied image, but `--dry` is specified  from="docker://busybox:uclibc" to="dir:/tmp/test/busybox:uclibc"
INFO[0001] Copying image ref 176/179                     from="docker://busybox:unstable" to="dir:/tmp/test/busybox:unstable"
INFO[0001] Would have copied image, but `--dry` is specified  from="docker://busybox:unstable" to="dir:/tmp/test/busybox:unstable"
INFO[0001] Copying image ref 177/179                     from="docker://busybox:unstable-glibc" to="dir:/tmp/test/busybox:unstable-glibc"
INFO[0001] Would have copied image, but `--dry` is specified  from="docker://busybox:unstable-glibc" to="dir:/tmp/test/busybox:unstable-glibc"
INFO[0001] Copying image ref 178/179                     from="docker://busybox:unstable-musl" to="dir:/tmp/test/busybox:unstable-musl"
INFO[0001] Would have copied image, but `--dry` is specified  from="docker://busybox:unstable-musl" to="dir:/tmp/test/busybox:unstable-musl"
INFO[0001] Copying image ref 179/179                     from="docker://busybox:unstable-uclibc" to="dir:/tmp/test/busybox:unstable-uclibc"
```
I think that the `Copying image ref 173/179 ...` logs shouldn't be printed for `--dry`.

@rhatdan PTAL

vrothberg:
Other than that, LGTM.

cmd/skopeo/sync.go
```diff
@@ -104,6 +105,7 @@ See skopeo-sync(1) for details.
 	flags.StringVarP(&opts.destination, "dest", "d", "", "DESTINATION transport type")
 	flags.BoolVar(&opts.scoped, "scoped", false, "Images at DESTINATION are prefix using the full source image path as scope")
 	flags.BoolVarP(&opts.all, "all", "a", false, "Copy all images if SOURCE-IMAGE is a list")
+	flags.BoolVar(&opts.dry, "dry", false, "Run without actually copying data")
```
mtrmac, line 108: Maybe `"dry-run"`? (Disclaimer: not a native speaker); AFAIK the idiom is “dry run”, and “dry“ does not usually have the “test-only” meaning outside of that idiom.

(I know, naming is hard and colors of the bike-shed…)

cmd/skopeo/sync.go
```diff
@@ -604,19 +606,22 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) error {
...
+			if opts.dry {
+				logrus.WithFields(fromToFields).Info("Would have copied image, but `--dry` is specified")
+			} else {
+				if err = retry.RetryIfNecessary(ctx, func() error {
+					_, err = copy.Image(ctx, policyContext, destRef, ref, &options)
+					return err
+				}, opts.retryOpts); err != nil {
+					return errors.Wrapf(err, "Error copying ref %q", transports.ImageName(ref))
+				}
+				imagesNumber++
```
mtrmac, line 623: Should this be not counted in the dry-run mode? AFAICS this will now say “Synced 0 images from $nonzero sources”, which is a bit strange.

mtrmac (commented):
> In the current form I find the logs too redundant:
> 
> ```
> INFO[0001] Would have copied image, but `--dry` is specified  from="docker://busybox:unstable-musl" to="dir:/tmp/test/busybox:unstable-musl"
> INFO[0001] Copying image ref 179/179                     from="docker://busybox:unstable-uclibc" to="dir:/tmp/test/busybox:unstable-uclibc"
> ```
> 
> I think that the `Copying image ref 173/179 ...` logs shouldn't be printed for `--dry`.

Also I don’t think every line needs to repeat that “--dry is specified`. Maybe just replace the `Copying image ref` message with `Would try copying image ref` in the dry-run mode?

mtrmac:
> @vrothberg You _could_ hit the [API Version Check endpoint](https://docs.docker.com/registry/spec/api/#api-version-check) on the registry to confirm it's a registry and potentially authn/z would play into that, but you're right: If that API isn't authenticated on the specific registry implementation, then we're only verifying that the defined repository is _probably_ a registry which is a worthwhile check, I think.

Yeah, I don’t think we currently know how to do this. IIRC on _some_ registries, the “get token for $username+$password accessing $scope” endpoint actually does fail if the credentials are valid but not allowed to use $scope (e.g. write to a specific repo), but on others this succeeds (and accessing the “version check” would work just fine) and only fails when actually trying to do an operation (i.e. actually trying write something to that repo). So we don’t know whether credentials for writing are valid until we actually try to write something, which is very incompatible with a “dry-run” concept.

Having a mode that tests the source enumeration / allows validation of a YAML with regexes and the like, is still quite valuable without testing credentials.

twexler:
> Having a mode that tests the source enumeration / allows validation of a YAML with regexes and the like, is still quite valuable without testing credentials.

This mode does test this, as it compiles the images to copy before any of the new code comes into play.

cmd/skopeo/sync.go
twexler: I'd argue it's nice to be explicit here of what _would_ have happened, but did not.

cmd/skopeo/sync.go
twexler: I agree with you, this makes total sense.

mtrmac:
Yup, I just wanted to say that I don’t thing not testing credentials is a reason to reject or delay accepting this PR.

cmd/skopeo/sync.go
mtrmac: So maybe change the summary message to say “Would have synced … images”? Or alternatively drop the `--dry-run` reference from all the other messages, and move it into this summary message?

(I know, tinkering with details, and I’m not even really around this week. Just things to possibly consider. If others think this is fine to merge as-is, that’s perfectly fine with me.)

twexler: I think I just addressed the second approach in 7635815. Let me know what you think!

docs/skopeo-sync.1.md
```diff
@@ -50,6 +50,10 @@ Path of the authentication file for the source registry. Uses path given by `--a
 
 Path of the authentication file for the destination registry. Uses path given by `--authfile`, if not provided.
 
+**--dry**
```
mtrmac, line 53: Please update this to `dry-run` as well.

cmd/skopeo/sync.go
```diff
@@ -39,6 +39,7 @@ type syncOptions struct {
 	destination         string         // Destination registry name
 	scoped              bool           // When true, namespace copied images at destination using the source repository name
 	all                 bool           // Copy all of the images if an image in the source is a list
+	dry                 bool           // Don't actually copy anything, just output what it would have done
```
mtrmac, line 42: `dryRun`?

mtrmac (commented):
ACK overall, but I’ll let others to try this out and review in detail, for this week.

cmd/skopeo/sync.go
```diff
@@ -604,19 +610,22 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) error {
...
-
-			if err = retry.RetryIfNecessary(ctx, func() error {
-				_, err = copy.Image(ctx, policyContext, destRef, ref, &options)
-				return err
-			}, opts.retryOpts); err != nil {
-				return errors.Wrapf(err, "Error copying ref %q", transports.ImageName(ref))
 			}
-			imagesNumber++
+			if opts.dry {
+				logrus.WithFields(fromToFields).Info("Would have copied image ref %d/%d", counter+1, len(srcRepo.ImageRefs))
```
mtrmac, line 618:
```suggestion
				logrus.WithFields(fromToFields).Infof("Would have copied image ref %d/%d", counter+1, len(srcRepo.ImageRefs))
```
per the failing tests

twexler:
I'll squash these commits once code review is complete.

cmd/skopeo/sync.go
mtrmac: I was thinking a warning at the end, because that’s right above the next CLI prompt; a warning at the beginning will scroll away in interactive use. It _is_ a good place for warnings in non interactive iterations, because it will be at the top of a log file, but I don’t think that’s our use: it makes little sense to me to have a `--dry-run` in something like `cron`, it seems likely to be used interactively when tuning the configuration.

Something like `In --dry-run mode, didn't sync %d images from %d sources` ?

OTOH I don’t feel strongly about this; what do others think?

twexler: I could see both ways. I'm happy to add an additional mention of dry-run at the end, to be thorough.

cmd/skopeo/sync.go
```diff
@@ -606,30 +613,37 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) error {
...
 			}
 			imagesNumber++
 		}
 	}
 
-	logrus.Infof("Synced %d images from %d sources", imagesNumber, len(srcRepoList))
+	logMsg := fmt.Sprintf("Synced %d images from %d sources", imagesNumber, len(srcRepoList))
+	if opts.dryRun {
+		logMsg = fmt.Sprintf("Would have %s", logMsg)
+	}
```
mtrmac, line 645: Something like
```suggestion
	if !opts.dryRun {
		logrus.Infof("Synced %d images from %d sources", imagesNumber, len(srcRepoList))
	} else {
		logrus.Infof("Would have synced %d images from %d sources", imagesNumber, len(srcRepoList))
	}
```

so that the capitalization is correct, please.

docs/skopeo-sync.1.md
```diff
@@ -50,6 +50,10 @@ Path of the authentication file for the source registry. Uses path given by `--a
 
 Path of the authentication file for the destination registry. Uses path given by `--authfile`, if not provided.
 
+**--dry-run**
```
mtrmac, line 53: Please add the flag to `completions/bash/skopeo` as well.

mtrmac (commented):
Looks good, some last nits. Also, please rebase the PR.

jackivanov:
Any progress on this?

mfriedenhagen:
Hi @twexler, it would be nice if you could rebase and integrate the remarks :-)

Issue #715, opened 2019-09-10
Documentation does not mention Windows binary of Skopeo.


tandeday:
A quick read of the documentation did not immediately reveal how I got a Windows (10 if that matters) binary of skopeo for me to use against our OpenShift cluster.  I do not have Go experience so I didn't just go ahead and build it.

It would be nice if the documentation clearly stated:

* If a binary is available for download (probably isn't).  
* How to create one if possible? (along with a difficulty estimage).
* If not possible within reason, then explicitly mention that.

Thanks

/Thorbjørn

vrothberg:
Hi @m86194, thanks for opening this issue.

We don't offer binaries for any OS. Skopeo is (build) tested against Linux and Mac OS but I am not sure how far we can come on Windows. I don't have access to a Windows machine for testing.

@mtrmac @rhatdan WDYT?

mtrmac:
We do test Windows cross-compilation of c/image, so it is probably buildable, but I wouldn’t be at all surprised if Skopeo were a pretty bad Windows citizen (WRT configuration file paths and the like), and not extremely surprised if it were broken due to different path syntax or the like (though there was at least one Windows-based user of the c/image library at one point).

`make binary-local` would be the default way to try to build it, adding some `BUILDTAGS` might be necessary.

Frankly, making Windows work, and work smoothly, and having everything well documented, would be mostly dependent on someone interested contributing that. We are definitely interested, at least enough not to sabotage that :) , but lack both access and time to do it ourselves.

Same for providing pre-built binaries; on Linux it is more or less an intentional choice not to, because a single binary is not portable to various Linuxes; on Windows it might be more practical, but again dependent on someone integrating that into the current build system.

xphyr:
SO, it very much seems possible. I was able to build a Windows binary just this evening (completely unrelated to @m86194 request). I made a small tweak to the Makefile (created a Windows build target) and it does compile. I will test it out tomorrow at work on a Windows machine and see if its usable. If you are curious about the addition to the Makefile take a look at my fork of this project.

m86194:
First of all, please notice I am not requesting a windows build (which would be nice, though) but that the _documentation_ is clearer about what can be expected from a new user coming to the project.  (I need a dockerless image mirrorer) because it is slowly showing up in google searches for this kind of utility.

[...]

Issue #563, opened 2018-10-18
Support MacOS keychain for ca's


MansM:
As a skopeo user within an enterprise I want to use skopeo
on my local mac talking to a registry from the company running an corporate ca signed cert
in order to do skopeo inspect/copy/etc 

the error message I got:
 x509: certificate signed by unknown authority

mtrmac:
Thanks for your report.

The docker/distribution client, at least, uses the system’s CAs as provided by Golang’s `x509.SystemCertPool`; AFAICS that does use the OS configuration on macOS (using `SecTrustSettingsCopyCertificates` per `src/crypto/x509/root_cgo_darwin.go`) — Go does not have its own list of CAs  it seems, and `skopeo` definitely does not have a separate one. So, something else might be going on.

Please provide _precise_ instructions how to reproduce this, and include the output of `skopeo --debug $your_command`.

mtrmac:
Thanks; that uses the default Golang TLS configuration, with only https://github.com/containers/image/blob/bd10b1b53b2976f215b3f2f848fb8e7cad779aeb/docker/docker_client.go#L137 changed; that doesn’t look like it should change things.

MansM:
ehm. It could be me (probably), but I have no idea what you are trying to say.  
Should something be fixed in skopeo or in docker_client?

mtrmac:
That’s still unknown; the discrepancy between the certificates you have configured in the macOS keychain and the certificates used by the Golang TLS stack needs to be debugged.

Honestly I’m completely unfamiliar with the macOS keychain API (or, really, debugging any macOS native code), so I’m afraid I’m not likely to be that much help with this, in the very near future at least. https://github.com/containers/skopeo/issues/563#issuecomment-431338006 is all the pointers I can provide for that right now.

Maybe if you can find some other Golang-implemented TLS client that does work, without extra configuration, with the CA trusted by the keychain, that could serve as an indication what needs to be done. But, again, AFAICT the Golang standard library intends to have the keychain roots of trust work by default, and `skopeo` does not do anything explicit to break that, so it seems likely that something non-trivial is going on.

rhatdan:
@MansM Is this still something you are interested in?  Could you help out with a PR?  This is not something a bunch of Linux programmers are likely to fix, so we need help from someone with Mac Experience.

MansM:
I wish I could, but I lack the knowledge on MacOS internals and go code skills 

[...]

PR #1464, opened 2021-09-28
Update installation doc with latest steps

lsm5:
Update installation doc with latest steps

- Remove Kubic repo suggestions where skopeo exists by default
- Include documentation about lack of Windows package
(RE: https://github.com/containers/skopeo/issues/715)

@vrothberg @mtrmac @rhatdan @TomSweeneyRedHat @lbscorpio PTAL

### DIFF
--- a/install.md
+++ b/install.md
@@ -15,29 +15,6 @@ sudo dnf -y install skopeo
 sudo dnf -y install skopeo
 ```
 
-Newer Skopeo releases may be available on the repositories provided by the
-Kubic project. Beware, these may not be suitable for production environments.
-
-on CentOS 8:
-
-```sh
-sudo dnf -y module disable container-tools
-sudo dnf -y install 'dnf-command(copr)'
-sudo dnf -y copr enable rhcontainerbot/container-selinux
-sudo curl -L -o /etc/yum.repos.d/devel:kubic:libcontainers:stable.repo https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/CentOS_8/devel:kubic:libcontainers:stable.repo
-sudo dnf -y install skopeo
-```
-
-on CentOS 8 Stream:
-
-```sh
-sudo dnf -y module disable container-tools
-sudo dnf -y install 'dnf-command(copr)'
-sudo dnf -y copr enable rhcontainerbot/container-selinux
-sudo curl -L -o /etc/yum.repos.d/devel:kubic:libcontainers:stable.repo https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/CentOS_8_Stream/devel:kubic:libcontainers:stable.repo
-sudo dnf -y install skopeo
-```
-
 ### RHEL/CentOS ≤ 7.x
 
 ```sh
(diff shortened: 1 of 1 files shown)


install.md
```diff
@@ -62,19 +39,23 @@ sudo apk add skopeo
 brew install skopeo
 ```
 
+You may need to add registry certificate to your macOS truststore.
+```
+sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain <registry_cert.crt>
```
mtrmac, line 44: This part doesn’t make sense to me.

- AFAICS this has nothing to do with installing Skopeo; just the `brew install` does that.
- WRT configuring a certificate trusted to connect to a registry, platform-agnostic paths like `--cert-dir` are still expected to work
- All of #563 ended up surprising to me. The original report was that that CAs from the macOS keychain were ignored, and the recommended fix is to … add the CA to the macOS keychain? Stated this way, it just does not follow at all, so there must be some subtlety to #563 that wasn’t written down.

Overall I think this part of the PR should just be dropped, and #563 either discussed further, or just closed as “2 years later, this was reported to work fine”.

mtrmac (commented):
ACK other than the macOS keychain part.

Thanks!

install.md
lsm5: ack, will do. thanks @mtrmac , and apologies if this bothered you on your PTO.

mtrmac: (No worries — I chose to take a little time to see what’s new, and I know I had the option not to. But I appreciate the thought all the same.)

lsm5:
removed macos part. PTAL

lsm5:
Also, added a note for Windows package absence. PTAL as well.

lsm5:
@vrothberg @rhatdan @TomSweeneyRedHat tests are green. PTAL

rhatdan:
LGTM

install.md
```diff
@@ -97,17 +73,8 @@ sudo apt-get -y update
...
-Checkout the [Kubic project page](https://build.opensuse.org/package/show/devel:kubic:libcontainers:stable/skopeo)
-for a list of supported Ubuntu version and
-architecture combinations. **NOTE:** The command `sudo apt-get -y upgrade`
-maybe required in some cases if Skopeo cannot be installed without it.
-The build sources for the Kubic packages can be found [here](https://gitlab.com/rhcontainerbot/skopeo/-/tree/debian/debian).
-
-CAUTION: On Ubuntu 20.10 and newer, we highly recommend you use Buildah, Podman and Skopeo ONLY from EITHER the Kubic repo
-OR the official Ubuntu repos. Mixing and matching may lead to unpredictable situations including installation conflicts.
+The [Kubic project](https://build.opensuse.org/package/show/devel:kubic:libcontainers:stable/skopeo)
+provides packages for active Ubuntu 20.04 (it should also work with direct derivatives like Pop!\_OS).
```
vrothberg, line 77:
```suggestion
provides packages for Ubuntu 20.04 (it should also work with direct derivatives like Pop!\_OS).
```

Not sure what "active" means.

vrothberg (commented):
I think that Ubuntu should get a dedicated section along with the info that skopeo is the main repositories starting with 20.10

install.md
TomSweeneyRedHat: is it just 20.24 or "20.24 and later"?

lsm5: > Not sure what "active" means.

@vrothberg ha, `active` was a poor choice, I wanted to mean something that's not EOL, and yes, we could just get rid of that word.



> is it just 20.24 or "20.24 and later"?

@TomSweeneyRedHat in the case of podman, we're not mentioning kubic repos if the package exists in the default distro repos. I'm hoping to do the same here.

lsm5:
> I think that Ubuntu should get a dedicated section along with the info that skopeo is the main repositories starting with 20.10

Ubuntu already has a dedicated section. Listed below debian and raspberry pi os. Can you check again and let me know if you meant something else?

vrothberg (commented):
LGTM, thanks!

vrothberg:
> > I think that Ubuntu should get a dedicated section along with the info that skopeo is the main repositories starting with 20.10
> 
> Ubuntu already has a dedicated section. Listed below debian and raspberry pi os. Can you check again and let me know if you meant something else?

Apologies, it looked like it was in the Debian section.

Issue #1466, opened 2021-09-30
skopeo sync aborts if any of the image tags has issues


jaikiran:
We use skopeo to mirror an external registry into our internal registry. The process of mirroring involves, our tool deciding which image to `sync`. A particular image has a large number of tags in the external registry from which we are syncing. For whatever reason, it appears that at least one such published tag of the image is corrupt, because when we run the following `sync` command, it results into an error:

```
sync -s docker -d docker --dest-no-creds --dest-tls-verify=false --src-tls-verify=false --src
-creds=user:pass foobarserver:5000/helloworld 192.168.1.2:5000
```
The error is as follows:

```
INFO[0010] Tag presence check                            imagename="foobarserver:5000/helloworld" tagged=false
INFO[0010] Getting tags                                  image="foobarserver:5000/helloworld"
INFO[0010] Copying image ref 1/157                       from="docker://foobarserver:5000/helloworld:release-i-v0.9.13" to="docker://192.168.1.2:5000/helloworld:release-i-v0.9.13"
Getting image source signatures
Skipping: image already present at destination
INFO[0010] Copying image ref 2/157                       from="docker://foobarserver:5000/helloworld:release_test_tag_i" to="docker://192.168.1.2:5000/helloworld:release_test_tag_i"
Getting image source signatures
Copying blob 1479d35649b2 skipped: already exists  
Copying blob b84b226fd20e skipped: already exists  
Copying blob 17dea04ab044 skipped: already exists  
Copying blob 1fd3936b9903 skipped: already exists  
Copying blob 67dc9985f582 skipped: already exists  
Copying blob 591b1660a002 skipped: already exists  
Copying blob c36ff2885a3f skipped: already exists  
Copying blob 3ac493f8fd0e [======================================] 403.5MiB / 403.7MiB
[...]

PR #1468, opened 2021-10-02
Introduce a --ignore option to allow "sync" command to continue syncing even after a particular image sync fails

jaikiran:
Fixes https://github.com/containers/skopeo/issues/1466

As noted in the linked issue, in the current form, the `sync` command aborts as soon as any image tag has issue during copy. This ends up being a blocker if any one of the image tag out of hundreds (for a particular image) has some issue.

As suggested in the linked issue, the commit in this PR introduces a `--ignore` option to the `sync` command which allows the `sync` to continue with the rest of the images/tags. When an error occurs during the image copy, that error is logged and when the sync finally finishes, such errors contribute to `sync` being marked as failed.

I have tested this change against our setup where we were running into this issue. With this change, when `--ignore` is used, the error is now logged and the next image tag is processed and finally when the process exits you see the failure message and `sync` returns with a non-zero exit code. This is the snippet from such a run:

(there are a total to 352 tags identified for this image and the 154th one is the one having the issue)
```
....
INFO[0017] Copying image ref 154/352                     from="docker://dockerserver:5000/helloworld:release-1.0.0-SNAPSHOT-2019_Jun_18_08_58_UTC" to="docker://192.168.1.2:5000/helloworld:release-1.0.0-SNAPSHOT-2019_Jun_18_08_58_UTC"
Copying blob fd48b0e2a9b9 skipped: already exists  
Copying blob 1fd3936b9903 skipped: already exists  
Copying blob a87126e1a510 skipped: already exists  
Copying blob 17dea04ab044 skipped: already exists  
Copying blob 44403e1d5128 [==================================>---] 66.9MiB / 72.4MiB
ERRO[0018] Error copying ref "docker://dockerserver:5000/helloworld:release-1.0.0-SNAPSHOT-2019_Jun_18_08_58_UTC"  error="writing blob: Patch http://192.168.1.2:5000/v2/helloworld/blobs/uploads/6fdba9fa-9367-4218-bccd-d33f38d7a833?_state=Pdq1xGqTyt00iB4TRvx8xypLgbyEj03Qnyi7PNSUfg57Ik5hbWUiOiJpZGVudGl0eSIsIlVVSUQiOiI2ZmRiYTlmYS05MzY3LTQyMTgtYmNjZC1kMzNmMzhkN2E4MzMiLCJPZmZzZXQiOjAsIlN0YXJ0ZWRBdCI6IjIwMjEtMTAtMDJUMDQ6MDc6MjAuMzE2NDYwNDhaIn0%3D: readfrom tcp 192.168.1.2:56274->192.168.1.2:5000: happened during read: Digest did not match, expected sha256:44403e1d5128211c7dcf598e85b608190560dc0ffac8d9df7f9ea1b270b60c12, got sha256:6b375acbd82c7ee62a88c06a4be0e1a7e4d0ba03450a792d8a932d7b3aa52fab"
```
That error gets reported and the process now moves on to next:
```
...
INFO[0018] Copying image ref 155/352                     from="docker://dockerserver:5000/helloworld:release_1.0.0s_Jun18_13_40_UTC" to="docker://192.168.1.2:5000/helloworld:release_1.0.0s_Jun18_13_40_UTC"
...
```
After the rest are synced, the final log message looks like:
```
INFO[0532] Synced 351 images from 1 sources             
FATA[0532] Sync failed due to previous reported error(s) for one or more images 
```
The exit code from that run is verified to be non-zero:
```
 echo $?
1
```

I thought of adding a test case, but it's not straightforward to trigger the error in first place (I can't yet think of a way). So I haven't added one.

P.S: This is my first time with Go. Is there any checkstyle/formatting rules that I can configure my IDE with? Everytime I did a certain change, it ended up formatting a huge part of the file where it removed certain whitespaces in various places. It's currently uses the IDE default formatter (I use GoLand from IntelliJ).

### DIFF
--- a/cmd/skopeo/sync.go
+++ b/cmd/skopeo/sync.go
@@ -39,6 +39,7 @@ type syncOptions struct {
 	destination         string         // Destination registry name
 	scoped              bool           // When true, namespace copied images at destination using the source repository name
 	all                 bool           // Copy all of the images if an image in the source is a list
+	keepGoing           bool           // Whether or not to abort the sync if there are any errors during syncing the images
 }
 
 // repoDescriptor contains information of a single repository used as a sync source.
@@ -104,6 +105,7 @@ See skopeo-sync(1) for details.
 	flags.StringVarP(&opts.destination, "dest", "d", "", "DESTINATION transport type")
 	flags.BoolVar(&opts.scoped, "scoped", false, "Images at DESTINATION are prefix using the full source image path as scope")
 	flags.BoolVarP(&opts.all, "all", "a", false, "Copy all images if SOURCE-IMAGE is a list")
+	flags.BoolVarP(&opts.keepGoing, "keep-going", "", false, "Do not abort the sync if any image copy fails")
 	flags.AddFlagSet(&sharedFlags)
 	flags.AddFlagSet(&deprecatedTLSVerifyFlags)
 	flags.AddFlagSet(&srcFlags)
@@ -568,7 +570,6 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) error {
 		return err
 	}
 
-	imagesNumber := 0
 	options := copy.Options{
 		RemoveSignatures:                      opts.removeSignatures,
 		SignBy:                                opts.signByFingerprint,
@@ -578,7 +579,8 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) error {
 		OptimizeDestinationImageAlreadyExists: true,
 		ForceManifestMIMEType:                 manifestType,
 	}
-
+	errorsPresent := false
+	imagesNumber := 0
 	for _, srcRepo := range srcRepoList {
 		options.SourceCtx = srcRepo.Context
 		for counter, ref := range srcRepo.ImageRefs {
@@ -614,12 +616,22 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) error {
 				_, err = copy.Image(ctx, policyContext, destRef, ref, &options)
 				return err
 			}, opts.retryOpts); err != nil {
-				return errors.Wrapf(err, "Error copying ref %q", transports.ImageName(ref))
+				if !opts.keepGoing {
+					return errors.Wrapf(err, "Error copying ref %q", transports.ImageName(ref))
+				}
+				// log the error, keep a note that there was a failure and move on to the next
+				// image ref
+				errorsPresent = true
+				logrus.WithError(err).Errorf("Error copying ref %q", transports.ImageName(ref))
+				continue
 			}
 			imagesNumber++
 		}
 	}
 
 	logrus.Infof("Synced %d images from %d sources", imagesNumber, len(srcRepoList))
-	return nil
+	if !errorsPresent {
+		return nil
+	}
+	return errors.New("Sync failed due to previous reported error(s) for one or more images")
 }
--- a/docs/skopeo-sync.1.md
+++ b/docs/skopeo-sync.1.md
@@ -88,6 +88,9 @@ Print usage statement.
 
 **--retry-times**  the number of times to retry, retry wait time will be exponentially increased based on the number of failed attempts.
 
+**--keep-going**
+If any errors occur during copying of images, those errors are logged and the process continues syncing rest of the images and finally fails at the end.
+
 ## EXAMPLES
 
 ### Synchronizing to a local directory
--- a/completions/bash/skopeo
+++ b/completions/bash/skopeo
@@ -97,6 +97,7 @@ _skopeo_sync() {
     --scoped
     --src-no-creds
     --src-tls-verify
+    --keep-going
     "
 
     local transports


cmd/skopeo/sync.go
```diff
@@ -104,6 +105,7 @@ See skopeo-sync(1) for details.
 	flags.StringVarP(&opts.destination, "dest", "d", "", "DESTINATION transport type")
 	flags.BoolVar(&opts.scoped, "scoped", false, "Images at DESTINATION are prefix using the full source image path as scope")
 	flags.BoolVarP(&opts.all, "all", "a", false, "Copy all images if SOURCE-IMAGE is a list")
+	flags.BoolVarP(&opts.continueOnError, "ignore", "", false, "Abort the sync if any image ref sync failed")
```
mtrmac, line 108: (Absolutely non-blocking: _After_ the option name settles, consider whether  the `continueOnError` field should be named similarly to the option. OTOH `continueOnError` is a pretty great internal name, it might be better than a short-hand external option name.)

docs/skopeo-sync.1.md
```diff
@@ -88,6 +88,9 @@ Print usage statement.
 
 **--retry-times**  the number of times to retry, retry wait time will be exponentially increased based on the number of failed attempts.
 
+**--ignore**
```
mtrmac, line 91: Please update `completions/bash/skopeo` as well.

cmd/skopeo/sync.go
```diff
@@ -578,7 +580,7 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) error {
 		OptimizeDestinationImageAlreadyExists: true,
 		ForceManifestMIMEType:                 manifestType,
 	}
-
+	errorsPresent := false
```
mtrmac, line 583: Non-blocking: the `imagesNumber` and `errorsPresent` values have the same lifetime and conceptual purpose, it would be nice for them to be initialized/updated/used together.

I.e., in this case, move the `imagesNumber := 0` initialization above just next to this newly added line.

mtrmac (commented):
Thanks for the PR!

(See https://github.com/containers/skopeo/issues/1466#issuecomment-932783234 , this might not be the best solution for that particular situation; let’s discuss that in there. This PR is a useful feature anyway, e.g. to make as much progress as possible on unreliable network links.)

The implementation looks good; as usual, naming is the hard part :)

> P.S: This is my first time with Go. Is there any checkstyle/formatting rules that I can configure my IDE with? Everytime I did a certain change, it ended up formatting a huge part of the file where it removed certain whitespaces in various places. It's currently uses the IDE default formatter (I use GoLand from IntelliJ).

We use the language’s `gofmt` with the `-s` option, IIRC currently using the Go 1.17 version. I don’t know how to set that up in GoLand, maybe someone else can help. Anyway, the PR tests do a formatting check and this PR looks good as is already.

docs/skopeo-sync.1.md
jaikiran: Done. Updated the PR with this change.

cmd/skopeo/sync.go
jaikiran: Done in latest update to this PR.

cmd/skopeo/sync.go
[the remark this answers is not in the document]
vrothberg: The description of the flag suggests the inverse.  Shouldn't it be `"Do not abort [...]"` given the default value is `false`?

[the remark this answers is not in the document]
vrothberg: The Podman CLI uses `--ignore` in a number of commands but the "what" is pretty clear there. For instance, `podman rm --ignore $CONTAINER` won't error if the container doesn't exist (anymore).

In that regard `--ignore` would make sense if we're trying to copy an image that doesn't exist and make that non-fatal.

`--continue-on-error` seems to cover more than that including transient network errors.

In other words, I concur with @mtrmac :^)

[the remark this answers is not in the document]
rhatdan: I would prefer --keep-going, less characters.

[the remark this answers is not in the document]
jaikiran: Updated the PR to use --keep-going

jaikiran: Now that the PR uses --keep-going, I think calling that internal variable `keepGoing` instead of `continueOnError` is slightly better. So I've updated the PR to rename this internal member.

cmd/skopeo/sync.go
```diff
@@ -104,6 +105,7 @@ See skopeo-sync(1) for details.
 	flags.StringVarP(&opts.destination, "dest", "d", "", "DESTINATION transport type")
 	flags.BoolVar(&opts.scoped, "scoped", false, "Images at DESTINATION are prefix using the full source image path as scope")
 	flags.BoolVarP(&opts.all, "all", "a", false, "Copy all images if SOURCE-IMAGE is a list")
+	flags.BoolVarP(&opts.keepGoing, "keep-going", "", false, "Do not abort the sync if any image ref sync failed")
```
mtrmac, line 108:
```suggestion
	flags.BoolVarP(&opts.keepGoing, "keep-going", "", false, "Do not abort the sync if any image sync failed")
```
the `ImageReference` type is a Go-level concept users don’t know (and don’t need to care) about.

Maybe “image copy” instead of “image sync”, to more clearly differentiate between “the sync” and “image sync”.

(And should that be “failed” or “fails”? I’m not a native speaker.)

---

Well, the existing output does say “ref %d/%d”… I guess I’d prefer for this new option description not to introduce another instance, and to keep the progress output as is in this PR (changing the progress output text is not what this PR is about; we can do that later separately).

(Or, any other reviewers: feel free to override me and merge as is.)

mtrmac (commented):
LGTM, one tiny wording nit.

(Or, others, feel free to merge as is.)

cmd/skopeo/sync.go
jaikiran: Done. I've updated this PR to use "image copy" and also changed "failed" to "fails" (I'm not a native English speaker either, but I think fails sounds more correct).

mtrmac (approved):
LGTM. Thanks again!

jaikiran:
Thanks everyone for the reviews and the inputs.

@mtrmac, do releases happen off the main branch? If so what versions get released off that branch? The reason I ask is to decide whether I should create a PR for the 1.4 release branch containing this change so that it becomes available to us whenever the next release happens.

rhatdan:
Releases happen off of the main branch.

mtrmac:
… and not on an extremely regular/frequent schedule, so, @jaikiran , if you need an “official” release to happen pretty soon, please speak up and we’ll tag one.

jaikiran:
Hello @mtrmac, this particular fix and the Makefile change that was merged some days back, both, will help us in our usage of skopeo in our product. So yes, a release will be really helpful. Thank you very much.

Issue #527, opened 2018-07-23
Allow supplying credentials with colons in usernames


grdryn:
I'm trying to use `skopeo` in a Jenkins pipeline on OpenShift, and copy an image from the built-in registry (using credentials for a ServiceAccount) to an external registry, but I get the following error saying there's an invalid username/password:

```
[...]

PR #1470, opened 2021-10-04
Introduce --username and --password to pass credentials

jaikiran:
The commit in this PR introduces new options to pass username and password for credentials while interacting with registries. This was requested as an enhancement in https://github.com/containers/skopeo/issues/527


rhatdan:
LGTM

lsm5 (approved):
LGTM

@mtrmac @vrothberg PTAL

cmd/skopeo/utils.go
```diff
@@ -180,12 +184,23 @@ func (opts *imageOptions) newSystemContext() (*types.SystemContext, error) {
 	if opts.credsOption.present && opts.noCreds {
```
mtrmac, line 184: Shouldn’t this now check for `opts.credsOption.present || opts.userName.present) && opts.noCreds` (with an appropriate error message update)?

Or in some other way ensure that of the three options, exactly one can be set.

cmd/skopeo/utils.go
```diff
@@ -180,12 +184,23 @@ func (opts *imageOptions) newSystemContext() (*types.SystemContext, error) {
...
 		return nil, errors.New("creds and no-creds cannot be specified at the same time")
 	}
+	if opts.credsOption.present && opts.userName.present {
+		return nil, errors.New("creds and username cannot be specified at the same time")
+	}
+	if opts.password.present && !opts.userName.present {
```
mtrmac, line 190: Should `--username` without `--password` be accepted? I don’t have a _strong_ opinion, but…

cmd/skopeo/utils.go
```diff
@@ -180,12 +184,23 @@ func (opts *imageOptions) newSystemContext() (*types.SystemContext, error) {
...
 	if opts.credsOption.present {
 		var err error
 		ctx.DockerAuthConfig, err = getDockerAuth(opts.credsOption.value)
 		if err != nil {
 			return nil, err
 		}
+	} else if opts.userName.present {
+		ctx.DockerAuthConfig = &types.DockerAuthConfig{
+			Username: opts.userName.value,
+			Password: opts.password.value,
```
mtrmac, line 202: … strictly speaking `opts.password.value` is not valid to use `if !opts.password.present`; so if the case of `--username` without `--password` should be accepted, this should probably add an explicit condition.

It just seems simpler to me to require `opts.userName.present == opts.password.present`, and if someone really has an account with an empty password, force them to use `--password ''` explicitly. That is simpler for us to implement, and even for the users it generalizes better.

cmd/skopeo/utils_test.go
```diff
@@ -197,6 +197,47 @@ func TestImageDestOptionsNewSystemContext(t *testing.T) {
 	assert.Error(t, err)
 }
 
+func TestImageOptionsUsernamePassword(t *testing.T) {
```
mtrmac, line 200: Non-blocking: Could this be turned into a table-driven test to reduce repetition?

Something approximately like
```go
for _, c := range []struct{
    args []string
    expected *types.DockerAuthConfig // data to expect, or nil if an error is expected
} {
    // test cases
} {
    opts := fakeImageDestOptions(t, "dest-", true, []string{}, c.args)
    res, err := opts.newSystemContext()
    if c.expected == nil {
        assert.Error(t, err)
    } else {
        require.NoError(t, err)
 	assert.Equal(t, &types.SystemContext{
 		DockerRegistryUserAgent: defaultUserAgent,
 		DockerAuthConfig:        c.expected,
 	}, res)
    }
}
```

docs/skopeo-copy.1.md
```diff
@@ -164,6 +164,22 @@ Bearer token for accessing the destination registry.
 
 The number of times to retry. Retry wait time will be exponentially increased based on the number of failed attempts.
 
+**--src-username**
+
+User name to access the source registry
```
mtrmac, line 169: In all the man page entries: It seems that the existing convention is to terminate each description with a full stop.

mtrmac (commented):
Thanks! This is great work.

vrothberg (commented):
LGTM once @mtrmac's comments are addressed.

Thanks for contributing!

cmd/skopeo/utils.go
```diff
@@ -122,6 +124,8 @@ func dockerImageFlags(global *globalOptions, shared *sharedImageOptions, depreca
 		fs.Var(newOptionalStringValue(&flags.authFilePath), flagPrefix+"authfile", "path of the authentication file. Default is ${XDG_RUNTIME_DIR}/containers/auth.json")
 	}
 	fs.Var(newOptionalStringValue(&flags.credsOption), flagPrefix+"creds", "Use `USERNAME[:PASSWORD]` for accessing the registry")
+	fs.Var(newOptionalStringValue(&flags.userName), flagPrefix+"username", "User name for accessing the registry")
```
TomSweeneyRedHat, line 127:
```suggestion
	fs.Var(newOptionalStringValue(&flags.userName), flagPrefix+"username", "Username for accessing the registry")
```

docs/skopeo-copy.1.md
TomSweeneyRedHat:
```suggestion
User name to access the source registry
```
```suggestion
The username to access the source registry.
```

docs/skopeo-copy.1.md
```diff
@@ -164,6 +164,22 @@ Bearer token for accessing the destination registry.
...
+
+**--src-password**
+
+Password to access the source registry
+
+**--dest-username**
+
+User name to access the destination registry
```
TomSweeneyRedHat, line 177:
```suggestion
Username to access the destination registry
```

docs/skopeo-delete.1.md
```diff
@@ -64,6 +64,14 @@ Directory to use to share blobs across OCI repositories.
 
 Require HTTPS and verify certificates when talking to the container registry or daemon. Default to registry.conf setting.
 
+**--username**
+
+User name to access the registry
+
+**--password**
+
+Password to access the registry
```
TomSweeneyRedHat, line 73: see prior

docs/skopeo-sync.1.md
```diff
@@ -88,6 +88,22 @@ Print usage statement.
...
+Password to access the source registry
+
+**--dest-username**
+
+User name to access the destination registry
+
+**--dest-password**
+
+Password to access the destination registry
+
```
TomSweeneyRedHat, line 106: see prior

cmd/skopeo/utils.go
mtrmac: That’s a good idea — but also a new feature we don’t yet have, and I don’t think this PR (targeted at allowing `:` in usernames) should block on implementing the prompting.

So for now, we should probably reject that combination, so that we can add prompting later.

cmd/skopeo/utils.go
jaikiran: Good catch. I had missed that. I've now updated this PR to fix that.

PR #1472, opened 2021-10-05
Update github.com/containerd/containerd to v1.5.7

mtrmac:
... to include a fix for https://github.com/advisories/GHSA-c2h3-6mxw-7mvq .

(Note that Skopeo doesn't depend on the vulnerable code, so this is primarily to avoid dependency checker warnings.)


rhatdan:
LGTM

Issue #785, opened 2019-09-12
skopeo inspect: a way how to avoid fetching all tags from repository


MartinBasti:
Hello,

in our use cases we need to get only information about image, like labels, but we don't need all `repoTags`. We have repositories with many tags (1700+) and it is resource consuming to get all tags from registry and just drop that.

Could you please provide a way how to avoid fetching repotags?

Thank you

lcarva:
This would be very useful if we want to use skopeo to simply determine the digest of the manfiest. Currently, I don't think there's a way to retrieve this information without also retrieving all the tags in the repository (which is very inefficient for repositories with many tags).

rhatdan:
Since this can not be done based on the registry protocol, I think this issue should be close.  Reopen if I am mistaken.

mtrmac:
Filtering can’t be done, not getting the data at all definitely can be done.

rittneje:
+1 on this. Since the `RepoTags` have nothing to do with the specific image that is being inspected (#1208), it's actually very confusing that it's in the output at all. I think it should be removed in a future major version of skopeo. Plus that list will continue to grow without bound over time and make things slow. (For reference the ubuntu registry currently has over 400 tags and counting.)

airadier:
Some registries like Public ECR don't seem to support listing tags, so Skopeo inspect will fail:
```
skopeo inspect docker://public.ecr.aws/jg/hello:latest
FATA[0004] Error determining repository tags: Error fetching tags list: invalid status code from registry 404 (Not Found)
```

Using --raw will work instead.

gsaslis:
+1 @rittneje 

I also found RepoTags *pretty* confusing when inspecting a specific **image**. Agree on removing it altogether because it belongs to a different entity: the Repository, not the Image. 

i.e. this seems like it is both slow and out-of-place.

mtrmac:
> Some registries like Public ECR don't seem to support listing tags

Thanks, let’s track that separately in #1230.

lsm5:
> I'm using `skopeo inspect --config <pullspec>` if I need information about labels. It's not fetching tags


@vrothberg @mtrmac is this sufficient to close this, perhaps accompanied with a manpage update?

ktdreyer:
[...]

PR #1480, opened 2021-10-08
skopeo inspect command - introduce a way to skip querying all available tags

jaikiran:
The commit in this PR introduces a new option to the `inspect` command which allows users to specify whether they want to disable querying and displaying all available tags for the image being inspected. This has been requested in https://github.com/containers/skopeo/issues/785

The new option is `--query-tags` which defaults to `true` (to preserve backward compatibility) and is an optional optional. When set to `--query-tags=false` the implementation skips querying the tags for the repository and the displayed output will contain an empty `RepoTags[]` property:

```
{
    "Name": "docker.io/library/python",
    "Digest": "sha256:5ca194a80ddff913ea49c8154f38da66a41d2b73028c5cf7e46bc3c1d6fda572",
    "RepoTags": [],
    ...
```

I couldn't find existing tests for this command so I haven't updated/added any. However, I've run a bunch of manual tests (with `--query-tags=false`, with `--query-tags`, with `--query-tags=true`, without `--query-tags`) and they all have gone fine.

Additionally, from a performance point of view, when not using or when not setting this to `false` and doing an inspect on a image which is expected to have a large number of tags (for example `docker.io/library/python`) results in a very noticable and big improvement. For example, the following command without this new option returns in around 15-16 seconds consistently (and displays all available tags):

```
time bin/skopeo inspect   --override-arch=amd64 --override-os=linux docker://docker.io/library/python
....
real	0m15.884s
user	0m0.173s
sys	0m0.131s
```
Whereas when run with `--query-tags=false` for the same image, it returns consistently in around 8-9 seconds (of course with `RepoTags: []`):

```
time bin/skopeo inspect --query-tags=false  --override-arch=amd64 --override-os=linux docker://docker.io/library/python
...

real    0m8.444s
user    0m0.153s
sys 0m0.120s
```


arewm:
I do not know if there is a precedence for this within skopeo, but is there a way to get the default value for the parameter from the environment so that I do not have to always remember to set the flag when calling it?

docs/skopeo-inspect.1.md
```diff
@@ -77,6 +77,10 @@ The username to access the registry.
 
 The password to access the registry.
 
+**--query-tags**=_bool_
+
+Query and include in the output all the available tags for the image. Defaults to `true`.
```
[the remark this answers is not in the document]
jaikiran: The updated PR has the suggested change, although it doesn't exactly match the previous sentence. So feel free to suggest any additional change if you think is needed.

jaikiran:
> @jaikiran it might be warranted to add a single-letter option to disable the tag listing, though (`-t`?`-n`? `-l`?) Absolutely non-blocking, it can be added any time later.

I decide to use `-l`. Actually, now that I'm typing this and notice that the lower case "l" (ell) looks like a upper case "I" (eye), I'm having second thoughts about it. I think "-t" is a better choice.

jaikiran:
> > @jaikiran it might be warranted to add a single-letter option to disable the tag listing, though (`-t`?`-n`? `-l`?) Absolutely non-blocking, it can be added any time later.
> 
> I decide to use `-l`. Actually, now that I'm typing this and notice that the lower case "l" (ell) looks like a upper case "I" (eye), I'm having second thoughts about it. I think "-t" is a better choice.

PR has now been updated to use `-t` for this new option.

mtrmac (approved):
LGTM. Thanks!

@vrothberg I’d appreciate a second pair of eyes.

docs/skopeo-inspect.1.md
```diff
@@ -77,6 +77,10 @@ The username to access the registry.
 
 The password to access the registry.
 
+**--no-tags**, **-n**=_bool_
+
+Do not list in the output, all tags available in the same repository. Defaults to `false` because of which the output will include all tags available.
```
TomSweeneyRedHat, line 82:
```suggestion
Do not list the available tags from the repository in the output. When invoked, the `RepoTags` array will be empty.  Defaults to `false`, which includes all available tags.
```

docs/skopeo-inspect.1.md
```diff
@@ -77,6 +77,10 @@ The username to access the registry.
 
 The password to access the registry.
 
+**--no-tags**, **-n**=_bool_
```
TomSweeneyRedHat, line 80: Just noticed on the page that the options are not listed in alpha order.  While you're in there, could you rearrange them please?

TomSweeneyRedHat:
@jaikiran a few nits and suggestions, but great change!  TYVM for bringing it forward.

cmd/skopeo/inspect.go
```diff
@@ -60,6 +61,7 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
 	flags.BoolVar(&opts.raw, "raw", false, "output raw manifest or configuration")
 	flags.BoolVar(&opts.config, "config", false, "output configuration")
 	flags.StringVarP(&opts.format, "format", "f", "", "Format the output to a Go template")
+	flags.BoolVarP(&opts.doNotListTags, "no-tags", "n", false, "Do not list in the output, all tags available in the same repository")
```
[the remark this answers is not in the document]
vrothberg: Please no trailing dot (for consistency)

docs/skopeo-inspect.1.md
jaikiran: Done. The updated version of this PR (almost) uses this text. The only minor change I additionally did was, instead of saying "When invoked...", I used "When `true`...". I hope that's fine.

docs/skopeo-inspect.1.md
jaikiran: Hello @TomSweeneyRedHat, 

I had a look at some other man files too and they too don't have it in alphabetical order. Given that, do you still want this to be changed for this doc content in this PR? If yes, let me know and I'll go ahead and do it.

docs/skopeo-inspect.1.md
```diff
@@ -77,6 +77,10 @@ The username to access the registry.
...
+
+Do not list in the output, all tags available in the same repository. Defaults to `false` because of which the output will include all tags available.
+
 ## EXAMPLES
 
 To review information for the image fedora from the docker.io registry:
```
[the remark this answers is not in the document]
jaikiran: Done. Added the example.

jaikiran:
Hello @vrothberg ,

> I'd also love to have a simple test exercising the new flag similar to https://github.com/containers/skopeo/blob/main/systemtest/010-inspect.bats#L67.

Is that some internal (skopeo) specific test framework. It looks like it's bash shell driven? Is there some way I can trigger these tests locally?

Either way, I have added a new test which I think should work, but I haven't been able to test it locally given my lack of knowledge on how to run those tests. I guess we will know shortly how it goes in the CI run.

jaikiran:
> Regarding the flag name: I would prefer `list-tags` or `query-tags` over `no-tags` since it feels more natural to turn a functionality off via `--do-something=false` rather than turning an off-switch on via `--do-not-do-something=true`.

I think the `--no-tags` is a bit more natural in this case because the current command by default lists those tags. So users who do want to disable that behaviour would be looking for "what do I pass to disable tags" and "pass --no-tags or -n" appears a more natural answer than "pass --list-tags=false or -l=false".

Having said that, I will let @mtrmac, you and others decide which one is better. I have let it stay `--no-tags` and `-n` for now in the latest PR. If there's an agreement to changing it to something else, please let me know and I'll go ahead and update the PR.

arewm:
This behavior _feels_ polarizing enough to potentially warrant is breaking change nature. What is the process to do a Major version bump to define the default behavior to not fetch tags? 

If this isn't the right time for a major version bump, can we make sure to keep this feature on the docket for a future major version bump?

mtrmac:
> What is the process to do a Major version bump to define the default behavior to not fetch tags?

Well there isn’t a precedent for Skopeo :) But I also don’t think it’s necessary.

---

> If this isn't the right time for a major version bump, can we make sure to keep this feature on the docket for a future major version bump?

That’s a good point. We don’t have a formal tracker, so I guess a `// TO DO: Flip the default in a future major version bump` would be the least we could do.

mtrmac:
> What about `--no-repo-tags` instead of `--no-tags`?

\*shrug\* I have a strong opinion at all. Because there is a single-letter option, we can probably afford a more accurate but longer option; OTOH some people hate typing.

> "no tags" would imply that I don't care about tags at all, but I do care about one tag - the one I'm inspecting

Note that `skopeo inspect` is only providing a tag value in the `Tag` field on old schema1 images, and it’s either not reliable to be any particular tag value (if the registry doesn’t enforce tag matching), or it is exactly the tag used to write to the registry (if the registry does enforce it), i.e. something the caller has quite possibly provided on the command line — and the value is not present on newer images at all. I _can_ imagine processes where querying this schema1 field is important, in principle, but it seems increasingly unlikely now that default distribution/distribution configuration is to reject schema1.

Issue #1478, opened 2021-10-07
Publish binary releases to GitHub


jawnsy:
This is an issue to discuss adding binaries to GitHub releases for skopeo. I've had good experience with using goreleaser for this, and since skopeo builds with the regular Go toolchain (e.g. `go install github.com/containers/skopeo/cmd/skopeo@latest` works great), it should be relatively straightforward to add these builds.

Would you accept a pull request to add release builds?

rhatdan:
I think we go through this statically linking idea every year or two. Just 10 days ago we closed.

https://github.com/containers/skopeo/issues/755
https://github.com/containers/skopeo/pull/932
https://github.com/containers/skopeo/pull/670

Please read these issues, before asking any questions.

We prefer distro based versions of the skopeo or to run a containerized version of skopeo from your favorite distro.

podman run quay.io/skopeo/stable:latest ...

We had a nix version which was eventually removed.

lsm5:
@mtrmac podman only releases the linux remote client as a static binary, https://github.com/containers/podman/releases/tag/v3.4.0 . Buildah doesn't publish any static assets https://github.com/containers/podman/releases/tag/v3.4.0 . 

I see a comment in podman's Makefile saying podman doesn't work well w/o CGO_ENABLED, so that's why only the `podman-remote-static` target is built without CGO and published.

@jawnsy If you're signing up to maintain the binary release part going forward and have reasonable certainty of goreleaser not breaking/blocking release, I'm cool with it.

The skopeo Makefile does use a `DISABLE_CGO` variable, but I'd defer to @mtrmac @vrothberg @rhatdan on its impact.

mtrmac:
> I would also appreciate a lot to have static binaries (or pseudo-static binaries that only depend on the glibc[1]), for the following use-cases:
> 
> * teaching (in a short workshop it's easier if folks can download a binary that "just works" rather than wasting 10-15 minutes for everyone to figure out the steps specific to their environments

Don’t the containers serve that purpose, especially in a container-focused course?

> ; especially as there are no packages for Windows for now)

(FWIW with Windows-native, the binary compatibility is probably not an issue, but to my knowledge no-one has worked on making Skopeo a _good_ Windows citizen, e.g. looking for configuration files in the right places.)

> * CI/CD tooling images (it's best if each tool can be installed as a single binary as well, to minimize image size and simplify the maintenance of these images)

Presumably the CI/CD system is not a personal Linux-from-scratch, so can it use a distribution package?

[...]

PR #1483, opened 2021-10-18
Add instructions to generate static binaries

jpetazzo:
Following the discussion in #1478, we don't want to provide
(and maintain) static binaries, but giving instructions to
produce such builds (with appropriate warnings around these
instructions) was considered acceptable, so - here we go!


install.md
```diff
@@ -203,3 +203,42 @@ Finally, after the binary and documentation is built:
...
+
+That being said, if you would like to build Skopeo statically, you might be able to do it by:
+- exporting environment variable `CGO_ENABLED=0` (disabling CGO causes Go to prefer native libraries when possible, instead of dynamically linking against system libraries)
+- passing the `BUILDTAGS=containers_image_openpgp` Make flag (this remove the dependency on `libgpgme` and its companion libraries)
+- commenting out the `GO_DYN_FLAGS` line in the Makefile (this flag seems to force the creation of a dynamic executable)
+
+The following command implements these steps to produce a static binary in the `bin` subdirectory of the repository:
+
+```bash
+sed 's/GO_DYN_FLAGS=/#/' Makefile |
```
mtrmac, line 221: Hum, (assuming the other question doesn’t makes this moot), I’d rather have this as a `Makefile` variable or target instead of a document for how to edit the `Makefile`; this could extremely easily break without anyone noticing.

---

In my testing, just `make … GO_DYN_FLAGS=` seems to work fine to override the variable, just like `make BUILDTAGS=…` works. Does that work for you?

install.md
```diff
@@ -203,3 +203,41 @@ Finally, after the binary and documentation is built:
 ```bash
 sudo make install
 ```
+
+### Building a static binary
+
+There have been efforts in the past to produce and maintain static builds, but the maintainers prefer to run Skopeo using distro packages or within containers. This is because static builds of Skopeo tend to be unreliable and functionally restricted. Specifically:
+- some features of Skopeo depend on non-Go libraries like `libgpgme` and `livdevmapper`,
```
TomSweeneyRedHat, line 210:
```suggestion
- some features of Skopeo depend on non-Go libraries like `libgpgme` and `livdevmapper`.
```

install.md
```diff
@@ -203,3 +203,41 @@ Finally, after the binary and documentation is built:
...
+- [#1336] - Unable to run skopeo on Fedora 30 (due to dyn lib dependency)
+- [#1478] - Publish binary releases to GitHub (request+discussion)
+
+[#391]: https://github.com/containers/skopeo/issues/391
+[#669]: https://github.com/containers/skopeo/issues/669
+[#670]: https://github.com/containers/skopeo/issues/670
+[#755]: https://github.com/containers/skopeo/issues/755
+[#932]: https://github.com/containers/skopeo/issues/932
+[#1336]: https://github.com/containers/skopeo/issues/1336
+[#1478]: https://github.com/containers/skopeo/issues/1478
```
TomSweeneyRedHat, line 243: Well, I must say this is an md trick that I've not seen before but I like!

mtrmac (approved):
Thanks again!

PR #1488, opened 2021-10-29
main: Error out if an unrecognized subcommand is provided

cgwalters:
Surprisingly, the spf13/cobra CLI parsing logic, when presented
with an unknown subcommand outputs usage to stdout
and *exits successfully*.

This is bad for both users and scripts.  Cargo cult some code
I found in podman to handle this.

Motivated by https://github.com/containers/containers-image-proxy-rs/pull/1

### DIFF
--- a/cmd/skopeo/main.go
+++ b/cmd/skopeo/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	commonFlag "github.com/containers/common/pkg/flag"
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
LGTM
IDK, if @mtrmac might prefer the new function in a separate file like Podman's (https://github.com/containers/podman/blob/056f492f59c333d521ebbbe186abde0278e815db/cmd/podman/validate/args.go), I'll let him weigh in on that.

cgwalters:
> IDK, if @mtrmac might prefer the new function in a separate file like Podman's 

Note that as of right now, skopeo does not have nested subcommands; i.e. it has just `skopeo copy` but not a depth of 2 or or more like `podman image push`.  I think sharing the code across the latter motivates splitting it out of `main.go`.

cgwalters:
Though I guess I could try moving it to containers/common like https://github.com/containers/skopeo/commit/7885162a35d9ac89cdae970884d651ec0674a7db ?

lsm5:
> Though I guess I could try moving it to containers/common like [7885162](https://github.com/containers/skopeo/commit/7885162a35d9ac89cdae970884d651ec0674a7db) ?

sgtm, @mtrmac has CZ holiday today (afaik) and @vrothberg is pto this week.

mtrmac (commented):
Thanks, great catch!

---

> IDK, if @mtrmac might prefer the new function in a separate file like Podman's 

I personally don’t feel strongly about this at all, `main.go` is just fine for this single user.  `utils.go` would probably the other candidate — but that location would probably need a “unit” test (comparable to `TestManifestDigest`), to maintain the pressure to keep test coverage of parsers in that file high. So there’s a benefit to keeping this in `main.go` 😉 

---

> Though I guess I could try moving it to containers/common like 7885162 ?

Realistically I don’t think we are going to keep the two in sync in any other way (at the very least not without a corresponding comment in the Podman counterpart). OTOH I also personally care very little about Skopeo/Podman doing _exactly_ the same thing in this situation (the error strings don’t look exactly the same already), so I’m not sure it’s worth the extra effort.

Ultimately, I think The Right Way to share this behavior would be to make this a Cobra feature directly. For the record, I found https://github.com/spf13/cobra/issues/706 , which suggests that it might not be happening soon.

---

Do note the `--help` artifact in https://github.com/spf13/cobra/issues/706#issuecomment-592022331 ; it applies to this PR as well, and the only workaround I see (setting a custom usage template) is so ugly that I don’t think it’s worth it.

That `--help` artifact also suggests to me that a c/common helper is unlikely to be the right long-term solution, so that’s another minor argument against doing it even short-term.

mtrmac (commented):
LGTM if tests pass (I triggered a re-run of one failure that looked unrelated). Thanks!

lsm5 (approved):
LGTM

lsm5:
merging this before v1.5.1

PR #1494, opened 2021-11-04
Cirrus: Fix only_if error

cevich:
The logic parser for `only_if` and `skip` is a little weird in that it
wants a boolean, but doesn't support using them.  In this case, using `1
== 0` results in an error from the parser:

`Error while parsing tasks: INVALID_ARGUMENT: internal boolevator error:
parsing error: 1 == 0 :1:1 - 1:2 unexpected Int while scanning
extensions!`

However, since the goal is to block this particular task from running
(essentially) indefinitely, it's easier to just remove it.  We can rely
on git recording it's history in case it needs to be recovered in the
future.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -64,23 +64,6 @@ validate_task:
     script: make validate-local
 
 
-cross_task:
-    only_if: 1 == 0
-    macos_instance:
-        image: catalina-xcode
-    setup_script: |
-        export PATH=$GOPATH/bin:$PATH
-        brew install gpgme go go-md2man
-        go get -u golang.org/x/lint/golint
-    test_script: |
-        export PATH=$GOPATH/bin:$PATH
-        go version
-        go env
-        make validate-local test-unit-local bin/skopeo
-        sudo make install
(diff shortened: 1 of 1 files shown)


mtrmac (commented):
`success_task` `depends_on` `cross`; is that a problem?

The “Build parsing results” step did succeed but also complains about this.

cevich:
> success_task depends_on cross; is that a problem?

It's not suppose to be be a problem (I confirmed this years ago with them).  The `depends_on` is suppose to properly handle the case of children being skipped for any reason.

Properly as in: skipped == pass when evaluating "dependency was met".

So if this ever breaks, we need to report it to them ASAP (lots of our CI depends on this behavior).

cevich:
> The “Build parsing results” step

FWIW: I just ignore that thing, it complains about a lot of things we do.

cevich:
Uggg, wait, no, my bad.  This is a real problem.  Fixing....

mtrmac:
Yeah, I didn’t see any test results in the GitHub box with the previous version. This LGTM, feel free to merge after tests pass.

PR #1495, opened 2021-11-08
proxy: Add `GetConfig`, add manifest list support, add an integration test

cgwalters:
proxy: Add an API to fetch the config upconverted to OCI

While the caller could fetch this today as a blob, it'd be in
either docker or oci schema.  In keeping with the model of having
this proxy only expose OCI, add an API which uses the c/image logic
to do the conversion.

This is necessary for callers to get the diffIDs, and in general
to implement something like an external `skopeo copy`.

---

proxy: Add a helper to return a byte array

Since this is shared between the manifest and config paths.

---

proxy: Use float → int helper for pipeid

Just noticed while scrolling past the code.

---

tests/integration/proxy_test: New test that exercises `proxy.go`

I debated adding "reverse dependency testing" using
https://crates.io/crates/containers-image-proxy
but I think it's easier to reuse the test infrastructure here.

This also starts fleshing out a Go client for the proxy (not
that this is going to be something most Go projects would want
versus vendoring c/image...but hey, maybe it'll be useful).

Now what I hit in trying to use the main test images is currently
the proxy fails on manifest lists, so I'll need to fix that.

---

proxy: Add support for manifest lists

We need to support manifest lists. I'm not sure how I missed this
originally.  At least now we have integration tests that cover this.

The issue here is fairly subtle - the way c/image works right now,
`image.FromUnparsedImage` does pick a matching image from a list
by default.  But it also overrides `GetManifest()` to return the
original manifest list, which defeats our goal here.

Handle this by adding explicit manifest list support code.  We'll
want this anyways for future support for `GetRawManifest` or so
which exposes OCI manifest lists to the client.

---

proxy_test: Add a helper method to call without fd

To verify in one place.

---

proxy_test: Add helper to read all from a reply

Prep for testing `GetConfig`.

---

proxy_test: Test `GetConfig`

Now that we have a test suite, let's use it more.

---

### DIFF
--- a/cmd/skopeo/proxy.go
+++ b/cmd/skopeo/proxy.go
@@ -491,11 +491,11 @@ func (h *proxyHandler) FinishPipe(args []interface{}) (replyBuf, error) {
 
 	var ret replyBuf
 
-	pipeidf, ok := args[0].(float64)
-	if !ok {
-		return ret, fmt.Errorf("finishpipe: expecting pipeid, not %T", args[0])
+	pipeidv, err := parseUint64(args[0])
+	if err != nil {
+		return ret, err
 	}
-	pipeid := uint32(pipeidf)
+	pipeid := uint32(pipeidv)
 
 	f, ok := h.activePipes[pipeid]
 	if !ok {
@@ -507,7 +507,7 @@ func (h *proxyHandler) FinishPipe(args []interface{}) (replyBuf, error) {
 	// And only now do we close the write half; this forces the client to call this API
 	f.w.Close()
 	// Propagate any errors from the goroutine worker
-	err := f.err
+	err = f.err
 	delete(h.activePipes, pipeid)
 	return ret, err
 }
--- a/cmd/skopeo/proxy.go
+++ b/cmd/skopeo/proxy.go
@@ -83,9 +83,11 @@ import (
 
 // protocolVersion is semantic version of the protocol used by this proxy.
 // The first version of the protocol has major version 0.2 to signify a
-// departure from the original code which used HTTP.  The minor version is 1
-// instead of 0 to help exercise semver parsers.
-const protocolVersion = "0.2.1"
+// departure from the original code which used HTTP.
+//
+// 0.2.1: Initial version
+// 0.2.2: Added support for fetching image configuration as OCI
+const protocolVersion = "0.2.2"
 
 // maxMsgSize is the current limit on a packet size.
 // Note that all non-metadata (i.e. payload data) is sent over a pipe.
@@ -380,6 +382,53 @@ func (h *proxyHandler) GetManifest(args []interface{}) (replyBuf, error) {
 	return ret, nil
 }
 
+// GetConfig returns a copy of the image configuration, converted to OCI format.
+func (h *proxyHandler) GetConfig(args []interface{}) (replyBuf, error) {
+	h.lock.Lock()
+	defer h.lock.Unlock()
+
+	var ret replyBuf
+
+	if h.sysctx == nil {
+		return ret, fmt.Errorf("Must invoke Initialize")
+	}
+	if len(args) != 1 {
+		return ret, fmt.Errorf("invalid request, expecting: [imgid]")
+	}
+	imgref, err := h.parseImageFromID(args[0])
+	if err != nil {
+		return ret, err
+	}
+
+	ctx := context.TODO()
+	config, err := imgref.img.OCIConfig(ctx)
+	if err != nil {
+		return ret, err
+	}
+	serialized, err := json.Marshal(&config.Config)
+	if err != nil {
+		return ret, err
+	}
+
+	piper, f, err := h.allocPipe()
+	if err != nil {
+		return ret, err
+	}
+
+	go func() {
+		// Signal completion when we return
+		defer f.wg.Done()
+		_, err = io.Copy(f.w, bytes.NewReader(serialized))
+		if err != nil {
+			f.err = err
+		}
+	}()
+
+	ret.fd = piper
+	ret.pipeid = uint32(f.w.Fd())
+	return ret, nil
+}
+
 // GetBlob fetches a blob, performing digest verification.
 func (h *proxyHandler) GetBlob(args []interface{}) (replyBuf, error) {
 	h.lock.Lock()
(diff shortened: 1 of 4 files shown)


cgwalters:
> vet: integration/proxy_test.go:186:3: unknown field Pdeathsig in struct literal

Blah.

integration/proxy_test.go
```diff
@@ -0,0 +1,231 @@
...
+	"strings"
+	"syscall"
+
+	"gopkg.in/check.v1"
+
+	"github.com/containers/image/v5/manifest"
+)
+
+// This image is known to be x86_64 only right now
+const knownNotManifestListedImage_x8664 = "docker://quay.io/coreos/11bot"
```
cgwalters, line 21: I'm sure there's a better image for this, maybe something in `quay.io/libpod`?

vrothberg:
Thanks, @cgwalters!  I will have a look at this PR tomorrow morning.

cmd/skopeo/proxy.go
```diff
@@ -380,6 +382,53 @@ func (h *proxyHandler) GetManifest(args []interface{}) (replyBuf, error) {
...
 
+// GetConfig returns a copy of the image configuration, converted to OCI format.
+func (h *proxyHandler) GetConfig(args []interface{}) (replyBuf, error) {
+	h.lock.Lock()
+	defer h.lock.Unlock()
+
+	var ret replyBuf
+
+	if h.sysctx == nil {
+		return ret, fmt.Errorf("Must invoke Initialize")
```
vrothberg, line 445: Go errors are not capitalized by convention.  It may be helpful to prefix it with `internal error`.

vrothberg (commented):
Only a minor nit.  Could you sign (`git commit -s`) your commits?  I am not sure why the CI isn't barking on that but our contributing guidelines require that.

cmd/skopeo/proxy.go
cgwalters: OK, fixed that.

cgwalters:
> Only a minor nit. Could you sign (git commit -s) your commits?

Done!

vrothberg (commented):
LGTM
@lsm5 @mtrmac @rhatdan PTAL

cgwalters:
Of these I think https://github.com/containers/skopeo/pull/1495/commits/982aa510a1bee685207f192e9f829f439e7dabb0 could use specific review.  I find the c/image APIs here a bit confusing.

integration/proxy_test.go
```diff
@@ -0,0 +1,293 @@
...
+		fetchchan <- byteFetch{
+			content: manifestBytes,
+			err:     err,
+		}
+	}()
+	_, err = self.callNoFd("FinishPipe", []interface{}{fd.id})
+	if err != nil {
+		return
+	}
+	fetchRes := <-fetchchan
```
mtrmac, line 157: Non-blocking: This being the primary tests for the feature, some kind of timeout to visibly catch breakage would be nice. OTOH we can build that only if this ever breaks :)

integration/proxy_test.go
```diff
@@ -0,0 +1,293 @@
...
+	if err != nil {
+		return err
+	}
+
+	v, configBytes, err := p.callReadAllBytes("GetConfig", []interface{}{imgid})
+	if err != nil {
+		return err
+	}
+	var config imgspecv1.ImageConfig
+	err = json.Unmarshal(configBytes, &config)
```
mtrmac, line 268: Non-blocking: `json.Unmarshal` is going to accept almost everything. Maybe as a smoke check, verify that some field (`Config.Entrypoint`?) is set,

mtrmac (commented):
Thanks! Looks good overall.

Do note the DiffID comment.

mtrmac:
> Of these I think [982aa51](https://github.com/containers/skopeo/commit/982aa510a1bee685207f192e9f829f439e7dabb0) could use specific review. I find the c/image APIs here a bit confusing.

Looks fine to me.

integration/proxy_test.go
```diff
@@ -0,0 +1,293 @@
...
+	_, err = p.callNoFd("CloseImage", []interface{}{imgid})
+
+	return nil
+}
+
+func (s *ProxySuite) TestProxy(c *check.C) {
+	p, err := newProxy()
+	c.Assert(err, check.IsNil)
+
+	err = runTestGetManifestAndConfig(p, knownNotManifestListedImage_x8664)
```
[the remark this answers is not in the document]
cgwalters: Personally...I like to try to keep test code looking as much as possible like production code.  I know it's not the Go style.  We recently debated this for Rust, where it is much easier because one can just have a test function return `Result<>`.

cmd/skopeo/proxy.go
```diff
@@ -360,24 +431,43 @@ func (h *proxyHandler) GetManifest(args []interface{}) (replyBuf, error) {
...
-	go func() {
-		// Signal completion when we return
-		defer f.wg.Done()
-		_, err = io.Copy(f.w, bytes.NewReader(serialized))
-		if err != nil {
-			f.err = err
-		}
-	}()
+	ctx := context.TODO()
+	config, err := img.OCIConfig(ctx)
```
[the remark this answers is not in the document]
cgwalters: The caller is going to invoke `GetManifest()` first in all cases I think, so we'd fail on docker v2s1 there.

cmd/skopeo/proxy.go
```diff
@@ -300,15 +297,82 @@ func (h *proxyHandler) allocPipe() (*os.File, *activePipe, error) {
...
+	ret.pipeid = uint32(f.w.Fd())
+	return ret, nil
+}
+
+// cacheTargetManifest is invoked when GetManifest or GetConfig is invoked
+// the first time for a given image.  If the requested image is a manifest
+// list, this function resolves it to the image matching the calling process'
+// operating system and architecture.
+//
+// TODO: Add GetRawManifest or so that exposes manifest lists
```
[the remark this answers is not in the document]
cgwalters: > Alternatively, it might make sense to actually design for multi-arch, in some way, at least enough to see what the future would look like. ...
>  It just feels like a kind of thing where a little thought advance could save some pain in the future.

Yeah. I don't have an immediate need to support non-current arch images but it definitely makes sense to consider it.

> For context, the c/image approach which completely silently converts multi-arch manifest lists into current-arch images, with the weird warts like still returning the raw multi-arch manifest list, was created in a sort of panic after Docker Hub started serving clients multi-arch overnight. It’s not necessarily a well-thought-out model to follow — but then we never had the option to consider completely ripping that approach out, so I also don’t have a strong opinion on what would be a good model.)

Ahh, I see.  That historical context helps explain things.

integration/proxy_test.go
cgwalters: Good call, I was actually returning the image and not the config!  Fixed.

cgwalters:
CI failure looks unrelated, is there a way to retrigger or should I just repush?

vrothberg:
I retriggered, let's see what's going on.

vrothberg:
And one more time

PR #1503, opened 2021-11-11
Fix bug that prevented useful diagnostics on registry fail

cevich:
Sigh. 'expr 1 - 1' yields 0 (correctly) but also exits 1. This
is even documented in the man page, but I didn't know it. And
thus, on the final iteration, when timeout reached 0, BATS
errored out on the expr instead of continuing to the 'podman logs'
or the 'die' message.

Solution is super trivial: use $(( ... )) instead of expr.

### DIFF
--- a/systemtest/helpers.bash
+++ b/systemtest/helpers.bash
@@ -356,7 +356,7 @@ start_registry() {
             return
         fi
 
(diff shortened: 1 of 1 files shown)


lsm5 (approved):
LGTM

PR #1507, opened 2021-11-16
[WIP] Makefile: set CONTAINERSCONFDIR based on the OS

lsm5:
This will put config dir under $PREFIX for FreeBSD.

Continuing PR#1298 by @mateuszkwiatkowski.

Fixes: #787

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,12 @@ GPGME_ENV := CGO_CFLAGS="$(shell gpgme-config --cflags 2>/dev/null)" CGO_LDFLAGS
 # The following variables very roughly follow https://www.gnu.org/prep/standards/standards.html#Makefile-Conventions .
 DESTDIR ?=
 PREFIX ?= /usr/local
-CONTAINERSCONFDIR ?= /etc/containers
+# MacOS and FreeBSD install configs under $PREFIX
+ifeq ($(shell uname -s),$(filter $(shell uname -s),Darwin FreeBSD))
+	CONTAINERSCONFDIR ?= $(PREFIX)/etc/containers
+else
+	CONTAINERSCONFDIR ?= /etc/containers
+endif
 REGISTRIESDDIR ?= ${CONTAINERSCONFDIR}/registries.d
 SIGSTOREDIR ?= /var/lib/containers/sigstore
 BINDIR ?= ${PREFIX}/bin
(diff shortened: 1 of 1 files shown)


lsm5:
@glensc @mateuszkwiatkowski would either of you have the time to give this a try please?

glensc:
PR #1298 - i like links

glensc:
@lsm5 LGTM:
- https://github.com/Homebrew/homebrew-core/pull/89524#issuecomment-970721779

Makefile
```diff
@@ -91,7 +96,12 @@ CONTAINER_RUN ?= $(CONTAINER_CMD) --security-opt label=disable -v $(CURDIR):$(CO
 GIT_COMMIT := $(shell git rev-parse HEAD 2> /dev/null || true)
 
 EXTRA_LDFLAGS ?=
-SKOPEO_LDFLAGS := -ldflags '-X main.gitCommit=${GIT_COMMIT} $(EXTRA_LDFLAGS)'
+SKOPEO_LDFLAGS := -ldflags '-X main.gitCommit=${GIT_COMMIT} \
+				  -X github.com/containers/image/v5/docker.systemRegistriesDirPath=${REGISTRIESDDIR} \
+				  -X github.com/containers/image/v5/signature.systemDefaultPolicyPath=${CONTAINERSCONFDIR}/policy.json \
+				  -X github.com/containers/image/v5/pkg/sysregistriesv2.systemRegistriesConfPath=${CONTAINERSCONFDIR}/registries.conf \
+				  -X github.com/containers/image/v5/internal/tmpdir.unixTempDirForBigFiles=/var/tmp \
```
mtrmac, line 103: This overrides the hard-coded path with the same hard-coded path, AFAICS; why?

mtrmac (commented):
This doesn’t seem to resolve the concerns in https://github.com/containers/skopeo/pull/1298#issuecomment-849529062 ; just silently switching a default on an upgrade, and having defaults different from Buildah/Podman (ideally from the c/image builtins), is not desirable.

Making that change exclusive to macOS and FreeBSD makes it less problematic for Linux, but still problematic for the two, I think.

Makefile
lsm5: @mtrmac so 

> This doesn’t seem to resolve the concerns in [#1298 (comment)](https://github.com/containers/skopeo/pull/1298#issuecomment-849529062) ; just silently switching a default on an upgrade, and having defaults different from Buildah/Podman (ideally from the c/image builtins), is not desirable.
> 
> Making that change exclusive to macOS and FreeBSD makes it less problematic for Linux, but still problematic for the two, I think.

@mtrmac so would you prefer the same change be done across the board? If macos and freebsd prefer it in /usr/local, then I imagine this change has to be done at some point, yes?

The extra ldflag lines were a copy-paste from the older PR. I could make those conditional for homebrew / FreeBSD if they prefer to have them. @glensc @mateuszkwiatkowski wdyt?

mtrmac: > If macos and freebsd prefer it in /usr/local, then I imagine this change has to be done at some point, yes?

No?

Users who use Brew or other build systems that already override the paths can continue to use those build systems. Users who use `make install` would be broken on upgrade if we just changed the default with no compatibility mechanism; so we shouldn’t just change the default.

My view of #1298 is that adding reliable variables (like `CONTAINERSCONFDIR` or whatever), which would help Brew reliably work, is obviously useful and desirable; changing what the variable point to, by default, is not.

mtrmac: Alternatively, we could work to actually provide a compatibility mechanism (i.e. support at least two paths in c/image, and configure them to… something in there and possibly here). I don’t see that that work and complexity helps any known existing user — though it would help users who cross-grade between packaging systems like Brew and manual `make install`.

glensc:
Can this be finished? so i don't have to bump brew PR to keep it active:
- https://github.com/Homebrew/homebrew-core/pull/89524#issuecomment-1008857642

rhatdan:
@glensc @lsm5 is on PTO until the end of the month, I believe.  If you want this to move forward, I suggest you take it over and open a new PR, following @mtrmac recommendations.

mtrmac:
Changes in the meantime: #1642 + https://github.com/containers/image/pull/1544 will move the config files to `/usr/local/etc` on FreeBSD (but not macOS like this PR also does); the transition will be an incompatible break but at least it will (eventually) be consistent with other c/image users.

At least the other aspect of this PR, making the directory overrides turn into c/image built-in path overrides, is still valuable and desirable for #787 .

PR #1510, opened 2021-11-19
[WIP] Bump github.com/opencontainers/image-spec to v1.0.2

lsm5:
Fix for GHSA-77vh-xpmg-72qh

Vulnerable versions: < 1.0.2
Patched version: 1.0.2

`Impact`
In the OCI Image Specification version 1.0.1 and prior, manifest
and index documents are not self-describing and documents with
a single digest could be interpreted as either a manifest or an
index.

`Patches`
The Image Specification will be updated to recommend that both
manifest and index documents contain a mediaType field to
identify the type of document.

`Workarounds`
Software attempting to deserialize an ambiguous document may
reject the document if it contains both “manifests” and “layers”
fields or “manifests” and “config” fields.

`References`
GHSA-mc8v-mgrf-8f4m


lsm5:
currently fails to build

vendor/github.com/opencontainers/image-spec/specs-go/v1/index.go
```diff
@@ -21,6 +21,9 @@ import "github.com/opencontainers/image-spec/specs-go"
 type Index struct {
 	specs.Versioned
 
+	// MediaType specificies the type of this document data structure e.g. `application/vnd.oci.image.index.v1+json`
+	MediaType string `json:"mediaType,omitempty"`
+
```
lsm5, line 26: actual fix for GHSA-77vh-xpmg-72qh

vendor/github.com/opencontainers/image-spec/specs-go/v1/manifest.go
```diff
@@ -20,6 +20,9 @@ import "github.com/opencontainers/image-spec/specs-go"
 type Manifest struct {
 	specs.Versioned
 
+	// MediaType specificies the type of this document data structure e.g. `application/vnd.oci.image.manifest.v1+json`
+	MediaType string `json:"mediaType,omitempty"`
+
```
lsm5, line 25: actual fix for GHSA-77vh-xpmg-72qh

mtrmac (changes requested):
NAK
- None of this is actually a fix, it just introduces a field, which doesn’t (usually) change any behavior. Our fix is https://github.com/containers/image/pull/1409 .
- This fails to build because it removes post-1.0 MediaType values that we depend on. It should instead vendor from the post-1.0 branch, as https://github.com/mtrmac/image/tree/ambiguous does.

Basically Skopeo needs to make a release soon, on top of c/image 5.17.0. We can vendor this image-spec version separately, and we want to rely on that field (per that WIP branch) in a future version but not in a rushed security fix one.

PR #1511, opened 2021-11-21
Add an option to allow copying image indexes alone

Jamstah:
This new --list-processing option deprecates the --all option and, when
an image index is copied, allows the user to select between copying the
image associated with the system platform, all images in the index, or
just the index itself without attempting to copy the images.

### DIFF
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -32,6 +32,7 @@ type copyOptions struct {
 	format              commonFlag.OptionalString // Force conversion of the image to a specified format
 	quiet               bool                      // Suppress output information when copying images
 	all                 bool                      // Copy all of the images if the source is a list
+	multiArch           commonFlag.OptionalString // How to handle multi architecture images
 	encryptLayer        []int                     // The list of layers to encrypt
 	encryptionKeys      []string                  // Keys needed to encrypt the image
 	decryptionKeys      []string                  // Keys needed to decrypt the image
@@ -72,6 +73,7 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
 	flags.StringSliceVar(&opts.additionalTags, "additional-tag", []string{}, "additional tags (supports docker-archive)")
 	flags.BoolVarP(&opts.quiet, "quiet", "q", false, "Suppress output information when copying images")
 	flags.BoolVarP(&opts.all, "all", "a", false, "Copy all images if SOURCE-IMAGE is a list")
+	flags.Var(commonFlag.NewOptionalStringValue(&opts.multiArch), "multi-arch", `How to handle multi-architecture images (system, all, or index-only)`)
 	flags.BoolVar(&opts.removeSignatures, "remove-signatures", false, "Do not copy signatures from SOURCE-IMAGE")
 	flags.StringVar(&opts.signByFingerprint, "sign-by", "", "Sign the image using a GPG key with the specified `FINGERPRINT`")
 	flags.StringVar(&opts.digestFile, "digestfile", "", "Write the digest of the pushed image to the specified file")
@@ -80,6 +82,27 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
 	flags.IntSliceVar(&opts.encryptLayer, "encrypt-layer", []int{}, "*Experimental* the 0-indexed layer indices, with support for negative indexing (e.g. 0 is the first layer, -1 is the last layer)")
 	flags.StringSliceVar(&opts.decryptionKeys, "decryption-key", []string{}, "*Experimental* key needed to decrypt the image")
 	return cmd
+}
+
+// parseMultiArch parses the list processing selection
+// It returns the copy.ImageListSelection to use with image.Copy option
+func parseMultiArch(multiArch string) (copy.ImageListSelection, error) {
+	switch multiArch {
+	case "system":
+		return copy.CopySystemImage, nil
+	case "all":
+		return copy.CopyAllImages, nil
+	// There is no CopyNoImages value in copy.ImageListSelection, but because we
+	// don't provide an option to select a set of images to copy, we can use
+	// CopySpecificImages.
+	case "index-only":
+		return copy.CopySpecificImages, nil
+	// We don't expose CopySpecificImages other than index-only above, because
+	// we currently don't provide an option to choose the images to copy. That
+	// could be added in the future.
+	default:
+		return copy.CopySystemImage, fmt.Errorf("unknown multi-arch option %q. Choose one of the supported options: 'system', 'all', or 'index-only'", multiArch)
+	}
 }
 
 func (opts *copyOptions) run(args []string, stdout io.Writer) error {
@@ -143,7 +166,17 @@ func (opts *copyOptions) run(args []string, stdout io.Writer) error {
 	if opts.quiet {
 		stdout = nil
 	}
+
 	imageListSelection := copy.CopySystemImage
+	if opts.multiArch.Present() && opts.all {
+		return fmt.Errorf("Cannot use --all and --multi-arch flags together")
+	}
+	if opts.multiArch.Present() {
+		imageListSelection, err = parseMultiArch(opts.multiArch.Value())
+		if err != nil {
+			return err
+		}
+	}
 	if opts.all {
 		imageListSelection = copy.CopyAllImages
 	}
--- a/docs/skopeo-copy.1.md
+++ b/docs/skopeo-copy.1.md
@@ -66,6 +66,17 @@ MANIFEST TYPE (oci, v2s1, or v2s2) to use in the destination (default is manifes
 
 Print usage statement
 
+**--multi-arch**
+
+Control what is copied if _source-image_ refers to a multi-architecture image. Default is system.
+
+Options:
+- system: Copy only the image that matches the system architecture
+- all: Copy the full multi-architecture image
+- index-only: Copy only the index
+
+The index-only option usually fails unless the referenced per-architecture images are already present in the destination, or the target registry supports sparse indexes.
+
 **--quiet**, **-q**
 
 Suppress output information when copying images.
--- a/integration/copy_test.go
+++ b/integration/copy_test.go
@@ -123,10 +123,10 @@ func (s *CopySuite) TestCopyAllWithManifestListRoundTrip(c *check.C) {
 	dir2, err := ioutil.TempDir("", "copy-all-manifest-list-dir")
 	c.Assert(err, check.IsNil)
 	defer os.RemoveAll(dir2)
-	assertSkopeoSucceeds(c, "", "copy", "--all", knownListImage, "oci:"+oci1)
-	assertSkopeoSucceeds(c, "", "copy", "--all", "oci:"+oci1, "dir:"+dir1)
-	assertSkopeoSucceeds(c, "", "copy", "--all", "dir:"+dir1, "oci:"+oci2)
-	assertSkopeoSucceeds(c, "", "copy", "--all", "oci:"+oci2, "dir:"+dir2)
+	assertSkopeoSucceeds(c, "", "copy", "--multi-arch=all", knownListImage, "oci:"+oci1)
+	assertSkopeoSucceeds(c, "", "copy", "--multi-arch=all", "oci:"+oci1, "dir:"+dir1)
+	assertSkopeoSucceeds(c, "", "copy", "--multi-arch=all", "dir:"+dir1, "oci:"+oci2)
+	assertSkopeoSucceeds(c, "", "copy", "--multi-arch=all", "oci:"+oci2, "dir:"+dir2)
 	assertDirImagesAreEqual(c, dir1, dir2)
 	out := combinedOutputOfCommand(c, "diff", "-urN", oci1, oci2)
 	c.Assert(out, check.Equals, "")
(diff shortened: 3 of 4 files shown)


cmd/skopeo/copy.go
```diff
@@ -72,16 +73,36 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
...
 	flags.StringSliceVar(&opts.decryptionKeys, "decryption-key", []string{}, "*Experimental* key needed to decrypt the image")
+	flags.MarkDeprecated("all", "please use --list-processing=CopyAllImages instead")
 	return cmd
 }
 
+// parseListProcessing parses the list processing selection
+// It returns the copy.ImageListSelection to use with image.Copy option
+func parseListProcessing(listProcessing string) (copy.ImageListSelection, error) {
+	switch listProcessing {
+	case "CopySystemImage":
```
mtrmac, line 92: I think it would be more convenient for the Skopeo option names to be shorter, and probably not capitalized. E.g. in this case, just `system` might work.

(The Go enum values have the Copy/Image words because the values are in the full module namespace, not scoped to the enum type, like in many other languages; and it’s not that much of an issue because IDEs have pretty good autocompletion, unlike the shell, at least right now. [But compare #1500 , eventually.])

cmd/skopeo/copy.go
```diff
@@ -72,16 +73,36 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
...
+		return copy.CopySystemImage, nil
+	case "CopyAllImages":
+		return copy.CopyAllImages, nil
+	// There is no CopyNoImages value in copy.ImageListSelection, but because we
+	// don't provide an option to select a set of images to copy, we can use
+	// CopySpecificImages as CopyNoImages.
+	case "CopySpecificImages", "CopyNoImages":
```
mtrmac, line 99: Providing `CopySpecificImages` without a way to set the list doesn’t seem useful. We can just not implement it, comment here that it is not supported, and if someone ends up needing it (I’m not quite sure that will ever happen, BTW — it seems more useful when integrated into a complex tool in Go), we can add it later.

cmd/skopeo/copy.go
```diff
@@ -72,16 +73,36 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
 	flags.StringSliceVar(&opts.additionalTags, "additional-tag", []string{}, "additional tags (supports docker-archive)")
 	flags.BoolVarP(&opts.quiet, "quiet", "q", false, "Suppress output information when copying images")
 	flags.BoolVarP(&opts.all, "all", "a", false, "Copy all images if SOURCE-IMAGE is a list")
+	flags.StringVar(&opts.listProcessing, "list-processing", "CopySystemImage", "How to process image lists. Choose one of CopySystemImage, CopyAllImages, or CopyNoImages")
```
mtrmac, line 76: The option name can probably be better.

`list` is not really how users think of multi-arch images; I don’t think `processing` adds any information at all.

Maybe `skopeo copy --multi-arch={system-only|all|none)`? Others can probably suggest better wording.

cmd/skopeo/copy.go
```diff
@@ -72,16 +73,36 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
...
+		return copy.CopySystemImage, nil
+	case "CopyAllImages":
+		return copy.CopyAllImages, nil
+	// There is no CopyNoImages value in copy.ImageListSelection, but because we
+	// don't provide an option to select a set of images to copy, we can use
+	// CopySpecificImages as CopyNoImages.
+	case "CopySpecificImages", "CopyNoImages":
+		return copy.CopySpecificImages, nil
+	default:
+		return copy.CopyAllImages, fmt.Errorf("unknown image list processing option %q. Choose one of the supported options: 'CopySystemImage', 'CopyAllImages', or 'CopyNoImages'", listProcessing)
```
mtrmac, line 102: (Non-blocking: This might be possible to implement as a map[string]copy.ImageListSelection`; that would also allow constructing the list of supported options here and in the flag definition in a way that doesn’t require updating three places.

But, sure, this works well enough, and `parseManifestFormat` does the same thing.)

mtrmac (commented):
Thanks!

Do I understand correctly that the primary goal is to provide the “CopyNoImages” value?

---

> I went with deprecating the --all flag because I thought it was cleaner, but open to suggestions on UX. My other thoughts were to leave them both in and throw an error if they are used together, or to add a `--list-only` to just copy the list.

I think the `--all` flag is useful enough, and used frequently enough that we should leave this short option name available as the primary documented approach. Refusing an invalid combination would work well.

WRT `--list-only` vs. (some name of) `--list-processing=…`, I don’t have a strong preference at this point; adding a single option with alternatives, like this PR does, instead of several top-level options with no obvious relationship to each other, seems a bit more future-proof / easier to scale to new features.

Jamstah:
> Do I understand correctly that the primary goal is to provide the “CopyNoImages” value?

Yes, exactly. I am looking for a solution for mirroring images from official registries for running software offline that are digest-stable but don't need all architectures to be mirrored.

cmd/skopeo/copy.go
```diff
@@ -143,7 +163,17 @@ func (opts *copyOptions) run(args []string, stdout io.Writer) error {
 	if opts.quiet {
 		stdout = nil
 	}
+
 	imageListSelection := copy.CopySystemImage
+	if opts.multiArch.Present() && opts.all {
+		return fmt.Errorf("Cannot use all and multi-arch flags together")
```
mtrmac, line 169:
```suggestion
		return fmt.Errorf("Cannot use --all and --multi-arch flags together")
```

cmd/skopeo/copy.go
Jamstah: I'm not sure familiar with cobra/pflags, I couldn't find a simple "Here's how to do a multiple choice", so I copied the manfiest format.

If you know of any doc I'd be happy to try it out :)

cmd/skopeo/copy.go
Jamstah: True, I was lost in the API layer. I like your suggestion but I think `none` should be `index-only` - because we're not copying nothing.

Refactored.

cmd/skopeo/copy.go
```diff
@@ -72,6 +73,7 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
 	flags.StringSliceVar(&opts.additionalTags, "additional-tag", []string{}, "additional tags (supports docker-archive)")
 	flags.BoolVarP(&opts.quiet, "quiet", "q", false, "Suppress output information when copying images")
 	flags.BoolVarP(&opts.all, "all", "a", false, "Copy all images if SOURCE-IMAGE is a list")
+	flags.Var(commonFlag.NewOptionalStringValue(&opts.multiArch), "multi-arch", `How to handle multi-architecture indexes (system, all, or index-only)`)
```
mtrmac, line 76:
```suggestion
	flags.Var(commonFlag.NewOptionalStringValue(&opts.multiArch), "multi-arch", `How to handle multi-architecture images (system, all, or index-only)`)
```

cmd/skopeo/copy.go
Jamstah: Yes, I was just trying to match up the options from image, but really you're right - this should be skopeo UX not image internals.

docs/skopeo-copy.1.md
```diff
@@ -66,6 +66,15 @@ MANIFEST TYPE (oci, v2s1, or v2s2) to use in the destination (default is manifes
 
 Print usage statement
 
+**--multi-arch**
+
+If _source-image_ refers to an index of images for multiple architectures, this option controls what is copied. Default is system.
```
mtrmac, line 71:
```suggestion
Control what is copied if _source-image_ refers to a multi-architecture image. Default is system.
```

docs/skopeo-copy.1.md
```diff
@@ -66,6 +66,15 @@ MANIFEST TYPE (oci, v2s1, or v2s2) to use in the destination (default is manifes
...
+
+Options:
+- system: Copy the image that matches the system architecture instead of the index
+- all: Copy all of the images in the index and the index itself
+- index-only: Copy only the index itself
```
mtrmac, line 76: This should likewise talk about images instead of the details of the index/manifest list / … terminology (except for the index-only case).

mtrmac, line 76: Add something like “… . Note that this usually fails unless the referenced per-architecture images are already present in the destination.”?

cmd/skopeo/copy.go
Jamstah: Yes, autocompletion would be a good reason to have the options recognised by cobra. If you know how I'm happy to try it.

Jamstah:
> I think the --all flag is useful enough, and used frequently enough that we should leave this short option name available as the primary documented approach. Refusing an invalid combination would work well.

Yes, I was probably too hasty deprecating it, its commonly used enough and does no harm.

docs/skopeo-copy.1.md
```diff
@@ -66,6 +66,15 @@ MANIFEST TYPE (oci, v2s1, or v2s2) to use in the destination (default is manifes
 
 Print usage statement
 
+**--multi-arch**
```
mtrmac, line 69: Also add the item to `completions/bash/skopeo`, please.

cmd/skopeo/copy.go
mtrmac: At this point I think we can do without; only after we have the basic infrastructure established it would be reasonable to ask new flags to conform to established practice.

Of course if you want to figure out how this is done, that would be welcome as well :)

cmd/skopeo/copy.go
mtrmac: Non-blocking: Maybe just add a comment above the `default`:
> We don't expose CopySpecificImages other than index-only above, because we currently don't provide an option to choose the images to copy.  That could be added in the future.

This PR is already a good enough record of that thought process; a comment directly in the code would make it a bit easier to find.

cmd/skopeo/copy.go
rhatdan: I am not sure what system (Native arch?) means, but -only is implied.
`--multi-arch={system|all|none)` would make more sense to me.

Jamstah: To me, --multi-arch is "what do we do if we're copying a multi-arch (image index) manifest not a standard image.

system = replace it with whatever image from the index matches the running system

all = copy the index and all the architecture

none to me would mean "don't do anything", when actually, the option we want is "copy the index, but don't copy any of the images in the index".

That's why I went with index-only, but happy to change - the doc will explain it clearly anyway.

Thoughts?

cmd/skopeo/copy.go
Jamstah: Technically, its not a multi-architcture image, its an index of images (or manifest list, or fat manifest).

Do you think for skopeo users it would be better UX to gloss over that? I'm happy to go with it if you're sure :)

Jamstah: Looking at your further comments, you seem sure, I'll go with it :)

integration/copy_test.go
```diff
@@ -103,13 +103,20 @@ func (s *CopySuite) TestCopyWithManifestList(c *check.C) {
...
 
-func (s *CopySuite) TestCopyAllWithManifestList(c *check.C) {
+func (s *CopySuite) TestCopyAllFlagWithManifestList(c *check.C) {
 	dir, err := ioutil.TempDir("", "copy-all-manifest-list")
 	c.Assert(err, check.IsNil)
 	defer os.RemoveAll(dir)
 	assertSkopeoSucceeds(c, "", "copy", "--all", knownListImage, "dir:"+dir)
 }
 
+func (s *CopySuite) TestCopyAllWithManifestList(c *check.C) {
```
mtrmac, line 113: BTW I’m not sure we need this; remote image fetches are somewhat costly, and this is just a minimal test that the command doesn’t report failure without any verification of the contents.

There is a command with `--all` above, and `--multi-arch=all` elsewhere, so that’s probably sufficient.

(Do leave the new `TestCopyNoneWithManifestList` test around, it’s much cheaper and it’s unique.)

cmd/skopeo/copy.go
mtrmac: I agree index-only (metadata-only ? top-level-only? Docker schema2 doesn’t name it “index”) is better than “none”.

I think it’s going to be used fairly rarely, and in unusual situations, so being accurate might be a bit more valuable than being easy to type.

cmd/skopeo/copy.go
mtrmac: I was going to say “we don’t document `skopeo copy` to copy manifest/config/layers individually, it’s a tool to copy container images and the rest are implementation details” — except that the man page does happen to talk about the details :)

Overall, yes, I don’t think most users care, nor should they have to care. A multi-arch/single-arch image is a clear enough concept. Talking about OCI indices / schema2 manifest lists introduces terminology most users don’t care about, especially when those items can’t even stand alone (registries usually don’t allow uploading indices without the per-arch images).

Users of the index-only option do care; I don’t think we should optimize for that case, making the simpler cases harder to understand.

mtrmac (commented):
LGTM.

(I’d appreciate a second pair of yes on the naming and documentation.)

integration/copy_test.go
Jamstah: Agreed on the extra test, when I first added it I didn't realise how expensive they are.

Jamstah:
@mtrmac I also want to say thanks for the specific, thoughtful and speedy reviews (here and over in containers/image), it's always more fun coding with a friend!

Jamstah:
Are the tests unstable? These failures don't look related.

integration/test/version
```diff
@@ -0,0 +1 @@
+Directory Transport Version: 1.1
```
mtrmac, line 1: @Jamstah Is this file necessary? I can’t see any user.

If it is, please move it into `integration/fixtures`, as `integration/fixtures/some-descriptive-name`.

mtrmac:
@vrothberg PTAL when you get a chance.

integration/test/version
Jamstah: Sorry, that's a file created by running the integration tests. Should probably be added to .gitignore.

mtrmac:
> @vrothberg PTAL when you get a chance.

… but #1517 should be merged first.

vrothberg (commented):
Code LGTM but I am not convinced by the hardest part: the naming :^)

`--multi-arch` restricts an image index to only be used for providing multiple architectures and while they are used like that in most cases, there may also be embedded indexes or OCI artifacts (e.g., certain signatures, helm charts, source images, etc.).

Could we name the flag `--index` or `--image-index`?

mtrmac:
> there may also be embedded indexes or OCI artifacts (e.g., certain signatures, helm charts, source images, etc.).

`copy.Image` handles none of these structures right now:
- `skopeo copy` is not the right tool for copying a collection of images like a full repo
- If indices are used to point at signatures (e.g. the Notary v2 “reverse reference” idea), I’d expect `skopeo copy` to do that transparently as an implementation detail of signatures.
- (I’m rather unclear on what nested OCI indexes are supposed to mean. It’s undefined/non-interoperable, AFAICT.)
- What does `--index=system` mean for helm charts / source images?

I’m thinking it’s preferable to focus on user-relevant outcomes over internal implementation: “Index” is a somewhat weird jargon for a collection of images, especially it _already_ isn’t a perfect match because it’s not what v2s2 calls the corresponding blob.

But then again the code already has an `index-only` option value, so it would be consistent, at least.

So I don’t feel strongly about this — we can always add one more (mutually exclusive) option. It just feels more natural to me have an ergonomic option for the case we do have, rather than try to predict a universal option for use cases we can’t yet support and don’t fully understand.

vrothberg:
> What does --index=system mean for helm charts / source images?

I don't have a definite answer to that.  Maybe a future `--index=artifact` (possibly with another option to further discriminate).

Note that I do not feel strongly about that but `--index` would be a minor preference.  Extending in the future if needed SGTM

Jamstah:
I'm happy with either, but if we only do multi arch now kind of lean to staying with that. We can design the clean options for a --index when it happens and deprecate this one.

mtrmac:
In the back of my head is a future feature like `--multi-arch=[amd64]` to do partial copies per-arch, instead of the caller manually scripting multi-arch digest parsing. That works better when the flag name is more constrained. Also `--index=system` reads weird (but then why would anyone type that, when it’s the default?).

OTOH the one case we actually care about, `--index=index-only` really isn’t _only_ multi-arch specific, so that would actually be a better name.

Let’s give the US folks a few hours to catch up?

But really I’d rather have the PR with _any_ name than delay it until we have perfect clarity, that seems unlikely to happen soon. So if `--index` gets that done, I’m perfectly fine with it.

Jamstah:
Well I'm new here, so will follow guidance of the project. LMK when you think there's enough consensus from maintainers and I'm happy to adjust the PR to suit.

rhatdan:
I like --multi-arch=[amd64]

Jamstah:
We'll need a containers/image PR to add the ability to select architectures, at the moment to copy multiple images you have to specifically provide the digests instead of the architectures: https://github.com/containers/image/blob/main/copy/copy.go#L138

But I agree that the extension to the `--multi-arch` flag of passing an array is a good one. I'll find time to make the update to containers/image to support it in a future skopeo pr.

mtrmac:
No no no, that was meant just an imagined future feature to motivate a specific approach to naming the _current_ option; I didn’t want to imply that the scope of this work needs to be extended.

Jamstah:
> No no no, that was meant just an imagined future feature to motivate a specific approach to naming the _current_ option; I didn’t want to imply that the scope of this work needs to be extended.

Yes, I meant for a future pr. Its an option I'd find useful, but wasn't planning to try and fit it in here :)

mtrmac (approved):

mtrmac:
I’ll just make the call so that this gets in: Merging `--multi-arch`. (There’s still some, unspecified, time left, to change it before the next release.)

@Jamstah thanks again!

PR #1518, opened 2021-11-26
Stop test producing output in source directory

Jamstah:
Use a dynamic temp dir for test

This test was incorrectly assuming that nothing would be made on disk,
but it was putting files into the source directory.


mtrmac (commented):
Thanks for taking this on.

I think tests should not be writing into the code directories in the first place; I’d prefer to fix that instead.

It seems that this is `TestCopyNoPanicOnHTTPResponseWithoutTLSVerifyFalse` making a no-longer-correct assumption that a source failure makes the destination completely irrelevant; would it work to change that to use a temporary directory in `ioutil.TempDir`, like all other tests, instead of hard-coding `dir:test`?

Jamstah:
Sure, clean engineering ftw.

mtrmac (approved):
LGTM. Thanks!

PR #1520, opened 2021-12-01
Add option to preserve digests on copy and sync

Jamstah:
When enabled, if digests can't be preserved an error will be raised.

### DIFF
--- a/docs/skopeo-copy.1.md
+++ b/docs/skopeo-copy.1.md
@@ -54,6 +54,10 @@ Directory to use to share blobs across OCI repositories.
 
 After copying the image, write the digest of the resulting image to the file.
 
+**--preserve-digests**
+
+Preserve the digests during copying. Fail if the digest cannot be preserved.
+
 **--encrypt-layer** _ints_
 
 *Experimental* the 0-indexed layer indices, with support for negative indexing (e.g. 0 is the first layer, -1 is the last layer)
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -33,6 +33,7 @@ type copyOptions struct {
 	quiet               bool                      // Suppress output information when copying images
 	all                 bool                      // Copy all of the images if the source is a list
 	multiArch           commonFlag.OptionalString // How to handle multi architecture images
+	preserveDigests     bool                      // Preserve digests during copy
 	encryptLayer        []int                     // The list of layers to encrypt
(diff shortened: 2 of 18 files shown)


docs/skopeo-copy.1.md
```diff
@@ -54,6 +54,10 @@ Directory to use to share blobs across OCI repositories.
 
 After copying the image, write the digest of the resulting image to the file.
 
+**--preserve-digests**
+
+Preserve the digests during copying. Fail if the digest cannot be preserved.
```
mtrmac, line 59: Tentative suggestion, I’m not at all sure: `--preserve-digests` is a bit of typing. Maybe `--exact`?

OTOH users who do care about this almost certainly know about digests.

@vrothberg WDYT?

mtrmac (commented):
Thanks! Implementation LGTM.

docs/skopeo-copy.1.md
vrothberg: I am very unhelpful since I like both :^)

vrothberg (commented):
LGTM

Jamstah:
Rebased for conflicts.

vrothberg (commented):

@mtrmac @rhatdan PTanotherL

mtrmac:
Oh — @Jamstah , could you add this to `skopeo sync` as well, please? It’s even more likely to be useful there. (Just a CLI option is fine for now.)

mtrmac:
WRT naming, I think let’s wait for a day or so and then merge as proposed if no strong opinion comes up.

Jamstah:
> Oh — @Jamstah , could you add this to `skopeo sync` as well, please? It’s even more likely to be useful there. (Just a CLI option is fine for now.)

Not sure what you meant by Just a CLI option, but done, just because I like you :)

I considered it for multi-arch too, but I feel like that wouldn't be useful for sync until we can specify a list of platforms we care about.

mtrmac:
> Not sure what you meant by Just a CLI option

Exactly this (i.e. not a new field in the YAML input format). Thanks.

---

> I considered it for multi-arch too, but I feel like that wouldn't be useful for sync until we can specify a list of platforms we care about.

I agree — a single `skopeo copy` can reasonably be a building block for a script building a partial image, but `skopeo sync` works on a large number of images, so it needs to handle each single image without much external help.

mtrmac:
Thanks again!

PR #1523, opened 2021-12-06
proxy: Add a GetFullConfig method

cgwalters:
Sadly...I swear I had tested this at one point, but it was
*definitely* not the intention that we just return the container
runtime configuration.

I need a method to return the full image configuration.  At some point
I must have accidentally added a redundant `.Config`.

This whole new method `GetFullConfig` is like `GetConfig` but
returns the whole image configuration.  A specific motivation
here is that it's only in the image configuration that we can
stick arbitrary metadata (labels) that will survive a round trip through
docker schema v2.

### DIFF
--- a/cmd/skopeo/proxy.go
+++ b/cmd/skopeo/proxy.go
@@ -430,7 +430,45 @@ func (h *proxyHandler) GetManifest(args []interface{}) (replyBuf, error) {
 	return h.returnBytes(digest, serialized)
 }
 
-// GetConfig returns a copy of the image configuration, converted to OCI format.
+// GetFullConfig returns a copy of the image configuration, converted to OCI format.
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
LGTM

rhatdan:
LGTM

cgwalters:
When can we turn the crack on a new skopeo release?  I'd like to be able to use this code.

rhatdan:
Everything is being lined up to be released with Podman, we like to coordinate the release of containers/image, containers/storage and containers/common.  Bottom line, I think later this week.
Plan is to release Buildah on Wednesday, so I would figure the same for skopeo.

mtrmac:
A release should finally happen in #1561  .

Issue #1498, opened 2021-11-09
Performance improvements for large sync jobs


adriangb:
I recently had the need to sync multiple images across registries, and this seems like the best tool for the job. Thank you to all of the developers!

For my use case, I have several images, each with dozens to hundreds of tags.
Currently, syncing is basically impossible because:

1. The initial sync takes a long time because there are a lot of images/tags to sync. This could be mitigated by parallelizing as suggested in https://github.com/containers/skopeo/pull/1445#issuecomment-919953326

2. Subsequent syncs take a very long time even if nothing is copied in the end. I believe this is because after getting the tags from the source, each one is checked against the destination individually. I wonder if it would be possible to get the tags from the destination and pre-compute the missing ones so that a no-op sync would be almost instant.

Unfortunately I am not a Gopher and know almost nothing about registry APIs, so I don't think I can implement this myself, but I wanted to leave the suggestion.

Thanks!

mtrmac:
Thanks for your report.

Just to be sure, please make sure to use Skopeo ≥ 1.2.3, to benefit from #1189.

>  I wonder if it would be possible to get the tags from the destination and pre-compute the missing ones so that a no-op sync would be almost instant.

That wouldn’t work because listing tags just lists the tag names, but finding out what image the tag points to requires another per-image round-trip.

---

The underlying image client code is primarily single-image focused and it could almost certainly be more efficient when dealing with repos that has many tags (approximately going from 3 HTTP requests to 1).

Beyond that, actually measuring and profiling the performance might well uncover more optimization possibilities.

adriangb:
> Just to be sure, please make sure to use Skopeo ≥ 1.2.3, to benefit from #1189.

I was using v1.4.1

> That wouldn’t work because listing tags just lists the tag names, but finding out what image the tag points to requires another per-image round-trip.

Just to be clear, you're referring to the situation where the tag exists but the image is completely different? I guess I hadn't thought of that, we use tags as if they were immutable and never duplicate them.

I guess maybe I can manually list tags for both the source and destination, compute the difference and then launch multiple processes to do the sync on each image.

mtrmac:
> Just to be clear, you're referring to the situation where the tag exists but the image is completely different?

Yes.

> I guess I hadn't thought of that, we use tags as if they were immutable and never duplicate them.

The `:latest` convention, at least, is _very_ widespread.

[...]

PR #1531, opened 2021-12-30
feature: add --scoped-append-registry flag

art-shutter:
Introduces a new flag to simplify work with YAML source as referenced in https://github.com/containers/skopeo/issues/1072

Signed-off-by: Artem Kamenev <artem@kamenev.pw>

### DIFF
--- a/cmd/skopeo/sync.go
+++ b/cmd/skopeo/sync.go
@@ -42,6 +42,7 @@ type syncOptions struct {
 	source                   string                    // Source repository name
 	destination              string                    // Destination registry name
 	scoped                   bool                      // When true, namespace copied images at destination using the source repository name
+	scopedAppendRegistry     bool                      // When true, omits repository name from final namespace when using skoped
 	all                      bool                      // Copy all of the images if an image in the source is a list
 	dryRun                   bool                      // Don't actually copy anything, just output what it would have done
 	preserveDigests          bool                      // Preserve digests during sync
@@ -111,7 +112,8 @@ See skopeo-sync(1) for details.
 	flags.VarP(commonFlag.NewOptionalStringValue(&opts.format), "format", "f", `MANIFEST TYPE (oci, v2s1, or v2s2) to use when syncing image(s) to a destination (default is manifest type of source, with fallbacks)`)
 	flags.StringVarP(&opts.source, "src", "s", "", "SOURCE transport type")
 	flags.StringVarP(&opts.destination, "dest", "d", "", "DESTINATION transport type")
-	flags.BoolVar(&opts.scoped, "scoped", false, "Images at DESTINATION are prefix using the full source image path as scope")
+	flags.BoolVar(&opts.scoped, "scoped", false, "Images at DESTINATION are prefixed using the full source image path as scope")
+	flags.BoolVar(&opts.scopedAppendRegistry, "scoped-append-registry", true, "Add root registry name to DESTINATION image prefix when using --scoped")
 	flags.BoolVarP(&opts.all, "all", "a", false, "Copy all images if SOURCE-IMAGE is a list")
 	flags.BoolVar(&opts.dryRun, "dry-run", false, "Run without actually copying data")
 	flags.BoolVar(&opts.preserveDigests, "preserve-digests", false, "Preserve digests of images and lists")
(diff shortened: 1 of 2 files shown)


mtrmac (changes requested):
Thanks, but as discussed in https://github.com/containers/skopeo/issues/854#issuecomment-604903514 , I don’t think it makes sense to continue adding individual options for every mirror organization options a user might choose to design.

mtrmac:
For YAML, contrary to the existing `images` and `images-by-tag-regex`, we need a way to add more per-repo options.

Maybe something vaguely like (based on the man page example):
```yaml
registry.example.com:
    repos:
        busybox:
            destination: mirror.local/busybox
            assume-mutable-tags: latest
        redis:
            destination: mirror.local/redis
            mirror-tags: ["1.0", "2.0"]
            mirror-digests: ["sha256:0000000000000000000000000000000011111111111111111111111111111111"]
        nginx:
            destination: mirror.local/completely/different/policy/nginx
            mirror-by-tag-regex: ^1\.13\.[12]-alpine-perl$
    credentials:
        username: john
        password: this is a secret
    tls-verify: true
    cert-dir: /home/John/certs
```
with a per-repo `destination` directly in the config file, not on the command line.

This would also allow adding per-repo `assume-mutable-tags` options discussed in #1498 , and other options over time.

art-shutter:
I would agree that supplying complete destination in the YAML itself looks better.
I suggest the following:

- add global settings section for repos
- add two destination-related keys
  - `custom-destination` with value of any custom path
  -  `destination-path-style` with these possible values: `"full", "full-no-registry", "flat"`
- `custom-destination` would take precedence over `destination-path-style`
- default behavior would be as it is now, `destination-path-style: "flat"`

I would also raise the question of deprecating/refactoring the `--scoped` flag. As of now, if would leave it and when parsing YAML it would, if neither destination-related keys are specified, switch default behavior from `destination-path-style: "flat"` to `destination-path-style: "full"`

So, the resulting YAML might look something like 

```
registry.example.com:
    global-settings:
        destination-path-style: full
        mirror-tags: ["latest"]
        assume-mutable-tags: ["latest"]
    repos:
        busybox:
            custom-destination: mirror.local/busybox
            assume-mutable-tags: ["best-tag"]
        redis:
            destination-path-style: flat
            mirror-tags: ["1.0", "2.0"]
            mirror-digests: ["sha256:0000000000000000000000000000000011111111111111111111111111111111"]
        nginx:
            mirror-by-tag-regex: ^1\.13\.[12]-alpine-perl$
    credentials:
        username: john
        password: this is a secret
    tls-verify: true
    cert-dir: /home/John/certs
```


However, I would leave `assume-mutable-tags` functionality outside the scope of this PR.

@mtrmac What do you think?

anisimovdk:
I agree with @art-shutter.
In our case we need to sync a lot of images to a bunch of registries, that why defining `destination` per image looks bad.
Global settings much more preferable or defining destination registry by argument from cli.

ToroNZ:
any chance we can have something while the ideal solution is worked on? this kinda ruins the tidyness of using the YAML source option.

DavidSpek:
I'm also needing this feature and it'd be great if this PR can be merged.

art-shutter:
since there hasn't been much happening in terms of updating the yaml to the new format, I have updated changes here with the default branch and have also proposed another per-repo solution in https://github.com/containers/skopeo/pull/1769

mhowell24:
I would be interested in working on the updated yaml format mentioned above if nobody else already is. @mtrmac would you accept the PR if I were to put one together?

mtrmac:
I do think it is valuable to finally solve the `sync` repo mapping deficiencies, but I’m afraid it’s also not something I can spend sufficient time on in the next few weeks or likely months.

@vrothberg do you have the time to help make this happen?

art-shutter:
@mhowell24 I would definitely like to help, be it reviewing/testing or anything else.
@mtrmac are we ok with introducing breaking changes to manifests?

mtrmac:
If you are talking about the YAML format read by `skopeo copy`, I think breaking existing scripts is very undesirable, and easily avoidable — if nothing else, `sync --source yaml2` allows us to define a new format without affecting the old one at all.

mhowell24:
Brilliant, thanks @art-shutter and @mtrmac I will try and put something together this week

mhowell24:
Hi @art-shutter and @mtrmac I have opened #1792 as a draft to implement the yaml config from above, let me know what you think

PR #1536, opened 2022-01-06
Update github.com/containerd/containerd to 1.5.9

mtrmac:
```
go get github.com/containerd/containerd@latest
make vendor
```
    
... because 1.5.9 contains a vulnerability fix, and we want to silence scanners.
    
NOTE: Skopeo DOES NOT use the vulnerable code that was fixed in containerd 1.5.9, so it is NOT vulnerable to GHSA-mvff-h3cj-wj9c .

---

This PR also updates opencontainers/image-spec to fix using c/image along with other Go modules that reference the 1.0.2 release tag, like https://github.com/containers/buildah/pull/3678 (that includes this containerd update).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/containers/skopeo
 go 1.12
 
 require (
-	github.com/containerd/containerd v1.5.8 // indirect
+	github.com/containerd/containerd v1.5.9 // indirect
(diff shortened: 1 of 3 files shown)


TomSweeneyRedHat:
Cast ye away evil scanners!  ;^)
LGTM

rhatdan:
LGTM

Issue #1261, opened 2021-04-23
Automating the  gpg  Passphrase for signing images in pipelines


akshaykumar-vijapur:
Hi team,

I am using this `gpg2 --pinentry-mode=loopback  --passphrase ***  --import private.key`  to automatically load the privatekey into keyring without user intervention for password. 

I am trying to sign the images using the Skopeo 
```
 skopeo copy docker://"${image}" docker://"${image}" --dest-creds iamapikey:"${CLOUD_API_KEY}" --src-creds iamapikey:"${CLOUD_API_KEY}" --sign-by "signer@company.com"
```

This is able to sign my image correctly. But it is asking password in the prompt. Is there any way I can automate this password entering in Skopeo?

TomSweeneyRedHat:
@akshaykumar-vijapur which password is it prompting you for? For the registry server that you're copying to?  If so, can you do a `skopeo login` to there first?  Or is it a credentials getting requested from you gpg2?

akshaykumar-vijapur:
@TomSweeneyRedHat It is not about registry. I'm taking about  GPG password prompt which will be prompted to enter GPG password which was used to create the GPG keys. 

That prompt can be overridden by using this command 
```
gpg2 --pinentry-mode=loopback --passphrase *** 
```

But not sure weather this is supported by `skopeo copy`.

TomSweeneyRedHat:
Thanks, @akshaykumar-vijapur for the info.  I don't believe there's a way to do that yet but sounds like a nice RFE if not.

akshaykumar-vijapur:
@TomSweeneyRedHat  Please let me know if I need to open another issue for this RFE ?

rhatdan:
@akshaykumar-vijapur Interested in opening a PR to make this happen?

TomSweeneyRedHat:
@akshaykumar-vijapur no need for an additional issue, I'll just note RFE in the title for this one and will label it so.

akshaykumar-vijapur:
Any timeline for this feature ?

mtcolman:
Hi, we are hitting this limitation too...

mtcolman:
You can provide it with `skopeo copy <params/flags> | cat /path/to/password.txt` however the password will show on screen.

mtcolman:
Linking to this issue: https://github.com/containers/image/issues/1303

mtcolman:
Hi, a colleague (@SionArtingstall) and I have finally found a way round this issue:

[...]

PR #1540, opened 2022-01-19
prompt-less signing via passphrase file

vrothberg:
To support signing images without prompting the user, add CLI flags for
providing a passphrase file.


systemtest/050-signing.bats
```diff
@@ -12,6 +12,8 @@ function setup() {
     export GNUPGHOME=$TESTDIR/skopeo-gpg
     mkdir --mode=0700 $GNUPGHOME
 
+    export CONTAINERS_SIGNATURE_PASSPHRASE="MakeTomHappy"
```
mtrmac, line 15: Hum… this is convenient, and it’s not crossing any security boundary AFAICT, but it also puts a copy of that secret data to every single process executed in this shell. Consider Heartbleed, that kind of unnecessary exposure is not desirable.

Of course it’s _possible_ to use the environment variable in a restricted context but it’s also much easier to do just what this code does.

So I think it should at least be _possible_ to pass the passphrase in a file (OTOH then we need to worry about including, or stripping, a trailing newline — probably do what `gpg2 --passphrase-file` does), and I’m not sure if we should allow environment variables at all.

vrothberg: Sure, we can restrict it to a file directly.

OTOH I am not sure if it would change much if used insecurely. Exporting a variable as done in the test is nice for testing but absolutely dangerous in production.

I can alter the c/image logic and turn it into `CONTAINERS_SIGNATURE_PASSPHRASE_FILE` if you prefer that.

mtrmac: I was primarily thinking a `skopeo copy --passphrase-file=…` option; direct and explicit.

But I don’t feel strongly against a …`_FILE` environment variable. We do support a few environment variables, so, \*shrug\*.

TomSweeneyRedHat:
Does this potentially address #1261 ?

TomSweeneyRedHat:
I should have read deeper into my inbox, per @vrothberg this will address #1261

systemtest/050-signing.bats
akshayvijapur: @mtrmac @vrothberg Kindly provide an option to add the passphrase directly in the skopeo copy command 

```
  skopeo copy --pinentry-mode=loopback --passphrase ***
```

mtrmac: That makes the passphrase visible to _every single user_ on a system via `/proc/*/cmdline`.

I’m not saying there are absolutely no legitimate use cases, but if we ever add this, it would be named something like `--dangerous-publicly-visible-passphrase=`.

vrothberg:
@mtrmac updated:
* using github.com/proglottis/gpgme
* the env variable is now pointing to a passphrase file instead

I am cool using an env var. This way, all consumers support it by simply vendoring c/image.

mtrmac:
> I am cool using an env var. This way, all consumers support it by simply vendoring c/image.

I am unhappy with c/image providing options that are _only_ configurable using environment variables. (To be fair, there are a few such variables like HOME, but at least those are clearly not expected to differ within a process; that’s not the case with the passphrase.)

We don’t do that for every single option in `SystemContext`. What makes this one different? Most of the precedents like `--authfile` at least _allow_ using a CLI option.

vrothberg:
@mtrmac PTanotherL, now with CLI options.  Once we're cool with the code, I will take care of updating the docs as well.  Tests have been extended accordingly.

TomSweeneyRedHat:
Could we add a `--gpg-passphrase-file` or `--gpg-file` option too?  I'm not married to either name.  That might be handy in some circumstances.  Would that allay your concerns @mtrmac ?

vrothberg:
> Could we add a `--gpg-passphrase-file` or `--gpg-file` option too? I'm not married to either name. That might be handy in some circumstances. Would that allay your concerns @mtrmac ?

That's what the PR already does. There's a `--sign-passphrase-file` for copy/syn and a `--passphrase-file` for standalone-sign

cmd/skopeo/copy.go
```diff
@@ -78,6 +79,7 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
 	flags.BoolVar(&opts.preserveDigests, "preserve-digests", false, "Preserve digests of images and lists")
 	flags.BoolVar(&opts.removeSignatures, "remove-signatures", false, "Do not copy signatures from SOURCE-IMAGE")
 	flags.StringVar(&opts.signByFingerprint, "sign-by", "", "Sign the image using a GPG key with the specified `FINGERPRINT`")
+	flags.StringVar(&opts.signPassphraseFile, "sign-passphrase-file", "", "Path pointing to a passphrase file when signing")
```
mtrmac, line 82: Non-blocking? “Path of a file” might be sufficient, `--help` benefits from shorter text. “When signing” doesn’t say what it _is_ … “for the --sign-by key”? That’s fairly long.

In all the copies.

mtrmac (commented):
_Eventually_ this will need man page and bash completion updates, but let’s settle the c/image implementation first.

cmd/skopeo/copy.go
mtrmac: Well, or just something like “file that contains a passphrase for the --sign-by key”. The concept of paths can probably be safely assumed to be understood :)

vrothberg:
* vendored c/image/v5@main
* updated tests and added one with a wrong passphrase-file
* updated docs
* addressed previous comments on phrasing

vrothberg:
@TomSweeneyRedHat @mtrmac PTanotherL

cmd/skopeo/signing.go
```diff
@@ -46,7 +49,17 @@ func (opts *standaloneSignOptions) run(args []string, stdout io.Writer) error {
...
 	}
 	defer mech.Close()
-	signature, err := signature.SignDockerManifest(manifest, dockerReference, mech, fingerprint)
+
+	passphrase, err := cli.ReadPassphraseFile(opts.passphraseFile)
+	if err != nil {
+		return err
+	}
+
+	options := signature.SignOptions{
```
mtrmac, line 58: Absolutely non-blocking: Consider inlining this. This does work perfectly fine.

docs/skopeo-copy.1.md
```diff
@@ -93,6 +93,10 @@ Do not copy signatures, if any, from _source-image_. Necessary when copying a si
 
 Add a signature using that key ID for an image name corresponding to _destination-image_
 
+**--sign-passphrase-file**=_path_
+
+Passphare to use when signing with the key ID from `--sign-by`. Only the first line will be read. A passphrase stored in a file is of questionable security if other users can read this file. Do not use this option if you can avoid.
```
mtrmac, line 98: `s/if you can avoid it/& it/` ?

(Everywhere)

mtrmac (commented):
Code LGTM (and the added tests are great).

Please update `completions/bash/skopeo` as well.

vrothberg:
Updated :heavy_check_mark:

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.12
 
 require (
 	github.com/containers/common v0.47.0
-	github.com/containers/image/v5 v5.18.0
+	github.com/containers/image/v5 v5.18.1-0.20220125143129-bd97b5826c07
```
rhatdan, line 7: Can  you revendor v5.19.0?

vrothberg: Done :+1:

vrothberg:
> Please update completions/bash/skopeo as well.

Sigh ... I still need to do that.

vrothberg:
OK, now everything's in order.

mtrmac (approved):

docs/skopeo-copy.1.md
```diff
@@ -93,6 +93,10 @@ Do not copy signatures, if any, from _source-image_. Necessary when copying a si
 
 Add a signature using that key ID for an image name corresponding to _destination-image_
 
+**--sign-passphrase-file**=_path_
+
+Passphare to use when signing with the key ID from `--sign-by`. Only the first line will be read. A passphrase stored in a file is of questionable security if other users can read this file. Do not use this option if you can avoid it.
```
TomSweeneyRedHat, line 98: Sorry @vrothberg, you're breaking our "no pronouns" rule for man pages. Plus a tpyo on the passphrase
```suggestion
The passphrase to use when signing with the key ID from `--sign-by`. Only the first line will be read. A passphrase stored in a file is of questionable security if other users can read this file. Do not use this option if at all avoidable.
```

docs/skopeo-standalone-sign.1.md
```diff
@@ -25,6 +25,10 @@ Print usage statement
 
 Write signature to _output file_.
 
+**--passphrase-file**=_path_
+
+Passphare to use when signing with the specified key ID. Only the first line will be read. A passphrase stored in a file is of questionable security if other users can read this file. Do not use this option if you can avoid it.
+
```
TomSweeneyRedHat, line 31: ditto prior

docs/skopeo-sync.1.md
```diff
@@ -68,6 +68,8 @@ Print usage statement.
 
 **--sign-by**=_key-id_ Add a signature using that key ID for an image name corresponding to _destination-image_.
 
+**--sign-passphrase-file**=_path_ Passphare to use when signing with the key ID from `--sign-by`. Only the first line will be read. A passphrase stored in a file is of questionable security if other users can read this file. Do not use this option if you can avoid it.
+
```
TomSweeneyRedHat, line 72: ditto prior

TomSweeneyRedHat:
A few man page nits to pick before merging.

vrothberg:
Good to go :checkered_flag:

mrmadira:
What version of docker is compatible with skopeo 1.6.0?

When we installed latest docker CE (3:20.10.14-3.el8),  it bring along skopeo 1.4.0. When i upgrade skopeo - it is uninstalling Docker! 

This is on CentOS8.

PR #1550, opened 2022-01-26
add a SIF systemtest

vrothberg:
To make sure that the basic functionality is exercised in Skopeo and
c/image CI.


mtrmac (commented):
Thanks!

I’m not _too_ happy about having a 774 kB binary in the repo, but I don’t know about an alternative.

I suppose we could add it to https://github.com/containers/automation_images, but that would make the test significantly harder to follow, and 775 kB is not _that_ much.

---

I wonder a bit about the `podman run` part — wouldn’t it break on non-x86 architectures?

(Alternatively, maybe we could just copy `sif:` to `dir:`, and inspect the tar file to see that a specific file exists.)

Cc: @edsantiago

vrothberg:
> I wonder a bit about the podman run part — wouldn’t it break on non-x86 architectures?
>
> (Alternatively, maybe we could just copy sif: to dir:, and inspect the tar file to see that a specific file exists.)

Yes, the `run` indeed was already causing issues.  I'll go with dir + inspect.  We can re-iterate in the future.

vrothberg:
My workday is about to end, so I used a shortcut to only look for the architecture.  The digests (also of the layer) are different on each copy which makes testing a bit harder.

systemtest/020-copy.bats
```diff
@@ -144,6 +144,14 @@ function setup() {
     diff -urN $shareddir $dir2/blobs
 }
 
+@test "copy: sif image" {
+    local localimg=dir:$TESTDIR/dir
+
+    run_skopeo copy sif:systemtest/testdata/busybox_latest.sif $localimg
```
edsantiago, line 150: Danger: this will fail in gating tests, because those run while cd'ed to a different directory (which is and should be a perfectly fine thing to do). Suggestion:
```diff
diff --git a/systemtest/020-copy.bats b/systemtest/020-copy.bats
index 5d7ad6d5..d29a91ec 100644
--- a/systemtest/020-copy.bats
+++ b/systemtest/020-copy.bats
@@ -147,7 +147,7 @@ function setup() {
 @test "copy: sif image" {
     local localimg=dir:$TESTDIR/dir

-    run_skopeo copy sif:systemtest/testdata/busybox_latest.sif $localimg
+    run_skopeo copy sif:${TEST_SOURCE_DIR}/testdata/busybox_latest.sif $localimg
     run_skopeo inspect $localimg --format "{{.Architecture}}"
     expect_output "amd64"
 }
diff --git a/systemtest/helpers.bash b/systemtest/helpers.bash
index 2f57b215..a81c646e 100644
--- a/systemtest/helpers.bash
+++ b/systemtest/helpers.bash
@@ -1,6 +1,10 @@
 #!/bin/bash

-SKOPEO_BINARY=${SKOPEO_BINARY:-$(dirname ${BASH_SOURCE})/../skopeo}
+# Directory containing system test sources
+TEST_SOURCE_DIR=${TEST_SOURCE_DIR:-$(dirname ${BASH_SOURCE})}
+
+# Skopeo executable
+SKOPEO_BINARY=${SKOPEO_BINARY:-${TEST_SOURCE_DIR}/../bin/skopeo}

 # Default timeout for a skopeo command.
 SKOPEO_TIMEOUT=${SKOPEO_TIMEOUT:-300}
```
Download patch: [skopeo-1550.diff.txt](https://github.com/containers/skopeo/files/7943838/skopeo-1550.diff.txt)

IMPORTANT: carefully note and review the added `bin` in `SKOPEO_BINARY`. Looks like some time since the last I looked here, `make` writes into `bin`, not the root dir.

edsantiago (changes requested):
Heads up to @lsm5 and @jnovy that this will change the `skopeo-tests` package from `noarch` to `<arch>`. I don't think specfile changes will be necessary, because I see `cp -pav` in both Fedora and RHEL, but please be alert on the next build.

systemtest/020-copy.bats
vrothberg: Thank you!

vrothberg:
> Heads up to @lsm5 and @jnovy that this will change the `skopeo-tests` package from `noarch` to `<arch>`. I don't think specfile changes will be necessary, because I see `cp -pav` in both Fedora and RHEL, but please be alert on the next build.

I think it can remain `noarch` since the sif image is only copied, not being run in a container.

vrothberg:
Green.  @edsantiago @mtrmac PTanotherL

contrib/cirrus/runner.sh
```diff
@@ -61,6 +61,9 @@ _run_setup() {
     # VM's come with the distro. skopeo package pre-installed
     dnf erase -y skopeo
 
+    # Required for testing the SIF transport
+    dnf install -y fakeroot squashfs-tools
```
vrothberg, line 65: @cevich, could we add these two packages to the CI images?

edsantiago:
LGTM

rhatdan:
LGTM

contrib/cirrus/runner.sh
cevich: I can but...the images are very bloated as-is, and these are tools only needed for one test in one repo. I think in this case it might be worth a compromise to just live with some runtime installs...as much as it pains me.  (The bloat really is bad).

PR #1556, opened 2022-01-31
Update vendor of containers/storage and containers/common

rhatdan:


mtrmac (approved):
LGTM.

(Do we care about vendoring the updated c/storage into c/image, and updating that everywhere as well? I can’t think of a reason why would _have_ to — primarily I’m asking to know for sure when we are ready for a release.)

vrothberg:
@mtrmac, the primary reason is to not have to worry whether we need to care or not, I guess.

TomSweeneyRedHat:
FWIW, I do prefer bumping c/storage, getting that into c/image, then dancing up from there.  Especially for a big release.  It probably shouldn't matter, but I've too many scars from relying on probably not happening.

PR #1557, opened 2022-01-31
Improve the documentation of boolean flags

### LINKED ISSUE #1435: Does skopeo copy --dest-compress accept/require boolean value?
https://github.com/containers/skopeo/blob/main/docs/skopeo-copy.1.md says, that the `--dest-compress ` parameter accepts a boolean value.  `skopeo copy --help` says nothing about the boolen value.

Likewise for `--dest-decompress`.

mtrmac:
The Go behavior of boolean flags is as follows:

Accepted values are `--flag`, which is the same as `--flag=true`, and `--flag=false`, which is the default (except for `OptionalBoolFlag`). `--flag {false,true}` is parsed as `--flag=true` with a non-option {`false`,`true`} argument.

So, for almost all flags, document them just as `--flag`, not mentioning the [`=`{`false`,`true`}] part, because users can just
omit `=true`, or the whole flag instead of `=false`.

OTOH, for `tls-verify`, document only the `tls-verify=`{`true`,`false`} variant, because the primary use is `tls-verify=false`, and because `tls-verify` is not “the default”, but equivalent to an explicit `tls-verify=true` (overriding `registries.conf`).

Fixes #1435 .

### DIFF
--- a/docs/skopeo-copy.1.md
+++ b/docs/skopeo-copy.1.md
@@ -113,15 +113,15 @@ Key to be used for decryption of images. Key can point to keys and/or certificat
 
 Credentials for accessing the source registry.
 
-**--dest-compress** _bool-value_
+**--dest-compress**
 
 Compress tarball image layers when saving to directory using the 'dir' transport. (default is same compression type as source).
 
-**--dest-decompress** _bool-value_
+**--dest-decompress**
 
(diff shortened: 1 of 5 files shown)


rhatdan:
LGTM

vrothberg (commented):
LGTM

PR #1564, opened 2022-02-07
tests: skip sif test on RHEL

edsantiago:
(or, more precisely, if fakeroot binary not in $PATH).

Solves RHEL gating-test failure.

Also, fix rootless cleanup bug (need unshare, to avoid EPERM)

### DIFF
--- a/systemtest/020-copy.bats
+++ b/systemtest/020-copy.bats
@@ -125,6 +125,10 @@ function setup() {
     run podman --root $TESTDIR/podmanroot images
     expect_output --substring "mine"
 
+    # rootless cleanup needs to be done with unshare due to subuids
+    if [[ "$(id -u)" != "0" ]]; then
+        run podman unshare rm -rf $TESTDIR/podmanroot
+    fi
 }
 
 # shared blob directory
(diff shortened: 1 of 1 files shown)


lsm5 (approved):
LGTM. @mtrmac @vrothberg @rhatdan PTAL

edsantiago:
Dan suggested also checking for `squashfs-tools`, but on f35 that only provides binaries in `/usr/sbin` (which is not in default $PATH), and none of those binaries appear in a string search in the skopeo tree anyway, so I'm not sure what a sane check would look like.

rhatdan:
@vrothberg @mtrmac would know best.

mtrmac:
> Dan suggested also checking for `squashfs-tools`, but on f35 that only provides binaries in `/usr/sbin` (which is not in default $PATH), and none of those binaries appear in a string search in the skopeo tree anyway

The string I would expect to be found is `"unsquashfs -d %s -f %s && tar --acls --xattrs -C %s -cpf %s ./"`, and AFAICT `/usr/sbin` is in the default `PATH`, at least per https://pagure.io/setup/blob/master/f/profile#_42 . Is there some other mechanism that removes it again?

If `unsquashfs` isn’t in the default path, we should change the transport to find it in the default installation instead of forcing users to deal with that.

mtrmac (approved):
LGTM. Thanks!

edsantiago:
Ah yes - I see `unsquashfs` now. I must've fat-fingered it. Anyhow, thanks, if we see further failures I'll add that check.

TomSweeneyRedHat:
Thx for running these down to the ground @edsantiago !

Issue #1554, opened 2022-01-31
skopeo inspect returns digest of manifest list, not the resolved manifest


alexlarsson:
For example, the current `quay.io/centos/centos` points to an image manifest list called `sha256:8301d100...`, which points to an arm64 manifest called `sha256:7723d6b5...`.

But, If i run `skopeo --override-arch=arm64 inspect docker://quay.io/centos/centos` it only prints:
```
{
    "Name": "quay.io/centos/centos",
    "Digest": "sha256:8301d100...",
...
```

I expected to get `sha256:7723d6b5...`, either as "Digest" itself, or at least available *somewhere*.  Especially since the other information printed, like the layer digests are clearly from the image specific manifest.

alexlarsson:
Also, while on this topic, i would also like a way to get the digest of the config without having to do `skopeo inspect --config --raw | sha256sum`

mtrmac:
Thanks for your report.

> skopeo inspect returns digest of manifest list, not the resolved manifest 

Yes. That `skopeo inspect` contains a mix of manifest list data and per-arch data for the current arch is an unfortunate artifact of having to add multi-arch support in a hurry. Sadly we can’t very well change that without risking breaking existing users.

---

Adding per-arch-digest and config-digest fields to the output is a good idea.

mtrmac:
#1555 will at least make the documentation a bit clearer.

queeup:
Can't get correct digest hash.

```bash
$ podman manifest inspect quay.io/fedora/fedora-silverblue:40
{
    "schemaVersion": 2,
    "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
    "manifests": [
        {
            "mediaType": "application/vnd.oci.image.manifest.v1+json",
            "size": 33153,
            "digest": "sha256:36c453c465763adafbe4d1a19fd1837389172fd7cabe356a316bdae3be298150",
            "platform": {
                "architecture": "amd64",
                "os": "linux"
            }
        },
        {
            "mediaType": "application/vnd.oci.image.manifest.v1+json",
            "size": 32907,
            "digest": "sha256:7cc3ff1f666c3bfdf6705c0c4c8ab5841ca63336172c0e2a01c1909e7bb2b6eb",
            "platform": {
                "architecture": "arm64",
                "os": "linux"
            }
        },
        {
            "mediaType": "application/vnd.oci.image.manifest.v1+json",
            "size": 33523,
[...]

PR #1572, opened 2022-02-15
Don't expect the config blob to be listed in (skopeo inspect)

mtrmac:
... because it currently isn't.

(Eventually #1554, but need working CI in order to provide a c/image feature that can then be exposed in Skopeo.)

### DIFF
--- a/systemtest/010-inspect.bats
+++ b/systemtest/010-inspect.bats
@@ -34,9 +34,10 @@ load helpers
     # Each SHA-named layer file (but not the config) must be listed in the output of 'inspect'.
     # As of Skopeo 1.6, (skopeo inspect)'s output lists layer digests,
     # but not the digest of the config blob ($config_digest), if any.
+    layers=$(jq -r '.Layers' <<<"$inspect_local")
     for sha in $(find $workdir -type f | xargs -l1 basename | egrep '^[0-9a-f]{64}$'); do
         if [ "sha256:$sha" != "$config_digest" ]; then
-            expect_output --from="$inspect_local" --substring "sha256:$sha" \
+            expect_output --from="$layers" --substring "sha256:$sha" \
                         "Locally-extracted SHA file is present in 'inspect'"
         fi
     done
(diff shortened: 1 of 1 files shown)


mtrmac:
@cevich / @edsantiago PTAL.

systemtest/010-inspect.bats
```diff
@@ -27,11 +27,16 @@ load helpers
...
     inspect_local=$output
+    run_skopeo inspect --raw dir:$workdir
+    inspect_local_raw=$output
+    config_digest=$(jq -r '.config.digest' <<<"$inspect_local_raw")
 
     # Each SHA-named file must be listed in the output of 'inspect'
     for sha in $(find $workdir -type f | xargs -l1 basename | egrep '^[0-9a-f]{64}$'); do
-        expect_output --from="$inspect_local" --substring "sha256:$sha" \
-                      "Locally-extracted SHA file is present in 'inspect'"
+        if [ "sha256:$sha" != "$config_digest" ]; then
```
edsantiago, line 36: As a favor for ignorami like me, could you pretty-please add a comment explaining the circumstances in which this will trigger?

systemtest/010-inspect.bats
```diff
@@ -27,11 +27,16 @@ load helpers
...
+            expect_output --from="$inspect_local" --substring "sha256:$sha" \
```
edsantiago, line 37: As long as you're touching this code, would it make sense to tighten this up a little, something like
```bash
    layers=$(jq -r '.Layers' <<<"$inspect_local")
    expect_output --from="$layers" ....
```
It might make it clearer to a future maintainer. (Or maybe not; maybe I've misunderstood).

edsantiago (commented):
Tentative LGTM, two suggestions as favors for future maintainers

systemtest/010-inspect.bats
mtrmac: Added

edsantiago:
/lgtm

Thank you!

PR #1578, opened 2022-02-25
do not recommend upgrading all packages

slhck:
The command to install skopeo for Ubuntu 20.04 includes a forced upgrade step for all packages.

Installing skopeo does not require the upgrade step, and it could lead to possible issues completely unrelated to the project.


mtrmac (commented):
Thanks!

This looks reasonable to me. @lsm5 , PTAL, there might be something I’m missing.

rhatdan:
LGTM

Issue #1029, opened 2020-08-27
Add support for docker-archive: to skopeo list-tags


guanxingfeng:
command: skopeo inspect docker-archive:/root/busybox.tar
{
    "Digest": "sha256:d8db71d650ed60da14a6c6d46ea1103d253a1117586e846020b82726e71d3b19",
    "RepoTags": [],
    "Created": "2020-06-29T20:21:41.42102751Z",
    "DockerVersion": "18.09.7",
    "Labels": null,
    "Architecture": "amd64",
    "Os": "linux",
    "Layers": [
        "sha256:50761fe126b6e4d90fa0b7a6e195f6030fe250c016c2fc860ac40f2e8d2f2615"
    ],
    "Env": [
        "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
    ]
}

expect display right ,eg "RepoTags":["busybox:latest"]!!

mtrmac:
Thanks for your report.

Conceptually, `skopeo inspect` exists to provide information about an _image_; the way it also provides repository tags, _only_ for docker/distribution registries, does not really generalize (and is often _actively unwanted_ in `inspect` due to the extra overhead). It’s more or less a historical accident preserved for compatibility, and not something that we want to further enhance.

This functionality would be a good fit for `skopeo list-tags`, though.

rhatdan:
@guanxingfeng Does list-tags satify your needs?

mtrmac:
@rhatdan `skopeo list-tags` doesn’t currently implement `docker-archive:`.

guanxingfeng:
@rhatdan I want to get tags from a image tar，skopeo list-tags doesn’t currently implement docker-archive:

rhatdan:
Would that change make more sense? Using skopeo list-tags on a docker-archive?

guanxingfeng:
I wang to implement upload multiple docker tars to image hub using UI，like this，

![image](https://user-images.githubusercontent.com/25924966/97379789-8d861080-1900-11eb-8423-f0769cac5fc0.png)

It will be very useful and convenient to get the image tags automatically instead of manual input.

mtrmac:
FWIW https://github.com/containers/image/pull/991 has built the infrastructure necessary to implement this.

zhangguanzhang:
> FWIW [containers/image#991](https://github.com/containers/image/pull/991) has built the infrastructure necessary to implement this.

[...]

PR #1581, opened 2022-03-02
Add support for docker-archive: to skopeo list-tags

zhangguanzhang:
Fixes: https://github.com/containers/skopeo/issues/1029
Signed-off-by: zhangguanzhang <zhangguanzhang@qq.com>


cmd/skopeo/list_tags.go
```diff
@@ -39,12 +40,12 @@ func tagsCmd(global *globalOptions) *cobra.Command {
 		retryOpts: retryOpts,
 	}
 	cmd := &cobra.Command{
-		Use:   "list-tags [command options] REPOSITORY-NAME",
-		Short: "List tags in the transport/repository specified by the REPOSITORY-NAME",
-		Long: `Return the list of tags from the transport/repository "REPOSITORY-NAME"
+		Use:   "list-tags [command options] transport",
```
vrothberg, line 43: How about `SOURCE-IMAGE`?  `transport` is strictly speaking not accurate as we need `transport+image`.

cmd/skopeo/list_tags.go
```diff
@@ -113,23 +114,48 @@ func (opts *tagsOptions) run(args []string, stdout io.Writer) (retErr error) {
...
-
 	var repositoryName string
 	var tagListing []string
-	if err = retry.RetryIfNecessary(ctx, func() error {
-		repositoryName, tagListing, err = listDockerTags(ctx, sys, imgRef)
-		return err
-	}, opts.retryOpts); err != nil {
-		return err
+
+	if transport.Name() == docker.Transport.Name() {
```
vrothberg, line 120: Could you turn this into a `switch` statement?

cmd/skopeo/list_tags.go
```diff
@@ -113,23 +114,48 @@ func (opts *tagsOptions) run(args []string, stdout io.Writer) (retErr error) {
...
+				repoTags, err := tarReader.ManifestTagsForReference(ref)
+				if err != nil {
+					return err
+				}
+				tagListing = append(tagListing, repoTags...)
+			}
+		}
+
+	} else {
+		return fmt.Errorf("Unsupported transport '%v' for tag listing. Only supported: '%s','%s' ",
```
vrothberg, line 157:
```suggestion
		return fmt.Errorf("Unsupported transport '%v' for tag listing. Only supported: %q, %q ",
```

cmd/skopeo/list_tags.go
```diff
@@ -113,23 +114,48 @@ func (opts *tagsOptions) run(args []string, stdout io.Writer) (retErr error) {
...
+			return err
+		}
+		if err = retry.RetryIfNecessary(ctx, func() error {
+			repositoryName, tagListing, err = listDockerTags(ctx, sys, imgRef)
+			return err
+		}, opts.retryOpts); err != nil {
+			return err
+		}
+	} else if transport.Name() == archive.Transport.Name() {
+		parts := strings.SplitN(args[0], ":", 2)
```
vrothberg, line 133: While parsing works like that just fine, I'd prefer using the transport's API directly.

```go
ref, err := dockerArchiveTransport.ParseReference(args[0])         
if err != nil {                                                 
        return err
}                            
archivePath := ref.StringWithinTransport()
```

vrothberg (commented):
Thanks!  Besides the minor nits, would you add a test?  We have a number of .tar's for testing in containers/common.  We could add https://github.com/containers/common/blob/main/libimage/testdata/docker-two-images.tar.xz here and add an integration test for it.

zhangguanzhang:
all had done

docs/skopeo-list-tags.1.md
```diff
@@ -72,6 +72,9 @@ This commands refers to repositories using a _transport_`:`_details_ format. The
         "docker.io/myuser/myimage:v1.0"
         "docker.io/myuser/myimage@sha256:f48c4cc192f4c3c6a069cb5cca6d0a9e34d6076ba7c214fd0cc3ca60e0af76bb"
 
+## Docker archive
+
+An image is stored in the docker save formatted file.
```
vrothberg, line 77:
```suggestion
One or more images are stored in an `docker-archive` (see containers-transports(1)).
```
Plus add containers-transports(1) to the `# See also` section.

vrothberg (commented):
To fix the failing CI, have a look at `hack/man-page-checker`

The man page seems to refer to it as `TRANSPORT` while the help message states `SOURCE-IMAGE`.  Both should be consistent.

> skopeo list-tags:         man='TRANSPORT' help='SOURCE-IMAGE'

systemtest/070-list-tags.bats
```diff
@@ -0,0 +1,48 @@
...
+    output=$(run_skopeo list-tags --tls-verify=false docker://localhost:5000/busybox)
+    echo $output | grep "${img_tag}"
+}
+
+# list from a local docker-archive file
+@test "list-tags: from a docker-archive file" {
+    local file_name=docker-two-images.tar.xz
+    local output=
+
+    curl -o $file_name https://raw.githubusercontent.com/containers/common/main/libimage/testdata/docker-two-images.tar.xz
```
vrothberg, line 34: Could you move the file here?  This way, the Skopeo tests are independent.

vrothberg (commented):
One final nit from my side. Thank you, great work!

@edsantiago @mtrmac PTAL

systemtest/070-list-tags.bats
zhangguanzhang: > Could you move the file here? This way, the Skopeo tests are independent.

I will do this

docs/skopeo-list-tags.1.md
```diff
@@ -121,8 +124,21 @@ $ skopeo list-tags docker://localhost:5000/fedora
 
 ```
 
+### Docker-archive Transport
+
+```sh
+skopeo list-tags docker-archive:/root/busybox.tar.gz
```
edsantiago, line 130: Suggestion: include prompt on all sample commands. Since there's no reason to make this root-only, I would encourage you to use `$` and to change the path as well (to `/tmp/` instead of `/root`).

systemtest/070-list-tags.bats
```diff
@@ -0,0 +1,47 @@
...
+
+# list from registry
+@test "list-tags: remove image from registry" {
+    local remote_image=docker://quay.io/libpod/busybox:latest
+    local img_tag=unsigned
+    local localimg=docker://localhost:5000/busybox:${img_tag}
+    local output=
+
+    run_skopeo copy --dest-tls-verify=false $remote_image $localimg
+    output=$(run_skopeo inspect --tls-verify=false --raw $localimg)
```
edsantiago, line 22: Please don't do this. `run_skopeo` already sets `$output`, you do not need to do the `output=` thing.

systemtest/070-list-tags.bats
```diff
@@ -0,0 +1,47 @@
...
+    echo $output | grep "vnd.docker.distribution.manifest.v2+json"
```
edsantiago, line 23: Please don't do this. If it fails, the error message will be completely worthless:
```
    `echo $output | grep "vnd.docker.distribution-manifest.v2+json"' failed         
```
(it gives no indication of what the problem was). Instead, use the existing `expect_output` mechanism:
```
    run_skopeo inspect --tls-verify=false --raw $localimg
    mediatype=$(jq -r '.mediaType' <<<"$output")
    expect_output --from="$mediatype" \
                  "application/vnd.docker.distribution.manifest.v2+json"
```
(the `jq` tightens up the test, making it less likely to false-positive on something else in the `inspect` output)

systemtest/070-list-tags.bats
```diff
@@ -0,0 +1,47 @@
...
+
+    output=$(run_skopeo list-tags --tls-verify=false docker://localhost:5000/busybox)
+    echo $output | grep "${img_tag}"
```
edsantiago, line 26: See above

systemtest/070-list-tags.bats
```diff
@@ -0,0 +1,47 @@
...
+}
+
+# list from a local docker-archive file
+@test "list-tags: from a docker-archive file" {
+    local file_name=./systemtest/testdata/docker-two-images.tar.xz
+    local output=
+    pwd
+    echo $TEST_DIR
+    ls -l $TEST_DIR
```
edsantiago, line 35: These are debug statements, no longer needed

systemtest/070-list-tags.bats
```diff
@@ -0,0 +1,47 @@
...
+    output=$(run_skopeo list-tags docker-archive:$file_name)
+    echo $output | grep "example.com/empty"
```
edsantiago, line 37: See above notes

edsantiago (changes requested):

docs/skopeo-list-tags.1.md
```diff
@@ -121,8 +124,21 @@ $ skopeo list-tags docker://localhost:5000/fedora
 
 ```
 
+### Docker-archive Transport
+
+```sh
+skopeo list-tags docker-archive:/tmp/busybox.tar.gz
```
edsantiago, line 130: Don't forget the prompt :)

systemtest/070-list-tags.bats
```diff
@@ -0,0 +1,47 @@
...
+    expect_output --substring "$img_tag"
+}
+
+# list from a local docker-archive file
+@test "list-tags: from a docker-archive file" {
+    local file_name=./systemtest/testdata/docker-two-images.tar.xz
+    local output=
+
+    run_skopeo list-tags docker-archive:$file_name
+    expect_output --substring "example.com/empty.+"
```
edsantiago, line 38: You don't need (or want) the `.+`. `--substring` takes care of that.

edsantiago (commented):
Minor nits still

cmd/skopeo/list_tags.go
mtrmac: Well, “image” also seems quite inappropriate, when the input is some kind of aggregate (repo/archive) and the output is a _list_ of images.

cmd/skopeo/list_tags.go
```diff
@@ -113,23 +114,51 @@ func (opts *tagsOptions) run(args []string, stdout io.Writer) (retErr error) {
...
+			return err
+		}
+		if err = retry.RetryIfNecessary(ctx, func() error {
+			repositoryName, tagListing, err = listDockerTags(ctx, sys, imgRef)
+			return err
+		}, opts.retryOpts); err != nil {
+			return err
+		}
+	case archive.Transport.Name():
+		ref, err := archive.ParseReference(strings.TrimPrefix(args[0], archive.Transport.Name()+":"))
```
mtrmac, line 134: Maybe do `alltransports.ParseImageName`, (along with an assertion that the transport didn’t change??) to minimize the amount of custom parsing code.

See also below about refusing references that contain more than a path.

cmd/skopeo/list_tags.go
```diff
@@ -113,23 +114,51 @@ func (opts *tagsOptions) run(args []string, stdout io.Writer) (retErr error) {
...
+				repoTags, err := tarReader.ManifestTagsForReference(ref)
+				if err != nil {
+					return err
+				}
+				tagListing = append(tagListing, repoTags...)
+			}
+		}
+	default:
+		return fmt.Errorf("Unsupported transport '%v' for tag listing. Only supported: %q,%q ",
+			transport.Name(), docker.Transport.Name(), archive.Transport.Name())
```
mtrmac, line 161: Non-blocking?

Define a
```go
transportHandlers map[transportName string]
    func(userInput string) (repositoryName string, tagListing []string, error)
```
, and split handling of the different transports into separate functions. That way it scales better than one large function.

Then, in this (“no map entry found”) case, we can programmatically iterate the map keys, and construct a sorted list of the supported transports in a way that never gets out of sync.

OTOH at 2 cases this is not urgent, I’d insist on this at about 3 cases.

cmd/skopeo/list_tags.go
```diff
@@ -113,23 +114,51 @@ func (opts *tagsOptions) run(args []string, stdout io.Writer) (retErr error) {
...
+			return err
+		}
+
+		for _, items := range imageRefs {
+			for _, ref := range items {
+				repoTags, err := tarReader.ManifestTagsForReference(ref)
+				if err != nil {
+					return err
+				}
+				tagListing = append(tagListing, repoTags...)
```
mtrmac, line 156: @vrothberg , others RFC: The above, AFAICS, returns _nothing_ for untagged images. Should we include something (like the `@source-index` syntax)?

docs/skopeo-list-tags.1.md
```diff
@@ -1,14 +1,14 @@
 % skopeo-list-tags(1)
 
 ## NAME
-skopeo\-list\-tags - List tags in the transport-specific image repository.
+skopeo\-list\-tags - List tags in the transport-specific image repository or a local docker-archive file.
```
mtrmac, line 4: Non-blocking: How about something generic like “List image names in a transport-specific collection of images” ?

docs/skopeo-list-tags.1.md
```diff
@@ -72,6 +72,9 @@ This commands refers to repositories using a _transport_`:`_details_ format. The
         "docker.io/myuser/myimage:v1.0"
```
mtrmac, line 72: Above, the text “Currently, only the Docker transport is supported.” needs to be updated (probably just deleted).

docs/skopeo-list-tags.1.md
```diff
@@ -72,6 +72,9 @@ This commands refers to repositories using a _transport_`:`_details_ format. The
...
         "docker.io/myuser/myimage@sha256:f48c4cc192f4c3c6a069cb5cca6d0a9e34d6076ba7c214fd0cc3ca60e0af76bb"
 
+## Docker archive
```
mtrmac, line 75: This should follow format of the `docker://` paragraph above, an in particular precisely document the accepted syntax.

docs/skopeo-list-tags.1.md
```diff
@@ -121,8 +124,23 @@ $ skopeo list-tags docker://localhost:5000/fedora
...
 ```
 
+### Docker-archive Transport
+
+To list the tags in a local docker-archive file:
+
+```sh
+skopeo list-tags docker-archive:/tmp/busybox.tar.gz
+{
+    "Repository": "",
```
mtrmac, line 134: Hum, should this field be filled in? Or hidden if empty?

I guess we can start with the latter, and fill the field in later if we find a need to do so.

systemtest/070-list-tags.bats
```diff
@@ -0,0 +1,47 @@
...
+    local localimg=docker://localhost:5000/busybox:${img_tag}
+    local output=
+
+    run_skopeo copy --dest-tls-verify=false $remote_image $localimg
+
+    run_skopeo inspect --tls-verify=false --raw $localimg
+    mediatype=$(jq -r '.mediaType' <<<"$output")
+    expect_output --from="$mediatype" \
+        "application/vnd.docker.distribution.manifest.v2+json"
+
```
mtrmac, line 27: Why is all of this a part of the test? AFAICS this is not necessary to test `list-tags`.

systemtest/070-list-tags.bats
```diff
@@ -0,0 +1,47 @@
...
+load helpers
+
+function setup() {
+    standard_setup
+
+    start_registry --enable-delete=true reg
+}
+
+# list from registry
+@test "list-tags: remove image from registry" {
```
mtrmac, line 15:
```suggestion
@test "list-tags: remote repository on a registry" {
```
???

systemtest/070-list-tags.bats
```diff
@@ -0,0 +1,47 @@
...
+#
+# list-tags tests
+#
+
+load helpers
+
+function setup() {
+    standard_setup
+
+    start_registry --enable-delete=true reg
```
mtrmac, line 11: After the registry test is simplified, this shouldn’t be necessary.

docs/skopeo-list-tags.1.md
zhangguanzhang: > Hum, should this field be filled in? Or hidden if empty?
> 
> I guess we can start with the latter, and fill the field in later if we find a need to do so.
may be there are more images in the file
```
$ ./bin/skopeo list-tags  docker-archive:docker-two-images.tar.xz
{
    "Repository": "",
    "Tags": [
        "example.com/empty:latest",
        "example.com/empty/but:different"
    ]
}
```

systemtest/070-list-tags.bats
zhangguanzhang: this had done

systemtest/070-list-tags.bats
zhangguanzhang: this had done

systemtest/070-list-tags.bats
zhangguanzhang: this had done

docs/skopeo-list-tags.1.md
zhangguanzhang: this had done

docs/skopeo-list-tags.1.md
zhangguanzhang: this had done

docs/skopeo-list-tags.1.md
zhangguanzhang: this had done

cmd/skopeo/list_tags.go
```diff
@@ -39,14 +40,14 @@ func tagsCmd(global *globalOptions) *cobra.Command {
...
+		Use:   "list-tags [command options] SOURCE-IMAGE",
+		Short: "List tags in the transport/repository specified by the SOURCE-IMAGE",
+		Long: `Return the list of tags from the transport/repository "SOURCE-IMAGE"
 
 Supported transports:
-docker
+docker docker-archive
 
-See skopeo-list-tags(1) section "REPOSITORY NAMES" for the expected format
+See skopeo-list-tags(1) section "EXAMPLES" for the expected format
```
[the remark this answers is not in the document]
zhangguanzhang: this had done

cmd/skopeo/list_tags.go
zhangguanzhang: I'll update this part later when I have time

zhangguanzhang: > Maybe do `alltransports.ParseImageName`, (along with an assertion that the transport didn’t change??) to minimize the amount of custom parsing code.
> 
> See also below about refusing references that contain more than a path.

I'll update this part later when I have time

systemtest/070-list-tags.bats
```diff
@@ -0,0 +1,32 @@
...
+    local output=
+
+    run_skopeo list-tags --tls-verify=false $remote_image
+    expect_output --substring "quay.io/libpod/alpine_labels"
+    expect_output --substring "latest"
+}
+
+# list from a local docker-archive file
+@test "list-tags: from a docker-archive file" {
+    local file_name=./systemtest/testdata/docker-two-images.tar.xz
```
edsantiago, line 24: Tests are not guaranteed to run while cd'ed to this directory. You must use an absolute path for this:
```bash
    local file_name=${TEST_SOURCE_DIR}/testdata/docker-two-images.tar.xz
```
And actually, since this string is only used in one place, there isn't really a need for it to be a variable.

systemtest/070-list-tags.bats
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bats
+#
+# list-tags tests
+#
+
+load helpers
+
+function setup() {
+    standard_setup
+}
```
edsantiago, line 10: This is now a NOP, please remove

systemtest/070-list-tags.bats
```diff
@@ -0,0 +1,32 @@
...
+
+# list from registry
+@test "list-tags: remote repository on a registry" {
+    local remote_image=docker://quay.io/libpod/alpine_labels
+    local output=
+
+    run_skopeo list-tags --tls-verify=false $remote_image
```
edsantiago, line 17: There's no longer a need for `--tls-verify=false`

systemtest/070-list-tags.bats
```diff
@@ -0,0 +1,32 @@
...
+    expect_output --substring "quay.io/libpod/alpine_labels"
+    expect_output --substring "latest"
+}
+
+# list from a local docker-archive file
+@test "list-tags: from a docker-archive file" {
+    local file_name=./systemtest/testdata/docker-two-images.tar.xz
+    local output=
```
edsantiago, line 25: Not needed, please remove

systemtest/070-list-tags.bats
```diff
@@ -0,0 +1,32 @@
...
+
+    run_skopeo list-tags docker-archive:$file_name
+    expect_output --substring "example.com/empty"
```
edsantiago, line 28: Is there any reason you're not checking for `:latest` and `but:different`? I don't know enough to make this a requirement, but it seems like a good idea at least superficially

edsantiago (changes requested):
Tests will not work in gating-test environment; that needs to be fixed. Plus a few less-critical comments.

Thank you for persevering on this; I realize you've spent a lot of time.

systemtest/070-list-tags.bats
zhangguanzhang: > Tests are not guaranteed to run while cd'ed to this directory. You must use an absolute path for this:
> 
> ```shell
>     local file_name=${TEST_SOURCE_DIR}/testdata/docker-two-images.tar.xz
> ```
> 
> And actually, since this string is only used in one place, there isn't really a need for it to be a variable.

done

systemtest/070-list-tags.bats
zhangguanzhang: > This is now a NOP, please remove

done

systemtest/070-list-tags.bats
zhangguanzhang: > Is there any reason you're not checking for `:latest` and `but:different`? I don't know enough to make this a requirement, but it seems like a good idea at least superficially

done

docs/skopeo-list-tags.1.md
```diff
@@ -72,6 +72,9 @@ This commands refers to repositories using a _transport_`:`_details_ format. The
         "docker.io/myuser/myimage:v1.0"
         "docker.io/myuser/myimage@sha256:f48c4cc192f4c3c6a069cb5cca6d0a9e34d6076ba7c214fd0cc3ca60e0af76bb"
 
+  **docker-archive:**_path_[**:**_docker-reference_]
```
mtrmac, line 75: Using the `[…docker-reference]` would AFAICS actually not work (it would not be a valid path the `Reader` could consume); so remove that part.

docs/skopeo-list-tags.1.md
mtrmac: What I meant is that I think the output should be
```
$ ./bin/skopeo list-tags  docker-archive:docker-two-images.tar.xz
{
    "Tags": [
        "example.com/empty:latest",
        "example.com/empty/but:different"
    ]
}
```
without the pointless `Repository` line. That can be achieved using a `json:",omitempty"` annotation.

docs/skopeo-list-tags.1.md
```diff
@@ -1,14 +1,14 @@
...
 
 ## SYNOPSIS
-**skopeo list-tags** [*options*] _repository-name_
+**skopeo list-tags** [*options*] _source-image_
 
-Return a list of tags from _repository-name_ in a registry.
+Return a list of tags from _source-image_ in a registry or a local docker-archive file.
 
-  _repository-name_ name of repository to retrieve tag listing from
+  _source-image_ name of repository to retrieve tag listing from or a local docker-archive file.
```
TomSweeneyRedHat, line 11:
```suggestion
  _source-image_ name of the repository to retrieve a tag listing from or a local docker-archive file.
```

TomSweeneyRedHat: I know it's old but would appreciate the touch-up if you make other changes.

docs/skopeo-list-tags.1.md
```diff
@@ -121,8 +124,23 @@ $ skopeo list-tags docker://localhost:5000/fedora
 
 ```
 
+### Docker-archive Transport
+
+To list the tags in a local docker-archive file:
+
+```sh
+skopeo list-tags docker-archive:/tmp/busybox.tar.gz
```
TomSweeneyRedHat, line 132: Ed's suggestion I think...
```suggestion
$ skopeo list-tags docker-archive:/tmp/busybox.tar.gz
```

zhangguanzhang: done

cmd/skopeo/list_tags.go
zhangguanzhang: I tried to change to this https://paste.debian.net/1232877/ ,  But it doesn't look good

mtrmac: Oh, the functions in `transportHandlers` would of course be separate named functions, not in-line lambdas.

```go
var transportHandlers = map[string]func(…) (…) {
    docker.Transport.Name(): listDockerTags,
    archive.Transport.Name(): listDockerArchiveTags,
}
```

cmd/skopeo/list_tags.go
```diff
@@ -39,12 +46,12 @@ func tagsCmd(global *globalOptions) *cobra.Command {
...
-		Use:   "list-tags [command options] REPOSITORY-NAME",
-		Short: "List tags in the transport/repository specified by the REPOSITORY-NAME",
-		Long: `Return the list of tags from the transport/repository "REPOSITORY-NAME"
+		Use:   "list-tags [command options] SOURCE-IMAGE",
+		Short: "List tags in the transport/repository specified by the SOURCE-IMAGE",
+		Long: `Return the list of tags from the transport/repository "SOURCE-IMAGE"
 
 Supported transports:
-docker
+docker docker-archive
```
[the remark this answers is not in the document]
zhangguanzhang: > Compute this from `transportHandlers`?

all had done

cmd/skopeo/list_tags.go
```diff
@@ -95,40 +102,97 @@ func listDockerTags(ctx context.Context, sys *types.SystemContext, imgRef types.
...
+		i, err = strconv.Atoi(parts[2][1:])
+		if err != nil {
+			err = errors.Wrapf(err, "Invalid source index %s", parts[2])
+			return
+		}
+		if i+1 > len(tagListing) {
+			err = fmt.Errorf("Invalid source index @%d, only %d tags items available", i, len(tagListing))
+			return
+		}
+		tagListing = []string{tagListing[i]}
```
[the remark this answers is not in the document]
zhangguanzhang: done

zhangguanzhang:
@mtrmac PTAL

cmd/skopeo/list_tags.go
```diff
@@ -95,40 +102,97 @@ func listDockerTags(ctx context.Context, sys *types.SystemContext, imgRef types.
...
-		return fmt.Errorf("Invalid %q: does not specify a transport", args[0])
+		return "", nil, fmt.Errorf("Invalid %q: does not specify a transport", userInput)
 	}
 
-	if transport.Name() != docker.Transport.Name() {
-		return fmt.Errorf("Unsupported transport '%v' for tag listing. Only '%v' currently supported", transport.Name(), docker.Transport.Name())
+	if val, ok := transportHandlers[transport.Name()]; ok {
+		return val(ctx, sys, opts, userInput)
 	}
+	err = fmt.Errorf("Unsupported transport '%q' for tag listing", transport.Name())
```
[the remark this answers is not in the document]
zhangguanzhang: > Compute the supported transports from `transportHandlers`, and list them here, like the previous version did?

already done it

cmd/skopeo/list_tags.go
```diff
@@ -28,6 +29,20 @@ type tagsOptions struct {
...
+	archive.Transport.Name(): listDockerArchiveTags,
+}
+
+// getSupportTransport return all the supported transports
+func getSupportTransport(joinStr string) string {
+	supportTranspors := make([]string, 0, len(transportHandlers))
+	for handlerName := range transportHandlers {
+		supportTranspors = append(supportTranspors, handlerName)
+	}
+	return strings.Join(supportTranspors, joinStr)
```
mtrmac, line 43: `sort.Strings(supportedTransports)`, otherwise the iteration order of the hash is not guaranteed to be consistent over time.

cmd/skopeo/list_tags.go
```diff
@@ -28,6 +29,20 @@ type tagsOptions struct {
...
 }
 
+var transportHandlers = map[string]func(ctx context.Context, sys *types.SystemContext, opts *tagsOptions, userInput string) (repositoryName string, tagListing []string, err error){
+	docker.Transport.Name():  listDockerRepoTags,
+	archive.Transport.Name(): listDockerArchiveTags,
+}
+
+// getSupportTransport return all the supported transports
+func getSupportTransport(joinStr string) string {
+	supportTranspors := make([]string, 0, len(transportHandlers))
```
mtrmac, line 39:
```suggestion
// getSupportedTransports returns all the supported transports
func getSupportedTransports(joinStr string) string {
	supportedTransports := make([]string, 0, len(transportHandlers))
```

---

Absolutely non-blocking: Alternatively, the function could just be called `supportedTransports`, and the local variable maybe just `res`.

cmd/skopeo/list_tags.go
```diff
@@ -38,13 +53,17 @@ func tagsCmd(global *globalOptions) *cobra.Command {
 		image:     imageOpts,
 		retryOpts: retryOpts,
 	}
+	supportTranspors := make([]string, 0, len(transportHandlers))
+	for handlerName := range transportHandlers {
+		supportTranspors = append(supportTranspors, handlerName)
+	}
```
mtrmac, line 59: This now seems unused.

cmd/skopeo/list_tags.go
```diff
@@ -95,6 +114,61 @@ func listDockerTags(ctx context.Context, sys *types.SystemContext, imgRef types.
...
+	)
+	for _, items := range imageRefs {
+		for _, ref := range items {
+			repoTags, err = tarReader.ManifestTagsForReference(ref)
+			if err != nil {
+				return
+			}
+			// handle for each untagged image
+			if len(repoTags) == 0 {
+				repoTags = []string{fmt.Sprintf("@%d", untaggedIndex)}
```
mtrmac, line 162: The `@` indices include tagged values, i.e. they are just indices into `imageRefs`. So counting the `untaggedIndex` is unnecessary, and ultimately incorrect.

Above, do `for imageIndex, items := range imageRefs`.

mtrmac (commented):
Thanks! This looks very good.

One thing I’m somewhat unhappy is naming the input “source-image”, per https://github.com/containers/skopeo/pull/1581#discussion_r817448150 . @vrothberg , would something vague like “collection” work?

---

@edsantiago Could you review the system tests parts once more, please?

cmd/skopeo/list_tags.go
zhangguanzhang: > How about `SOURCE-IMAGE`? `transport` is strictly speaking not accurate as we need `transport+image`.

There doesn't seem to be a suitable word to describe it, what do you think about it @vrothberg

zhangguanzhang:
@vrothberg PTAL

systemtest/070-list-tags.bats
```diff
@@ -0,0 +1,28 @@
...
+}
+
+# list from a local docker-archive file
+@test "list-tags: from a docker-archive file" {
+    local file_name=${TEST_SOURCE_DIR}/testdata/docker-two-images.tar.xz
+
+    run_skopeo list-tags docker-archive:$file_name
+    expect_output --substring "example.com/empty:latest"
+    expect_output --substring "example.com/empty/but:different"
+
```
vrothberg, line 24: Non-blocking nit: remove empty trailing line

vrothberg (commented):
LGTM

@mtrmac PTanotherL

vrothberg:
Nice work all!

mtrmac (commented):
One more tiny nit, code LGTM otherwise.

---

@vrothberg 
> One thing I’m somewhat unhappy is naming the input “source-image”, per https://github.com/containers/skopeo/pull/1581#discussion_r817448150 . @vrothberg , would something vague like “collection” work?

zhangguanzhang:
@vrothberg  PTAL on [#1581](https://github.com/containers/skopeo/pull/1581#discussion_r817448150)

cmd/skopeo/list_tags.go
vrothberg: LGTM

vrothberg (approved):
LGTM, thanks a lot!

Issue #1432, opened 2021-08-26
skopeo delete command deletes by reference even when we provide tag 


shimish2:
when we issue skopeo delete command to delete a tag instead of sending delete request by tag to image repository server skopeo sending delete by reference request. 

``` json 
skopeo --debug delete --tls-verify=false docker://localhost:8080/skopeo-delete-test:1.0.0
DEBU[0000] Credentials not found                        
DEBU[0000] Using registries.d directory /etc/containers/registries.d for sigstore configuration 
DEBU[0000]  Using "default-docker" configuration        
DEBU[0000]   Using file:///var/lib/containers/sigstore  
DEBU[0000] Looking for TLS certificates and private keys in /etc/docker/certs.d/localhost:8080 
DEBU[0000] Loading registries configuration "/etc/containers/registries.conf" 
DEBU[0000] Loading registries configuration "/etc/containers/registries.conf.d/shortnames.conf" 
DEBU[0000] GET https://localhost:8080/v2/               
DEBU[0000] Ping https://localhost:8080/v2/ err Get "https://localhost:8080/v2/": http: server gave HTTP response to HTTPS client (&url.Error{Op:"Get", URL:"https://localhost:8080/v2/", Err:(*errors.errorString)(0xc000379d50)}) 
DEBU[0000] GET http://localhost:8080/v2/                
DEBU[0000] Ping http://localhost:8080/v2/ status 200    
DEBU[0000] GET http://localhost:8080/v2/skopeo-delete-test/manifests/1.0.0 
DEBU[0000] DELETE http://localhost:8080/v2/skopeo-delete-test/manifests/sha256:6ffade1dd9371ad14429e1564cdc189ef1fb92a44a6ddc84d5f9ef2194ce8943 
DEBU[0000] Deleting /var/lib/containers/sigstore/skopeo-delete-test@sha256=6ffade1dd9371ad14429e1564cdc189ef1fb92a44a6ddc84d5f9ef2194ce8943/signature-1
```

```skopeo --version 
skopeo version 1.4.1 commit: 130f32f047b7bf9b1fb4f95c4c5b1da6255e4829
```

rgl:
FYI, `regctl tag rm` (see https://github.com/regclient/regclient) can remove a docker hub image tag, e.g.:

```bash
function regctl {
  docker container run --rm --net host \
    -u "$(id -u):$(id -g)" -e HOME -v $HOME:$HOME \
[...]

PR #1597, opened 2022-03-18
Improve the (skopeo delete) man page

mtrmac:
… per https://github.com/containers/image/issues/1010 , at least as a short-term stopgap. Compare that issue, and #1432, for much more about possible code improvements.

- Actually add a DESCRIPTION heading.
- Warn about deleting by digest —  it affects an unknown set of tags.
- Warn about deleting by tag — it currently works by the resolved digest.

### DIFF
--- a/docs/skopeo-delete.1.md
+++ b/docs/skopeo-delete.1.md
@@ -6,17 +6,27 @@ skopeo\-delete - Mark the _image-name_ for later deletion by the registry's garb
 ## SYNOPSIS
 **skopeo delete** [*options*] _image-name_
 
-Mark _image-name_ for deletion.  To release the allocated disk space, you must login to the container registry server and execute the container registry garbage collector. E.g.,
+## DESCRIPTION
+
+Mark _image-name_ for deletion.
+The effect of this is registry-specific; many registries don’t support this operation, or don’t allow it in some circumstances / configurations.
+
+**WARNING**: If _image-name_ contains a digest, this affects the referenced manifest, and may delete all tags (within the current repository?) pointing to that manifest.
+
+**WARNING**: If _image-name_ contains a tag (but not a digest), in the current version of Skopeo this resolves the tag into a digest, and then deletes the manifest by digest, as described above (possibly deleting all tags pointing to that manifest, not just the provided tag). This behavior may change in the future.
+
+
+When using the github.com/distribution/distribution registry server:
+To release the allocated disk space, you must login to the container registry server and execute the container registry garbage collector. E.g.,
 
 ```
 /usr/bin/registry garbage-collect /etc/docker-distribution/registry/config.yml
-
+```
 Note: sometimes the config.yml is stored in /etc/docker/registry/config.yml
 
 If you are running the container registry inside of a container you would execute something like:
-
+```
 $ docker exec -it registry /usr/bin/registry garbage-collect /etc/docker-distribution/registry/config.yml
-
 ```
 
 ## OPTIONS


docs/skopeo-delete.1.md
```diff
@@ -6,17 +6,27 @@ skopeo\-delete - Mark the _image-name_ for later deletion by the registry's garb
 ## SYNOPSIS
 **skopeo delete** [*options*] _image-name_
 
-Mark _image-name_ for deletion.  To release the allocated disk space, you must login to the container registry server and execute the container registry garbage collector. E.g.,
+## DESCRIPTION
+
+Mark _image-name_ for deletion.
+The effect of this is registry-specific; many registries don’t this operation, or don’t allow it in some circumstances / configurations.
```
vrothberg, line 12:
```suggestion
The effect of this is registry-specific; many registries don’t support this operation, or don’t allow it in some circumstances / configurations.
```

docs/skopeo-delete.1.md
```diff
@@ -6,17 +6,27 @@ skopeo\-delete - Mark the _image-name_ for later deletion by the registry's garb
...
+
+**WARNING**: If _image-name_ contains a digest, this affects the referenced manifest, and may delete all tags (within the current repository?) pointing to that manifest.
```
vrothberg, line 14:
```suggestion
**WARNING**: If _image-name_ contains a digest, this affects the referenced manifest, and may delete all tags pointing to that manifest.
```
I am quite confused by that.  In case it's important, could you rephrase?

docs/skopeo-delete.1.md
```diff
@@ -6,17 +6,27 @@ skopeo\-delete - Mark the _image-name_ for later deletion by the registry's garb
...
+
+**WARNING**: If _image-name_ contains a tag (but not a digest), currently this resolves the tag into a digest, and then deletes the manifest by digest, as described above (possibly deleting all tags pointing to that manifest, not just the provided tag.) This behavior may change in the future.
```
graywolf-at-work, line 16:
```suggestion
**WARNING**: If _image-name_ contains a tag (but not a digest), currently this resolves the tag into a digest, and then deletes the manifest by digest, as described above (possibly deleting all tags pointing to that manifest, not just the provided tag). This behavior may change in the future.
```

TomSweeneyRedHat, line 16: Minor tweak for consideration
```suggestion
**WARNING**: If _image-name_ contains a tag (but not a digest), this resolves the tag into a digest and then deletes the manifest by digest, as described above (possibly deleting all tags pointing to that manifest, not just the provided tag). This behavior may change in the future.
```

docs/skopeo-delete.1.md
mtrmac: After pushing the same image (same digest) to
- repo1:tag1A
- repo1:tag1B
- repo2:tag2A
- repo2:tag2B

the distribution/distribution implementation of deleting repo1@digest is to untag repo1:tag1A and repo1:tag1B, but not repo2:tag2A and repo2:tag2B.

The specification does not describe any of this in detail.

---

Should the text say something like “all tags within the relevant repository that point to the manifest in question” without the parentheses and the question mark? That’s _probably_ the most reasonable behavior, I just don’t have any data about what various registries do.

docs/skopeo-delete.1.md
mtrmac: The intent of “currently” was to give us an opportunity to change this problematic design. I have replaced that with “in the current version of Skopeo”, is that clearer?

rhatdan:
LGTM

PR #1610, opened 2022-03-30
Add option to specify the identity for signing

### LINKED ISSUE #1588: Support copy --sign-by where the needed docker reference and push domain are different.
A feature proposal. 

The scenario is that the image needs to be pushed to one domain e.g. one with write capability which is different to the domain that clients will pull the image from e.g. a read-only mirrored registry. 
There could also be different variations on the docker reference (in repository / tag) such that the image can physically transit other system before becoming available for signature verified consumption at the intended destination.  

While that can be achieved today with `standalone-sign` and a temporary directory store, from the documentation "This is primarily a debugging tool, useful for special cases, and usually should not be a part of your normal operational workflow; use skopeo copy --sign-by instead to publish and sign an image in one step." 

We are looking for something that is supported as "part of our normal workflow". 

I envisage this could be provided using a new option on `copy --sign-by` such as `--docker-reference "client.domain.io/my/product/image:tag"`  which could perhaps be provided multiple times to create multiple signatures.

Jamstah:
This enables pushing to registries where the push and pull uris may be
different, for example where pushed images are mirrored to a read only
replica for distribution.

Closes #1588

### DIFF
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -30,6 +30,7 @@ type copyOptions struct {
 	removeSignatures    bool                      // Do not copy signatures from the source image
 	signByFingerprint   string                    // Sign the image using a GPG key with the specified fingerprint
 	signPassphraseFile  string                    // Path pointing to a passphrase file when signing
+	signIdentity        string                    // Identity of the signed image, must be a fully specified docker reference
 	digestFile          string                    // Write digest to this file
 	format              commonFlag.OptionalString // Force conversion of the image to a specified format
 	quiet               bool                      // Suppress output information when copying images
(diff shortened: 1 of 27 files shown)


cmd/skopeo/copy.go
```diff
@@ -231,11 +233,17 @@ func (opts *copyOptions) run(args []string, stdout io.Writer) (retErr error) {
 		return err
 	}
 
+	identity, err := reference.ParseNamed(opts.signIdentity)
```
mtrmac, line 236: This fails if `opts.signIdentity` is `""`; so this parsing needs to be conditional on that.

cmd/skopeo/copy.go
```diff
@@ -231,11 +233,17 @@ func (opts *copyOptions) run(args []string, stdout io.Writer) (retErr error) {
...
+	if err != nil {
+		return fmt.Errorf("Could not parse sign identity: %v", err)
```
mtrmac, line 238: Maybe
```suggestion
		return fmt.Errorf("Could not parse --sign-identity: %v", err)
```

cmd/skopeo/copy.go
mtrmac, line 236: Non-blocking:
```suggestion
	signIdentity, err := reference.ParseNamed(opts.signIdentity)
```

mtrmac (commented):
Thanks!

Please update also `docs/skopeo-copy.1.md` and `completions/bash/skopeo`.

Jamstah:
@mtrmac apologies, I opened this pr too soon and was late to switch to draft!

Thanks for the comments so far, will make sure they all go in for when its ready for review.

mtrmac (approved):
LGTM. Thanks!

PR #1613, opened 2022-03-31
[release-1.6] Tag 1.6.2

mtrmac:


mtrmac:
@lsm5 @jnovy FYI. (The release should be tagged in the middle of this PR.)

rhatdan:
I did not think we did -dev tags in released branches?

mtrmac:
It’s quite possible, in that case I’m just unaware of that convention. In the Skopeo context, historical practice varies.

(I think _some_ marker to make sure the 1.6.2 release is unambiguous, even if we latter use the branch, is valuable. I don’t care about the particular name, if `-dev` is perceived to imply there is active development going on on that branch; it could be `-unreleased` or `-release-branch` or whatever.)

mtrmac:
@jnovy After this PR, there  would be a version `v1.6.2` tagged either way.

This would only affect about the version reported in any future cherry-pick, _if_ it happened without tagging a post-v1.6.2 release. From the upstream perspective, it is valuable to be able to differentiate the “real” v1.6.2 from a cherry-pick of a different tree.

rhatdan:
I think it is fine to leave it, but we almost always do a back port and cut a release.  We don't have long lasting changes to a release branch.

lsm5 (approved):
LGTM

mtrmac:
Merging so that this doesn’t block packaging work; we can adjust the post-release name later.

PR #1615, opened 2022-04-06
Improve error messages when system image missing

Jamstah:
When inspecting an image index where the platform image for the current
architecture is missing, be clearer in the error message about what is
wrong.

### DIFF
--- a/systemtest/010-inspect.bats
+++ b/systemtest/010-inspect.bats
@@ -102,7 +102,7 @@ END_EXPECT
 
     # By default, 'inspect' tries to match our host os+arch. This should fail.
     run_skopeo 1 inspect $img
-    expect_output --substring "parsing manifest for image: choosing image instance: no image found in manifest list for architecture $arch, variant " \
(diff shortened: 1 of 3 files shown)


Jamstah:
For background, see https://github.com/distribution/distribution/pull/3536

mtrmac (commented):
Thanks!

- A change like this, which needs to be coordinated between c/image and Skopeo, is a bit convoluted to do. See https://github.com/containers/skopeo/blob/main/CONTRIBUTING.md#dependencies-management , essentially we’ll need a “finished” c/image PR, along with a WIP Skopeo test update PR.
- WRT the “error” words, those were deliberately removed in https://github.com/containers/image/pull/1240 . I’m not saying these must not be re-added — just that there probably should be some argument for why these cases are unusual. Cc: @rhatdan 
- WRT the “found target platform image” change, sure, that looks useful.

mtrmac:
No worries – and we _will_ need a Skopeo PR for the test update anyway.

Issue #1619, opened 2022-04-11
fea: `skopeo sync` support use src or dest  use oci format


tanguofu:
oci format did support the `dest_shared_blobs` which can reduce the images size when there are lots of images.

oci format also support the `src_shared_blobs` which can reduce the unnecessary downloads.

mtrmac:
(There’s more discussion about the tradeoffs in #1620; sharing blobs does sound desirable, but using the OCI format is probably not the right way todo it.)

PR #1620, opened 2022-04-11
fea(#1619): skopeo sync support use src or dest use oci format

tanguofu:


cmd/skopeo/sync.go
```diff
@@ -186,6 +198,16 @@ func destinationReference(destination string, transport string) (types.ImageRefe
...
 		}
 		imageTransport = directory.Transport
+	case oci.Transport.Name():
+		dir := path.Dir(destination)
+		if err := os.MkdirAll(destination, 0755); err != nil {
+			return nil, errors.Wrapf(err, "Error creating directory:%s for image %s", dir, destination)
+		}
+		if ref, err := reference.ParseNormalizedNamed(destination); err != nil {
+			tagref, _ := reference.TagNameOnly(ref).(reference.Tagged)
+			destination = fmt.Sprintf("%s:%s", reference.Path(ref), tagref)
```
mtrmac, line 208: This is pre-existing, but now is a good time to clean up:

Let’s stop formatting `imageTransport` and `destination` strings only to parse them below in `imageTransport.ParseReference`; instead, let’s have every case create `destRef` directly (e.g. in this case using `oci.NewReference`).

mtrmac, line 208: Isn’t this building $top-dir/$repo directories, one for each repository?

Why build an entirely new directory structure on top of OCI, when we can just build one large OCI repository?

cmd/skopeo/sync.go
```diff
@@ -531,19 +619,23 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) (retErr error) {
...
 	}
 
-	if opts.source == opts.destination && opts.source == directory.Transport.Name() {
-		return errors.New("sync from 'dir' to 'dir' not implemented, consider using rsync instead")
+	if opts.source == opts.destination && contains(opts.source, []string{directory.Transport.Name(), oci.Transport.Name()}) {
+		return errors.Errorf("sync from '%s' to '%s' not implemented, consider using rsync instead", opts.source, opts.destination)
+	}
+
+	if opts.preserveDigests && opts.destination == oci.Transport.Name() {
+		return errors.New("oci DESTINATION transports not support --preserve-digests options")
```
mtrmac, line 638:
```suggestion
		return errors.New("using the oci transport as a destination does not support --preserve-digests")
```

mtrmac, line 638: Note to self: All of this looks increasingly clumsy; take a deeper look.

mtrmac (commented):
Thanks!

For now, just a quick look:
- You have already identified the downside: OCI can’t represent the original format of all images. Still, that might not be an issue for some.
- I feel rather strongly that the code to parse/list images from an OCI repository should exist as a public API in c/image, not here.
- Is there a specific reason to use this layout (one OCI repo per registry repo)? It seems to unnecessary limit data sharing, which was the motivation for this feature in the first place.

cmd/skopeo/sync.go
mtrmac: (To clarify, I meant the existing code structure — it feels like a collection of special cases disconnected from the actual implementation, and can get out of sync.)

tanguofu:
thanks for review,  my usecase is distribute  lots of docker images from public online registry to private  offline registry which no internet access.

sun as the `sync.yaml`:
```
"docker.io":
  "images":
    "library/mysql":
    - "8.0.27"
    - "8.0.25"
    "library/registry":
    - "latest"
"quay.io":
  "images":
    "minio/mc":
    - "RELEASE.2021-11-16T20-37-36Z"
    "minio/minio":
    - "RELEASE.2021-11-24T23-19-33Z"
```
i will export  the images of `sync.yaml` by :  
````
skopeo sync --src yaml --dest oci --format oci --src-shared-blob-dir=cachedir --dest-shared-blob-dir=./images/blobs ../test.yaml images
tar zcvf image.tar.gz   images
````
when came to offline  registry `127.0.0.1:5000` , import the images by
```
skopeo sync --src oci --dest docker --format oci --src-shared-blob-dir=./images/blobs  --dest-tls-verify=false  --scoped  ./images/ 127.0.0.1:5000 
```
so it is very nessary that:
1. use oci can  benefits from the `--src-shared-blob-dir` which did not copy already download blobs
2. use oci can  benefits from the `---dest-shared-blob-dir` which can shared the blobs to one dirs, reduce the `image.tar.gz` size. as there  hundreds of image, the sum size is hundreds of GB.
3. use `library/mysql` `library/registry`  as dir name,  then it will put the offline  registry `127.0.0.1:5000` with same docker image namespace. keep the images name not change also can avoid same  image;tag name conflict.

i hope this case can show your the necessity of  to support oci fomart @mtrmac 

as your mention that `OCI can’t represent the original format of all images`, could u explain or give some case that oci 
 incompatible with docker image format. 

thank you very much.

File: cmd/skopeo/utils.go [hunk 3]
```diff
@@ -354,3 +412,19 @@ func adjustUsage(c *cobra.Command) {
...
+	_, err := os.Stat(destination)
+	if err == nil {
+		return errors.Errorf("Refusing to overwrite destination directory %q", destination)
+	}
+	if !os.IsNotExist(err) {
+		return errors.Wrap(err, "Destination directory could not be used")
+	}
+	// the directory holding the image must be created here
+	if err = os.MkdirAll(destination, 0755); err != nil {
+		return errors.Wrapf(err, "Error creating directory for image %s", destination)
```
mtrmac, line 426: Given the “Refusing to overwrite” behavior and this text saying “for image %s”, this seems unlikely to be reused for anything else — it is rather specific to the `dir:` transport; so I’m not sure that splitting this out is worthwhile.

I guess it depends on if we can find a good clear name for this behavior. `checkExistAndMkdir` is not that — that feels like two completely separate operations collected together, and `checkExist` doesn’t indicate anything about the consequences. So, for now it seems to me that leaving this in-line is _much_ easier than finding a good name for a separate function, but there might well be a great name / way to express this that I’m missing.

docs/skopeo-sync.1.md
```diff
@@ -19,14 +19,15 @@ One of the problems of prefixing a destination with its transport is that, the r
 Available _source_ transports:
  - _docker_ (i.e. `--src docker`): _source_ is a repository hosted on a container registry (e.g.: `registry.example.com/busybox`).
  If no image tag is specified, skopeo sync copies all the tags found in that repository.
- - _dir_ (i.e. `--src dir`): _source_ is a local directory path (e.g.: `/media/usb/`). Refer to skopeo(1) **dir:**_path_ for the local image format.
+ - _dir_ (i.e. `--src dir`): _source_ is a local c path (e.g.: `/media/usb/`). Refer to skopeo(1) **dir:**_path_ for the local image format.
```
mtrmac, line 22: This change seems to be a typo.

cmd/skopeo/sync.go
```diff
@@ -76,25 +83,27 @@ type sourceConfig map[string]registrySyncConfig
...
 	}
 
 	cmd := &cobra.Command{
 		Use:   "sync [command options] --src TRANSPORT --dest TRANSPORT SOURCE DESTINATION",
 		Short: "Synchronize one or more images from one location to another",
 		Long: `Copy all the images from a SOURCE to a DESTINATION.
 
-Allowed SOURCE transports (specified with --src): docker, dir, yaml.
-Allowed DESTINATION transports (specified with --dest): docker, dir.
+Allowed SOURCE transports (specified with --src): docker, dir, yaml, oci.
```
mtrmac, line 105: Please sort the values in alphabetical order.

cmd/skopeo/sync.go
```diff
@@ -477,7 +513,24 @@ func imagesToCopy(source string, transport string, sourceCtx *types.SystemContex
...
 		descriptors = append(descriptors, desc)
+	case oci.Transport.Name():
+		desc := repoDescriptor{
+			Context: sourceCtx,
+		}
 
+		if _, err := os.Stat(source); err != nil {
+			return descriptors, errors.Wrap(err, "Invalid source directory specified")
+		}
+		desc.DirBasePath = filepath.Clean(source)
```
mtrmac, line 524: `dirBasePath` is only documented to apply to `dir:` — and AFAICS isn’t used in the PR as is anyway.

cmd/skopeo/sync.go
```diff
@@ -609,22 +674,27 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) (retErr error) {
...
-				// docker -> dir or docker -> docker
-				destSuffix = ref.DockerReference().String()
+				// docker -> dir,docker,oci, trim domain
+				dockerRef := ref.DockerReference()
+				tagref, _ := reference.TagNameOnly(dockerRef).(reference.Tagged)
+				destSuffix = fmt.Sprintf("%s:%s", reference.Path(dockerRef), tagref.Tag())
+
 			case directory.Transport:
-				// dir -> docker (we don't allow `dir` -> `dir` sync operations)
+				// dir -> docker or oci -> docker(we don't allow `dir` -> `dir` sync operations)
```
mtrmac, line 683: This seems incorrect — `ref` refers to the source, and `oci.Transport` is handled below.

cmd/skopeo/sync.go
```diff
@@ -283,6 +289,36 @@ func imagesToCopyFromDir(dirPath string) ([]types.ImageReference, error) {
...
+	}
+	defer indexJSON.Close()
+
+	index := &imgspecv1.Index{}
+	if err := json.NewDecoder(indexJSON).Decode(index); err != nil {
+		return sourceReferences, errors.Wrapf(err, "Cannot decode index.json for transport %q and Path: %q", oci.Transport.Name(), dirPath)
+	}
+
+	for _, md := range index.Manifests {
+		ref, err := oci.NewReference(dirPath, md.Annotations[imgspecv1.AnnotationRefName])
```
mtrmac, line 312: (`AnnotationRefName` can legitimately be not set. Also https://github.com/containers/common/blob/31e8cc4aeeabbb50538fe3948d1fa3d7506a733a/libimage/pull.go#L176 . Either way, this should be in c/image.)

cmd/skopeo/sync.go
```diff
@@ -166,36 +176,32 @@ func parseRepositoryReference(input string) (reference.Named, error) {
 // destinationReference creates an image reference using the provided transport.
```
mtrmac, line 176: Something like
```suggestion
// destinationReference creates an image reference, combining the provided transport, the user-specified destination, and destSuffix containing a part of the source image’s name that should be used in the destination name.
```
otherwise the purpose of `destSuffix` is not obvious.

Possibly the order of arguments should then be changed to `transport, destination, destSuffix` so that it matches the usual order of the output.

cmd/skopeo/sync.go
```diff
@@ -166,36 +176,32 @@ func parseRepositoryReference(input string) (reference.Named, error) {
...
 	default:
 		return nil, errors.Errorf("%q is not a valid destination transport", transport)
 	}
-	logrus.Debugf("Destination for transport %q: %s", transport, destination)
 
-	destRef, err := imageTransport.ParseReference(destination)
-	if err != nil {
-		return nil, errors.Wrapf(err, "Cannot obtain a valid image reference for transport %q and reference %q", imageTransport.Name(), destination)
+	if destRefErr != nil {
+		return nil, errors.Wrapf(destRefErr, "Cannot obtain a valid image reference for transport %q and reference %q", transport, destString)
```
mtrmac, line 200: `destString` is not actually used on the OCI path … I guess this should list all of `transport, destination, destSuffix` individually..

cmd/skopeo/sync.go
mtrmac: Yeah, let’s split this up, to remove the transport list duplication, and so that `imagesToCopy` and `destinationReference` don’t grow very large.

Turn `imagesToCopy` into a
```go
var sourceImageListers map[string]func(…)… = {
   docker.Transport.Name(): listDockerSourceImages,
   …
}
```
(`imagesToCopy` has almost no shared code, so that can be just carved up)

and similarly `destinationReference` should turn into
```go
var destinationReferenceConstructors map[string]func(…)… = {
   docker.Transport.Name(): createDockerDestinationReference,
   …
}
```

Then these `contains` checks turn into
```go
sourceImageLister, ok := sourceImageListers[opts.source]
if !ok { fail }
```

---

That feels like a necessary but arguably unrelated cleanup — I can work on that in a separate PR, if you’d like.

mtrmac (commented):
A full review now; https://github.com/containers/skopeo/pull/1620#discussion_r847331828 is still the major structural blocker.

(Also note that tests are failing; see the logs, something like `gofmt -s -w cmd/skopeo/copy.go` might help.)

PR #1626, opened 2022-04-22
Update vendor of containers/(common,storage,image)

rhatdan:


vrothberg (commented):
LGTM

vendor/modules.txt
```diff
@@ -238,7 +238,7 @@ github.com/golang/protobuf/ptypes/timestamp
 github.com/google/go-intervals/intervalset
 # github.com/google/uuid v1.3.0
 github.com/google/uuid
-# github.com/gorilla/mux v1.8.0
+# github.com/gorilla/mux v1.7.4
```
mtrmac, line 241: What has happened here?

Trying to reproduce, a `go get github.com/containers/common@v0.47.5-0.20220421111103-112a47964ddb` adds an
```diff
 +       github.com/gorilla/mux v1.8.0 // indirect
```
entry to `go.mod` (and several others like that), so no downgrade happens. Was this PR done in some other way that doesn’t protect against downgrades?

Ultimately, AFAICS that doesn’t change the contents of the vendored code, so it doesn’t matter, but I’m worried that next time such a downgrade _could_ make a difference.

mtrmac (commented):
I mean, sure — but could we record why is this being done (this way), please?

What makes the three updates simultaneously so important that we want to upgrade (all of them), _and_ so unimportant that they aren’t tagged as upstream releases? All three at the same time?!

rhatdan:
We are preparing for RC1 release of Podman, and are trying to avoid doing multiple vendor dances.  The goal was to get these three in podman, buildah and skopeo and make sure they pass through their CI systems.  We will be bumping releases of all three in the next couple of weeks and then releasing new versions of Podman, Buildah and Skopeo.

vendor/modules.txt
rhatdan: These vendors were done the same way every vendor was done.

make vendor-in-container

Nothing special.

mtrmac: `vendor-in-container` only updates based on a `go.mod`; the initial operation to point at a newer versions must have been done in some other way.

---

(Unrelated: `vendor-in-container` uses `quay.io/libpod/golang:1.16`; should that be updated? Podman is using 1.17 at least.)

mtrmac:
> The goal was to get these three in podman, buildah and skopeo and make sure they pass through their CI systems.

Thanks, that’s helpful.

PR #1634, opened 2022-05-02
vendor github.com/json-iterator/go@v1.1.12

vrothberg:
Fix a runtime segfault when being compiled with recent versions of Go.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2079759
Signed-off-by: Valentin Rothberg <vrothberg@redhat.com>

@TomSweeneyRedHat @baude @mtrmac PTAL


mtrmac (approved):
Sure, why not. (c/storage, which brings in the dependency, has been been updated.)

This is going to effectively disappear after updating to a new c/storage, isn’t it?

vrothberg:
> This is going to effectively disappear after updating to a new c/storage, isn’t it?

I think so yes.

mtrmac:
> This is going to effectively disappear after updating to a new c/storage, isn’t it?

Oh, this is actually on an older back ports branch…

PR #1642, opened 2022-05-10
Config files live in /usr/local/etc on FreeBSD (highlight in release notes!)

dfr:
Config files for non-base packages live in /usr/local/etc and **https://github.com/containers/image/pull/1544** changes the default paths in containers/image. This changes the skopeo Makefile to match.


mtrmac (approved):
LGTM.

Thanks!

rhatdan:
LGTM

mtrmac:
**NOTE**: This will only work if we vendor an updated c/image before a Skopeo release. Historically that’s consistently been the practice, just making a note here for posterity if something went wrong.

PR #1647, opened 2022-05-18
use spf13/cobra to generate shell completions 

Luap99:
Replaces #1500

see commits for details


Makefile
```diff
@@ -156,8 +159,18 @@ docs: $(MANPAGES)
 docs-in-container:
 	${CONTAINER_RUN} $(MAKE) docs $(if $(DEBUG),DEBUG=$(DEBUG))
 
+.PHONY: completions
+completions: bin/skopeo
+	# key = shell, value = completion filename
+	declare -A outfiles=([bash]=%s [zsh]=_%s [fish]=%s.fish [powershell]=%s.ps1);\
```
mtrmac, line 165: This seems `bash`-specific.

install.md
```diff
@@ -212,6 +212,65 @@ Building in a container is simpler, but more restrictive:
 $ make binary
 ```
 
+### Build and Install shell completion scripts
```
mtrmac, line 215: Does this need an extra step? AFAICS `make install` would, as of this PR already, cause generation of the completion scripts, and then install the completion scripts, just as it installs man pages or config files without having to be specifically instructed.

cmd/skopeo/main.go
```diff
@@ -113,6 +109,8 @@ func createApp() (*cobra.Command, *globalOptions) {
 		tagsCmd(&opts),
 		untrustedSignatureDumpCmd(),
 	)
+	// add the cobra completion command to auto generate shell completion scripts.
+	completion.AddCompletionCommand(rootCommand)
```
mtrmac, line 113: https://github.com/spf13/cobra 1.3.0 now provides
```go
rootCmd.CompletionOptions.HiddenDefaultCmd = true
```

Do we really need a special case command, different from the upstream one? I see some value in using the library as the upstream intends it (and, sure, some value in being consistent among all c/common users, but the latter seems easier to change.)

install.md
```diff
@@ -212,6 +212,65 @@ Building in a container is simpler, but more restrictive:
...
+By default the completion scripts will show extra descriptions as help:
+```
+$ skopeo l[TAB]
+list-tags  (List tags in the transport/repository specified by the SOURCE-IMAGE)  logout     (Logout of a container registry)
+login      (Login to a container registry)
+```
+
+To turn this of you can generate the scripts with the `skopeo completion --no-desc` option,
+see below for the individual shell instructions.
+
```
mtrmac, line 238: Do users care strongly enough about this to make everyone _think_ about making a decision? To an extent, the purpose of `install.md` is make installing Skopeo easy.

install.md
```diff
@@ -212,6 +212,65 @@ Building in a container is simpler, but more restrictive:
...
+If you want to install the files to the system directories (on linux) use:
+```bash
+$ sudo make install-completions
+```
+
+By default the completion scripts will show extra descriptions as help:
+```
+$ skopeo l[TAB]
+list-tags  (List tags in the transport/repository specified by the SOURCE-IMAGE)  logout     (Logout of a container registry)
```
mtrmac, line 232: (A new line before `logout`?)

install.md
```diff
@@ -212,6 +212,65 @@ Building in a container is simpler, but more restrictive:
...
+To make it available for all fish sessions run:\
+`skopeo completion -f ~/.config/fish/completions/skopeo.fish fish`
+
+#### POWERSHELL
+To load the completion script into the current session run:
+`skopeo completion powershell | Out-String | Invoke-Expression`
+
+To make it available in all powershell sessions that a user has, write the
+completion output to a file and source that to the user's powershell profile.
+More information about profiles is available with **Get-Help about_Profiles**.
```
mtrmac, line 272: Why all these individual commands?

`make install{-completions}` already puts the scripts into auto-loaded locations, so I’d expect all of it to work without any further action (on the next bash session restart).

Documenting how to load the completion within the current session feels potentially useful, but a bit too specific —  the users would probably benefit from a generic knowledge how to reload all completions into the session. (We also run into the inconsistency of hiding the `skopeo completion` subcommand and not having a man page, but documenting how to use it here.)

Isn’t documenting how to _individually_ write the config files via `completion -f` completely redundant with the `make install{-completions}` command?

---

I think this could all be just dropped; or, alternatively, _after_ the `make install`, add some note that completions will only work on new shell instances, and point at upstream *-completion documentation for how to reload updated/added completions into existing shells.

mtrmac (commented):
Thanks for taking this on!

Just a minimal skim on the external aspects; I haven’t currently taken the time to (re)learn how the completion mechanism / configurability / extensibility works.

cmd/skopeo/main.go
Luap99: The upstream command was only added after I already had implemented it in podman, unfortunately the command is not compatible with the podman completion command. The podman command offers the `--file` and `--no-desc` option while the upstream command only has `--no-description`. Also the podman one adds an extra line that the file was generated automatically: https://github.com/containers/common/blob/1bc8353fc42e4a766da619a0b4134a2cfbcd9652/pkg/completion/command.go#L93

I would like to use the upstream command but I also do not want to break podman. I cannot modify the upstream command to accept the current podman options.
We do not have this problem in skopeo of course but I think having a consistent command across skopeo,buildah,podman is valuable.
Alternatively we can start using the upstream command here and change the podman command with v5.0. This would reduce at least some maintenance.

install.md
Luap99: I cannot answer that. I personally do not know anybody using it but this doesn't say anything.
When the description support was first introduced in cobra I had to use this option because of many bugs with incorrect rendering of the descriptions. I am not aware of any problems with the current cobra completion scripts for all shells so I no longer use it.
But is is a personal choice and upstream provides this option so I see no reason to not expose this option.

install.md
Luap99: When you have enough width the shell will place them on the same line.

install.md
Luap99: So I agree the almost all users will not care since the package manger (or make install) should install the scripts for them.
However I think the are also cases where this does not apply:

- I do not know what the status of windows support is but powershell users would have to do it manually so they need documentation on how to do it. (Also you can use powershell on linux too)
- What if skopeo is distributed as single static binary. This seems to be common for many go applications. In this case the user has to manually load the script so they need instructions.

I think pointing at the upstream docs would be good enough. At least it would prevent us from having to manually update it when something changes.

install.md
Luap99: yes I think make install covers it but there can be cases where you would want to only generated the script.
Not everyone is using the make rules. Given that this removes the current script from the repo it would be good to have some doc on how to recreate it.

mtrmac: `install.md` guides users through what we think is the best way to install Skopeo. It’s not a reference documentation for `Makefile` documenting every target and variable. (We don’t really have that anywhere… and I’m not sure we should, at least given the current shape of the `Makefile`: the targets and variables range from clearly end-user-targeted to clearly internal implementation details, with a lot, sadly, somewhere in the middle.)

> but there can be cases where you would want to only generated the script.

Like what? It seems to me that script and the binary should match, so if both are installed, it always makes sense to do both together. Is that not the case?

install.md
mtrmac: > What if skopeo is distributed as single static binary

Ugh. You are quite right, that could happen, and those users could still benefit from completions.

OTOH these tend to be users who don’t have root (making installation into `/etc/` or `/usr` unviable), or they want a static binary to run from a “we-don’t-care-what-distribution-it-is” container in some automated system. Actual interactive by users who just copy around binaries seems rarer — especially because we don’t provide such a binary.

BTW just copying `bin/skopeo` around basically doesn’t work for interactive use because Skopeo refuses to copy images without `policy.json` (or an opt-out CLI flag that needs to be typed every time). So… if anything, we could document the GNU standards-defined `make install DESTDIR=…` to create a complete set of files that {can, need to} be installed.

install.md
mtrmac: Ah, my fault for not actually trying that.

install.md
mtrmac: Given that the shortest installation method is just `make install`, I guess we’d want a Makefile variable to control that flag. (And should that be just `COMPLETION_GENERATION_FLAGS=arbitrary_string`, or actually `COMPLETIONS_NO_DESCRIPTIONS=true`? The latter is more work to implement and document, OTOH invariant to using the upstream or c/common completion generation command.)

cmd/skopeo/main.go
Luap99: I will change it to the upstream one. This is better in the long run and more consistent with other applications which use cobra. I can change podman with the next major release and I doubt that anybody will care if the option is renamed.

The better thing about using the upstream one is that `skopeo completion bash --help` would already show information on how to load the script. I think in this case I could just add this as note in the install guide.

install.md
Luap99: Updated to just point at `skopeo completion --help`

install.md
Luap99: Not sure if it is worth the added complexity. I would wait until someone actually requests this.

mtrmac: Works for me.

Luap99:
@mtrmac OK I think this is ready from my side.
I left some pointers in the last commit on how to make the completion experience better. For reference here are the podman completion functions: https://github.com/containers/podman/blob/main/cmd/podman/common/completion.go

If skopeo has a format flag which accepts go templates we could move the `AutocompleteFormat()` function to c/common so that skopeo can use this functionality as well.

cmd/skopeo/completion/completions.go
```diff
@@ -0,0 +1,16 @@
+package completion
+
+import (
+	"github.com/containers/image/v5/transports"
+	"github.com/spf13/cobra"
+)
+
+// AutocompleteSupportedTransports list all supported transports with the colon suffix.
+func AutocompleteSupportedTransports(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
mtrmac, line 9: For extra peace of mind (I’d rather not make an unexpected Go API stability promise), it would be nice not to make this available as a function that could be imported by third parties.

Probably simplest would be to move this to `cmd/skopeo.autocomplete…`. Alternatively, `cmd/skopeo/internal.Autocomplete…` or `cmd/skopeo/internal/completion.Autocomplete`. I don’t care at all which one is chosen.

mtrmac (commented):
Thanks!

Code LGTM, just a few more nits about the infrastructure.

cmd/skopeo/completion/completions.go
Luap99: Is this actually a concern? Is there anywhere stated that skopeo code can be used as library and that it will keep the interfaces stable?
Also that function signature is already defined by the cobra library (since we have to set that function to the ValidArgsFunction) so it cannot be changed until upstream changes it.

mtrmac: It’s a tiny concern, noting that https://github.com/containers/skopeo/blob/main/cmd/skopeo/inspect/output.go was explicitly defined as an external API, and it is side by side with this new subpackage. (Compare `go doc ./cmd/skopeo` (nothing) and `go doc ./cmd/skopeo/inspect`.)

Hypothetically, assuming we manage to make this completion code much smarter to also autocomplete the image names (for some transports), I can easily imagine e.g. Podman wanting to share the code — and making it private would be a reminder to move it to c/common instead.

Ultimately I feel fairly comfortable with telling anyone “this is not a supported API, I’m sorry, go away”, but we can make that situation impossible right now, at a trivial cost, so I think we might just as well do that.

--

The function’s _signature_ (and name) is not a problem at all, I’m talking purely about the functions’ external visibility. E.g. move the existing file to `cmd/skopeo/completions.go`, globally `s/AutocompleteSupported/autocompleteSupported/g`.

Makefile
```diff
@@ -176,9 +187,14 @@ ifneq ($(DISABLE_DOCS), 1)
...
-install-completions:
-	install -m 755 -d ${DESTDIR}${BASHCOMPLETIONSDIR}
-	install -m 644 completions/bash/skopeo ${DESTDIR}${BASHCOMPLETIONSDIR}/skopeo
+install-completions: completions
+	install ${SELINUXOPT} -d -m 755 ${DESTDIR}${BASHINSTALLDIR}
+	install ${SELINUXOPT} -m 644 completions/bash/skopeo ${DESTDIR}${BASHINSTALLDIR}
+	install ${SELINUXOPT} -d -m 755 ${DESTDIR}${ZSHINSTALLDIR}
+	install ${SELINUXOPT} -m 644 completions/zsh/_skopeo ${DESTDIR}${ZSHINSTALLDIR}
+	install ${SELINUXOPT} -d -m 755 ${DESTDIR}${FISHINSTALLDIR}
+	install ${SELINUXOPT} -m 644 completions/fish/skopeo.fish ${DESTDIR}${FISHINSTALLDIR}
```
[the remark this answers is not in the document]
Luap99: I removed it.

cmd/skopeo/completion/completions.go
Luap99: done

rhatdan:
LGTM

PR #1649, opened 2022-05-19
Stop calling gpgme-config

mtrmac:
Note: This vendors in unreleased c/image, and all of this by no means urgent. It might well make sense to defer this PR until a proper c/image release.

As of the just-updated `github.com/proglottis/gpgme` 0.1.2, the `gpgme` subpackage uses CGo’s native `#cgo pkg-config` support to find the relevant libraries, and we no longer need to manually set `CGO_CFLAGS` and `CGO_LDFLAGS`. So stop doing that.

Note that the `proglottis/gpgme` update (implied by vendoring c/image) means the minimal supported version of GPGME is 1.13.0.

Also remove a workaround for Glibc < 2.24.


vrothberg (approved):
LGTM

rhatdan:
LGTM

PR #1652, opened 2022-05-25
[CI:DOCS] Pin actions to a full length commit SHA

cevich:
This corresponds to similar changes made in Buildah and Podman: https://github.com/containers/podman/pull/13564 https://github.com/containers/buildah/pull/3872

+ Pin actions to a full length commit SHA is currently the only way
  to use an action as an immutable release. Pinning to a particular SHA
  helps mitigate the risk of a bad actor adding a backdoor to the action's
  repository, as they would need to generate a SHA-1 collision for a valid
  Git object payload. Ref:
  https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

+ Explicitly set permissions for actions to minimum required.  The
  defaults are (unfortunately) overly permissive: Ref:
  https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

### DIFF
--- a/.github/workflows/check_cirrus_cron.yml
+++ b/.github/workflows/check_cirrus_cron.yml
@@ -18,6 +18,9 @@ on:
     # Debug: Allow triggering job manually in github-actions WebUI
     workflow_dispatch: {}
 
+permissions:
+    contents: read
+
 env:
     # Debug-mode can reveal secrets, only enable by a secret value.
     # Ref: https://help.github.com/en/actions/configuring-and-managing-workflows/managing-a-workflow-run#enabling-step-debug-logging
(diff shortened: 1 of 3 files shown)


mtrmac (commented):
ACK.

Should we use Dependabot to keep updating the actions? Per https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot it seems to need to be explicitly opted into.

Alternatively, do we just risk using stale actions? Looking at other projects, the Dependabot updates to actions seem to be pretty frequent, and if we got into the habit of approving the Dependabot updates without reading the diffs, the whole point of this PR would be undone.

cevich:
> Should we use Dependabot

In theory it's a good idea, and I appreciate you're asking/suggesting.  My preference is to use what we have that works, until it stops working or we hear about some new horrible security problem.  Right now our use of github actions is pretty limited, and low-risk.  The only things which have any kind of "write-access" are:

+ Multi-arch image builder (replacement being tested in podman, rollout soon to other repos.)
+ Cirrus-Cron job checker, has access to send e-mail but only runs from 'main' so pretty low-risk.
+ Stale issue management.  Any abuse will immediately be noticed through notifications.
+ PR title changing for branches. Abuse will immediately be noticed.

So going forward, I'm in favor of the "leave it alone" and "limit use of github actions" for anything complex/, critical, or consequential.

mtrmac:
Works for me; please merge whenever you think this is ready.

TomSweeneyRedHat:
Not a SME by any means, but LGTM

PR #1653, opened 2022-05-25
Updated skopeo logo with new artwork

mairin:


TomSweeneyRedHat:
Rather than replace, could we rename the old one to *_old.svg?

rhatdan:
Git will recall the old one.

PR #1656, opened 2022-05-26
Update to gopkg.in/yaml.v3 v3.0.0

mtrmac:
... to include a fix for CVE-2022-28948 = GHSA-hp87-p4gw-j4gq .

Note that the package is only used for Skopeo’s tests, so Skopeo’s users can't reach the vulnerable code.


rhatdan:
LGTM
@vrothberg PTAL

go.mod
```diff
@@ -21,4 +21,5 @@ require (
 	github.com/syndtr/gocapability v0.0.0-20200815063812-42c35b437635
 	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c
 	gopkg.in/yaml.v2 v2.4.0
```
vrothberg, line 23: Shall we update skopeo-sync to use v3.0 as well?

mtrmac: That’s a good thought. I decided not to do that here:
- That wouldn’t immediately get rid of the `v2` dependency (via `ghodss/yaml` and `ocicrypt` right now.
- It feels somewhat out of scope of this “trivially silence Dependabot” PR; the two seem to implement different versions of YAML, although at least the top-level README does suggest that the library tires to be compatible. The upgrade is almost certainly worth doing, but it deserves at least a cursory investigation.

vrothberg: SGTM, thanks!

vrothberg (approved):
LGTM

go.mod
rhatdan: https://github.com/containers/ocicrypt/pull/63 Is merged.

rhatdan: @mtrmac Does the attached patch make sense?

[diff.patch](https://github.com/containers/skopeo/files/8818377/diff.patch.txt)

mtrmac: - I don’t expect that to compile, in the first place: `gopkg.in/yaml.v3` doesn’t have a `YAMLToJSON` method.
- I don’t immediately know (the “deserves at least a cursory investigation” part). Why do _you_ think this does make sense, so that I don’t need to re-do all your analysis from scratch?
  - At minimum, the README of `ghodss/yaml` claims that that package uses `json:` field tags, while `gopkg.in/yaml` documents `yaml:` field tags only. If the two really differ, then the field tags would need adjusting.
  - (Asking about the YAML versions and semantics differences would probably border on just being a pest, because I’m inclined not to worry about that either as long as tests pass… still, if there _is_ a known concern, it would be good to know now.)
- Right now, `github.com/sigstore/rekor/pkg/client/yaml.go` also imports `ghodss/yaml`. So immediately, we would benefit, but not necessarily afterwards. Of course Rekor could also be adjusted…

rhatdan: I was doing a cursory inspection and did not see a compilation issue, so I thought I would ask.

PR #1661, opened 2022-06-03
[CI:BUILD] Cirrus: Migrate multiarch build off github actions

cevich:
The github actions workflow for this operation is complex and difficult
to maintain.  For several months now a replacement has been running well
in the podman repository.  It's scripts/components are centralized,
versioned, unit, and integration tested.  Add cirrus tasks to run the
build, and another task to allow test builds in a PR.

This also adds support for a new magic CI string: `[CI:BUILD]`.
With this string in the PR title, automation will only do basic build
verification, and enable testing of the multi-arch build process.

Otherwise, many tasks were updated to not be created when running the
cirrus-cron multi-arch image builds, since this would simply be a waste
of time and invitation for flakes.

Lastly, since only native tooling is used in the new build process,
rename all the recipes to `Containerfile`.


cevich:
This is ready for review/merge.  Note: The contrib/skopeoimage/upstream build takes nearly 1:30 hours.  We should seriously consider fixing that somehow.  Maybe by installing pre-build packages based on `main` instead of compiling it all from scratch?

mtrmac:
> The contrib/skopeoimage/upstream build takes nearly 1:30 hours.

And the others almost an hour, which feels even more curious to me, considering that it’s basically just some RPM installations.

Hopefully the computers have the patience for that.

> We should seriously consider fixing that somehow. Maybe by installing pre-build packages based on main instead of compiling it all from scratch?

Where would the pre-built packages come from? I’d expect them to come from some cron job :)

Any idea where is most of the time spent? I can maybe imagine something like pre-building a build container every week, amortizing the time to install RPMs and maybe to pull some version of the repo (with the daily builds just pulling the last commits and actually building), but that’s more complexity and it could mean older = less secure RPMs in the image.

mtrmac:
> pre-built packages 

Uh, are we talking about RPMs or about `/usr/bin/skopeo`?

cevich:
> And the others almost an hour

This is fairly normal, `dnf` itself is running in emulation mode across 3 foreign architectures.  It's not fast (and neither was the old github-action workflow).

> Hopefully the computers have the patience for that.

Lol, they do.  And people are protected: The job will only run as part of cirrus-cron, or if manually triggered *AND* `[CI:BUILD]` is in the PR subject line.

> Where would the pre-built packages come from? I’d expect them to come from some cron job :)

Yeah exactly, that would be the thing to solve, and Buildah has the exact same challenge.  The thing I'm most worried about is Cirrus has a hard-coded max of 2-hours runtime.  So spreading that out with pre-built RPMs or binaries elsewhere would be good.

Lokesh has a fancy thing setup for podman-next in COPR, but no buildah or skopeo yet.  @lsm5 is this on your radar at all for "the future" sometime?

Another option is to add a job to the cirrus-cron for main that builds multi-arch binaries.  Then just consume those in the Containerfile (would work for manual image builds too).

cevich:
Clarification: I'm not super concerned about the 1:30 build time, only slightly mildly concerned :grinning:

cevich:
Note to me: Once this merges, need to add a cirrus-cron job named 'multiarch'.

lsm5:
> > And the others almost an hour
> 
> This is fairly normal, `dnf` itself is running in emulation mode across 3 foreign architectures. It's not fast (and neither was the old github-action workflow).
> 
> > Hopefully the computers have the patience for that.
> 
> Lol, they do. And people are protected: The job will only run as part of cirrus-cron, or if manually triggered _AND_ `[CI:BUILD]` is in the PR subject line.
> 
> > Where would the pre-built packages come from? I’d expect them to come from some cron job :)
> 
> Yeah exactly, that would be the thing to solve, and Buildah has the exact same challenge. The thing I'm most worried about is Cirrus has a hard-coded max of 2-hours runtime. So spreading that out with pre-built RPMs or binaries elsewhere would be good.
> 
> Lokesh has a fancy thing setup for podman-next in COPR, but no buildah or skopeo yet. @lsm5 is this on your radar at all for "the future" sometime?

If we're talking about building the latest stuff merged upstream, then `rhcontainerbot/podman-next` would be the copr to use. I don't have packages for skopeo and buildah there yet, but I can work on those within the next few days.

Would we need this for ubuntu tests as well? I can try to enable bleeding-edge ubuntu packages as well.

cevich:
> Would we need this for ubuntu tests as well? I can try to enable bleeding-edge ubuntu packages as well.

I don't really see a need there yet, but thanks for asking.

cevich:
Clarification: I believe this is ready to go.

cevich:
No worries, this is replacing something that's "working" so not much rush.

cevich:
@mtrmac is it tomorrow yet? :grin: 
(No worries, I'm sure you're busy)

contrib/skopeoimage/stable/Containerfile
```diff
@@ -0,0 +1,44 @@
...
+# This image can be used to create a secured container
+# that runs safely with privileges within the container.
+#
+FROM registry.fedoraproject.org/fedora:latest
+
+# Don't include container-selinux and remove
+# directories used by dnf that are just taking
+# up space.
+RUN dnf -y update && \
+    rpm --setcaps shadow-utils 2>/dev/null && \
```
mtrmac, line 15: Do you happen to have a reference to why this is necessary? I could only find RHEL bugs and at least https://bugzilla.redhat.com/show_bug.cgi?id=1995337#c3 (private) suggests this should have been fixed, for RHEL.

(Absolutely non-blocking, we didn’t have a reference before.)

mtrmac (commented):
> @mtrmac is it tomorrow yet? 😁

It is not tomorrow (that will be tomorrow), but it _is_ totally my fault. I’m sorry.

LGTM, and I very much appreciate the improved error detection in the Containerfile scripts.

On the shadow-utils question, feel free to just merge as is if it is not something you happen to know about.

contrib/skopeoimage/stable/Containerfile
cevich: This is something Dan knows about, we do the exact same thing in the podman and buildah images.  It replaces an old `restorecon` command we had for a long while.  I think I might be able to dig up a reference discussion at least (which may be important for your bugzilla finding)...

cevich: ...ahh it was Nalin, https://github.com/containers/podman/pull/14437#discussion_r887295318  and is Fedora specific.  Though on second reading, we probably don't need this in a skopeo image.  I'll remove it to save anybody from wondering or looking for related issues/documentation.

cevich:
Force-push: Rebased, removed `rpm --setcaps` lines as [nobody should ever need to actually run a nested container](https://github.com/containers/podman/pull/14437#pullrequestreview-992645503) inside a skopeo image.

contrib/skopeoimage/stable/Containerfile
mtrmac: … and looking at https://src.fedoraproject.org/container/fedora-container-image/blob/rawhide/f/image-build.conf (although the F30 references are _quite_ suspect, so I might be looking at the wrong place), the necessary tar options seem to be missing. Thanks, that explains things.

---

As a maintainer, my primary concern with a workaround is being able to maintain it over time. I.e. “how do I know that the workaround broke, and how can I verify that it was fixed again”. This thread definitely helps.

mtrmac:
> Force-push: Rebased, removed `rpm --setcaps` lines as [nobody should ever need to actually run a nested container](https://github.com/containers/podman/pull/14437#pullrequestreview-992645503) inside a skopeo image.

Skopeo does have a case (copies from/to `containers-storage:` when the process doesn’t have enough capabilities) where it would trigger a rootless run, AFAICS needing `newuidmap` and the like.

I have no data to suggest whether it’s used frequently, rarely, or not at all.

But actually a fairly large part of the `Containerfile` deals with making local storage usable (configuring FUSE, and especially setting up /etc/subuid) — so dropping just the `newuidmap` part seems inconsistent. Admittedly that’s not quite evidence that it was intended to be used, maybe it was just copy&pasted…

Still, maybe, now that we know what it was for, it might be easier to leave the workaround in than to worry too much about whether we need it, and to re-visit that every time the three projects’ Containerfiles are compared.

contrib/skopeoimage/stable/Containerfile
mtrmac: @jnovy do you know if https://bugzilla.redhat.com/show_bug.cgi?id=1995337#c3 still affects Fedora, or where to confirm?

mtrmac:
At this point, I’m indifferent between putting that back, and waiting for https://github.com/containers/skopeo/pull/1661#discussion_r909929133 so that we know whether it’s necessary (and possibly to nudge Fedora WRT fixing this in the images). Up to you, I don’t know how merging much this PR makes your work easier/harder.

cevich:
> I don’t know how merging much this PR makes your work easier/harder.

Keeping the Containerfiles consistent with podman & buildah is probably my only argument for putting it back.  Certainly adding a comment with references also seems important.  Also, since you raised there is a case where a rootless container is run (news to me).

mtrmac (approved):

mtrmac:
For future reference, at least https://www.redhat.com/sysadmin/how-run-skopeo-container documents using `containers-storage:` with the Skopeo container, but _not_ relying on the rootless mode feature; instead, the container is run with `--privileged` to avoid that mode.

PR #1669, opened 2022-06-09
Add flag to log to stdout

iNoahNothing:
This PR simply adds a flag that allows users to configure if logrus logs to `stderr` (the default) or `stdout`. 

Logrus currently [cannot log to different locations based on the log levels](https://github.com/sirupsen/logrus/issues/403) and defaults to logging all levels to `stderr`. This causes issues when systems are collecting these logs and reporting INFO logs as errors because they are logged to `stderr`. This provides a simple mechanism for changing that behavior.

Signed-off-by: Noah Krause <krausenoah@gmail.com>

### DIFF
--- a/cmd/skopeo/main.go
+++ b/cmd/skopeo/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"os"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


mtrmac:
Thanks. Why is this extra flag necessary instead of just using shell’s redirection (`2>&1`, `&>`, `|&`)?

TomSweeneyRedHat:
Changes LGTM
But if we go forward with this, we'll also need a man page change, I think here:https://github.com/containers/skopeo/blob/main/docs/skopeo.1.md

PR #1670, opened 2022-06-10
[release-1.3] Change a repo used for sync tests

mtrmac:
The `k8s.gcr.io/coredns/coredns` repo now contains an OCI artifact, which we can't copy; so, use a different repo to test syncing.


TomSweeneyRedHat:
LGTM
@vrothberg ?

PR #1672, opened 2022-06-11
Change a repo used for sync tests

mtrmac:
The `k8s.gcr.io/coredns/coredns` repo now contains an OCI artifact, which we can't copy; so, **temporarily** use a different
repo to test syncing.


mtrmac:
#1673 track reverting this.

rhatdan:
LGTM

Issue #1673, opened 2022-06-11
Implement OCI artifacts and revert #1672


mtrmac:
#1672 changes a repo we use for testing `skopeo sync`, because that repo has added a Cosign signature, and the code now fails (the immediate cause is that it’s trying to compress the image and we don’t have a mapping for the signature blob’s MIME type).

(Note that the signature is not quite an OCI artifact: it uses an ordinary image‘s config MIME type, just an invalid layer MIME type. Still, OCI artifact support is basically a superset of handling this signature.)

We should teach `skopeo sync` to handle this image, and then revert #1672.

Independent parts necessary for this:
- Implement OCI artifact support in c/image. https://github.com/containers/image/pull/1574 , currently mostly works but with FIXMEs about the API and feature set. This should be split into several smaller PRs.
- The registry we test copying _to_, on `localhost`, is a very old version that doesn’t support OCI images at all. https://github.com/containers/automation_images/blob/main/skopeo_cidev/Containerfile#L15 probably needs updating, and then that needs to propagate to Skopeo (and hopefully doesn’t break anything).
- (Eventually we should to think more about `skopeo sync`’s behavior in repos with Cosign signatures. It can copy signatures along with the individual signed images, or it can ignore the signature relationships and copy each OCI tag completely independently. It would be _weird_ and inefficient(although not quite broken) if it ended up doing both.)

mtrmac:
This was done by #1680.

> * (Eventually we should to think more about `skopeo sync`’s behavior in repos with Cosign signatures. It can copy signatures along with the individual signed images, or it can ignore the signature relationships and copy each OCI tag completely independently. It would be _weird_ and inefficient(although not quite broken) if it ended up doing both.)

We will end up in the weird situation after #1701, for users that opt into `use-sigstore-attachments`.

PR #1676, opened 2022-06-13
DO NOT MERGE: Test using registry v2.8.1

mtrmac:
See #1673 . The actual change is going to happen in the automation_images repo, do a smoke-check before doing that.


mtrmac:
The registry was updated in #1679.

PR #1681, opened 2022-06-16
CoPR: Autobuild rpm on rhcontainerbot/podman-next

lsm5:
The new file `skopeo.spec.rpkg` along with a webhook will automatically
build rpms on every PR merge on the main branch.

Run `rpkg local` or `make rpm` to generate the rpm.

Known issue: Doesn't yet build for EL8 environments.

@rhatdan @vrothberg @mtrmac PTAL

I have added the webhook already, so merging this should trigger an rpm build on the copr.

### DIFF
--- a/skopeo.spec.rpkg
+++ b/skopeo.spec.rpkg
@@ -1,0 +1,126 @@
+# For automatic rebuilds in COPR
+
+# The following tag is to get correct syntax highlighting for this file in vim text editor
+# vim: syntax=spec
+
+%global gomodulesmode GO111MODULE=on
+%global with_debug 1
+
+%if 0%{?with_debug}
+%global _find_debuginfo_dwz_opts %{nil}
+%global _dwz_low_mem_die_limit 0
+%else
+%global debug_package %{nil}
+%endif
+
+%if ! 0%{?gobuild:1}
+%define gobuild(o:) go build -buildmode pie -compiler gc -tags="rpm_crashtraceback ${BUILDTAGS:-}" -ldflags "${LDFLAGS:-} -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \\n') -extldflags '-Wl,-z,relro -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld '" -a -v -x %{?**};
+%endif
+
+Name: {{{ git_dir_name }}}
+Epoch: 101
+Version: {{{ git_dir_version }}}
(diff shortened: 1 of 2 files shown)


skopeo.spec.rpkg
```diff
@@ -0,0 +1,126 @@
+# For automatic rebuilds in COPR
```
mtrmac, line 1: A totally non-blocking it: Would it be (_without extra hassle_) possible to move this somewhere to `contrib`? We don’t really want users to build their own RPMs, and that’s where the container image descriptions are (admittedly `contrib` is not a great name for “our own build infrastructure that other users shouldn’t need to touch”).

OTOH I see `podman.spec.rpkg` in the same place, and that consistency is certainly valuable. So entirely up to you.

mtrmac (commented):
Thanks.

Is this related to #1661 ? I’m rather confused about how that is supposed to work — the original issue is that building cross-architecture containers using qemu-user-static takes too long, and the improvement is to first build RPMs and then build containers using those RPMs, in two separate stages and systems?

How is that faster? Is there some significant infrastructure difference, like the RPM builders being native, or the hardware being faster/cheaper?

Or is this completely unrelated to #1681?

lsm5:
> Thanks.
> 
> Is this related to #1661 ? I’m rather confused about how that is supposed to work — the original issue is that building cross-architecture containers using qemu-user-static takes too long, and the improvement is to first build RPMs and then build containers using those RPMs, in two separate stages and systems?
> 
> How is that faster? Is there some significant infrastructure difference, like the RPM builders being native, or the hardware being faster/cheaper?
> 
> Or is this completely unrelated to #1681?

Completely unrelated, unless @cevich wants it to be. This is mainly a way to let users to fetch RPMs built using the latest merged upstream code. With the current setup, this will only build RPMs `post-merge`. There might be ways to trigger it pre-merge for testing purposes but I don't yet have a good way to do it without polluting the CoPR with a ridiculous amount of builds.

skopeo.spec.rpkg
lsm5: Let me check if that's doable. IIRC, this file is expected in the main dir, but let me confirm..

mtrmac:
> This is mainly a way to let users to fetch RPMs built using the latest merged upstream code. 

Thanks. I’ll entirely defer to your expertise, then — feel free to merge as appropriate.

lsm5:
Thanks @mtrmac. Merging ..

PR #1688, opened 2022-06-21
Use bytes.ReplaceAll instead of bytes.Replace(..., -1)

mtrmac:
... for a trivial improvement in readability.

Should not change behavior.


rhatdan:
LGTM

PR #1698, opened 2022-06-30
Error handling cleanups, and drop pkg/errors

mtrmac:
This is motivated by https://github.com/containers/podman/issues/14784 ’s effort to remove pkg/errors.

Also includes other related cleanups, see individual commit messages for details.


mtrmac:
Note to self: There’s probably a more reliable argument as to whether replacing `errors.Cause` is safe (if we know whether the type we are looking for implements `Cause()` or `Unwrap()`).

vrothberg (approved):
LGTM

rhatdan:
LGTM

PR #1706, opened 2022-07-13
Fix sync command description

gbonnefille:
As far as I understand, the `sync` command only works at a repository level. It is not able to sync a whole registry, iterating over the catalog.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ $ skopeo copy oci:busybox_ocilayout:latest dir:existingemptydirectory
 $ skopeo delete docker://localhost:5000/imagename:latest
 ```
 
-## Syncing registries
+## Syncing repositories
 ```console
 $ skopeo sync --src docker --dest dir registry.example.com/busybox /media/usb
 ```
@@ -207,7 +207,7 @@ Please read the [contribution guide](CONTRIBUTING.md) if you want to collaborate
 | [skopeo-manifest-digest(1)](/docs/skopeo-manifest-digest.1.md)    | Compute a manifest digest for a manifest-file and write it to standard output.   |
 | [skopeo-standalone-sign(1)](/docs/skopeo-standalone-sign.1.md)    | Debugging tool - Publish and sign an image in one step.                                                         |
 | [skopeo-standalone-verify(1)](/docs/skopeo-standalone-verify.1.md)| Verify an image signature.                                                    |
-| [skopeo-sync(1)](/docs/skopeo-sync.1.md)           | Synchronize images between container registries and local directories.                       |
+| [skopeo-sync(1)](/docs/skopeo-sync.1.md)           | Synchronize images between container repositories and local directories.                       |
 
 License
 -


README.md
```diff
@@ -207,7 +207,7 @@ Please read the [contribution guide](CONTRIBUTING.md) if you want to collaborate
 | [skopeo-manifest-digest(1)](/docs/skopeo-manifest-digest.1.md)    | Compute a manifest digest for a manifest-file and write it to standard output.   |
 | [skopeo-standalone-sign(1)](/docs/skopeo-standalone-sign.1.md)    | Debugging tool - Publish and sign an image in one step.                                                         |
 | [skopeo-standalone-verify(1)](/docs/skopeo-standalone-verify.1.md)| Verify an image signature.                                                    |
-| [skopeo-sync(1)](/docs/skopeo-sync.1.md)           | Synchronize images between container registries and local directories.                       |
+| [skopeo-sync(1)](/docs/skopeo-sync.1.md)           | Synchronize images between container repositories and local directories.                       |
```
mtrmac, line 210: @TomSweeneyRedHat I’m not sure about this part – the emphasis is on local storage vs. servers, “container repository” stand-alone might not imply a server. I’d appreciate an opinion from a native speaker.

mtrmac, line 210: This text actually comes from `skopeo.1.md`, `skopeo-sync.1.md`; I guess all of that should be updated in sync.

(Alternatively, `skopeo sync --help` just sys `Synchronize one or more images from one location to another`, which is always precise… but also rather less helpful.)

TomSweeneyRedHat: Well, first off, I think the `sync` man [page](https://github.com/containers/skopeo/blob/main/docs/skopeo-sync.1.md) might need some tweaking.  Based on my knowledge (full disclosure, I'm awful with terminology), a "registry" is a collection of "repositories".  Whereas a "repository" is a collection of related images with the same name but different tags.  i.e., `quay.io` is a registry, while `quay.io/containers/podman` is a repository.

The sync man page says, "Synchronize images between container registries and local directories."  I think that's wrong because it won't, at least to my knowledge, synchronize all of quay.io to a local directory if asked right?  Whatever we decide upon here probably ought to be changed on the sync page too.

I think I'd chicken out and make this particular line generic:

```suggestion
| [skopeo-sync(1)](/docs/skopeo-sync.1.md)           | Synchronize images between container repositories and a specified destination.
```

mtrmac: > The sync man page says, "Synchronize images between container registries and local directories." I think that's wrong because it won't, at least to my knowledge, synchronize all of quay.io to a local directory if asked right? 

That’s a fair point.

---

> I think I'd chicken out and make this particular line generic:

I think that’s making the uncontroversial part generic.

---

Either way, this PR should include man page updates, and needs a correct sign-off.

rhatdan:
@gbonnefille are you still working on this?

gbonnefille:
> @gbonnefille are you still working on this?

Not a lot of time for this currently.
Feel free to fix the documentation in an other PR and close this one.

mtrmac:
#1734 is another take on this problem, trying “registry repositories” (and updating all three files).

mtrmac:
#1734 was merged. Thanks!

Issue #1584, opened 2022-03-04
(make docs) and (make binary) commands fail on Ubuntu


jameskim0987:
I'm trying to do a clean install skopeo, Building from Source (**in** a container), but at `make docs` and `make binary` step seems to fail.

- Environment: `Ubuntu 18.04`
- Installed Go Version: `go version go1.17.7 linux/amd64`
- root@ubuntu1804:~# make --version
GNU Make 4.1
Built for x86_64-pc-linux-gnu
Copyright (C) 1988-2014 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
- root@ubuntu1804:~# podman --version
podman version 3.0.1

Some issues:
- root@ubuntu1804:~# go-md2man version
    **Output hangs**
- root@ubuntu1804:~# podman info
**Error: failed to mount overlay for metacopy check with "nodev,metacopy=off" options: invalid argument**

```
root@ubuntu1804:~/workspace/skopeo# make docs
make: command: Command not found
sed -e 's/\((skopeo.*\.md)\)//' -e 's/\[\(skopeo.*\)\]/\1/' docs/skopeo.1.md  | /bin/go-md2man -in /dev/stdin -out docs/skopeo.1
make: command: Command not found
/bin/sh: 1: /bin/go-md2man: not found
Makefile:147: recipe for target 'docs/skopeo.1' failed
make: *** [docs/skopeo.1] Error 127
```
and

```
root@ubuntu1804:~/workspace/skopeo# make binary
make: command: Command not found
Error: failed to mount overlay for metacopy check with "nodev,metacopy=off" options: invalid argument
docker run --rm -i -e TESTFLAGS="-v -check.v -test.timeout=15m" -e CI= -e SKOPEO_CONTAINER_TESTS=1 -t --security-opt label=disable -v /root/workspace/skopeo:/src/github.com/containers/skopeo -w /src/github.com/containers/skopeo  make bin/skopeo  BUILDTAGS='  libdm_no_deferred_remove '
make: command: Command not found
Unable to find image 'make:latest' locally
docker: Error response from daemon: pull access denied for make, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
See 'docker run --help'.
Makefile:135: recipe for target 'binary' failed
make: *** [binary] Error 125
```

[...]

PR #1730, opened 2022-08-12
Added fallback for systems that only have command as a built-in

grafail:
This solves #1584

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,8 +30,8 @@ endif
 
 # Multiple scripts are sensitive to this value, make sure it's exported/available
 # N/B: Need to use 'command -v' here for compatibility with MacOS.
-export CONTAINER_RUNTIME ?= $(if $(shell command -v podman),podman,docker)
-GOMD2MAN ?= $(if $(shell command -v go-md2man),go-md2man,$(GOBIN)/go-md2man)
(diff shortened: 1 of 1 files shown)


mtrmac (commented):
Thanks for the PR.

So I ended up actually examining how this works / doesn’t work.
- On Ubuntu FIXME, the current version just running `command` works fine; so no change is necessary
- On Ubuntu 20.04, following https://github.com/containers/skopeo/issues/1584#issuecomment-1075358336 , this patch does _help_, but _not_ by running `which`; it causes `make` to run `/bin/sh -c …` instead of avoiding the shell and searching for `command` in `PATH`.

So, adding `which` here would mostly obscure what is actually going on.

Searching further, the original Ubuntu 20.04 problem seems to be a variant of https://savannah.gnu.org/bugs/index.php?57625 (actually fixed separately in https://git.savannah.gnu.org/cgit/make.git/commit/job.c?h=4.3&id=1af314465e5dfe3e8baa839a32a72e83c04f26ef ), and the recommended workaround is to add a `;` at the end of the command, which has the same effect of forcing a shell to be used. That seems better-targeted to me.

So, would #1733 work? It seems so, for me, but I have only tested that on the Ubuntu 20.04 version.

mtrmac:
#1733 was merged, please reopen if it is insufficient.

PR #1731, opened 2022-08-15
Stop using docker/docker/pkg/homedir in tests

mtrmac:
`c/storage/pkg/homedir`, which we need anyway for other purposes, should work just as well.


rhatdan:
LGTM

PR #1732, opened 2022-08-15
Pin Go to 1.18

mtrmac:
1.19 has changed the expected gofmt format, and we don't want to follow such changes on the stable branch.

`go@1.18` is "keg-only", i.e. not installed by Brew to `/usr/local/bin`, so we need to change `PATH` to point at it (as the installation instructs us to).

---

This is based on commit 1b6d7f79a521bdcf5947f14363211d3d3ca92c99 from the 1.4 branch.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -87,12 +87,13 @@ osx_task:
     macos_instance:
         image: catalina-xcode
     setup_script: |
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

PR #1733, opened 2022-08-16
Workaround GNU Make 4.2.1 incorrectly executing shell commands

mtrmac:
Fix looking for commands with GNU make 4.2.1
    
Before https://git.savannah.gnu.org/cgit/make.git/commit/job.c?h=4.3&id=1af314465e5dfe3e8baa839a32a72e83c04f26ef , `make` was incorrectly trying to avoid running a shell for `command -v`.  Use the workaround recommended in https://savannah.gnu.org/bugs/index.php?57625 .

Fixes one half of (the reported problems in) #1584.

Compare #1730 .

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,9 +29,14 @@ GOBIN := $(GOPATH)/bin
 endif
 
 # Scripts may also use CONTAINER_RUNTIME, so we need to export it.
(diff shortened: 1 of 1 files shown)


vrothberg (approved):
LGTM

lsm5 (approved):
/lgtm

PR #1744, opened 2022-08-31
[DO NOT MERGE] Check if Cirrus image bump is to blame

lsm5:

verify if podman-gvproxy removal from downloaded packages in the cirrus image bump is to blame.

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,3 +213,6 @@ License
 -
 skopeo is licensed under the Apache License, Version 2.0. See
 [LICENSE](LICENSE) for the full license text.
+
+
+###### TEST CIRRUS


lsm5:
@cevich so it definitely is an image issue and not an issue with the test itself.

lsm5:
both skopeo and c/image tests fail with the image bump.

cevich:
Ya, I think we knew this mostly.  The trick is finding what in the image changed to cause problems.  And sometimes, the problems are due to BOTH the image contents being updated and the test(s) expecting different behavior.  In other words, it's somewhat rare but it could be both!

PR #1750, opened 2022-09-12
default.yaml should have all options commented

rhatdan:
Rely on the hard coded defaults in libraries rather then overriding in the yaml file.

### DIFF
--- a/default.yaml
+++ b/default.yaml
@@ -13,7 +13,7 @@
 # This is the default signature write location for docker registries.
 default-docker:
 #  lookaside: file:///var/lib/containers/sigstore
-  lookaside-staging: file:///var/lib/containers/sigstore
+#  lookaside-staging: file:///var/lib/containers/sigstore
 
 # The 'docker' indicator here is the start of the configuration
 # for docker registries.


rhatdan:
@vrothberg @mtrmac PTAL

vrothberg (commented):
LGTM
Is there an issue behind the fix?

rhatdan:
It was noticed by @mtrmac on my system when the directory became unreadable.  Rootless mode and rootful mode work differently and we should just use the defaults.  This file might be better off in the containers/image repo versus in skopeo BTW.  Bottom line we should keep it commented out.

mtrmac:
Oh, _now_ I remember.

The trouble with this is that it can break existing setups, because signatures no longer land at the previously-expected place.

I guess that’s fine for upstream, but this shouldn’t happen in an existing RHEL branch. Cc: @jnovy .

---

> This file might be better off in the containers/image repo versus in skopeo BTW.

If we were to move this, I think a repo corresponding to the containers-common RPM would conceptually be best.

But, pragmatically, a missing file and a file with everything commented out have almost the same effect (apart from failing on conflicts), so … c/image is fine as is?

PR #1753, opened 2022-09-19
Fix documentation in the default registries.d content.

mtrmac:
… following up on #1750.

### DIFF
--- a/default.yaml
+++ b/default.yaml
@@ -1,8 +1,8 @@
 # This is a default registries.d configuration file.  You may
 # add to this file or create additional files in registries.d/.
 #
-# lookaside: indicates a location that is read and write
-# lookaside-staging: indicates a location that is only for write
+# lookaside: for reading/writing simple signing signatures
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@vrothberg PTAL

default.yaml
```diff
@@ -1,26 +1,28 @@
 # This is a default registries.d configuration file.  You may
 # add to this file or create additional files in registries.d/.
 #
-# lookaside: indicates a location that is read and write
-# lookaside-staging: indicates a location that is only for write
+# lookaside: a location simple signing signatures are read from
```
vrothberg, line 4:
```suggestion
# lookaside: a location where simple signing signatures are read from
```
? Same below

mtrmac: That change was actually wrong, we do write to lookaside: . So I rewrote it completely differently again.

vrothberg (approved):
LGTM

PR #1757, opened 2022-09-24
proxy: Add `OpenImageOptional`

cgwalters:
In some code I'm writing I want to be able to cleanly test if an image exists, as distinguished from other errors like authentication problems, network flakes etc.

As best I can tell, the containers/image abstraction doesn't offer a clean way to do this.

For now, I chose the route of adding the ugly string error matching here for the two cases I care about (docker v2s2 registry and oci directories), so my Rust code can operate in terms of clean `Option<Image>`.

### DIFF
--- a/cmd/skopeo/proxy.go
+++ b/cmd/skopeo/proxy.go
@@ -73,9 +73,12 @@ import (
 
 	"github.com/containers/image/v5/image"
 	"github.com/containers/image/v5/manifest"
+	ocilayout "github.com/containers/image/v5/oci/layout"
 	"github.com/containers/image/v5/pkg/blobinfocache"
 	"github.com/containers/image/v5/transports/alltransports"
 	"github.com/containers/image/v5/types"
+	dockerdistributionerrcode "github.com/docker/distribution/registry/api/errcode"
+	dockerdistributionapi "github.com/docker/distribution/registry/api/v2"
 	"github.com/opencontainers/go-digest"
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/spf13/cobra"
@@ -99,6 +102,9 @@ const maxMsgSize = 32 * 1024
 // We hard error if the input JSON numbers we expect to be
 // integers are above this.
 const maxJSONFloat = float64(uint64(1)<<53 - 1)
+
+// sentinelImageID represents "image not found" on the wire
+const sentinelImageID = 0
 
 // request is the JSON serialization of a function call
 type request struct {
@@ -197,6 +203,29 @@ func (h *proxyHandler) Initialize(args []interface{}) (replyBuf, error) {
 // OpenImage accepts a string image reference i.e. TRANSPORT:REF - like `skopeo copy`.
 // The return value is an opaque integer handle.
 func (h *proxyHandler) OpenImage(args []interface{}) (replyBuf, error) {
+	return h.openImageImpl(args, false)
+}
+
+// isDockerManifestUnknownError is a copy of code from containers/image,
+// please update there first.
+func isDockerManifestUnknownError(err error) bool {
+	var ec dockerdistributionerrcode.ErrorCoder
+	if !errors.As(err, &ec) {
+		return false
+	}
+	return ec.ErrorCode() == dockerdistributionapi.ErrorCodeManifestUnknown
+}
+
+// isNotFoundImageError heuristically attempts to determine whether an error
+// is saying the remote source couldn't find the image (as opposed to an
+// authentication error, an I/O error etc.)
+// TODO drive this into containers/image properly
+func isNotFoundImageError(err error) bool {
+	return isDockerManifestUnknownError(err) ||
+		errors.Is(err, ocilayout.ImageNotFoundError{})
+}
+
+func (h *proxyHandler) openImageImpl(args []interface{}, allowNotFound bool) (replyBuf, error) {
 	h.lock.Lock()
 	defer h.lock.Unlock()
 	var ret replyBuf
@@ -218,9 +247,15 @@ func (h *proxyHandler) OpenImage(args []interface{}) (replyBuf, error) {
 	}
 	imgsrc, err := imgRef.NewImageSource(context.Background(), h.sysctx)
 	if err != nil {
+		if allowNotFound && isNotFoundImageError(err) {
+			ret.value = sentinelImageID
+			return ret, nil
+		}
 		return ret, err
 	}
 
+	// Note that we never return zero as an imageid; this code doesn't yet
+	// handle overflow though.
 	h.imageSerial++
 	openimg := &openImage{
 		id:  h.imageSerial,
@@ -230,6 +265,13 @@ func (h *proxyHandler) OpenImage(args []interface{}) (replyBuf, error) {
 	ret.value = openimg.id
 
 	return ret, nil
+}
+
+// OpenImage accepts a string image reference i.e. TRANSPORT:REF - like `skopeo copy`.
+// The return value is an opaque integer handle.  If the image does not exist, zero
+// is returned.
+func (h *proxyHandler) OpenImageOptional(args []interface{}) (replyBuf, error) {
+	return h.openImageImpl(args, true)
 }
 
 func (h *proxyHandler) CloseImage(args []interface{}) (replyBuf, error) {
@@ -277,6 +319,9 @@ func (h *proxyHandler) parseImageFromID(v interface{}) (*openImage, error) {
 	imgid, err := parseImageID(v)
 	if err != nil {
 		return nil, err
+	}
+	if imgid == sentinelImageID {
+		return nil, fmt.Errorf("Invalid imageid value of zero")
 	}
 	imgref, ok := h.images[imgid]
 	if !ok {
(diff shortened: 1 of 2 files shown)


cmd/skopeo/proxy.go
```diff
@@ -232,6 +233,56 @@ func (h *proxyHandler) OpenImage(args []interface{}) (replyBuf, error) {
 	return ret, nil
 }
 
+func isNotFoundImageError(err error) bool {
+	msg := err.Error()
+	// TODO: Drive a correct API for this into containers/image
+	return strings.HasSuffix(msg, "manifest unknown") ||
+		strings.Contains(msg, ": no descriptor found for reference")
```
mtrmac, line 240: Not a full review, just a quick note for now:

You’re right, there isn‘t a good abstraction for this (primarily because the registries behave quite differently and we just didn’t do the work to make a survey).

Currently, I’d recommend searching for `docker/distribution` error types, similarly to https://github.com/containers/image/blob/62be088831a0836823930f0c226152b164ff4241/docker/docker_client.go#L984 . Precedent for making that assumption outside of c/image: https://github.com/containers/common/blob/fd9763624b9c81d128589058e05460323988bf53/pkg/retry/retry.go#L52 .

But, also, note the changes to `isManifestUnknownError` proposed in https://github.com/containers/image/pull/1299 . So the integration test is quite necessary to keep such code (whether it relies on string matching or undocumented d/distribution type usage) working.

cgwalters: > Currently, I’d recommend searching for docker/distribution error types, 

That'd pull a direct dependency on docker/distribution into skopeo; but I dunno, perhaps that doesn't matter.

Maybe a local minima of effort/hackiness is `s/isManifestUnknownError/BestEffortIsManifestUnknownError/` in c/image (i.e. make it public but with documented caveats)?

In the c/image OCI case, we clearly can add a reliable API for this.  I can perhaps look at that, just trying to solve a higher level problem at the moment.

cmd/skopeo/proxy.go
```diff
@@ -232,6 +233,56 @@ func (h *proxyHandler) OpenImage(args []interface{}) (replyBuf, error) {
...
+		}
+		return ret, err
+	}
+
+	h.imageSerial++
+	openimg := &openImage{
+		id:  h.imageSerial,
+		src: imgsrc,
+	}
+	h.images[openimg.id] = openimg
```
mtrmac, line 280: This does not avoid the `ret.value = 0` value. Maybe factor ~this part out of the two `OpenImage` methods, and have it always skip `0`.

cmd/skopeo/proxy.go
mtrmac: A direct dependency (on the _error definitions_, not all of d/distribution) is fine; we already have the indirect one.

---

WRT making `isManifestUnknownError` public:

(Note that function is more or less _known_ not to be sufficient. In the worst case, IIRC some registries (in some, possibly very old, versions?) return a HTTP 404 without the API-expected response at all. c/image/docker only uses it behind an opt-in, so that users with unhandled registries can still get the other functionality. That generally applies to the idea of this PR — callers can’t assume it will _reliably_ report “no error, image not present”.[1])

Defining the c/image API, and for not documenting it as a best-effort thing, is a good idea — not something I’d block this PR on, but more valuable than either of the hacks discussed so far.

Just one thing: I’d prefer the API to be closer to the final shape, `errors.As(err, c/image/docker.SomeNewErrorType)` rather than a function.


[1] Note that “no error, image not present” is somewhat ill-defined. At least in the past, the Docker Hub would report “permission denied” on references to any missing repo, on the theory that users who don’t have the permission to view a repo don’t have the permission to know whether it exists.

---

If you are talking about the OCI format on registries, I don’t think that pulling v2s2 vs. pulling OCI makes any difference. If you are talking about the `oci` and `oci-archive` formats, yes, we clearly can add a reliable API in that case.

rhatdan:
@cgwalters can  you update your PR, we are getting ready for a release.

cmd/skopeo/proxy.go
cgwalters: I'm not sure I understand; note that we increment `h.imageSerial` before the success return path.  Oh, are you worried about overflow?  Getting up to 4 billion images seems...unlikely?

cgwalters: Ugh ok actually this overflow thing gave me flashbacks of https://gitlab.gnome.org/GNOME/glib/-/commit/b98a1c8df30f9e24588a48331dacf01e49760549 so...hmm, I think we could bump it to a `uint64` actually...

mtrmac: It’s a fair point that this is unlikely — actually if we did overflow, there is nothing in the existing code that avoids re-allocating a still used ID, so that would be broken anyway.

Really I was focusing on the _documented semantics_ of the method. If `0` means “image does not exist”, then the image ID space should clearly exclude `0`. Alternatively, instead of a reserved image ID value, add an extra `failedBecauseImageDidNotExist` field, and then the whole image ID space can be used without restrictions.

(It seemed to me that consolidating the “create `openImage` + return an ID” code was also a bit easier to maintain than the two copies, and adding three lines to explicitly skip `0` (or all non-positive values?) would have been cheap enough. But if we don’t avoid duplicates — and I agree that’s really unnecessary work at this point — then there’s not that much benefit.)

I guess I’d be fine with just a comment to the effect of “Note that if imageID ever overflows, we start allocating `0`, which has special meaning, and already-allocated values, and this completely falls apart.

---

int64 is a bit risky in JSON, where numbers are IIRC float with 53-bit precision unless users explicitly do some non-default parsing. Also note that the existing tests explicitly coerce to `uint32`.

mtrmac:
I’d still like to insist on `isNotFoundImageError` using Go types instead of string matches; the past experience with such “short-term things to be cleaned up” in `inspect.go` has been painful.

cgwalters:
> I’d still like to insist on isNotFoundImageError using Go types instead of string matches; the past experience with such “short-term things to be cleaned up” in inspect.go has been painful.

OK yep fair; to be clear please do go ahead and do a skopeo release without this PR, this PR is not in any kind of critical path and I may not have time to circle back for a bit.

cgwalters:
OK, cleaned this up a bit more, I think comments should be addressed.  But I do feel guilty for not trying to dive in to actually adding proper c/image APIs for this...

cgwalters:
BTW while I'm here...it's not related strictly to this PR and I think you know this but:

I want to highlight that we are using this proxy code to update the operating system by default in OCP 4.12 https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/CZ7UNVVXXOFIMOPD6YCCU6D7DSX2NBDJ/

cmd/skopeo/proxy.go
```diff
@@ -197,6 +203,34 @@ func (h *proxyHandler) Initialize(args []interface{}) (replyBuf, error) {
 // OpenImage accepts a string image reference i.e. TRANSPORT:REF - like `skopeo copy`.
 // The return value is an opaque integer handle.
 func (h *proxyHandler) OpenImage(args []interface{}) (replyBuf, error) {
+	return h.openImageImpl(args, false)
+}
+
+// isManifestUnknownError is a copy of code from containers/image,
+// please update there first.
+func isManifestUnknownError(err error) bool {
```
mtrmac, line 211:
```suggestion
func isDockerManifestUnknownError(err error) bool {
```
or update the comment to point out that this is intended to handle errors from the c/image/docker transport.

---

(Implementation of this LGTM.)

cmd/skopeo/proxy.go
```diff
@@ -197,6 +203,34 @@ func (h *proxyHandler) Initialize(args []interface{}) (replyBuf, error) {
...
+	return ec.ErrorCode() == dockerdistributionapi.ErrorCodeManifestUnknown
+}
+
+// isNotFoundImageError heuristically attempts to determine whether an error
+// is saying the remote source couldn't find the image (as opposed to an
+// authentication error, an I/O error etc.)
+// TODO drive this into containers/image properly
+func isNotFoundImageError(err error) bool {
+	return isManifestUnknownError(err) ||
+		strings.Contains(err.Error(), ": no descriptor found for reference")
```
mtrmac, line 230: This really should use a Go type created in c/image. Something like https://github.com/containers/image/pull/1675 ? (Warning: Absolutely untested.)

mtrmac: https://github.com/containers/image/pull/1675 is now merged in c/image, and vendored into Skopeo.

cgwalters: OK, updated to use that, thanks!

mtrmac: Thanks.

Tests are now failing, that’s because https://github.com/containers/image/pull/1299 (as mentioned earlier) changes what `isDockerManifestUnknownError` needs to be looking for; I’m afraid I forgot about that part. Just update the code to match the `c/image/docker/docker_client.go` variant, please.

mtrmac (approved):
Thanks!

cgwalters:
Binding in https://github.com/containers/containers-image-proxy-rs/pull/36

PR #1767, opened 2022-10-02
Adds `--append-suffix` flag to sync command

Phuurl:
Adds the option to append the sync timestamp to the destination image tag. 

This is useful for syncing source containers which have non-immutable tags to a destination which does have immutable tags, and allows a point-in-time view of the moving source tag.

For example - source `ubuntu:bionic` would be synced to `<host>/ubuntu:bionic-1664722433` with the `--append-timestamp` flag specified.

### DIFF
--- a/docs/skopeo-sync.1.md
+++ b/docs/skopeo-sync.1.md
@@ -66,7 +66,7 @@ Print usage statement.
 
 **--scoped** Prefix images with the source image path, so that multiple images with the same name can be stored at _destination_.
 
-**--append-suffix** _tag-suffix_ String to append to destination tags
+**--append-suffix** _tag-suffix_ String to append to destination tags.
 
 **--preserve-digests** Preserve the digests during copying. Fail if the digest cannot be preserved. Consider using `--all` at the same time.
 
@@ -186,6 +186,16 @@ Destination registry content:
 ```
 REPO                              TAGS
 registry.local.lan/repo/busybox   1-glibc, 1-musl, 1-ubuntu, ..., latest
+```
+
+### Synchronizing to a container registry with tag suffix
+```
+skopeo sync --src docker --dest docker --append-suffix '-mirror' registry.example.com/busybox my-registry.local.lan
+```
+Destination registry content:
+```
+REPO                         TAGS
+registry.local.lan/busybox   1-glibc-mirror, 1-musl-mirror, 1-ubuntu-mirror, ..., latest-mirror
 ```
 
 ### YAML file content (used _source_ for `**--src yaml**`)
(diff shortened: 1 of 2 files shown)


mtrmac:
Thanks!

The use case does make sense; I’m a bit worried about adding a specialized option for this. (In general, `skopeo sync` does not have a good comprehensive design of the name mapping between sources and destinations.)

E.g. other users would very likely want a human-readable date. And it seem useful to at tag all images from a single sync run with the same date, so that they can be correlated. So, at least, wouldn’t something like `--append-suffix=`_string_ work?

But even that seems like a special case where we might be able to provide a more general feature… of _some_ unknown design. Go template strings? Regexp pattern + replacement? OTOH we do have to ultimately constrain it to some language simpler than Go :)

art-shutter:
I believe appending metadata or anything else via cli would be a good idea, but doing it at least on a per-repo basis (as well as with a cli flag globally) would bring more configurability to the yaml.

TomSweeneyRedHat:
No movement on this one in over a month and it needs a rebase.  @Phuurl thoughts on going forward?

Phuurl:
@TomSweeneyRedHat Sorry completely forgot about this 🤦‍♂️ 
Should get chance this weekend to sort it

Phuurl:
@mtrmac @TomSweeneyRedHat PR updated 👍 

Totally agree that `--append-suffix` is a better approach as way more versatile - if users need to append the timestamp, they can just do that with a shell call to `date` passed in as the suffix

docs/skopeo-sync.1.md
```diff
@@ -66,6 +66,8 @@ Print usage statement.
 
 **--scoped** Prefix images with the source image path, so that multiple images with the same name can be stored at _destination_.
 
+**--append-suffix** _tag-suffix_ String to append to destination tags
```
TomSweeneyRedHat, line 69:
```suggestion
**--append-suffix** _tag-suffix_ String to append to destination tags.
```
It would be nice to have a small example with a sample command and then an expected result in the examples section.

Phuurl: Sure - example and missing `.` added

docs/skopeo-sync.1.md
```diff
@@ -186,6 +188,16 @@ REPO                              TAGS
...
+### Synchronizing to a container registry with tag suffix
+```
+skopeo sync --src docker --dest docker --append-suffix '-mirror' registry.example.com/busybox my-registry.local.lan
+```
+Destination registry content:
+```
+REPO                         TAGS
+registry.local.lan/busybox   1-glibc-mirror, 1-musl-mirror, 1-ubuntu-mirror, ..., latest-mirror
+```
+
```
TomSweeneyRedHat, line 200: nice!  TYVM

TomSweeneyRedHat (commented):
LGTM

TomSweeneyRedHat:
I've re-run the Skopeo test, fingers crossed.

rhatdan:
@mtrmac PTAL

mtrmac (approved):
LGTM. Thanks!

PR #1769, opened 2022-10-03
feat: add a per-repo destination-path-style setting for yaml source

art-shutter:
Introduces a new per-repo yaml setting to simplify work with YAML source as referenced in https://github.com/containers/skopeo/issues/1072 and in https://github.com/containers/skopeo/pull/1531

The change is backward-compatible and doesn't attempt to completely overhaul the format of the yaml manifest as suggested in https://github.com/containers/skopeo/pull/1531, but still adds a usable and requested functionality.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,4 +7,3 @@ updates:
     time: "10:00"
     timezone: Europe/Berlin
   open-pull-requests-limit: 10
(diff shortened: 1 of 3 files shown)


TomSweeneyRedHat:
@art-shutter any thoughts on moving this forward?

PR #1776, opened 2022-10-13
proxy: Bump semver for OpenImageOptional

cgwalters:
I should have done this in the previous commit, it's how clients can discover that we have the API.


cgwalters:
Good point, done!

mtrmac (approved):
Thanks! LGTM.

PR #1777, opened 2022-10-13
proxy: Change the imgid to uint64

cgwalters:
In PR review for a different issue, the question of what happens if we hit overflow for the imageid serial was hit.  This feels pretty unlikely; if I did the math right, it'd require opening an average of 136 images per second to overflow it in a year. Nevertheless, in practice what we're sending on the wire is just a JSON number, and if we extend this to the "max safe JSON number" of 2^53, it'd take 285,616,414 images per second to overflow in a year, going from implausible to probably impossible.

With a bit more work of course, we could make this a sparse mapping and reuse freed numbers, but eh.

### DIFF
--- a/cmd/skopeo/proxy.go
+++ b/cmd/skopeo/proxy.go
@@ -156,7 +156,7 @@ type activePipe struct {
 // openImage is an opened image reference
 type openImage struct {
 	// id is an opaque integer handle
-	id        uint32
+	id        uint64
 	src       types.ImageSource
 	cachedimg types.Image
 }
@@ -171,9 +171,9 @@ type proxyHandler struct {
 	cache  types.BlobInfoCache
 
 	// imageSerial is a counter for open images
-	imageSerial uint32
+	imageSerial uint64
 	// images holds our opened images
-	images map[uint32]*openImage
+	images map[uint64]*openImage
 	// activePipes maps from "pipeid" to a pipe + goroutine pair
 	activePipes map[uint32]*activePipe
 }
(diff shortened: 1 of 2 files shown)


mtrmac (commented):
Does this break any existing users? And if it does, do we need to worry about version negotiation and/or caller opt-in by calling newer versions of the methods? Ugh.

I think you are maintaining the known user / all the known users, so that’s really up to you.

---

Apart from the above, implementation LGTM.

cgwalters:
I don't think so, though I didn't actually yet cross-check with my code.  Marking draft until I do so.  (and I may look at adding some proper reverse dependency testing)

My assumption here is that these uint32/uint64 things are only internal to the proxy - what gets sent on the wire is JSON, and so we're just expanding the space of internal values.

Now, the Rust client code here also hardcodes uint32, so it would fail if one of the values actually *did* get larger, but that can be fixed later.

mtrmac:
That’s a fair point, this should be completely invisible on the wire until we actually overflow 32 bits — and once we do, both the old and new version would presumably fail. The only worry would be a consumer that connects to the new version, and casts the float to 32 bits, _silently_ discarding upper bits, and then referring to the wrong handle.

cgwalters:
Added test code here https://github.com/containers/skopeo/pull/1781 - let's merge that, then use this PR as a test for the tests to validate that this change doesn't break the API.

TomSweeneyRedHat:
Just noting that #1781 was closed in favor of https://github.com/cevich/skopeo/commit/f4b4eb043a4820de3e258ea9ed6dddc118e848b4, which looks to have merged.  Can this be rebased and re-tested at this point or is it "hurry up and wait" for something else now?

cgwalters:
I've rebased :surfer: this now that we have proper revdep testing.  But let's merge https://github.com/containers/skopeo/pull/2029 first.

mtrmac (approved):
Thanks!

Issue #1756, opened 2022-09-22
sync all images from a starting version


SamStenton:
It would be beneficial to be able to define a minimum version to sync against so that all versions above that initial definition are copied.

We're running Skopeo on a schedule to move images from one repo to another, I don't want to copy all images ever created but instead copy all images created that are greater than a defined version. 

This might look like this when using the yaml config: 

```yaml
docker.io:
    images:
        busybox:
            - "~1.0.0" # Sync all patch tags
            - "^1.0.0" # Sync all minor and patch tags
            - ">1.1.5" # Sync all tags above and including 1.1.5
```

I know this can be achieved using the regex method however the readability of that begins to get pretty difficult and importantly it's really easy to miss edge cases. 

```yaml
docker.io:
    images-by-tag-regex:
        jetstack/cert-manager-webhook: (v1\.[8-9]\.[0-9]+$)|v([2-9]+)\.([0-9]+)\.[0-9]+$ # match anything above 1.8.0 (various edge cases that get missed, like v1.10.0
        jetstack/cert-manager-webhook: (v1\.[8-9]\.[0-9]+$)|v([2-9]+)\.([0-9]+)\.[0-9]+$
        istio/proxyv2: ([0-9])\.([1-9][3-9]|[2-9][0-9])\.[0-9]*$ # match anything starting from 1.13.0.
```

mtrmac:
Thanks for reaching out.

This seems plausible in principle; the hard thing is finding a versioning standard that is _widely_ applicable (with things like patches, pre-releases and the like), to minimize the number of almost-but-not-quite-the-same options.

At some point there was some work on #965 using the “semantic versioning” syntax.

SamStenton:
#996 would suggest that made it in - but can't find this compare feature. Syntax in 965 would be great.

mtrmac:
I think there was some mix-up in #965 vs #996 , I’m not sure what exactly happened back then. Either way, the feature doesn’t exist right now.

k11h-de:
@SamStenton Have you found any working and user-friendly solution for now (apart from regex)?
Maybe we can use the initial work done in #965 and enhance / finish it. 
As my focus is in devops, my coding skills are not very advanced.

I would love to see this feature and happy to contribute whatever I can.
Cheers from Berlin
Karsten

mtrmac:
#2189 has implemented a semver filter.

PR #1779, opened 2022-10-17
feature: Add --top-refs flag

helobinvn:
Allow to only synchronize top n refs for image
Use cases:
- Only sync the numbers of the latest refs at the given time when the synchronization happens
- If you have a bunch of images managed by your custom tooling and want to migrate to `skopeo sync`, this will help you to simplify the complicated of determining regex* for each image to not reach the upstream registries' pull limit
- If you have some garbage collectors in your registry, this will also help reduce the cost to exclude the tags which are supposed to be cleaned up with the skopeo images' configuration

P/S: I'm new to Go, so feel free to discard this PR if it totally implemented in the wrong way, but we would like to have this feature.

### DIFF
--- a/cmd/skopeo/sync.go
+++ b/cmd/skopeo/sync.go
@@ -4,12 +4,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/hashicorp/go-version"
 	"io"
 	"io/fs"
 	"os"
 	"path"
 	"path/filepath"
 	"regexp"
+	"sort"
 	"strings"
 
 	commonFlag "github.com/containers/common/pkg/flag"
@@ -50,6 +52,7 @@ type syncOptions struct {
 	preserveDigests          bool                      // Preserve digests during sync
 	keepGoing                bool                      // Whether or not to abort the sync if there are any errors during syncing the images
 	appendSuffix             string                    // Suffix to append to destination image tag
+	topRefs                  int                       // Only get amount of top largest refs after sorted by Semantic versioning
 }
 
 // repoDescriptor contains information of a single repository used as a sync source.
@@ -118,6 +121,7 @@ See skopeo-sync(1) for details.
 	flags.StringVarP(&opts.destination, "dest", "d", "", "DESTINATION transport type")
 	flags.BoolVar(&opts.scoped, "scoped", false, "Images at DESTINATION are prefix using the full source image path as scope")
 	flags.StringVar(&opts.appendSuffix, "append-suffix", "", "String to append to DESTINATION tags")
+	flags.IntVar(&opts.topRefs, "top-refs", 0, "Only get amount of top largest refs after sorted by Semantic versioning")
 	flags.BoolVarP(&opts.all, "all", "a", false, "Copy all images if SOURCE-IMAGE is a list")
 	flags.BoolVar(&opts.dryRun, "dry-run", false, "Run without actually copying data")
 	flags.BoolVar(&opts.preserveDigests, "preserve-digests", false, "Preserve digests of images and lists")
@@ -647,6 +651,31 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) (retErr error) {
 
 	for _, srcRepo := range srcRepoList {
 		options.SourceCtx = srcRepo.Context
+		// only proceed for top refs if defined
+		if opts.topRefs > 0 {
+			// Check Refs follow Semantic Versioning
+			semverRefs, nonSemverRefs := filterOutRefsNonSemver(srcRepo.ImageRefs)
+			if len(nonSemverRefs) > 0 {
+				logrus.Warnf("Filtered out %d ref(s) are not follow Semver in total %d ref(s)", len(nonSemverRefs), len(srcRepo.ImageRefs))
+			}
+			// Sort semverRefs
+			sort.Slice(semverRefs, func(i, j int) bool {
+				v1, _ := version.NewVersion(semverRefs[i].DockerReference().(reference.Tagged).Tag())
+				v2, _ := version.NewVersion(semverRefs[j].DockerReference().(reference.Tagged).Tag())
+				return v1.LessThan(v2)
+			})
+			topRefs := opts.topRefs
+			totalRefs := len(srcRepo.ImageRefs)
+			totalSemverRefs := len(semverRefs)
+			if totalSemverRefs == 0 {
+				return errors.New("ERROR: Not found any refs follow Semantic versioning. Check your tag's regex configuration OR doesn't it follow Semantic versioning?")
+			}
+			if topRefs > totalSemverRefs {
+				topRefs = totalSemverRefs
+			}
+			logrus.Infof("Only sync top largest %d ref(s) follow Semver in %d ref(s) from the total %d ref(s)", topRefs, totalSemverRefs, totalRefs)
+			srcRepo.ImageRefs = semverRefs[totalSemverRefs-topRefs:]
+		}
 		for counter, ref := range srcRepo.ImageRefs {
 			var destSuffix string
 			switch ref.Transport() {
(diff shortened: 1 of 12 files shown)


mtrmac (changes requested):
Thanks.

AFAICS both in `imagesToCopyFromRepo` and `imagesToCopyFromDir`, the list of returned images has no specified order at all, so I can’t see how this could possibly work.

I also suspect that “latest image” is not quite as obvious to define. Last built? Last uploaded? Largest version number, using _some specific_ version semantics format?

---

Just to link related things together: #1756 .

helobinvn:
Thanks, @mtrmac for checking. I'm interested in only `sync` with src/dest's transport is`docker` only, also quickly tried to check the ordering of ref returned, and it looks already sorted by uploaded date, so maybe it is just a coincidence. In API spec https://docs.docker.com/registry/spec/api/#tags also is not mentioning anything about the ordering requirement.

r/d `latest image` term, I meant for the last uploaded, but I think it will be the same as the largest version if the image follows some semantic versioning.

IMO if we are able to implement https://github.com/containers/skopeo/issues/1756, then it can be easy to add this feature by sorting semantic versioning and just returning a subset of the largest version ones

helobinvn:
We have added sorting by Semantic versioning and getting the top largest tags. Maybe it could help someone who has the same use cases.

@mtrmac Do you think this request makes sense? It somehow relates to this https://github.com/containers/skopeo/issues/1756 as well, not sure if I should open this PR here, or move the discussion to that issue.

mtrmac:
Of the two features, I think limiting to a minimum version is probably more useful — limiting to N recent releases makes it non-obvious what will _actually_ be synced, especially for users who are on an older stable stream (suppose 3.1.0 exists for two years, there are dozens of releases ≥3.2.0, and 3.1.1 was just released).

OTOH that minimum version would need to be manually updated.

Strictly speaking, nothing prevents us from having both… though there’s some value to simplicity.

rhatdan:
@helobinvn are you still interested in this?

helobinvn:
Hi @rhatdan, yes. We're still using this patch on our end

rhatdan:
Could you rebase the PR then.

helobinvn:
@rhatdan I have rebased, but still, 1 test is failing, I guess it is unrelated to this PR? Please help to review, thx you!

rhatdan:
Did you modify the PR to match @mtrmac requests?

helobinvn:
I think I did, but not sure if this change covers enough cases or not, @mtrmac could you help to do CR again, pls?

cmd/skopeo/utils.go
```diff
@@ -409,3 +411,18 @@ func promptForPassphrase(privateKeyFile string, stdin, stdout *os.File) (string,
 	fmt.Fprintf(stdout, "\n")
 	return string(passphrase), nil
 }
+
+// Filter out refs which are not follow Semantic versioning
+func filterOutRefsNonSemver(imgRefs []types.ImageReference) ([]types.ImageReference, []types.ImageReference) {
```
mtrmac, line 416: (I’d keep this in `sync.go`; it seems unlikely to be useful for other subcommands.)

docs/skopeo-sync.1.md
```diff
@@ -68,6 +68,8 @@ Print usage statement.
 
 **--append-suffix** _tag-suffix_ String to append to destination tags.
 
+**--top-refs** Only synchronize for top latest refs after sorting by [Semver](https://semver.org/) provided by [go-version](https://github.com/hashicorp/go-version) lib. Fail if image's refs not follow Semantic versioning.
```
mtrmac, line 71: Users don‘t care about the implementation library.

---

If this actually only sorts by tags, it should say tags and not refs.

---

Further updates might be necessary depending on how exactly the scope of the filtering ends up being implemented.

cmd/skopeo/sync.go
```diff
@@ -647,6 +651,31 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) (retErr error) {
 
 	for _, srcRepo := range srcRepoList {
 		options.SourceCtx = srcRepo.Context
+		// only proceed for top refs if defined
+		if opts.topRefs > 0 {
```
mtrmac, line 655: Is doing it here the right place at all?

- This would trigger even on explicit single-tag sync configurations in the YAML format. That seems to only result in unnecessary noise.
- I imagine that users might very well want to specify a different limit for different repos. So the implementation should make that possible at least in principle. (Making that _actually_ possible would probably require changing the YAML format, compare https://github.com/containers/skopeo/pull/1792 .)
- (This filtering is enough code that moving it out of already pretty large `main` into a separate function seems worthwhile, either way. That would also allow invoking this code from the two parts of `imagesToCopyFromRegistry`, + `imagesToCopyFromDir`. And writing at least a smoke-test unit test for the filtering logic.)

cmd/skopeo/sync.go
```diff
@@ -118,6 +121,7 @@ See skopeo-sync(1) for details.
 	flags.StringVarP(&opts.destination, "dest", "d", "", "DESTINATION transport type")
 	flags.BoolVar(&opts.scoped, "scoped", false, "Images at DESTINATION are prefix using the full source image path as scope")
 	flags.StringVar(&opts.appendSuffix, "append-suffix", "", "String to append to DESTINATION tags")
+	flags.IntVar(&opts.topRefs, "top-refs", 0, "Only get amount of top largest refs after sorted by Semantic versioning")
```
mtrmac, line 124: “top” is fairly vague. Maybe `--semver-latest`? (That would naturally pair with a hypothetical `--semver-since=1.7.0`)

What do others think?

---

(That would then propagate everywhere.)

cmd/skopeo/sync.go
```diff
@@ -647,6 +651,31 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) (retErr error) {
...
+			topRefs := opts.topRefs
+			totalRefs := len(srcRepo.ImageRefs)
+			totalSemverRefs := len(semverRefs)
+			if totalSemverRefs == 0 {
+				return errors.New("ERROR: Not found any refs follow Semantic versioning. Check your tag's regex configuration OR doesn't it follow Semantic versioning?")
+			}
+			if topRefs > totalSemverRefs {
+				topRefs = totalSemverRefs
+			}
+			logrus.Infof("Only sync top largest %d ref(s) follow Semver in %d ref(s) from the total %d ref(s)", topRefs, totalSemverRefs, totalRefs)
```
mtrmac, line 676: All of these messages would really work _much_ better if they indicated which registry/repo (or whatever the scope of the filter is) they are talking about. There can be dozens of entries in `srcRepoList`, and without including the relevant scope these messages would just be numbers with no indication what they relate to.

So I suppose `repoDescriptor` needs to have some kind of `sourceDescription string` field added… Or maybe this code needs to be involved before a `repoDescriptor` is even created, with the source description provided by the caller in context.

mtrmac (commented):
Primary points of concern:
- What is the right scope of the filtering?
- How do we deal with directory sources?

cmd/skopeo/sync.go
TomSweeneyRedHat: I'm struggling with top-refs too..  If we stick with it, I'd be OK with the option name of `top-refs`, but we should really spell out the other `refs` references as `references`.  FWIW, I like @mtrmac 's name suggestion.

TomSweeneyRedHat:
@helobinvn thanks for the work on this.  Looks like it's time for another rebase.

mtrmac:
@helobinvn please see #1918 — would such an approach work for your needs?

rhatdan:
@helobinvn still working on this?

helobinvn:
> @helobinvn please see #1918 — would such an approach work for your needs?

Yes, and plus 1 more option to only take the amount of the largest tags, then it would be exactly what my PR intended to do

helobinvn:
> @helobinvn still working on this?

Hi @rhatdan, sorry I was not able to work on this recently, I will do it sometime next month. Thanks for understanding!

PR #1781, opened 2022-10-17
CI: Add a test flow that verifies the proxy

cgwalters:
This does reverse-dependency testing, verifying `proxy.go` using
the ostree-rs-ext Rust code's unit tests.

NOTE: Out of conservatism, this test *does not run by default*.
If you like, we can enable that.  However today, to run it,
the PR must be labeled with `proxyapi`.

### DIFF
--- a/.github/workflows/proxyapi.yaml
+++ b/.github/workflows/proxyapi.yaml
@@ -1,0 +1,65 @@
+# Keep this in sync with https://github.com/containers/containers-image-proxy-rs/blob/a4223a034895bf27423878f3af2ba22d50fc7b36/.github/workflows/ci.yaml
+name: Proxy ostree-ext
+
+permissions:
+  actions: read
+
+on:
+  pull_request:
+    types: [labeled, opened, synchronize, reopened]
+  workflow_dispatch:
+
+jobs:
+  build:
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'proxyapi') || github.event_name == 'workflow_dispatch' }}
+    name: Build
+    runs-on: ubuntu-latest
+    container: quay.io/coreos-assembler/fcos-buildroot:testing-devel
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Mark git checkout as safe
+        run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
+      - name: Install deps
+        run: dnf builddep -y skopeo
+      - name: Compile
+        run: make
+      - name: Upload binary
+        uses: actions/upload-artifact@v3
+        with:
+          name: skopeo
+          path: bin/skopeo
+  proxy-ostree-ext:
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'proxyapi') || github.event_name == 'workflow_dispatch'}}
+    name: Proxy API
+    needs: build
+    runs-on: ubuntu-latest
+    container: quay.io/coreos-assembler/fcos-buildroot:testing-devel
+
+    steps:
+      - name: Install skopeo deps
+        run: dnf -y install skopeo && rpm -e skopeo
+      - name: Download skopeo build
+        uses: actions/download-artifact@v3
+        with:
+          name: skopeo
+      - name: Install skopeo
+        run: "install skopeo /usr/bin"
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          repository: ostreedev/ostree-rs-ext
+          ref: main
+      - name: Mark git checkout as safe
+        run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
+      - name: Build
+        run: cargo test --no-run
+      - name: Run tests
+        run: cargo test -- --nocapture --quiet


cgwalters:
>  However today, to run it, the PR must be labeled with proxyapi.

We also need to create this label.

cgwalters:
Demo of a successful run of this job over here https://github.com/cgwalters/skopeo/actions/runs/3268236269/jobs/5374445389

(For those who didn't know, a handy tip: you can test github actions by force-pushing the change to *your fork's* `main` branch, then you'll see the runs under the "Actions" tab; that's how I did this)

mtrmac (commented):
Thanks.

- Existing tests are mostly driven by `.cirrus.yml`. I’m not sure if there are any implications to using a GitHub workflow.
- Currently, GitHub actions use digest pinning. IIRC there’s a security rationale.

Anyway, I’m way out of my area of expertise.

@cevich PTAL.

cevich:
Yeah my preference would be to do this in Cirrus-CI if possible.  Mainly because it's simple to understand and maintain by non-experts.  I don't mind github actions for "idiot-checks" like DCO, but this seems much more than that.

What's the reason for not wanting to run this by default?

If it doesn't need to run in every PR, we have a mechanism where we can run tests on a cron-like schedule.  That would allow re-use of the podman-next COPR to install skopeo (pre-built from main) and avoid a compile.  The cirrus-cron jobs monitored for any failures.  Would something like that be a better fit than the PR-CI workflow?

In all cases, we very much prefer to not install dependencies at runtime.  Automation reliability has nearly always suffered anytime we do that.  We have rust-toolchain VM images we use for netavark/aardvark CI, that include podman as well.  Maybe those could be used here to avoid runtime installs?

cgwalters:
Well basically: I know Prow, Jenkins and Github Actions; I don't know Cirrus.

> In all cases, we very much prefer to not install dependencies at runtime.

Yeah, we could use shared images here I'm sure. 

>  We have rust-toolchain VM images we use for netavark/aardvark CI

Friends don't let friends install compilers on the *root filesystem* :wink: - builds should be done in a containers IMO.

I guess bottom line: I'm not sure I can say I have the bandwidth to push "learn Cirrus" onto my stack here to then land more proxy PRs.  It took an hour of my time to add and test this test.  Seems like we should try to figure out a longer term CI deduplication/sharing, but I think for now this test is going to have zero impact here except when I do proxy PRs and add the label to run it.

cevich:
Cirrus-CI is far simpler than all those you've named.  We have lots of developers and even drive-by contributors who have no issue contributing to it, I'm 100% certain it's not beyond your skills :wink:  Cirrus-CI is very narrow-minded and naive.  You get one VM or container per task, and it runs the shell commands or scripts you give it.  For anything more complex, I'm happy to help.

For the most part we avoid Cirrus container tasks, instead crafting our own podman-commands to run in VMs as it gives more control over arguments and output.  In most cases, we treat our VM images similar to container images.  The goal is to support developer-velocity and automation stability.  So very few (if any) mechanisms rely on references to "latest" or versioned inputs/tooling. 

In any case, let me see if I understand the goal here:

You want to ensure a new proxy PR or tag is compatible with Skopeo from `main`.  At the same time, future Skopeo PRs or branch changes remain compatible with proxy.   Have I got that right?

Idea:

We have a "podman-next" COPR that continuously builds Skopeo (among other things), presumably Proxy has something similar?  What if we split up the check, such that Proxy's CI continuously tested vs. the "podman-next" Skopeo.  Then similarly, Skopeo CI continuously tested vs. continuous builds of Proxy?  That should cover all angles, ya?

cgwalters:
> You want to ensure a new proxy PR or tag is compatible with Skopeo from main. At the same time, future Skopeo PRs or branch changes remain compatible with proxy. Have I got that right?

Mmm...the proxy code in here *is* part of skopeo.  

What I want to do is: on an open PR to skopeo (e.g. https://github.com/containers/skopeo/pull/1757 ) - I want to verify that it doesn't break https://github.com/containers/containers-image-proxy-rs/ (really, that it doesn't break https://github.com/ostreedev/ostree-rs-ext )

It would also make sense to verify that the proxy's reverse dependencies work when we e.g. update the vendored containers/image in skopeo.

And continuous verification of skopeo's main *also* makes sense.

I think at a practical level, the per-PR testing combined with periodic integration testing - specifically, continuously verifying that tip `main` skopeo and rpm-ostree work for OS updates would be good.

> Cirrus-CI is far simpler than all those you've named. We have lots of developers and even drive-by contributors who have no issue contributing to it, I'm 100% certain it's not beyond your skills wink

I'm sure it's not, but...we also have on our plate e.g. integration with [Fedora tmt and distro testing](https://docs.fedoraproject.org/en-US/ci/tmt/) - and I'm balking a bit at trying to learn that simultaneously with Cirrus.  Happy to have help in porting this, if you really feel the effort in doing so is valuable versus just using the CI code I already wrote and tested.

cevich:
Sorry it took a while to reply, I was on PTO last week and catching up all this week.  

Yep, I think can help get this into our Cirrus stack here.  It's helpful for us to stick with a common set of simple, predictable tools that most folks on this team are familiar and comfortable using (and debugging).

I don't have any hand in downstream Fedora testing stuffs, but Cirrus can fire even web-hooks if that helps in some way.  Or we can poke APIs from the Cirrus cron-jobs as well.

cevich:
Okay, [that PR should take care of the ostree_ext](https://github.com/cevich/skopeo/commit/f4b4eb043a4820de3e258ea9ed6dddc118e848b4) testing.

re: TMT, looking at the link you provided, I'm a bit confused.  Are there jobs already stood up somewhere that you're needing to integrate with (how?).  Or is the [setup/commands on that page](https://docs.fedoraproject.org/en-US/ci/tmt/) something you're looking to execute in skopeo branch and/or PR CI?

Assuming the later, (on quick-glance) it looks like something we could do in a Cirrus container task (unless the tests themselves need to run containers).  Though to improve performance and CI-reliability, it would be good to use images w/ pre-installed TMT-dependencies and tooling.  

I can help arrange for that in c/automation_images if they're not already built somewhere.    Same for VM-based testing and VM-images if containers need to be orchestrated by the tests.  Just LMK.

cgwalters:
Closing in favor of https://github.com/cevich/skopeo/commit/f4b4eb043a4820de3e258ea9ed6dddc118e848b4

PR #1786, opened 2022-10-20
Update to c/image main branch

mtrmac:
```sh
go get github.com/containers/image/v5@main
make vendor
```

... to make sure that we don't regress against Skopeo 1.9.3.

Note that the `release-1.10` branch currently stays without the backported c/image features.


vrothberg:
> Note that the release-1.10 branch currently stays without the backported c/image features.

Why?

vrothberg (commented):
LGTM

mtrmac:
> > Note that the release-1.10 branch currently stays without the backported c/image features.
> 
> Why?

For RHEL purposes, I don’t expect the 1.10 branch to be used.

Still, thanks for the nudge — it’s better to spend that extra time now just decrease the risk of unexpected last-minute surprises in the future. https://github.com/containers/skopeo/pull/1787

PR #1792, opened 2022-10-24
Added yaml2 sync config

### LINKED ISSUE #1072: Sync: Add option to keep full name on target when using YAML

mhowell24:
Adds the yaml2 sync configuration that allows rewriting of the repository path on a repo by repo basis. This largely follows the new yaml format suggested in #1531 and fixes #1072. I have written some testing but need to update the documentation to document the new configuration and its features.

Signed-off-by: Max Howell <maxhowell24@gmail.com>

### DIFF
--- a/cmd/skopeo/sync.go
+++ b/cmd/skopeo/sync.go
@@ -50,9 +50,10 @@ type syncOptions struct {
 
 // repoDescriptor contains information of a single repository used as a sync source.
 type repoDescriptor struct {
-	DirBasePath string                 // base path when source is 'dir'
-	ImageRefs   []types.ImageReference // List of tagged image found for the repository
-	Context     *types.SystemContext   // SystemContext for the sync command
+	DirBasePath    string                 // base path when source is 'dir'
+	DestinationRef types.ImageReference   // Destination
+	ImageRefs      []types.ImageReference // List of tagged image found for the repository
+	Context        *types.SystemContext   // SystemContext for the sync command
 }
 
 // tlsVerifyConfig is an implementation of the Unmarshaler interface, used to
@@ -74,6 +75,25 @@ type registrySyncConfig struct {
 // sourceConfig contains all registries information read from the source YAML file
 type sourceConfig map[string]registrySyncConfig
 
+// syncConfig contains all registries information read from the source YAML file for source yaml2
+type syncConfig map[string]registrySyncConfigV2
+
+// registrySyncConfigV2 contains all information read from the yaml2 sync config for a registry
+type registrySyncConfigV2 struct {
+	Repos       map[string]repoSyncConfig
+	Credentials types.DockerAuthConfig // Username and password used to authenticate with the registry
+	TLSVerify   tlsVerifyConfig        `yaml:"tls-verify"` // TLS verification mode (enabled by default)
+	CertDir     string                 `yaml:"cert-dir"`   // Path to the TLS certificates of the registry
+}
+
+// repoSyncConfig contains all information needed to sync an individual repository
+type repoSyncConfig struct {
+	DestinationRepo string   `yaml:"destination-repo"`    // Destination repository
+	MirrorTags      []string `yaml:"mirror-tags"`         // tags to sync
+	MirrorDigests   []string `yaml:"mirror-digests"`      // digests to sync
+	MirrorTagsRegex string   `yaml:"mirror-by-tag-regex"` // Regex to filter tags by
+}
+
 func syncCmd(global *globalOptions) *cobra.Command {
 	sharedFlags, sharedOpts := sharedImageFlags()
 	deprecatedTLSVerifyFlags, deprecatedTLSVerifyOpt := deprecatedTLSVerifyFlags()
@@ -94,7 +114,7 @@ func syncCmd(global *globalOptions) *cobra.Command {
 		Short: "Synchronize one or more images from one location to another",
 		Long: `Copy all the images from a SOURCE to a DESTINATION.
 
-Allowed SOURCE transports (specified with --src): docker, dir, yaml.
+Allowed SOURCE transports (specified with --src): docker, dir, yaml, yaml2.
 Allowed DESTINATION transports (specified with --dest): docker, dir.
 
 See skopeo-sync(1) for details.
@@ -142,6 +162,21 @@ func (tls *tlsVerifyConfig) UnmarshalYAML(unmarshal func(interface{}) error) err
 // It returns a new unmarshaled sourceConfig object and any error encountered.
 func newSourceConfig(yamlFile string) (sourceConfig, error) {
 	var cfg sourceConfig
+	source, err := os.ReadFile(yamlFile)
+	if err != nil {
+		return cfg, err
+	}
+	err = yaml.Unmarshal(source, &cfg)
+	if err != nil {
+		return cfg, fmt.Errorf("Failed to unmarshal %q: %w", yamlFile, err)
+	}
+	return cfg, nil
+}
+
+// newSourceConfig unmarshals the provided YAML file path to the sourceConfig type.
+// It returns a new unmarshaled sourceConfig object and any error encountered.
+func newSourceConfigYaml2(yamlFile string) (syncConfig, error) {
+	var cfg syncConfig
 	source, err := os.ReadFile(yamlFile)
 	if err != nil {
 		return cfg, err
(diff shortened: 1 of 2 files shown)


cmd/skopeo/sync.go
```diff
@@ -50,9 +50,10 @@ type syncOptions struct {
 
 // repoDescriptor contains information of a single repository used as a sync source.
```
mtrmac, line 51: “used as a sync source” is no longer quite true I guess?

cmd/skopeo/sync.go
```diff
@@ -50,9 +50,10 @@ type syncOptions struct {
...
 type repoDescriptor struct {
-	DirBasePath string                 // base path when source is 'dir'
-	ImageRefs   []types.ImageReference // List of tagged image found for the repository
-	Context     *types.SystemContext   // SystemContext for the sync command
+	DirBasePath    string                 // base path when source is 'dir'
+	DestinationRef types.ImageReference   // Destination
```
mtrmac, line 54: `types.ImageReference` inherently points at a _single_ image. It should not be abused as a repo/directory value.

One way to do this might be to move the `destinationReference` calls into `repoDescriptor` construction somehow, and to have `repoDescriptor` already contain fully-resolved (source reference, destination reference) pairs.

cmd/skopeo/sync.go
```diff
@@ -74,6 +75,25 @@ type registrySyncConfig struct {
 // sourceConfig contains all registries information read from the source YAML file
 type sourceConfig map[string]registrySyncConfig
 
+// syncConfig contains all registries information read from the source YAML file for source yaml2
+type syncConfig map[string]registrySyncConfigV2
```
mtrmac, line 79: - So is this `syncConfigV2`?

- One of the major lessons from v1 is that having a top-level `map[something]something` structure makes it impossible to add options at that level. So maybe the top-level should be a `struct` here as well.

cmd/skopeo/sync.go
```diff
@@ -74,6 +75,25 @@ type registrySyncConfig struct {
...
+
+// registrySyncConfigV2 contains all information read from the yaml2 sync config for a registry
+type registrySyncConfigV2 struct {
+	Repos       map[string]repoSyncConfig
+	Credentials types.DockerAuthConfig // Username and password used to authenticate with the registry
```
mtrmac, line 84: Are credentials really a per-`SyncConfig` thing?

Conceptually we need both source and destination credentials. How does that work here?

(I _guess_ credentials should be a separate top-level map but I’m not at all sure about that.)

cmd/skopeo/sync.go
```diff
@@ -74,6 +75,25 @@ type registrySyncConfig struct {
...
+	TLSVerify   tlsVerifyConfig        `yaml:"tls-verify"` // TLS verification mode (enabled by default)
```
mtrmac, line 85: This should be separate for source and destination. (And, like credentials, possibly separate from the repo configurations — it makes no sense for two repo destinations on the same registry to use a different value.)

cmd/skopeo/sync.go
```diff
@@ -418,12 +453,161 @@ func imagesToCopyFromRegistry(registryName string, cfg registrySyncConfig, sourc
 	return repoDescList, nil
 }
 
+// imagesToCopyFromRegistryV2 builds a list of repository descriptors from the images
+// in a registry configuration.
+// It returns a repository descriptors slice with as many elements as the images
+// found and any error encountered. Each element of the slice is a list of
+// image references, to be used as sync source.
+func imagesToCopyFromRegistryV2(registryName string, cfg registrySyncConfigV2, sourceCtx types.SystemContext, destination string) ([]repoDescriptor, error) {
```
mtrmac, line 461: Would it be practical at all, instead of duplicating this large logic, to instead convert v1 data into the v2 format, and have only one execution path?

cmd/skopeo/sync.go
```diff
@@ -624,27 +826,61 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) (retErr error) {
...
+
+					destRef, err = docker.ParseReference(fmt.Sprintf("//%s", destTaggedRef.String()))
+					if err != nil {
+						return err
+					}
+				default:
+					logrus.Errorf("destination ref type is unsupported for this sync")
+					return fmt.Errorf("destination ref type is unsupported for this sync")
+				}
+			} else {
```
mtrmac, line 861: Is it really valuable at all to allow specifying a destination on the command line with YAML2? I imagine it’s always going to be the same for the same YAML file, though I’m not quite sure.

(OTOH we don’t support this code path, we need to account for directory destinations in the YAML format.)

mtrmac (commented):
Thanks!

Just a _very_ quick skim, focusing basically only on the config format semantics.

mhowell24:
Thanks for the quick response @mtrmac. You make a good point about the config being a map removing some flexibility, I agree a struct would be better. Considering your comments from above, how about something along the lines of this for new config format:

``` yaml
syncs:
- source:
    registry: example.docker.io
    tls-verify: true
    cert-dir: /example/certs
    credentials:
      username: joeBlogs123
      password: password123
  destination:
    registry: my.other.registry
    tls-verify: false
    cert-dir: /example/other/certs
    credentials:
      username: joeBlogs456
      password: password456
  repos:
  - source-path: example/image
    destination-path: example/test/image
    images:
      - v1.0
      - v1.1
      - v2.0
      - sha256:0000000000000000000000000000000011111111111111111111111111111111
    images-by-tag-regex: ^v[0-9]\.0$
  - source-path: example/image2
    destination-path: image2
    images:
      - "1.0"
      - "1.1"
      - "2.0"
      - sha256:0000000000000000000000000000000011111111111111111111111111111111
    images-by-tag-regex: ^[0-9]\.1$
```

Each element in `syncs` allows us to define everything we need for syncing multiple repos from a single source registry to repos in a single destination registry, it also give the user control over the destination path on a repo per repo basis.

We could go further and seperate the registries out completly and then use them as either sources or destinations e.g.
``` yaml
registries:
- name: reg0
  domain: example.docker.io
  tls-verify: true
  cert-dir: /example/certs
  credentials:
    username: joeBlogs123
    password: password123
- name: reg1
  domain: my.other.registry
  tls-verify: false
  cert-dir: /example/certs
  credentials:
    username: joeBlogs456
    password: password456
- name: reg2
  domain: my.backup.registry
  tls-verify: false
  cert-dir: /example/certs
  credentials:
    username: joeBlogs789
    password: password789
repos:
- source: reg0
  destination: reg1
  source-path: example/image
  destination-path: example/test/image
  images:
  - v1.0
  - v1.1
  - v2.0
  - sha256:0000000000000000000000000000000011111111111111111111111111111111
  images-by-tag-regex: ^v[0-9]\.0$
- source: reg1
  destination: reg3
  source-path: example/image2
  destination-path: image2
  images:
  - "1.0"
  - "1.1"
  - "2.0"
  - sha256:0000000000000000000000000000000011111111111111111111111111111111
  images-by-tag-regex: ^[0-9]\.1$
```

mtrmac:
The second version (separate registries configuration from repo sets) seems preferable to me, because I’m _guessing_ that many users have a single private / disconnected mirror that contains copies from several internet repositories; in that case it’s useful to only have to specify the single-mirror credentials once.

But that’s a _guess_ without any data. What do actual users of `skopeo sync` think?

mtrmac:
- Also, looking at the above, I’m not sure the registries need an extra “name”; I think just identifying them using the `host[:port]` value would work well enough. Most importantly that would allow the _simple_ case, where the registry needs no configuration (because credentials are available in `auth.json` already) to be set up without writing a `registries` entry.
- I’m not sure how the source/destination/source-path/destination-path fields work for dir: locations. One possibility would be to have `source-registry-repo`/`source-dir`/`destination-registry-repo`/`destination-dir`, with each field having a single clearly-defined syntax — but I haven’t really taken any time to consider the alternatives or trade-offs.

cmd/skopeo/sync.go
mhowell24: It would be nice to try and render everything down to pairs of source and destination references regardless of how sync was called i.e. using specific transports from the command line or using one of the yaml configs. I think this is possible however it could be a rather large change

art-shutter:
I believe the option to abstract registries away could actually complicate things. If all I need is a simple case "paths on the source and destination should be the same", supplying paths and repos for every set of images becomes a task that needs a templating engine on top to actually generate the config.

rhatdan:
@mhowell24 Still working on this?

PR #1802, opened 2022-11-03
proxy: Add LayerInfoJSON API

RishabhSaini:
ref: https://github.com/ostreedev/ostree-rs-ext/pull/396

### DIFF
--- a/cmd/skopeo/proxy.go
+++ b/cmd/skopeo/proxy.go
@@ -92,7 +92,8 @@ import (
 // 0.2.2: Added support for fetching image configuration as OCI
 // 0.2.3: Added GetFullConfig
 // 0.2.4: Added OpenImageOptional
-const protocolVersion = "0.2.4"
+// 0.2.5: Added LayerInfoJSON
+const protocolVersion = "0.2.5"
 
 // maxMsgSize is the current limit on a packet size.
 // Note that all non-metadata (i.e. payload data) is sent over a pipe.
@@ -171,6 +172,14 @@ type proxyHandler struct {
 	images map[uint32]*openImage
 	// activePipes maps from "pipeid" to a pipe + goroutine pair
 	activePipes map[uint32]*activePipe
+}
+
+// convertedLayerInfo is the reduced form of the OCI type BlobInfo
+// Used in the return value of GetLayerInfo
+type convertedLayerInfo struct {
+	Digest    digest.Digest `json:"digest"`
+	Size      int64         `json:"size"`
+	MediaType string        `json:"media_type"`
 }
 
 // Initialize performs one-time initialization, and returns the protocol version
@@ -614,6 +623,56 @@ func (h *proxyHandler) GetBlob(args []interface{}) (replyBuf, error) {
 	return ret, nil
 }
 
+// GetLayerInfo returns data about the layers of an image, useful for reading the layer contents.
+//
+// This needs to be called since the data returned by GetManifest() does not allow to correctly
+// calling GetBlob() for the containers-storage: transport (which doesn’t store the original compressed
+// representations referenced in the manifest).
+func (h *proxyHandler) GetLayerInfo(args []interface{}) (replyBuf, error) {
+	h.lock.Lock()
+	defer h.lock.Unlock()
+
+	var ret replyBuf
+
+	if h.sysctx == nil {
+		return ret, fmt.Errorf("client error: must invoke Initialize")
+	}
+
+	if len(args) != 1 {
+		return ret, fmt.Errorf("found %d args, expecting (imgid)", len(args))
+	}
+
+	imgref, err := h.parseImageFromID(args[0])
+	if err != nil {
+		return ret, err
+	}
+
+	ctx := context.TODO()
+
+	err = h.cacheTargetManifest(imgref)
+	if err != nil {
+		return ret, err
+	}
+	img := imgref.cachedimg
+
+	layerInfos, err := img.LayerInfosForCopy(ctx)
+	if err != nil {
+		return ret, err
+	}
+
+	if layerInfos == nil {
+		layerInfos = img.LayerInfos()
+	}
+
+	var layers []convertedLayerInfo
+	for _, layer := range layerInfos {
+		layers = append(layers, convertedLayerInfo{layer.Digest, layer.Size, layer.MediaType})
+	}
+
+	ret.value = layers
+	return ret, nil
+}
+
 // FinishPipe waits for the worker goroutine to finish, and closes the write side of the pipe.
 func (h *proxyHandler) FinishPipe(args []interface{}) (replyBuf, error) {
 	h.lock.Lock()
@@ -736,6 +795,8 @@ func (h *proxyHandler) processRequest(readBytes []byte) (rb replyBuf, terminate 
 		rb, err = h.GetFullConfig(req.Args)
 	case "GetBlob":
 		rb, err = h.GetBlob(req.Args)
+	case "GetLayerInfo":
+		rb, err = h.GetLayerInfo(req.Args)
 	case "FinishPipe":
 		rb, err = h.FinishPipe(req.Args)
 	case "Shutdown":


cmd/skopeo/proxy.go
```diff
@@ -166,6 +167,11 @@ type proxyHandler struct {
 	activePipes map[uint32]*activePipe
 }
 
+type layersInfo struct {
+  Success string
```
mtrmac, line 171: There already is an error reporting mechanism. Why does this need to exist? Especially when it’s not really used.

cmd/skopeo/proxy.go
```diff
@@ -568,6 +574,65 @@ func (h *proxyHandler) GetBlob(args []interface{}) (replyBuf, error) {
...
+  if err != nil {
+    return ret, err
+  }
+  if infos == nil {
+    rawManifest, _, err := imgref.src.GetManifest(ctx, nil)
+    if err != nil {
+      return ret, err
+    }
+    var serManifest manifest.OCI1
+    if err = json.Unmarshal(rawManifest, &serManifest); err != nil {
```
mtrmac, line 609: This is not how `manifest.OCI` is supposed to be used. And I see no reason for this to hard-code the OCI1 assumption.

Really all of this should be done in another way:
- Get a `types.Image` — `cacheTargetManifest` I think
- Then call `LayerInfosForCopy`/`LayerInfos` for that.

cmd/skopeo/proxy.go
```diff
@@ -568,6 +574,65 @@ func (h *proxyHandler) GetBlob(args []interface{}) (replyBuf, error) {
...
+  serialized, err := json.Marshal(&serStruct)
+	if err != nil {
+		return ret, err
+	}
+  
+  return h.returnBytes(nil, serialized)
+  /*
+  ret.value = serialized
+  return ret, err
+  */
```
mtrmac, line 633: Why is this using the much more costly pipe mechanism instead of just returning the data as in the commented-out path?

integration/proxy_test.go
```diff
@@ -230,7 +230,13 @@ type byteFetch struct {
...
 
 func runTestGetManifestAndConfig(p *proxy, img string) error {
-	v, err := p.callNoFd("OpenImage", []interface{}{knownNotManifestListedImage_x8664})
+  
+  v, err := p.callNoFd("OpenImage", []interface{}{"containers-storage:docker.io/busybox"})
+	if err != nil {
+		return err
+	}
+
+  v, err = p.callNoFd("OpenImage", []interface{}{knownNotManifestListedImage_x8664})
```
mtrmac, line 239: This seems unlikely to do what was desired, whatever that was.

integration/proxy_test.go
```diff
@@ -241,6 +247,12 @@ func runTestGetManifestAndConfig(p *proxy, img string) error {
 	}
 	imgid := uint32(imgidv)
 
+  info, err := p.callNoFd("LayerInfoJSON", []interface{}{imgid})
+  if err == nil {
+    return err
+  }
+  fmt.Printf("Field: %+v\n", info)
```
mtrmac, line 254: `fmt.Printf` should not be used in tests. And this is not testing much.

(Ultimately, I guess the test would be to verify that the returned info is actually usable for pulling, OTOH that might be too costly.)

cmd/skopeo/proxy.go
```diff
@@ -166,6 +167,11 @@ type proxyHandler struct {
 	activePipes map[uint32]*activePipe
 }
 
+type layersInfo struct {
+  Success string 
+  Layers []types.BlobInfo
```
mtrmac, line 172: Consider not exporting all of the `BlobInfo` — The “operation/algorithm” fields were really a mistake.

mtrmac (commented):
Note the test failures as well, especially https://github.com/containers/skopeo/blob/main/CONTRIBUTING.md#sign-your-prs.

cmd/skopeo/proxy.go
```diff
@@ -568,6 +573,52 @@ func (h *proxyHandler) GetBlob(args []interface{}) (replyBuf, error) {
 	return ret, nil
 }
 
+func (h *proxyHandler) LayerInfoJSON(args []interface{}) (replyBuf, error) {
+  h.lock.Lock()
+	defer h.lock.Unlock()
```
cgwalters, line 578: You should run `gofmt`.

cmd/skopeo/proxy.go
```diff
@@ -568,6 +573,52 @@ func (h *proxyHandler) GetBlob(args []interface{}) (replyBuf, error) {
...
+  if err != nil {
+    return ret, err
+  }
+  
+  if infos == nil {
+     infos = img.LayerInfos()
+  }
+
+  var serStruct layersInfo
+  serStruct.Success = "successfull"
```
cgwalters, line 615: I think we can drop this; we already have an error reporting mechanism.  Then we can just do `ret.value = infos` which would seem to simplify things.

cmd/skopeo/proxy.go
```diff
@@ -568,6 +573,52 @@ func (h *proxyHandler) GetBlob(args []interface{}) (replyBuf, error) {
...
+	}
+  
+  imgref, err := h.parseImageFromID(args[0])
+  if err != nil {
+    return ret, err
+  }
+
+  ctx := context.TODO()
+
+  var infos []types.BlobInfo
```
cgwalters, line 597: I think we can drop this and then below...

cmd/skopeo/proxy.go
```diff
@@ -568,6 +573,52 @@ func (h *proxyHandler) GetBlob(args []interface{}) (replyBuf, error) {
...
+
+  err = h.cacheTargetManifest(imgref)
+	if err != nil {
+		return ret, err
+	}
+	img := imgref.cachedimg
+
+  infos, err = img.LayerInfosForCopy(ctx)
```
cgwalters, line 605: Should work to just do `infos, err := img.LayerInfosForCopy(ctx)`.

cgwalters (commented):
Thanks so much for working on this!

TomSweeneyRedHat:
LGTM, but definitely want other eyeballs on this one.

cmd/skopeo/proxy.go
```diff
@@ -614,6 +620,51 @@ func (h *proxyHandler) GetBlob(args []interface{}) (replyBuf, error) {
...
+		layerInfos = img.LayerInfos()
+	}
+
+	var custBlobArr []blobInfo
+	for _, blob := range layerInfos {
+		custBlobArr = append(custBlobArr, blobInfo{blob.Digest, blob.Size, blob.MediaType})
+	}
+
+	ret.value = custBlobArr
+	return ret, err
```
mtrmac, line 665: `return ret, nil` would be a bit clearer.

mtrmac (commented):
Thanks! Looks reasonable from a quick look, I’ll reread more carefully later.
- Please drop the `integration.test` binary from the PR.
- Do we want a new test of some kind? Or would this just rely on the `ostree-rs-ext` tests in #1808?

cgwalters:
I'd also retitle the PR to "proxy: Add LayerInfoJSON API" so that it shows up more nicely in a merge commit too.

cmd/skopeo/proxy.go
```diff
@@ -614,6 +621,51 @@ func (h *proxyHandler) GetBlob(args []interface{}) (replyBuf, error) {
...
+	layerInfos, err := img.LayerInfosForCopy(ctx)
+	if err != nil {
+		return ret, err
+	}
+
+	if layerInfos == nil {
+		layerInfos = img.LayerInfos()
+	}
+
+	var custBlobArr []blobInfo
```
mtrmac, line 660: Just `blobs`? Or `converted`? “Array” doesn’t need to be in the variable name, “cust”(om???) seems unnecessary.

I think names should either be fairly long and descriptive to be very self-documenting (and in that case abbreviating words is counterproductive), typically when used over long distances; or short and convenient to use, when they have very small scope and the purpose is obvious.

cmd/skopeo/proxy.go
```diff
@@ -614,6 +621,51 @@ func (h *proxyHandler) GetBlob(args []interface{}) (replyBuf, error) {
...
+	for _, blob := range layerInfos {
+		custBlobArr = append(custBlobArr, blobInfo{blob.Digest, blob.Size, blob.MediaType})
+	}
```
mtrmac, line 663: `…blob := … layer…`

That sadly exists in the `c/image` type names, for ~historical reasons, but this code is specifically concerned about _layers_ (as they exist in c/storage; and as opposed to generic blobs in OCI artifacts, which can‘t be in c/storage). So maybe name this array, the iteration variable, and the `blobInfo` type to specifically refer to layers, not blobs.

cmd/skopeo/proxy.go
```diff
@@ -173,6 +174,12 @@ type proxyHandler struct {
 	activePipes map[uint32]*activePipe
 }
 
+type blobInfo struct {
```
mtrmac, line 177: Following precedent, please document this type — at least enough to say that is used in the return value of `LayerInfoJSON`.

cmd/skopeo/proxy.go
```diff
@@ -614,6 +621,51 @@ func (h *proxyHandler) GetBlob(args []interface{}) (replyBuf, error) {
 	return ret, nil
 }
 
+func (h *proxyHandler) LayerInfoJSON(args []interface{}) (replyBuf, error) {
```
mtrmac, line 624: Please document the operation here. At the very least a short blurb, and the exceptional/unusual/non-obvious aspects. And there _is_ a non-obvious aspect that really should be documented: This exists because the manifest data as returned by `GetManifest` is not sufficient to correctly call `GetBlob` against c/storage sources.

mtrmac (commented):
Implementation LGTM, just some comments on naming and documentation.

mtrmac:
> * Do we want a new test of some kind? Or would this just rely on the `ostree-rs-ext` tests in #1808?

@cgwalters WDYT?

cgwalters:
I'm ok just relying on the reverse dependency tests for now...longer term, one thing I've been thinking we could do perhaps is implement `skopeo copy --internal-via-proxy` that would actually cause skopeo to fork itself and copy via the proxy code.  That would allow us to gain a whole lot of testing for free.  (To fully generalize it we'd need write support in the proxy too, but copying from anything to an `oci` directory should be easy)

mtrmac:
(Having a “proxy client” transport in c/image is potentially interesting. For a time, long ago, someone was repeatedly asking for a “copy image over ssh” feature… but I haven’t heard about that for a long time now, I don’t know whether that’s because the feature is not actually needed or because those users were motivated to find another implementation.

I would recommend keeping both the server and client restricted to a “source”, at least for a first version: the source interface is fairly straightforward and doesn’t change that much, while the destination has has _much_ more reason to be extended or outright changed when looking for ways to speed up copies.)

cgwalters:
Yeah, though the "over ssh" bit would require us to skip the local-only optimization of fd passing via pipe().  On that thread...I've been investigating Cap'n Proto more and thinking of using that for an "official" interface.

Also tangentially related to this is that I did float the idea at one point that skopeo could actually stop linking in the containers/storage code entirely and instead fork off `/usr/bin/podman` to be the backend for that, communicating via this protocol.  (The advantage of this would be a lot less code and fewer dependencies linked into the skopeo binary, fewer worries about version compat etc.)

cmd/skopeo/proxy.go
```diff
@@ -736,6 +788,8 @@ func (h *proxyHandler) processRequest(readBytes []byte) (rb replyBuf, terminate
 		rb, err = h.GetFullConfig(req.Args)
 	case "GetBlob":
 		rb, err = h.GetBlob(req.Args)
+	case "LayerInfoJSON":
```
cgwalters, line 791: A naming nit here...I think the fact that we transport via JSON is an implementation detail.  How about just "GetLayerInfo"?

mtrmac: Good idea.

mtrmac (commented):
I’m sorry, somehow this remained a pending review…

mtrmac (approved):
LGTM. Thanks!

PR #1808, opened 2022-11-08
Cirrus: Add reverse-deps. test to verify proxy with ostree-rs-ext

cevich:
This does reverse-dependency testing, verifying `proxy.go` using the ostree-rs-ext Rust code's unit tests.

Based on #1781 by @cgwalters

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -119,6 +119,49 @@ cross_task:
         "${GOSRC}/${SCRIPT_BASE}/runner.sh" cross
 
 
+ostree-rs-ext_task:
+    alias: proxy_ostree_ext
+    # Don't run for docs-only or multi-arch image builds.
+    # Also don't run on release-branches or their PRs,
+    # since base container-image is not version-constrained.
+    only_if: &not_docs_or_release_branch >-
+        ($CIRRUS_BASE_BRANCH == $CIRRUS_DEFAULT_BRANCH ||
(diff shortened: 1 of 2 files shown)


cevich:
@cgwalters PTAL, this should be ready to go.

cevich:
***Note:*** Ignore all the `...but their only_if conditions are different` annotations, they're harmless - just warning that bypass of a parent could cause child artifact-retrieval failure.  Since we're not passing artifacts, there is no problem.

cgwalters (approved):
Looks sane to me, thanks!

cevich:
Great.  I remembered that rust builds are fairly intensive.  If the ~7min time becomes a problem, there are some build-caching options available.  Unless you object, I'm content to leave this as-is until the runtime proves problematic (caching adds a lot of complexity).

mtrmac (commented):
ACK — I’ll fully defer to the two of you on specifics.

---

The original PR, “out of conservatism”, did not enable the test by default (and there was some discussion about running cron-like instead). Do we need to revisit that?

(If you are happy with running this on every PR, fine, let’s do that. I just wanted to make sure this was not forgotten.)

cevich:
Thanks for checking Miloslav.  Yeah, that was in lieu of running in PRs.  I think rather than try and predict where a test-failure will be useful, let's just run it the same as any other skopeo test.  It's possible that any PR or merge could break it, so this way we'll know.  

We've already got cirrus-cron daily running on `main` here, we can keep an eye out in case it breaks on `main` :grinning:

PR #1820, opened 2022-12-06
GHA/Cirrus-cron: Fix execution order

cevich:
Fairly universally, the last Cirrus-Cron job is set to fire off at 22:22 UTC.  However, the re-run of failed jobs GHA workflow was scheduled for 22:05, meaning it will never re-run the last cirrus-cron job should it fail.

Re-arrange the execution order so as to give plenty of time between the last cirrus-cron job starting, the auto-re-run attempt, and the final failure-check e-mail.


edsantiago (approved):
/lgtm

mtrmac (approved):
Thanks!

PR #1739, opened 2022-08-24
Cirrus: use image with fewer downloaded dependencies

lsm5:
The latest cirrus image bump will use fewer downloaded dependencies in
favor of installation from copr / distro repos.

Signed-off-by: Lokesh Mandvekar <lsm5@fedoraproject.org>

@cevich PTAL

PR #1821, opened 2022-12-06
Cirrus: Update to F37 CI VM Images

cevich:


cevich:
@mtrmac PTAL, this seems like a really odd error to get on a F36 -> F37 image update.  Any idea what I might have screwed up in the image?

cevich:
> Just to link things together, https://github.com/containers/skopeo/pull/1739 is failing similarly.

Good, so at least we know it's not related at all to the package removal changes in @lsm5 PR.  The issue _must_ be due to an updated package.  That diff will be smaller in @lsm5 PR than this one, but still could be large.  I'm thinking the best approach is probably to go hands-on with `get_ci_vm.sh`, reproduce it. Then start pressing random buttons until it either explodes or starts working :grin:

cevich:
> Purely as a lazy guess, something from

Hmmm, so I wonder...maybe the difference is in the container image, not the VM image.

cevich:
Force-push: newer images.  **Not** expecting a different result.

cevich:
Force-push: newer images. Not expecting a different result.

mtrmac:
(FWIW, I realize that this is important work that will inevitably need to happen. At this point I’m prioritizing the sigstore feature work, given upcoming deadlines; let me know if that’s not the right judgement call.)

cevich:
> At this point I’m prioritizing the sigstore feature work

Please do, and ignore this for now.  I think I know what the next step[*] is to move this and @lsm5 PR forward.  I'll ping you (likely next year) if I need help.

[*] Use get_ci_vm.sh and reproduce the failure, and try to expose what changed in the environment to cause it.  Specifically, paying attention to runc/crun stuffs.

cevich:
Now this is weird...when I run the system tests manually via `get_ci_vm.sh Skopeo Test` they all pass.  The only significant difference I'm aware of is that I did **not** first run the integration tests.  I'll play with this setup more, to see if maybe the integration tests are somehow breaking the system tests...

cevich:
...I cannot reproduce the problem manually on a `get_ci_vm.sh`.  The tests pass no matter how or what order I run (int. vs. sys.) them.   This is bringing back some memories now.  IIRC Lokesh mentioned this fact to me last year about his PR.  Maybe I'll have some luck using the 're-run w/ terminal' button...

cevich:
...random idea: updated to images w/ older kernels (minus the multiple-port bindings bug).

cevich:
Hmmm, so that change simply kicked the can down the road.  It def. seems like the setup is wonky, but I cannot explain why running this stuff manually works perfectly fine.  That's what has me really puzzled.

cevich:
@lsm5 @mtrmac uh huh, well, I don't know how or why 0e94474 did it, but the tests are fixed now :grin: 

P.S. The commit message I wrote came directly out of the hole which I normally sit upon.  It's a complete guess - I was never able to reproduce the test failures manually :confused:

mtrmac:
It’s weird that `podman system reset` reports `A storage.conf file exists at /usr/share/containers/storage.conf` but then `podman run` seems to complain that the config file doesn’t contain the standard options.

It’s also weird that the `podman run … /bin/true` does _not_ contain the error messages that were plaguing us before, on the first run of a `podman run` command. Does either of `--runtime runc` (if used) and `--cgroup-manager=cgroupfs` make a difference in this behavior?

---

Either way… if there isn’t a way to investigate further, and this works, let’s merge.

Just, please, explicitly write down (in the `runner.sh` comments and the commit message) that we don’t know why this works / is necessary, and in one of the places document what failures this is working around. Documenting the hypothesis is useful as well, but to me the important thing is that in 3 years we will know not to treat that hypothesis as a certainty.

mtrmac:
> it's not the `system reset that's new, that's always been there.

True; I meant that it seems to be saying something about the configuration, as interpreted at that point.

cevich:
Force-push: Rebased.

PR #1822, opened 2022-12-07
Use common library reporter

cblecker:
This uses the common library reporter template generation, so that we can use the DefaultFuncs in format output: https://github.com/containers/common/blob/main/pkg/report/template.go#L37

### DIFF
--- a/cmd/skopeo/inspect.go
+++ b/cmd/skopeo/inspect.go
@@ -6,8 +6,6 @@ import (
 	"fmt"
 	"io"
 	"strings"
-	"text/tabwriter"
-	"text/template"
 
 	"github.com/containers/common/pkg/report"
 	"github.com/containers/common/pkg/retry"
@@ -74,6 +72,7 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error) 
 		rawManifest []byte
 		src         types.ImageSource
 		imgInspect  *types.ImageInspectInfo
+		rpt         *report.Formatter
 		data        []any
 	)
 	ctx, cancel := opts.global.commandTimeoutContext()
@@ -158,9 +157,13 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error) 
 				fmt.Fprintf(stdout, "%s\n", string(out))
 			}
 		} else {
-			row := "{{range . }}" + report.NormalizeFormat(opts.format) + "{{end}}"
+			rpt, err = report.New(stdout, "skopeo inspect").Parse(report.OriginUser, opts.format)
+			if err != nil {
+				return err
+			}
+			defer rpt.Flush()
 			data = append(data, config)
-			err = printTmpl(stdout, row, data)
+			err = rpt.Execute(data)
 		}
 		if err != nil {
 			return fmt.Errorf("Error writing OCI-formatted configuration data to standard output: %w", err)
@@ -235,19 +238,12 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error) 
 		}
 		return err
 	}
-	row := "{{range . }}" + report.NormalizeFormat(opts.format) + "{{end}}"
-	data = append(data, outputData)
-	return printTmpl(stdout, row, data)
-}
 
-func printTmpl(stdout io.Writer, row string, data []any) error {
-	t, err := template.New("skopeo inspect").Parse(row)
+	rpt, err = report.New(stdout, "skopeo inspect").Parse(report.OriginUser, opts.format)
 	if err != nil {
 		return err
 	}
-	w := tabwriter.NewWriter(stdout, 8, 2, 2, ' ', 0)
-	if err := t.Execute(w, data); err != nil {
-		return err
-	}
-	return w.Flush()
+	defer rpt.Flush()
+	data = append(data, outputData)
+	return rpt.Execute(data)
 }
--- a/cmd/skopeo/inspect.go
+++ b/cmd/skopeo/inspect.go
@@ -51,8 +51,8 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
 `, strings.Join(transports.ListNames(), ", ")),
 		RunE: commandAction(opts.run),
 		Example: `skopeo inspect docker://registry.fedoraproject.org/fedora
-  skopeo inspect --config docker://docker.io/alpine
-  skopeo inspect  --format "Name: {{.Name}} Digest: {{.Digest}}" docker://registry.access.redhat.com/ubi8`,
+skopeo inspect --config docker://docker.io/alpine
+skopeo inspect --format "Name: {{.Name}} Digest: {{.Digest}}" docker://registry.access.redhat.com/ubi8`,
 		ValidArgsFunction: autocompleteSupportedTransports,
 	}
 	adjustUsage(cmd)
--- a/docs/skopeo-inspect.1.md
+++ b/docs/skopeo-inspect.1.md
@@ -42,6 +42,7 @@ Use docker daemon host at _host_ (`docker-daemon:` transport only)
 
 Format the output using the given Go template.
 The keys of the returned JSON can be used as the values for the --format flag (see examples below).
+Supports the Go templating functions available at https://pkg.go.dev/github.com/containers/common/pkg/report#hdr-Template_Functions
 
 **--help**, **-h**
 


mtrmac (changes requested):
Thanks. Is this just for the extra functions?
- If so, why not just use `DefaultFuncs` directly? The other manipulations of the format string are not obviously desirable or harmless. (They might well be an improvement, but this PR doesn’t make an argument either way.)
- The new functions usable in the format string need to be documented _somehow_, otherwise will not be something users can rely on. That might well be a link to some other documentation that already exists, just not _nothing_.

cblecker:
@mtrmac The goal here was to have output filters from `podman images inspect --format` work the same as `skopeo inspect --format`. I figured the fact that there was a common reporting library for use by the `github.com/containers` tools, and that other functions from that library are in use, that this was just a bug/oversight to not use this function.

The docs don't really outline these custom functions either way:
https://docs.podman.io/en/latest/markdown/podman-image-inspect.1.html
https://github.com/containers/skopeo/blob/main/docs/skopeo-inspect.1.md

The godoc has some documentation on it: https://pkg.go.dev/github.com/containers/common@v0.50.1/pkg/report

I'm happy to add additional documentation somewhere if you'd like.

mtrmac:
It’s more that it was never discussed (the history is https://github.com/containers/skopeo/pull/1070 ); so now we essentially get do define what needs to happen.

Speaking for Skopeo, “Go templates as they existing the Go standard library” is a simple story to tell (although that’s not quite what Skopeo does, sadly). Any other features do need to be documented _somehow_.

Linking to `pkg.go.dev` might be enough — if the c/common package at least formatted the documentation enough to allow linking to a specific section, and the list was a _bit_ formatted; using the Go 1.19 documentation comment formatting might allow both of those, I’m not sure.

cblecker:
Works for me.

I opened https://github.com/containers/common/pull/1260 to get some deep links available on pkg.go.dev. Should I add the local details of this to https://github.com/containers/skopeo/blob/main/docs/skopeo-inspect.1.md ?

mtrmac:
Thanks!

At the very least a link to the resulting page on `pkg.go.dev` would be nice. Alternatively, just duplicating the list of functions in a new section of the the `skopeo inspect` man page would work too — I guess the link to `pkg.go.dev` would be better, because it would be kept up-to-date.

cblecker:
@mtrmac Updated PR with docs.

The anchor in the link won't work until https://github.com/containers/common/pull/1260 is merged, but the format of it will be similar to https://pkg.go.dev/encoding/gob#hdr-Encoding_Details

TomSweeneyRedHat:
@cblecker looks like you need a rebase here.

cblecker:
@TomSweeneyRedHat Thanks for the ping. Rebased.

TomSweeneyRedHat:
LGTM
thanks @cblecker !

cblecker:
@mtrmac @TomSweeneyRedHat I've resolved the merge conflict again. Is it possible to get this merged?

TomSweeneyRedHat:
LGTM
@mtrmac ?

rhatdan:
LGTM.

One of the things that @edsantiago has done in Podman is to a table to help users with --format options.

```
  --format, -f=format
       Format the output using the given Go template.  The  keys  of  the  re‐
       turned  JSON can be used as the values for the --format flag (see exam‐
       ples below).

       Valid placeholders for the Go template are listed below:

       ┌─────────────────────┬───────────────────────────────┐
       │Placeholder          │ Description                   │
       ├─────────────────────┼───────────────────────────────┤
       │.AppArmorProfile     │ AppArmor profile (string)     │
       ├─────────────────────┼───────────────────────────────┤
       │.Args                │ Command-line arguments (array │
       │                     │ of strings)                   │
       ├─────────────────────┼───────────────────────────────┤
       │.BoundingCaps        │ Bounding  capability set (ar‐ │
       │                     │ ray of strings)               │
       ├─────────────────────┼───────────────────────────────┤
       │.Config ...          │ Structure with config info    │
       ├─────────────────────┼───────────────────────────────┤
       │.ConmonPidFile       │ Path to file containing  con‐ │
       │                     │ mon pid (string)              │
       ├─────────────────────┼───────────────────────────────┤
       │.Created             │ Container    creation    time │
       │                     │ (string, ISO3601)             │
       ├─────────────────────┼───────────────────────────────┤
       │.Dependencies        │ Dependencies    (array     of │
       │                     │ strings)                      │
```

docs/skopeo-inspect.1.md
```diff
@@ -42,6 +42,7 @@ Use docker daemon host at _host_ (`docker-daemon:` transport only)
 
 Format the output using the given Go template.
 The keys of the returned JSON can be used as the values for the --format flag (see examples below).
+Supports the templating functions available here: https://pkg.go.dev/github.com/containers/common/pkg/report#hdr-Template_Functions
```
TomSweeneyRedHat, line 45: Soft suggestion: 
```suggestion
Supports the Go templating functions available [here](https://pkg.go.dev/github.com/containers/common/pkg/report#hdr-Template_Functions).
```

TomSweeneyRedHat:
@cblecker sorry, had another suggestion for consideration.  Regardless if you take it or not, LGTM.

docs/skopeo-inspect.1.md
cblecker: Done!

TomSweeneyRedHat:
LGTM
and happy green test buttons!

docs/skopeo-inspect.1.md
mtrmac: FWIW using a Markdown link like that looks worse in a man page (`make docs/skopeo-inspect.1 && man docs/skopeo-inspect.1`):
> functions available here ⟨https://pkg.go.dev/github.com/containers/common/pkg/report#hdr-Template_Functions⟩.

So just an “available at https://…” might be a reasonable compromise, at least GitHub’s Markdown rendering turns the URL into a link without extra syntax.

cblecker: @mtrmac Updated per your suggestion

mtrmac:
> @mtrmac The goal here was to have output filters from `podman images inspect --format` work the same as `skopeo inspect --format`.

Well, that’s not actually what this PR does; Podman uses `pkg/report.New`, not `pkg/report.NewTemplate`. The two differ in handling of `\n` directives, and tabulating output using `tabwriter`.


Now… as it happens, the _net_ effect of this PR is actually that `pkg/report.NewTemplate().Parse()`… is equivalent to not using `NewTemplate` at all (because existing callers of `printTmpl` both use `NormalizeFormat` first); it ends up having the same result as a `template.New(…).Funcs(report.DefaultFuncs).Parse()`. And the latter is clearly backwards-compatible; so I propose doing just that.

Alternatively, Skopeo could be moved to use the `pkg.Formatter` from `pkg/report.New` in full; AFAICS that would be an incompatible change in principle, but not in practice; it differs for
- formats that start with `{{range .}}`, which don’t work with the data we print anyway
- for formats with `\t`, which are currently completely broken.

The version in this PR right now, using a function from `pkg/report` that _nothing else uses_ (any more, AFAICS), is not very attractive.

mtrmac:
> * for formats with `\t`, which are currently completely broken.

#1921, for the record.

mtrmac:
… actually, looking at the history, as of the original #1070 Skopeo’s `inspect` format handling really was essentially equivalent to Podman’s; it’s Podman that was updated, leaving Skopeo behind.

So updating to catch up, and to use `report.New`, might well make sense. (In a sense, we got very lucky with the #1921 bug, in that there is no compatibility tradeoff.)

mtrmac:
> looking at the history, as of the original #1070 Skopeo’s `inspect` format handling really was essentially equivalent to Podman’s

https://github.com/containers/podman/blob/v2.2.0/cmd/podman/inspect/inspect.go , so that I don’t need to look that up again

cblecker:
Thank you for the detailed review, @mtrmac. That totally makes sense.

It looks like https://github.com/containers/podman/pull/15673 was where they made changes to how reports were done in podman, and I pulled those changes over. Compiled locally, and it seems to work as intended.

cmd/skopeo/inspect.go
```diff
@@ -158,9 +157,14 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
 				fmt.Fprintf(stdout, "%s\n", string(out))
 			}
 		} else {
-			row := "{{range . }}" + report.NormalizeFormat(opts.format) + "{{end}}"
+			format := report.NormalizeFormat(opts.format)
```
mtrmac, line 160: This isn’t necessary; `Formatter.Parse` already calls `NormalizeFormat`. In fact doing so here explicitly interferes with the `\n` logic of https://github.com/containers/podman/pull/15673 , and the `table ` feature.

(in both cases)

mtrmac (commented):
Thanks!

One last change, please.

cmd/skopeo/inspect.go
cblecker: @mtrmac Updated!

mtrmac (approved):
LGTM. Thanks!

PR #1825, opened 2022-12-13
proxy: Ensure images are closed when proxy is shutting down

cgwalters:
This is a complementary fix for
https://github.com/coreos/rpm-ostree/issues/4213

Basically in the case of `oci-archive` we have a temporary directory that needs cleanup.

### DIFF
--- a/cmd/skopeo/proxy.go
+++ b/cmd/skopeo/proxy.go
@@ -75,12 +75,14 @@ import (
 	"github.com/containers/image/v5/manifest"
 	ocilayout "github.com/containers/image/v5/oci/layout"
 	"github.com/containers/image/v5/pkg/blobinfocache"
+	"github.com/containers/image/v5/transports"
 	"github.com/containers/image/v5/transports/alltransports"
 	"github.com/containers/image/v5/types"
 	dockerdistributionerrcode "github.com/docker/distribution/registry/api/errcode"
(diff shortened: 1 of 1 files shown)


vrothberg (approved):
LGTM
@mtrmac PTAL

cmd/skopeo/proxy.go
```diff
@@ -701,6 +702,17 @@ func (h *proxyHandler) FinishPipe(args []interface{}) (replyBuf, error) {
 	return ret, err
 }
 
+// close releases all resources associated with this proxy backend
+func (h *proxyHandler) close() {
+	for _, image := range h.images {
+		err := image.src.Close()
+		if err != nil {
+			// This shouldn't be fatal
+			logrus.Warnf("failed to close image: %v", err)
```
mtrmac, line 711: Non-blocking:
```suggestion
			logrus.Warnf("Failed to close image %s: %v", transports.ImageName(image.Reference()), err)
```
in case there are multiple images; but this should basically never happen anyway…

… and `CloseImage` is not reporting the error at all, so focusing on this case is not really warranted.

mtrmac (commented):
LGTM, do you want to merge as is or update?

mtrmac (approved):

mtrmac:
Thanks!

PR #1833, opened 2022-12-18
install.md: add Windows static build example

christian-korneck:
This adds an example to `install.md` how to build a static `skopeo.exe` on Windows. Remote registry functionality seems to be working fine on Windows.

Examples where this might be useful:
- There aren't many options to run Windows containers, one of the more popular ones is to use Docker Desktop which can either be in Linux or Windows mode. In Windows mode a Skopeo Linux container can't be used.
- when only using a Windows _Docker_ or _containerd_ build (which can only run Windows containers, so again no Skopeo Linux container).

### DIFF
--- a/install.md
+++ b/install.md
@@ -245,6 +245,18 @@ docker run -v $PWD:/src -w /src -e CGO_ENABLED=0 golang \
 make BUILDTAGS=containers_image_openpgp GO_DYN_FLAGS=
 ```
 
+On Windows something like this might work (even more unsupported):
+
+```batch
+REM get git ref
+@echo off && for /f "tokens=*" %i in ('git rev-parse --short HEAD') do set skgitref=%i && @echo on
+REM build bin\skopeo.exe
+set CGO_ENABLED=0
+set GOOS=windows
+go build -mod=vendor -buildmode=pie -ldflags "-X main.gitCommit=%skgitref%" -gcflags "" -tags "containers_image_openpgp" -o ./bin/skopeo.exe ./cmd/skopeo
+```
+
+
 Keep in mind that the resulting binary is unsupported and might crash randomly. Only use if you know what you're doing!
 
 For more information, history, and context about static builds, check the following issues:


TomSweeneyRedHat:
Looks OK to me, but I don't have a Windows box that I can easily test this on.

mtrmac (changes requested):
Thanks!

I don’t think it makes much sense to me to provide a `for /f …` script in the middle of installation instructions, _only_ in the non-recommended static case.

Most of #715 likely still applies — the default paths and the like are probably not a good fit for Windows. It seems to me that looking into the functionality and integration should happen before widely publicizing a way to install the binary.

I’m not at all sure what the functional status of Skopeo on Windows is, but I think we need to figure that out first (and to see what we can or can not document as working), before deciding whether it even makes sense to document installation, let alone a specific installation method.

---

If we are to have Windows installation instructions, I’d expect the documentation to first have Windows instructions for the “recommended” approach, in some long-term maintainable form that isn’t _too_ likely to break without anyone noticing.

And then the non-recommended approach (if it needs to be documented at all) should, to the extent possible, defer to the recommended-approach tooling.

#715 seems to have settled on using `mingw32-make` for that purpose (and that would naturally extend to just using `make … CGO_ENABLED=0 … ` for the “static build” case); I have no idea whether that is practical, but it at least seems less likely to break than a Windows-specific script that only exists as text within a document.

---

Also, placing this text in the middle of fairly Linux-specific text talks would make the impression that the Windows build have similar properties; e.g. the random crashes due to ABI incompatibilities might not be a concern on Windows (or at least not as much as on Linux?), and the links about context are not relevant to Windows either.

christian-korneck:
@mtrmac Thanks for the feedback.

> Most of https://github.com/containers/skopeo/issues/715 likely still applies — the default paths and the like are probably not a good fit for Windows. It seems to me that looking into the functionality and integration should happen before widely publicizing a way to install the binary.

Thanks for mentioning this, I wasn't aware of #715. I'm closing this PR as there's already Windows work happening.

> I don’t think it makes much sense to me to provide a for /f … script in the middle of installation instructions, _only_ in the non-recommended static case.

This isn't important going forward, just for completeness if someone ends up reading this thread: This was _only_ about making a static build. The code block just repeats the Linux static build instructions from above in a win `cmd` shell (essentially just a `go build ...` line and some some env vars). Intended as contrib info, not install instructions.

PR #1841, opened 2023-01-05
[release-1.4] [CI:BUILD] Cirrus: Migrate OSX task to M1

lsm5:
Migrate our OSX build to a M1 instance, since Cirrus is sunsetting Intel-based macOS instances.

Signed-off-by: Ashley Cui <acui@redhat.com>
(cherry picked from commit b5ac534960bd4188f7fd847cec3225f55714abc4)

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -98,7 +98,7 @@ osx_task:
         brew install gpgme go@1.16 go-md2man
         go get -u golang.org/x/lint/golint
     test_script: |
-        export PATH=$GOPATH/bin:/usr/local/opt/go@1.16/bin:$PATH
+        export PATH=$GOPATH/bin:/usr/local/opt/go@1.16/bin:/opt/homebrew/opt/go@1.16/bin:$PATH
         go version
         go env
         make validate-local test-unit-local bin/skopeo
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -93,12 +93,12 @@ osx_task:
         # /usr/local/opt/go@1.16 will be populated by (brew install go@1.16) below
         # /opt/homebrew is the expected path for anything installed with brew
         # on M1 macs
-        export PATH=$GOPATH/bin:/usr/local/opt/go@1.16/bin:/opt/homebrew/opt/go@1.16/bin:$PATH
+        export PATH=$GOPATH/bin:/usr/local/opt/go@1.16/bin:/opt/homebrew/opt/go@1.16/bin:/opt/homebrew/bin:$PATH
         brew update
         brew install gpgme go@1.16 go-md2man
         go get -u golang.org/x/lint/golint
     test_script: |
-        export PATH=$GOPATH/bin:/usr/local/opt/go@1.16/bin:/opt/homebrew/opt/go@1.16/bin:$PATH
+        export PATH=$GOPATH/bin:/usr/local/opt/go@1.16/bin:/opt/homebrew/opt/go@1.16/bin:/opt/homebrew/bin:$PATH
         go version
         go env
         make validate-local test-unit-local bin/skopeo
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -88,10 +88,12 @@ osx_task:
     depends_on:
         - validate
     macos_instance:
-        image: catalina-xcode
+        image: ghcr.io/cirruslabs/macos-ventura-base:latest
     setup_script: |
         # /usr/local/opt/go@1.16 will be populated by (brew install go@1.16) below
-        export PATH=$GOPATH/bin:/usr/local/opt/go@1.16/bin:$PATH
+        # /opt/homebrew is the expected path for anything installed with brew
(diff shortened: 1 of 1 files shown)


lsm5:
@edsantiago @mtrmac PTAL

mtrmac:
From the failing test:
>   echo 'export PATH="/opt/homebrew/opt/go@1.16/bin:$PATH"' >> ~/.profile

That does not match https://github.com/containers/skopeo/blob/022b6f54735ffbd6dd8363c83ed1eeb045dc1f5c/.cirrus.yml#L93-L94 , probably due to an intentional Homebrew change on M1: https://docs.brew.sh/FAQ#why-is-the-default-installation-prefix-opthomebrew-on-apple-silicon ).

Does changing the path to `/opt/homebrew` help? And if it does, we should probably update the more recent branches to also move to `/opt/homebrew`, or possibly (probably not?) drop that path override if it is completely unnecessary.

ashley-cui:
Oh yes, good catch, the path should be `opt/homebrew` for anything that is installed with brew on M1 macs

lsm5:
hmm, wonder why it can't find gpgme.h. Do the header file paths need to be changed for brew?

lsm5:
@ashley-cui do you know any place where I can see the list of files installed by brew? the homebrew page for gpgme doesn't show me anything (or maybe i'm looking in the wrong place)

ashley-cui:
AFAIK it shouldn't matter.. On main it works, but the path  is `export PATH=$GOPATH/bin:$PATH`, maybe try that?

lsm5:
> AFAIK it shouldn't matter.. On main it works, but the path is `export PATH=$GOPATH/bin:$PATH`, maybe try that?

The go: command not found is taken care of. But, now it's unable to find gpgme.h, would PATH handle that on brew?

.cirrus.yml
```diff
@@ -88,15 +88,17 @@ osx_task:
...
-        image: catalina-xcode
+        image: ghcr.io/cirruslabs/macos-ventura-base:latest
     setup_script: |
         # /usr/local/opt/go@1.16 will be populated by (brew install go@1.16) below
-        export PATH=$GOPATH/bin:/usr/local/opt/go@1.16/bin:$PATH
+        # /opt/homebrew is the expected path for anything installed with brew
+        # on M1 macs
+        export PATH=$GOPATH/bin:/usr/local/opt/go@1.16/bin:/opt/homebrew/opt/go@1.16/bin:$PATH
         brew update
         brew install gpgme go@1.16 go-md2man
```
ashley-cui, line 98: Weird since we install gpgme here using brew..

ashley-cui:
Oh yeah, path should handle that in brew, I think adding `/opt/homebrew/bin` to the path might do the trick

ashley-cui:
Weird that it's still failing, as the only difference from main is in main the path  is `export PATH=$GOPATH/bin:$PATH`. But the paths you've used should covere everything so it should all work, I'm also a little stumped here

lsm5:
> Weird that it's still failing, as the only difference from main is in main the path is `export PATH=$GOPATH/bin:$PATH`. But the paths you've used should covere everything so it should all work, I'm also a little stumped here

@ashley-cui  i guess this should fail the same way on main branch then. Let me check there ..

.cirrus.yml
```diff
@@ -88,15 +88,17 @@ osx_task:
...
         - validate
     macos_instance:
-        image: catalina-xcode
+        image: ghcr.io/cirruslabs/macos-ventura-base:latest
     setup_script: |
         # /usr/local/opt/go@1.16 will be populated by (brew install go@1.16) below
-        export PATH=$GOPATH/bin:/usr/local/opt/go@1.16/bin:$PATH
+        # /opt/homebrew is the expected path for anything installed with brew
+        # on M1 macs
+        export PATH=$GOPATH/bin:/usr/local/opt/go@1.16/bin:/opt/homebrew/opt/go@1.16/bin:/opt/homebrew/bin:$PATH
```
mtrmac, line 96: Per the other link, `/usr/local/` could possibly contain Intel binaries, and `/opt/homebrew` ARM ones, so `/opt/homebrew` should be preferred. (That might well not make any difference in practice.)

We don’t expect to go back to non-M1, so the `/usr/local/…` paths can maybe be dropped entirely.

(Perhaps modify the script for debugging purposes to `ls -la /usr/local/opt/go*/bin/* /opt/homebrew/opt/go*/bin/* /opt/homebrew/bin `; then we would see what we are dealing with in the failure logs.)

mtrmac (commented):
I don’t know what’s going on with the missing `gpgme`; maybe something like

```sh
set -x
which pkg-config
pkg-config --debug --libs gpgme
brew list gpgme
brew list pkg-config
```
could shed some light.

mtrmac:
Note to self: alternative to #1847.

cevich:
Closing in favor of https://github.com/containers/skopeo/pull/1851

Issue #1829, opened 2022-12-15
proxy: Add `OpenImageWithRequiredSignatures`


cgwalters:
In the bootc/ostree-container effort, I am trying to enforce signatures being enabled by default.  The thing is, we kind of say that e.g. `podman run <some image from docker hub or whatever>` is "secure" - in the sense I'm using the word, we can and do fix security problems we find (mostly in the kernel) in a relatively timely fashion.  

But *booting* a container (or running with `--privileged` as well as some more subtle options) completely change that.

As part of the proxy, I'd like to add an `OpenImageWithRequiredSignatures` API that requires that the remote image is signed in *some* way configured in `containers-policy.json` - IOW that the policy for fetching the image does not fall through to `insecureAcceptAnything`.

(I think it would make sense to also add `podman pull --sigpolicy=required` or so)

When I looked at this, it seemed feasible but would require some changes in c/image.  Let me know if you have any thoughts.

mtrmac:
Just to reply quickly, some unorganized thoughts:

My first thought: And would the caller of this API then have an override toggle to disable this, for debugging and troubleshooting? If so, is that so much different from allowing `insecureAcceptAnything` in the policy in the first place?

The RHEL default `policy.json` handles the Red Hat registries. Is the goal basically an extra safeguard against something/someone removing that?

… but then the OpenShift default is https://github.com/openshift/machine-config-operator/blob/9c6c2bfd7ed498bfbc296d530d1839bd6a177b0b/templates/worker/01-worker-container-runtime/_base/files/policy.yaml . So… that just wouldn’t work as is.

The specially-designed system that uses `bootc` (or OpenShift’s MCO) might be able to manage the `policy.json` to ensure there is an entry for the `bootc` repo. (Well… except that c/image doesn’t really have an “edit policy” API to modify an user-created file; there is an independent implementation in `podman image trust`).

For some use cases, running `skopeo --policy /some/bootc-specific/policy.json` might be appropriate, completely separating the “can sign the OS” and “can sign containers that run on the OS”. OTOH if this is supposed to just rely on the default vendor signatures (and just use a separate repo for those boot images), it does make a lot of sense to use the global policy instead of duplicating “is the vendor’s signature valid” policy in two places.

---

[...]

PR #1844, opened 2023-01-10
proxy: Add optional flags to `OpenImage`

cgwalters:
See https://github.com/containers/skopeo/issues/1829#issuecomment-1355220066

This is prep for adding another option like `require-signatures`.

### DIFF
--- a/cmd/skopeo/proxy.go
+++ b/cmd/skopeo/proxy.go
@@ -95,7 +95,8 @@ import (
 // 0.2.3: Added GetFullConfig
 // 0.2.4: Added OpenImageOptional
 // 0.2.5: Added LayerInfoJSON
-const protocolVersion = "0.2.5"
+// 0.2.6: OpenImage now accepts a second argument with an array of options
+const protocolVersion = "0.2.6"
 
 // maxMsgSize is the current limit on a packet size.
 // Note that all non-metadata (i.e. payload data) is sent over a pipe.
@@ -214,8 +215,18 @@ func (h *proxyHandler) Initialize(args []interface{}) (replyBuf, error) {
 
 // OpenImage accepts a string image reference i.e. TRANSPORT:REF - like `skopeo copy`.
 // The return value is an opaque integer handle.
+//
+// A second argument may be present; if so, it is an array of string-valued flags.
+//
+// - optional: Do not error if the image is not present; instead return a zero imageid.  Since: 0.2.6
 func (h *proxyHandler) OpenImage(args []interface{}) (replyBuf, error) {
-	return h.openImageImpl(args, false)
+	opts := openOptions{}
+	if len(args) > 1 {
+		if err := opts.parse(args[1]); err != nil {
+			return replyBuf{}, err
+		}
+	}
+	return h.openImageImpl(args, opts)
 }
 
 // isDockerManifestUnknownError is a copy of code from containers/image,
@@ -237,7 +248,32 @@ func isNotFoundImageError(err error) bool {
 		errors.Is(err, ocilayout.ImageNotFoundError{})
 }
 
-func (h *proxyHandler) openImageImpl(args []interface{}, allowNotFound bool) (replyBuf, error) {
+type openOptions struct {
+	optional bool
+}
+
+func (o *openOptions) parse(argsval interface{}) error {
+	args, ok := argsval.([]interface{})
+	if !ok {
+		return fmt.Errorf("expecting array for options, not %T", argsval)
+	}
+	for _, argv := range args {
+		arg, ok := argv.(string)
+		if !ok {
+			return fmt.Errorf("expecting string option, not %T", arg)
+		}
+		switch arg {
+		case "optional":
+			o.optional = true
+		default:
+			return fmt.Errorf("unknown option %s", arg)
+		}
+	}
+
+	return nil
+}
+
+func (h *proxyHandler) openImageImpl(args []interface{}, opts openOptions) (replyBuf, error) {
 	h.lock.Lock()
 	defer h.lock.Unlock()
 	var ret replyBuf
@@ -245,7 +281,12 @@ func (h *proxyHandler) openImageImpl(args []interface{}, allowNotFound bool) (re
 	if h.sysctx == nil {
 		return ret, fmt.Errorf("client error: must invoke Initialize")
 	}
-	if len(args) != 1 {
+	switch len(args) {
+	case 1:
+		// This is is the default
+	case 2:
+		// The second argument, if present should have been parsed by the caller
+	default:
 		return ret, fmt.Errorf("invalid request, expecting one argument")
 	}
 	imageref, ok := args[0].(string)
@@ -259,7 +300,7 @@ func (h *proxyHandler) openImageImpl(args []interface{}, allowNotFound bool) (re
 	}
 	imgsrc, err := imgRef.NewImageSource(context.Background(), h.sysctx)
 	if err != nil {
-		if allowNotFound && isNotFoundImageError(err) {
+		if opts.optional && isNotFoundImageError(err) {
 			ret.value = sentinelImageID
 			return ret, nil
 		}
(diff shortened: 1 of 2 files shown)


cmd/skopeo/proxy.go
```diff
@@ -237,15 +248,45 @@ func isNotFoundImageError(err error) bool {
...
 
 	if h.sysctx == nil {
 		return ret, fmt.Errorf("client error: must invoke Initialize")
 	}
-	if len(args) != 1 {
+	switch len(args) {
+	case 1:
+		// This is is the default
+	case 2:
+		// The second argument, if present should have been parsed by the caller
```
mtrmac, line 288: … but `OpenImageOptional` doesn’t do that…

I think enforcing `len(args)`, and turning `args` into better-typed data, should be collocated close to each other.

Probably in `OpenImage` (allowing `len(args) in {1,2}`) and `OpenImageOptional` (allowing `len(args) == 1`), calling already well-typed `openImageImpl(imageRef string, opts openOptions)`; or maybe have `openImageImpl` fully responsible for `args`, with the effect of `OpenImageOptional`  also allowing the new option argument. I’d prefer the former one but I don’t feel nearly as strongly about this choice as I do about having the the `len` check and the code that parses `argv` close.

cmd/skopeo/proxy.go
```diff
@@ -237,15 +248,45 @@ func isNotFoundImageError(err error) bool {
 		errors.Is(err, ocilayout.ImageNotFoundError{})
 }
 
-func (h *proxyHandler) openImageImpl(args []interface{}, allowNotFound bool) (replyBuf, error) {
+type openOptions struct {
+	optional bool
+}
+
+func (o *openOptions) parse(argsval interface{}) error {
+	args, ok := argsval.([]interface{})
```
mtrmac, line 256: I’d find `map[string]any` more natural here — it seems fairly likely we will want to pass non-boolean options in the future.

---

To go further, (preferably?), there probably is a way to use `json.Unmarshal` to deal with parsing the options into a Go struct, not writing any manual parser at all. Maybe something vaguely like
```go
type request struct {
	Method string `json:"method"`
	Args []interface{} `json:"args"`
	Options json.RawMessage
}

type requestTableEntry struct {
	handler func(options any /* sadly type-erased */ args []any) (replyBuf, error)
	allocOptions func() any
}
var requestTable map[string]requestTableEntry = {
	…
	"Open": requestTableEntry{
		handler: h.OpenImage, /* FIXME: this can't actually refer to h. here — refactor somehow */
		allocOptions: func() any { return &openOptions{} },
	},
}

…
func (h *proxyHandler) processRequest(readBytes []byte) (rb replyBuf, terminate bool, err error) {
	var req request, options any

	if err = json.Unmarshal(readBytes, &req); err != nil {
		err = fmt.Errorf("invalid request: %v", err)
		return
	}
	handler, ok := requestTable[req.Method]; if !ok { … }
	if handler.allocOptions != nil {
		options = handler.allocOptions()
		if err = json.Unmarshal(req.Options, options); err != nil { … }
	} else if req.Options != nil { /* fail */ }
	return handler.handler(options, req.Args)
}
```

or some even more elaborate option where the handers have well-typed, not type-erased `options`

… but that screams YAGNI.

mtrmac: … or, why the … not, we could architecture-astronaut this, and remove the `args`/`options` distinction, and all the manual casts

```go
type parsedRequest interface{
    argsDestinations() []any
   /* or have ^^^ even return (mandatoryArgs []any, optionalArgs []any), and then we could farm out validating `len(args)` to the shared `processRequest` entirely */
    run() (replyBuf, error)
}

type openImageRequest{
    imageName string
    opts openOptions /* a struct with named fields, automatically decoded by json.Unmarshal */
}

func (*r openImageRequest) argsDestinations() []any {
   return []any{&r.imageName, &r.opts}
}

func (*r openImageRequest) run() (replyBuf, error)
    /* use r.imageName, r.opts - no manual type checks or decoding */
}

var requestTable map[string]func ()*parsedRequest = {
   "OpenImage": func()*parsedRequest{return &openImageRequest{}
}

func (h *proxyHandler) processRequest(readBytes []byte) (rb replyBuf, terminate bool, err error) {
   /* unmarshal req.Args as []json.RawMessage */
   newRequestState, ok := requestTable[req.Method]; if !ok { … }
   rs := newRequestState()
   argsDest := rs.argsDestinations()
   if len(req.Args) > len(argsDest) { /* fail */ }
   for i := range req.Args {
      if err := json.Unmarshal(Args[i], argsDest[i]); err != nil { … } 
   }
   return rs.run()
}
```

Honestly I have no idea if this ends up actually shorter. There are not _that_ many operations that avoiding manual parsing of `args` is clearly a win, and the `parsedRequest` implementation are just another kind of boilerplate.

The above still feels clumsy, like I’m missing some obvious way to have the standard library do all the work.

(OTOH If we keep going like this, I’m sure we will eventually end up using something like Swagger…)

mtrmac: (`s/any/interface{}/g`, I think we don’t yet require Go 1.18, but that might change soon.)

mtrmac: > The above still feels clumsy, like I’m missing some obvious way to have the standard library do all the work.

One more thought: Get rid of the untyped `Args` entirely, use named parameters only (doesn’t help for any existing API with the `Args []any`)

```go
type sharedRequestHeader struct {
   Method string `json:"method"`
}

type parsedRequest interface{
    run() (replyBuf, error)
}

type openImageRequest{
    sharedRequestHeader   
    ImageName string `json:"imageName"`
    Opts openOptions `json:"opts"`
}

func (*r openImageRequest) run() (replyBuf, error)
    /* use r.imageName, r.opts - no manual type checks or decoding */
}

var requestTable map[string]func ()*parsedRequest = {
   "OpenImage2": func()*parsedRequest{return &openImageRequest{}
}

func (h *proxyHandler) processRequest(readBytes []byte) (rb replyBuf, terminate bool, err error) {
   var header sharedRequeestHeader
   if err := json.Unmarshal(readBytes, &header); err != nil { … }
   newRequestState, ok := requestTable[header.Method]; if !ok { … }
   rs := newRequestState()
   if err := json.Unmarshal(readBytes, rs); err != nil { … } // re-parse now that we know the request type
   return rs.run()
}
```

mtrmac: Doh, if we unmarshaled the whole request as a method-specific `struct`, we wouldn’t need the `openOptions` sub-struct at all (though it wouldn’t hurt).

cgwalters:
Thanks for the review!  I mentioned this in passing before but I have been vaguely mulling a switch to e.g. Cap'n Proto, which would obviate all of this.  Any opinions on that?  

If we did it in the obvious way it would involve sending the blob data over the primary channel, but maybe that's OK.  (As of lately cpnp does [support fd passing](https://github.com/capnproto/capnproto/pull/821), but support is not yet widespread in the bindings; at least not Rust which is mainly what I care about)

mtrmac:
You did mention that — it the website looks nice. I didn’t take a deep look and I don’t really have a strong opinion on RPC mechanisms, I know far too little about the various trade-offs.

https://capnproto.org/otherlang.html seems sufficient for interoperability.

I guess the main concern would be total binary size — that _should_ not be an issue in principle, but I’m just dealing with a Swagger generator that somehow adds a BSON encoder and a telemetry framework as non-avoidable dependencies.

---

I guess to be clearer: from the stream-of-consciousness thread https://github.com/containers/skopeo/pull/1844#discussion_r1066221998 , the only thing I’m really concerned is that the options argument should be a JSON object, not an array. Parsing that JSON object with a hand-crafted `parse` is, I think, good enough that it doesn’t warrant a complete refactoring.

The other things about that possible refactoring are not blocking this work. It’s just me slowly realizing that the `Args []any` approach is hard to automate…

mtrmac:
@cgwalters is this still relevant?

cgwalters:
> @cgwalters is this still relevant?

Eh...not worth the churn at the moment.

PR #1847, opened 2023-01-11
[release-1.4] [CI:BUILD] Cirrus: Migrate OSX task to M1

ashley-cui:
Signed-off-by: Ashley Cui <acui@redhat.com>

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -88,15 +88,15 @@ osx_task:
     depends_on:
         - validate
     macos_instance:
-        image: catalina-xcode
+        image: ghcr.io/cirruslabs/macos-ventura-base:latest
     setup_script: |
-        # /usr/local/opt/go@1.16 will be populated by (brew install go@1.16) below
-        export PATH=$GOPATH/bin:/usr/local/opt/go@1.16/bin:$PATH
+        export PATH="/opt/homebrew/opt/go@1.16/bin:$GOPATH/bin:$PATH"
+        echo GOPATH
         brew update
         brew install gpgme go@1.16 go-md2man
         go get -u golang.org/x/lint/golint
     test_script: |
-        export PATH=$GOPATH/bin:/usr/local/opt/go@1.16/bin:$PATH
+        export PATH="/opt/homebrew/opt/go@1.16/bin:$GOPATH/bin:$PATH"
         go version
         go env
         make validate-local test-unit-local bin/skopeo


ashley-cui:
Aha! Found out why it's failing. On M1, gpgme is installed in `/opt/homebrew` but the [Makefile](https://github.com/containers/skopeo/blob/022b6f54735ffbd6dd8363c83ed1eeb045dc1f5c/Makefile#L5) only looks for gpgme in `/usr/local` (the Intel path). We [removed this line in main](https://github.com/containers/skopeo/commit/ebc5573e83fc97d2896a619af6a19d2981c8de74), so that's why main builds correctly. Would the better solution be to link the files in Cirrus or to change the Makefile? @mtrmac , @cevich

mtrmac:
@ashley-cui AFAICS that doesn’t hard-code `/usr/local`: it executes a `gpgme-config` command which should be provided by the Brew-installed package.

Looking at `curl  -L -H "Authorization: Bearer QQ==" -o x.tar.gz https://ghcr.io/v2/homebrew/core/gpgme/blobs/sha256:09eb6844769020807f538ccd6ff6fd916a2428e7cbe495eab1d20c751849304e` (with the URL from `brew info --json gpgme`) suggests that `gpgme-config` is included, and should be using the right paths.

Just looking at the build output, I can’t understand how this fails:
- If `/opt/homebrew/bin` were missing from $PATH entirely, I’d expect much more breakage.
- If `/usr/local/bin/` were preferred to `/opt/homebrew/bin`, I wouldn’t be surprised, but why would there be an Intel `/usr/local/bin/gpgme-config`?

but that should still be possible to confirm for sure by printing the variables / contents.

---

Alternatively, we could just hard-code a working `GPGME_ENV`: `make GPGME_ENV=…`.

But AFAICS the Ventura package does include `gpgme-config`, so there should be away to make that work (assuming the output is not completely broken).

---

Backporting _just_ https://github.com/containers/skopeo/commit/ebc5573e83fc97d2896a619af6a19d2981c8de74 wouldn’t work alone; we would also need to update the `{proglottis,mtrmac}/gpgme` dependency to use `peg-config gpgme` instead of `gpgme-config`. That’s also possible but updating dependencies on stable branches is not very desirable.

mtrmac:
Note to self: alternative to #1841 .

cevich:
@ashley-cui and @mtrmac with my CI-maintainer hat on: My preference is to just disable this test for non-main branches, taking any lessons-learned into fix PRs as needed.  Deleting the YAML in the release branches is fine by me.

To preserve this policy going forward, an update on `main` could be made like:

```
osx_task:
    # Run for regular PRs and those with [CI:BUILD] but not [CI:DOCS]
    only_if: &not_docs_multiarch >-
        $DEST_BRANCH == 'main' &&
        $CIRRUS_CHANGE_TITLE !=~ '.*CI:DOCS.*' &&
        $CIRRUS_CRON != 'multiarch'
```

I can open a PR for that...easy enough.

ashley-cui:
I think that might be for the best, thanks @cevich

cevich:
Opened https://github.com/containers/skopeo/pull/1850

PR #1848, opened 2023-01-11
Correctly use the stdout parameter in some places

mtrmac:
Should not change behavior - it would matter for unit tests which don't exist.

Also, `promptForPassphrase` must continue to hard-code "real" `os.Stdin` and `os.Stdout`.

So, this is just a low-priority pedantic cleanup.


TomSweeneyRedHat:
LGTM

rhatdan:
LGTM

PR #1850, opened 2023-01-12
Cirrus: Skip OSX CI on release-branches

cevich:
This task does not make sense to maintain long-term on release
branches.  Its intent is always/only to test the latest/greatest code
and environment.  After release, it's simply too difficult to maintain
functioning CI with a constantly changing (Cirrus-managed) OSX environment.
Ensure the task only runs for PRs targeted at the default branch, or if
the current branch is the default branch.

See also:
* https://github.com/containers/skopeo/pull/1851
* https://github.com/containers/skopeo/pull/1852

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -75,8 +75,12 @@ doccheck_task:
       "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" doccheck
 
 osx_task:
-    # Run for regular PRs and those with [CI:BUILD] but not [CI:DOCS]
-    only_if: &not_docs_multiarch >-
+    # Don't run for docs-only or multi-arch image builds.
+    # Also don't run on release-branches or their PRs,
+    # since base container-image is not version-constrained.
+    only_if: &not_docs_or_release_branch >-
+        ($CIRRUS_BASE_BRANCH == $CIRRUS_DEFAULT_BRANCH ||
+         $CIRRUS_BRANCH == $CIRRUS_DEFAULT_BRANCH ) &&
         $CIRRUS_CHANGE_TITLE !=~ '.*CI:DOCS.*' &&
         $CIRRUS_CRON != 'multiarch'
     depends_on:
@@ -99,7 +103,9 @@ osx_task:
 
 cross_task:
     alias: cross
-    only_if: *not_docs_multiarch
+    only_if: >-
+        $CIRRUS_CHANGE_TITLE !=~ '.*CI:DOCS.*' &&
+        $CIRRUS_CRON != 'multiarch'
     depends_on:
         - validate
     gce_instance: &standardvm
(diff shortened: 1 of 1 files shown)


cevich:
***Note:*** The 'cross' task is getting pulled in to the skip-on-release because of a YAML-alias.  I can remove that if you think maintaining cross-compile-ability on release branches is important (I don't have a strong opinion and only a weak argument).

.cirrus.yml
```diff
@@ -76,7 +76,10 @@ doccheck_task:
 
 osx_task:
     # Run for regular PRs and those with [CI:BUILD] but not [CI:DOCS]
+    # Skip entirely for release-branches (including PRs for them).
     only_if: &not_docs_multiarch >-
```
mtrmac, line 80: There already is a `not_docs_or_release_branch`, with exactly the right contents if my tired eyes don’t deceive me.

So would it make sense to use that one, and keep the current `cross_task` condition?

mtrmac (commented):
ACK to the idea, per email conversation.

.cirrus.yml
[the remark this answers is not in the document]
cevich: Okay, even if weak, it's a stronger opinion than mine.  I'll re-enable it.

cevich: Durr, yep, my bad.  Good catch.

cevich:
Force-push: Restored the cross task to run as it did previously (including on release-branches).  Updated commit and PR description accordingly.

.cirrus.yml
```diff
@@ -75,8 +75,12 @@ doccheck_task:
       "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" doccheck
 
 osx_task:
-    # Run for regular PRs and those with [CI:BUILD] but not [CI:DOCS]
-    only_if: &not_docs_multiarch >-
+    # Don't run for docs-only or multi-arch image builds.
+    # Also don't run on release-branches or their PRs,
+    # since base container-image is not version-constrained.
+    only_if: &not_docs_or_release_branch >-
```
mtrmac, line 81: There are now two definitions of `not_docs_or_release_branch`. Is that necessary?

.cirrus.yml
```diff
@@ -99,7 +103,9 @@ osx_task:
 
 cross_task:
     alias: cross
-    only_if: *not_docs_multiarch
+    only_if: |
```
mtrmac, line 106: (I have no idea if the difference between `|` and `>-` matters here, just noting the change.)

.cirrus.yml
cevich: Ugh, nope not good, it's a bug.  I'll fix.

.cirrus.yml
cevich: Doesn't matter much in this context.  Both are valid but I'll put back the old one since it's slightly preferred.

FWIW: `>-` means "fold" the next indented block into a single-line string, replacing newlines with ' '.  Whereas `|` does the same thing, but keeps the newlines.

cevich:
oof more yaml problems...time to actually get my pre-commit setup properly on this repo...

cevich:
...there we go.

mtrmac (approved):
LGTM. Thanks!

PR #1862, opened 2023-01-19
Cirrus: Fix c/image CI testing

cevich:
The containers/image CI setup reuses the runner script from this repo to execute the skopeo tests.  However, an env. var. is being taken out of context in that environment, leading to failure.  Fix this by hard-coding an image-name which will always be available in both environments.


mtrmac (approved):
Thanks!

Issue #1222, opened 2021-03-11
copy_test.go failures outside a specific container image in --privileged mode


cevich:
At the time of filing this issue, the skopeo integration tests cannot pass outside [a `--privileged` quay.io auto-build image](https://quay.io/repository/skopeo/ci).  A dozen or so `copy_test.go` items fail similar to:

```
FAIL: copy_test.go:187: CopySuite.TestCopyWithManifestListStorage
Running skopeo copy docker://registry.fedoraproject.org/fedora-minimal containers-storage:[vfs@/tmp/copy-manifest-list-storage789817936/root+/tmp/copy-manifest-list-storage789817936/runroot]test
copy_test.go:198:
    assertSkopeoSucceeds(c, "", "copy", knownListImage, "containers-storage:"+storage+"test")
utils.go:62:
    c.Assert(err, check.IsNil, check.Commentf("%s", out))
... value *exec.ExitError = &exec.ExitError{ProcessState:(*os.ProcessState)(0xc0004c2a40), Stderr:[]uint8(nil)} ("exit status 1")
... time="2021-03-04T17:00:15Z" level=warning msg="error reading allowed ID mappings: error reading subuid mappings for user \"root\" and subgid mappings for group \"root\": No subuid ranges found for user \"root\" in /etc/subuid"
time="2021-03-04T17:00:15Z" level=warning msg="Found no UID ranges set aside for user \"root\" in /etc/subuid."
time="2021-03-04T17:00:15Z" level=warning msg="Found no GID ranges set aside for user \"root\" in /etc/subgid."
Getting image source signatures
Copying blob sha256:2fa61fedb54d576e17d9129a27fbd3c1ff8503b1e0c45622ba8de6a51fb6a9ef
Copying config sha256:fa011f8784baff6b77fc56152b5024c368809c0f4c6b1279dbd9b173f534028a
Writing manifest to image destination
Storing signatures
time="2021-03-04T17:00:19Z" level=error msg="Error while applying layer: ApplyLayer exit status 1 stdout:  stderr: potentially insufficient UIDs or GIDs available in user namespace (requested 0:12 for /var/spool/mail): Check /etc/subuid and /etc/subgid: lchown /var/spool/mail: invalid argument"
[...]

PR #1864, opened 2023-01-19
Fix storage.conf overrides in test-system in CI, update c/storage

mtrmac:
<s>A wild guess at what is happening in #1863 .</s> See below.


mtrmac:
This
- Fixes how `make test-system` unnecessarily overrides `storage.conf` in CI.
- Attempts to fix, apparently long-unused, `make test-integration` and `make test-system` running in containers, for developers’ on-workstation use. At least it fixes the failure in #1222, but I didn’t yet actually get this _working_. But apparently it’s not being used, so that’s not nearly as important as the CI.
- Updates c/storage to 1.45.3, the version that is failing without the changes, as a demonstration.

@rhatdan PTAL.

@cevich FYI (non-urgent), this points at a part of the reason of the recent Podman trouble - changing `storage.conf` under a pre-existing Podman database seems to make it unhappy. Now that we don’t do that, that workaround is no longer necessary.

mtrmac:
> Attempts to fix, apparently long-unused, `make test-integration` and `make test-system` running in containers, for developers’ on-workstation use.

This does fix `make test-integration`. `make test-system` does not work for me — there is no reported error, something about the Podman-in-Podman networking seems to just hang when connecting to the registry container. I didn’t investigate that further.

rhatdan (approved):
LGTM
@vrothberg PTAL

vrothberg (approved):
/lgtm

cevich:
> @cevich FYI (non-urgent), this points at a part of the reason of the recent Podman trouble - changing storage.conf under a pre-existing Podman database seems to make it unhappy. Now that we don’t do that, that workaround is no longer necessary.

Oh wow, and so you got rid of the "magic" workaround too, fantastic!  Thanks!

PR #1868, opened 2023-01-20
Fix `make test-system` when run as an unprivileged user (containerized)

mtrmac:
This, _together with #1864_, allows running `make test-system` as an unprivileged user. (It doesn’t work for me as `root` and I don’t currently think it’s worth spending time on that.)

That makes the “single Linux developer” use case working again… in case anyone cares.


rhatdan:
LGTM
@vrothberg PTAL

rhatdan (approved):

vrothberg (approved):
LGTM

PR #1894, opened 2023-02-02
Update to Go 1.18

mtrmac:
We already depend on Go 1.18, due to:
```
vendor/github.com/rivo/uniseg/properties.go:func propertySearch[E interface{ [3]int | [4]int }](dictionary []E, r rune) (result E) {
```

So, actually update Skopeo proper, and take advantage of the new features.


vrothberg (approved):
LGTM, nice!

PR #1900, opened 2023-02-09
Run codespell on codebase

rhatdan:

### DIFF
--- a/cmd/skopeo/generate_sigstore_key_test.go
+++ b/cmd/skopeo/generate_sigstore_key_test.go
@@ -62,7 +62,7 @@ func TestGenerateSigstoreKey(t *testing.T) {
 	// we have to trigger a write failure.
 
 	// Success
-	// Just a smoke-test, useability of the keys is tested in the generate implementation.
+	// Just a smoke-test, usability of the keys is tested in the generate implementation.
 	dir := t.TempDir()
 	prefix := filepath.Join(dir, "prefix")
 	passphraseFile := filepath.Join(dir, "passphrase")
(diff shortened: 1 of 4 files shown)


mtrmac (approved):
Thanks!

edsantiago:
Interesting. Today's flake *looks* like the unexpected-EOF one, but is 500 instead:
```
time="2023-02-09T12:19:25-06:00" level=fatal msg="copying system image from manifest list:
      writing manifest: uploading manifest unsigned to localhost:5006/myns/mirroring-primary: received
     unexpected HTTP status: 500 Internal Server Error"
```

mtrmac:
That’s unrelated, that is a Skopeo-CI-specific problem. `localhost:5006` is the OpenShift integrated registry from a _very old_ version of OpenShift we run (all as binaries within a single container).

The registry sometimes fails this way, and I never got around to trying to debug it.

(The whole idea of running an old version of OpenShift only to test the very-deprecated `atomic:` transport could be questioned.)

rhatdan:
Can we just drop the atomic transport?

mtrmac:
> Can we just drop the atomic transport?

I’m not at all a fan of breaking c/image API.

Either way, that would just break CI without removing the relevant tests. Dropping that part of the integration tests is a bit more attractive (IIRC it might fairly notably speed them up), and we could do that without actually removing the code.

mtrmac:
_Now_ we get `happened during read: unexpected EOF`. Time to actually update c/image here…

mtrmac:
>  Time to actually update c/image here…

https://github.com/containers/skopeo/pull/1901 for the record.

PR #1902, opened 2023-02-09
[release-1.11] Update to c/image 5.24.1

mtrmac:
... to include an unexpected EOF workaround.

(I have also created the `release-1.11` branch.)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.17
 
 require (
 	github.com/containers/common v0.51.0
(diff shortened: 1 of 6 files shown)


vrothberg (commented):
LGTM
Do you want to bump the version here directly?

mtrmac:
My first instinct is to wait for a coordinated update / vendor dance, if there will be one. But I don’t really care at all, there are enough integers to make another release afterwards if necessary.

PR #1905, opened 2023-02-13
[CI:BUILD] Packit: initial enablement

lsm5:
This commit will run COPR builds on every PR against all active releases of CentOS Stream and Fedora, thus allowing buildability checks before the PR merges.

Builds are done on a custom COPR project:
`rhcontainerbot/packit-builds`.
Ref: https://copr.fedorainfracloud.org/coprs/rhcontainerbot/packit-builds/

The build targets are set in the copr itself, so we don't need to explicitly mention them in `.packit.yaml`, making upstream configuration a lot simpler.

The `spec.rpkg` file meant for rpm builds post-pr-merge at `rhcontainerbot/podman-next` copr gets reused for packit builds, so the packit jobs are independent of Fedora / CentOS dist-git.

NOTE: The Packit copr_build tasks help to check if every commit builds on
supported Fedora and CentOS Stream arches. They do not block the current
Cirrus-based workflow.

### DIFF
--- a/.packit.sh
+++ b/.packit.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+# This script handles any custom processing of the spec file generated using the `post-upstream-clone`
+# action and gets used by the fix-spec-file action in .packit.yaml.
+
+set -eo pipefail
+
+# Get Version from HEAD
+VERSION=$(grep '^const Version' version/version.go | cut -d\" -f2 | sed -e 's/-/~/')
+
+# Generate source tarball
+git archive --prefix=skopeo-$VERSION/ -o skopeo-$VERSION.tar.gz HEAD
+
+# RPM Spec modifications
+
+# Update Version in spec with Version from Cargo.toml
+sed -i "s/^Version:.*/Version: $VERSION/" skopeo.spec
+
+# Update Release in spec with Packit's release envvar
+sed -i "s/^Release:.*/Release: $PACKIT_RPMSPEC_RELEASE%{?dist}/" skopeo.spec
+
+# Update Source tarball name in spec
+sed -i "s/^Source:.*.tar.gz/Source: skopeo-$VERSION.tar.gz/" skopeo.spec
+
+# Update setup macro to use the correct build dir
+sed -i "s/^%setup.*/%autosetup -Sgit -n %{name}-$VERSION/" skopeo.spec
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,0 +1,27 @@
+---
+# See the documentation for more information:
+# https://packit.dev/docs/configuration/
+
+# NOTE: The Packit copr_build tasks help to check if every commit builds on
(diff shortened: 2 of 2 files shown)


lsm5:
@vrothberg @mtrmac @edsantiago @cevich PTAL .

lsm5:
here's what the packit copr build looks like: https://copr.fedorainfracloud.org/coprs/rhcontainerbot/packit-builds/build/5521572/

edsantiago:
LGTM, but my review mostly consisted of confirming the `Version` extraction, comparing against your netavark PR, and skimming through the build logs. @mtrmac PTAL.

mtrmac:
Same kind of review here…

- Do I understand correctly that PR merging is driven by the Cirrus “Total success” tasks, i.e. that this does not block at all? Or is the plan to make this a merging condition?
- (I assume the cost of these 19 builds is acceptable; just mentioning that to make sure it was considered.)

cevich:
> driven by the Cirrus “Total success” task

That would be challenging to pull off (linking the two).  Making the packet jobs "required" is problematic for the same reason as many of the cirrus tasks: They include non-static numbers in their names.  Unless, @lsm5 is there a way to make a packit job who's success and status depends on all the other jobs passing (i.e. similar to the "Total Success" task)?

cevich:
Just finished skimming the docs, didn't find any kind of job-dependency system.

mtrmac:
To be very explicit, I have no opinion on whether this should be blocking, and if so, how that could be done. I just want the experts to discuss that, and to have a written record of the intent (and ideally the rationale).

Personally, I’d find a long CI turnaround time a hassle, so I’m very fine with not making this blocking :)

lsm5:
> Same kind of review here…
> 
> * Do I understand correctly that PR merging is driven by the Cirrus “Total success” tasks, i.e. that this does not block at all? Or is the plan to make this a merging condition?

@mtrmac currently packit does not block any of the cirrus tasks including the success task and also doesn't block the openshift-merge-robot's commands , so the usual way of reviewing and merging shouldn't be affected. Made a note in the yaml file as well as the commit message. 

I think it's better to keep it this way until at least @edsantiago and @cevich are fully convinced of packit :) . I can check with the packit devs later if we want to make these jobs blocking.

> * (I assume the cost of these 19 builds is acceptable; just mentioning that to make sure it was considered.)

So far I've seen the packit tasks finish within the total time taken by all cirrus tasks and any infra costs AFAICT are borne by the copr people :) .

edsantiago (approved):

cevich:
>  I can check with the packit devs later if we want to make these jobs blocking.

To be clear, it's **us** that makes the jobs blocking.  It's a setting under the branch-protection.  You pick a job-name out of a list of recently run jobs - that's why having names changing is a problem.

In any case, I'm fine with this as-is for the time being.  Nice work.

/hold
/lgtm

mtrmac:
Thanks!

PR #1911, opened 2023-02-15
Update c/image after https://github.com/containers/image/pull/1842

mtrmac:
... so that Renovate doesn't keep proposing a downgrade.

See #1903 and https://github.com/containers/image/pull/1842 for the history of these attempts.


vrothberg (approved):
/lgtm

PR #1913, opened 2023-02-15
Replace gopkg.in/check.v1 by github.com/stretchr/testify/suite/

mtrmac:
`gopkg.in/check.v1` hasn't had any commit since Nov 2020. That's not an immediate issue for a test-only dependency, but because it hides access to the standard library `*testing.T`, eventually it will become limiting.

Also, using the same framework for unit and integration tests seems practical.

This is mostly a batch copy&paste job, with a fairly high risk of unexpected breakage.

Also, I didn't take much time at all to carefully choose between `assert.*` and `require.*`; we can tune that as failures show up.


mtrmac:
One significant advantage seems to be that using `testify/suite` (i.e. a `testing.T.Run()`) collects logs from all levels, in particular when a single test fails, the test output contains full logs from the registries and OpenShift serving that test.

And one disadvantage is that the same `-v` option both turns on logging the list of tests as they execute, and printing the full log output, including all the registry/OpenShift noise even on success.

So I have turned `-v` off; that gives us very little output on success, and probably all relevant output (logs of the relevant `suite`) on failure.

rhatdan:
LGTM

rhatdan:
@vrothberg PTAL

vrothberg (approved):
LGTM

PR #995, opened 2020-08-02
A couple of minor code cleanups.

rhatdan:
We want to use containers/podman for podman repo and make clean was blowing up.

PR #1918, opened 2023-02-17
Sync images with tags matching a semantic version constraint

### LINKED ISSUE #1756: sync all images from a starting version

kubasobon:
We are running `skopeo` to ensure our own registries contain specific images/tags. We do not want to sync every tag, but rather limit them to versions we utilize. This gets increasingly difficult for images using semantic versioning in their tags. See @SamStenton's reasoning in #1756 for examples.

This PR aims at adding an option to synchronize images with tags matching a particular semantic version constraint. Below is an example of YAML config that can be used with code added by this PR:
```yaml
docker.io:
  images-by-semver:
    alpine:
      - '3.12 - 3.13' # Sync all images in the range
      - '>= 3.17' # Sync all patch, minor, and major releases after 3.17.0
    ubuntu:
      - '~22' # Sync all releases >= 22.0 and < 23.0: 22.04, 22.10
```

This PR uses [github.com/Masterminds/semver v3](https://github.com/Masterminds/semver/tree/v3.2.0) package to parse and match semantic versions. This package respects the upstream (https://semver.org/) definition of semantic versioning.

Closes #1756


cmd/skopeo/sync.go
```diff
@@ -489,7 +556,7 @@ func imagesToCopy(source string, transport string, sourceCtx *types.SystemContex
 			return descriptors, err
 		}
 		for registryName, registryConfig := range cfg {
-			if len(registryConfig.Images) == 0 && len(registryConfig.ImagesByTagRegex) == 0 {
+			if len(registryConfig.Images) == 0 && len(registryConfig.ImagesByTagRegex) == 0 && len(registryConfig.ImagesBySemver) == 0 {
```
mtrmac, line 559: This is pretty far from the actual uses of those values, and risks getting out of sync — moving this whole check into `imagesToCopyFromRegistry` would be nice.

cmd/skopeo/sync.go
```diff
@@ -422,6 +424,71 @@ func imagesToCopyFromRegistry(registryName string, cfg registrySyncConfig, sourc
 			Context:   serverCtx})
 	}
 
+	// handle tags matching semver
+	for imageName, semverConstraints := range cfg.ImagesBySemver {
```
mtrmac, line 428: This is really quite a bit of duplicated code between (`cfg.Images`,),  `cfg.ImagesByTagRegex` and `cfg.ImagesBySemver`, and the function is now almost 200 lines.

Ignoring `cfg.Images`, please split the shared part, listing tags + filtering based on a predicate, into a separate function, parametrized by a filtering predicate; broadly similar to what #965 did.

(Alternatively, it would make sense to process each repo once, i.e. to extract the list of repos from all three `cfg.Images*` options, and then to iterate per-repo, implementing all three options in an inner loop. But I think that’s a larger refactoring than necessary for _this_ PR, and unfair to impose on this PR as a requirement.)

docs/skopeo-sync.1.md
```diff
@@ -235,6 +239,7 @@ This will copy the following images:
 - Repository `registry.example.com/redis`: images tagged "1.0" and "2.0" along with image with digest "sha256:0000000000000000000000000000000011111111111111111111111111111111".
 - Repository `registry.example.com/nginx`: images tagged "1.13.1-alpine-perl" and "1.13.2-alpine-perl".
 - Repository `quay.io/coreos/etcd`: images tagged "latest".
+- Repository `registry.example.com/alpine`: all images with tags satisfying either the "3.12 - 3.13" condition ("3.12.0", "3.12.1"...) or the ">= 3.17" ("3.17.5", "3.19.0", "4.0.0"...)
```
mtrmac, line 242: This should, somehow, point at the full relevant documentation of the specific constraint format. Maybe just a link to the website / repo.

mtrmac (commented):
Thanks! The use of constraint expressions like this seems very versatile and results in elegant configuration.

Compare #1918 — this is likely to be a more general solution.

---

One thing that we need to get “right” now is the constraint syntax — apparently https://semver.org doesn’t define anything like that as a part of the standard.

So, in addition to https://github.com/Masterminds/semver, there is also https://github.com/blang/semver (used in Podman, although not in a user-visible way), and https://github.com/hashicorp/go-version (used in previous #995 ); each seems to have a different syntax and/or feature set.

I don’t really know anything about the trade-offs, and in general the the PR author’s choice is the preferred one by default — still, I’d very much like others to weigh in here, if there are any issues to consider.

cmd/skopeo/sync.go
```diff
@@ -422,6 +424,71 @@ func imagesToCopyFromRegistry(registryName string, cfg registrySyncConfig, sourc
...
+					repoLogger.Errorf("Internal error, reference %s does not have a tag, skipping", sReference.DockerReference())
+					continue
+				}
+				tagVersion, err := semver.NewVersion(tagged.Tag())
+				if err != nil {
+					repoLogger.Tracef("Tag %q cannot be parsed as semver, skipping", tagged.Tag())
+					continue
+				}
+				if constraint.Check(tagVersion) {
+					uniqueSourceReferences[sReference.DockerReference().String()] = sReference
```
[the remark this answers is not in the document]
kubasobon: I have gone for the second solution. The code is moved into `semverFilterCollection()` now.

cmd/skopeo/sync.go
```diff
@@ -422,6 +424,71 @@ func imagesToCopyFromRegistry(registryName string, cfg registrySyncConfig, sourc
...
+				}
+			}
+		}
+
+		var sourceReferences []types.ImageReference
+		for _, sReference := range uniqueSourceReferences {
+			sourceReferences = append(sourceReferences, sReference)
+		}
```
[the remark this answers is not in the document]
kubasobon: The map is gone :)

cmd/skopeo/sync.go
kubasobon: I've attempted to do just that with my latest commits.

docs/skopeo-sync.1.md
kubasobon: Agreed, added a link to upstream documentation for version comparison and actual range syntax.

kubasobon:
> Thanks! The use of constraint expressions like this seems very versatile and results in elegant configuration.
> 
> Compare #1918 — this is likely to be a more general solution.
> 

Thank you for taking the time to review my PR!

> One thing that we need to get “right” now is the constraint syntax — apparently https://semver.org doesn’t define anything like that as a part of the standard.
> 
> So, in addition to https://github.com/Masterminds/semver, there is also https://github.com/blang/semver (used in Podman, although not in a user-visible way), and https://github.com/hashicorp/go-version (used in previous #995 ); each seems to have a different syntax and/or feature set.
> 
> I don’t really know anything about the trade-offs, and in general the the PR author’s choice is the preferred one by default — still, I’d very much like others to weigh in here, if there are any issues to consider.

I have done a bit of research on the side and created a test suite to compare the three libraries you mentioned. You can find it [here](https://github.com/kubasobon/semver-comparison/). My main goal was highlighting differences between the libraries and how well they handle edge cases, like incomplete versions, logical operators, etc. From my point of view https://github.com/Masterminds/semver offers the most features and covers every use case one might want/need from the other two. This hopefully satisfies the broadest spectrum of `skopeo` users.

cmd/skopeo/sync.go
```diff
@@ -368,61 +378,158 @@ func imagesToCopyFromRegistry(registryName string, cfg registrySyncConfig, sourc
...
+func tagRegexFilterCollection(collection map[string]string) (filterCollection, error) {
+	filters := filterCollection{}
+
+	for imageName, tagRegex := range collection {
+		pattern, err := regexp.Compile(tagRegex)
+		if err != nil {
+			return nil, err
+		}
+
+		f := func(logger *logrus.Entry, sourceReferences []types.ImageReference) []types.ImageReference {
```
mtrmac, line 468: This would be simpler if the filter only worked on a single reference, maybe `func(*logrus.Entry, types.ImageReference) bool`. The filtering loop could then exist only in `filterSourceReferences`

(It’s fine to lose the detailed “Start filtering using the regular expression: $value” to get this simplification; or alternatively `filterCollection` could be a `map[string]struct{filter func …, message string}`, but I don’t think that’s really necessary.

Really the “filtering” phase should take negligible time and I don’t think it needs a separate _progress_ log message at all. If the purpose of that log message were just to log the configuration that was used, that kind of logging could happen during the filter creation — but I wouldn’t bother with that without a specific indication of user need either.)

cmd/skopeo/sync.go
```diff
@@ -368,61 +378,158 @@ func imagesToCopyFromRegistry(registryName string, cfg registrySyncConfig, sourc
 			Context:   serverCtx})
 	}
 
-	for imageName, tagRegex := range cfg.ImagesByTagRegex {
-		repoLogger := logrus.WithFields(logrus.Fields{
+	// include repository descriptors for cfg.ImagesByTagRegex
+	{
```
mtrmac, line 382: Non-blocking: I don’t think the `{ }` blocks are that necessary — and the comments don’t say anything new either.

(Absolutely non-blocking: Possibly `s/filterCollection/filters/g`, `s/additionRepoDescList/descs/`; the variables have short enough scope that short names are sufficient. But this works just fine.)

mtrmac (commented):
Thanks!

@vrothberg (others), WDYT about the semver filter syntax?

vrothberg (commented):
> @vrothberg (others), WDYT about the semver filter syntax?

LGTM
I like the idea quite a lot and the dependency looks good.

Thanks a lot for contributing!

@rhatdan PTAL

cmd/skopeo/sync.go
kubasobon: That simplifies matters, indeed. I also think it's worth departing from logging all the details if it makes the code more difficult to navigate.

SamStenton:
Awesome stuff @kubasobon! Would love to start making use of this in some of our internal tooling - when can we expect this to be reviewed and merged?

rhatdan:
Well this PR needs to be rebased to even consider it.

derbauer97:
Is there any chance this will be merged? We would be happy to have such a feature it is exactly what we need :)

husseinferas:
Hey @kubasobon 
Thank you for the changes. I am looking forward to using this feature; is there any plan to move this forward and merge the PR in the near future?

dmitri-lerko:
Came here to propose a similar PR only to find it implemented and approved. How can I help to move this forward?

MShekow:
We also have been using the code of this PR for a couple of months, and would really love to get this merged.

What this PR is still lacking (but would be very sensible to add) is the ability to somehow additionally pre-filter tags that the semantic version parser library (here: [semver](https://github.com/Masterminds/semver/)) cannot possibly parse. The core problem is that version tags (in the Docker world) are a "wild west". For instance, look what the LinkerD-team has [done](https://github.com/orgs/linkerd/packages/container/proxy/versions?filters%5Bversion_type%5D=tagged): make up weird tag structures like `stable-<version>`. And similarly, cosign signatures may end up with tags such as `sha256-744544149f18ef2f80a2f8b1324d0ec6a05252d2b6613e60310881ecf476bcfa.sig`.

For this reason, I propose that there could be an additional _optional_ configuration option to configure this, e.g. using a named capture group inside a regular expression.

Example:

```yaml
images-by-semver:
  ghcr.io/linkerd/proxy:
    version-regex-filters:
    - "stable-(?P<semver>\\d+\\.\\d+\\.\\d)"
    version-ranges:
      - "3.12 - 3.13"
      - ">= 3.17"
```
The idea would be that _if_ any `version-regex-filters` are defined, only the content of the `semver` capture group is forwarded to the [semver](https://github.com/Masterminds/semver/) parser.

husseinferas:
Hello, everyone 
I've created a new PR with the same changes. I rebased it with the main branch 
https://github.com/containers/skopeo/pull/2189

cmd/skopeo/sync.go
```diff
@@ -368,61 +378,158 @@ func imagesToCopyFromRegistry(registryName string, cfg registrySyncConfig, sourc
...
+	}
+
+	return repoDescList, nil
+}
+
+// filterFunc is a function used to limit the initial set of image references
+// using tags, patterns, semver, etc.
+type filterFunc func(*logrus.Entry, []types.ImageReference) []types.ImageReference
+
+// filterCollection is a map of image names to filter functions.
```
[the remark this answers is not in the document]
husseinferas: are you sure about that? when I checked `type ImageReference` in `types.go` "ImageReference is an abstracted way to refer to an image location, namespaced within an ImageTransport." 
in the yaml file
```
registry.io:
  images-by-semver:
    image: [tags]
``` 
do you mean the "image" here should be renamed to repository? 
I think image name would be equal to repository name in this case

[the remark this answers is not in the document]
husseinferas: @mtrmac 
are you sure about that? when I checked `type ImageReference` in `types.go` "ImageReference is an abstracted way to refer to an image location, namespaced within an ImageTransport." 
in the yaml file
```
registry.io:
  images-by-semver:
    image: [tags]
``` 
do you mean the "image" here should be renamed to repository? 
I think image name would be equal to repository name in this case

mtrmac:
#2189 was merged. @kubasobon , @husseinferas thanks for all your effort!

PR #1921, opened 2023-02-22
Fix tabelating output in (skopeo inspect --format)

mtrmac:
`tabwriter` buffers lines that contain `\t` in memory, and only writes them out on a `.Flush()`. So actually call that.

Without this, things like `--format 'name\tdigest\tlabels\n{{.Name}}\t{{.Digest}}\t{{.Labels}}\n'` result in no output at all.


vrothberg (approved):
LGTM

PR #1927, opened 2023-02-27
Use io.WriteString

mtrmac:
... mostly so that I get practice and remember this exists in the future.

(This saves one allocation & copy when the target implements `io.StringWriter`. And that makes absolutely no relevant difference on this path.)


rhatdan:
LGTM
@vrothberg PTAL

vrothberg (approved):
LGTM

PR #1950, opened 2023-03-24
Verify signatures from a list of public keys

Jamstah:
Add the ability to use an on-disk trust store to verify signatures. Also allow the user to trust any known fingerprint instead of having to specify one.

The reason here is that we now have a number of different keys we have used to sign our images over time as they have been rotated. Being able to supply someone with a keystore they can use to verify any of our images with is valuable to us.

### DIFF
--- a/cmd/skopeo/signing.go
+++ b/cmd/skopeo/signing.go
@@ -6,11 +6,11 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 
 	"github.com/containers/image/v5/pkg/cli"
 	"github.com/containers/image/v5/signature"
 	"github.com/spf13/cobra"
-	"golang.org/x/exp/slices"
 )
 
 type standaloneSignOptions struct {
@@ -74,11 +74,11 @@ type standaloneVerifyOptions struct {
 func standaloneVerifyCmd() *cobra.Command {
 	opts := standaloneVerifyOptions{}
 	cmd := &cobra.Command{
-		Use:   "standalone-verify MANIFEST DOCKER-REFERENCE KEY-FINGERPRINT SIGNATURE",
+		Use:   "standalone-verify MANIFEST DOCKER-REFERENCE KEY-FINGERPRINTS SIGNATURE",
 		Short: "Verify a signature using local files",
 		Long: `Verify a signature using local files
 
-KEY-FINGERPRINT can be an exact fingerprint, or "any" if you trust all the keys in the public key file.`,
+KEY-FINGERPRINTS can be a comma separated list of fingerprints, or "any" if you trust all the keys in the public key file.`,
 		RunE: commandAction(opts.run),
 	}
 	flags := cmd.Flags()
@@ -93,10 +93,10 @@ func (opts *standaloneVerifyOptions) run(args []string, stdout io.Writer) error 
 	}
 	manifestPath := args[0]
 	expectedDockerReference := args[1]
-	expectedFingerprint := args[2]
+	expectedFingerprints := strings.Split(args[2], ",")
 	signaturePath := args[3]
 
-	if opts.publicKeyFile == "" && expectedFingerprint == "any" {
+	if opts.publicKeyFile == "" && len(expectedFingerprints) == 1 && expectedFingerprints[0] == "any" {
 		return fmt.Errorf("Cannot use any fingerprint without a public key file")
 	}
 	unverifiedManifest, err := os.ReadFile(manifestPath)
@@ -109,13 +109,13 @@ func (opts *standaloneVerifyOptions) run(args []string, stdout io.Writer) error 
 	}
 
 	var mech signature.SigningMechanism
-	var fingerprints []string
+	var publicKeyfingerprints []string
 	if opts.publicKeyFile != "" {
 		publicKeys, err := os.ReadFile(opts.publicKeyFile)
 		if err != nil {
 			return fmt.Errorf("Error reading public keys from %s: %w", opts.publicKeyFile, err)
 		}
-		mech, fingerprints, err = signature.NewEphemeralGPGSigningMechanism(publicKeys)
+		mech, publicKeyfingerprints, err = signature.NewEphemeralGPGSigningMechanism(publicKeys)
 	} else {
 		mech, err = signature.NewGPGSigningMechanism()
 	}
@@ -124,23 +124,16 @@ func (opts *standaloneVerifyOptions) run(args []string, stdout io.Writer) error 
 	}
 	defer mech.Close()
 
-	if opts.publicKeyFile != "" && expectedFingerprint == "any" {
-		_, expectedFingerprint, err = mech.Verify(unverifiedSignature)
-		if err != nil {
-			return fmt.Errorf("Could not determine fingerprint from signature: %w", err)
-		}
-		if !slices.Contains(fingerprints, expectedFingerprint) {
-			// This is theoretically impossible because mech.Verify only works if it can identify the key based on the signature
-			return fmt.Errorf("Signature fingerprint not found in public key file: %s", expectedFingerprint)
-		}
+	if len(expectedFingerprints) == 1 && expectedFingerprints[0] == "any" {
+		expectedFingerprints = publicKeyfingerprints
 	}
 
-	sig, err := signature.VerifyDockerManifestSignature(unverifiedSignature, unverifiedManifest, expectedDockerReference, mech, expectedFingerprint)
+	sig, verificationFingerprint, err := signature.VerifyImageManifestSignatureUsingKeyIdentityList(unverifiedSignature, unverifiedManifest, expectedDockerReference, mech, expectedFingerprints)
 	if err != nil {
 		return fmt.Errorf("Error verifying signature: %w", err)
 	}
 
-	fmt.Fprintf(stdout, "Signature verified using fingerprint %s, digest %s\n", expectedFingerprint, sig.DockerManifestDigest)
+	fmt.Fprintf(stdout, "Signature verified using fingerprint %s, digest %s\n", verificationFingerprint, sig.DockerManifestDigest)
 	return nil
 }
 
--- a/cmd/skopeo/signing.go
+++ b/cmd/skopeo/signing.go
@@ -10,6 +10,7 @@ import (
 	"github.com/containers/image/v5/pkg/cli"
 	"github.com/containers/image/v5/signature"
 	"github.com/spf13/cobra"
+	"golang.org/x/exp/slices"
 )
 
 type standaloneSignOptions struct {
@@ -41,12 +42,12 @@ func (opts *standaloneSignOptions) run(args []string, stdout io.Writer) error {
 
 	manifest, err := os.ReadFile(manifestPath)
 	if err != nil {
-		return fmt.Errorf("Error reading %s: %v", manifestPath, err)
+		return fmt.Errorf("Error reading %s: %w", manifestPath, err)
 	}
 
 	mech, err := signature.NewGPGSigningMechanism()
 	if err != nil {
-		return fmt.Errorf("Error initializing GPG: %v", err)
+		return fmt.Errorf("Error initializing GPG: %w", err)
 	}
 	defer mech.Close()
 
@@ -57,17 +58,17 @@ func (opts *standaloneSignOptions) run(args []string, stdout io.Writer) error {
 
 	signature, err := signature.SignDockerManifestWithOptions(manifest, dockerReference, mech, fingerprint, &signature.SignOptions{Passphrase: passphrase})
 	if err != nil {
-		return fmt.Errorf("Error creating signature: %v", err)
+		return fmt.Errorf("Error creating signature: %w", err)
 	}
 
 	if err := os.WriteFile(opts.output, signature, 0644); err != nil {
-		return fmt.Errorf("Error writing signature to %s: %v", opts.output, err)
+		return fmt.Errorf("Error writing signature to %s: %w", opts.output, err)
 	}
 	return nil
 }
 
 type standaloneVerifyOptions struct {
-	publickeyfile string
+	publicKeyFile string
 }
 
 func standaloneVerifyCmd() *cobra.Command {
@@ -81,7 +82,7 @@ KEY-FINGERPRINT can be an exact fingerprint, or "any" if you trust all the keys 
 		RunE: commandAction(opts.run),
 	}
 	flags := cmd.Flags()
-	flags.StringVar(&opts.publickeyfile, "public-key-file", "", `File containing public keys. If not specified, will use local GPG keys. When using a public key file, you can specify "any" as the fingerprint to trust any key in the public key file.`)
+	flags.StringVar(&opts.publicKeyFile, "public-key-file", "", `File containing public keys. If not specified, will use local GPG keys.`)
 	adjustUsage(cmd)
 	return cmd
 }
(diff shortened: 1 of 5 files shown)


cmd/skopeo/signing.go
```diff
@@ -98,17 +104,34 @@ func (opts *standaloneVerifyOptions) run(args []string, stdout io.Writer) error
...
+	} else {
+		mech, err = signature.NewGPGSigningMechanism()
+	}
 	if err != nil {
 		return fmt.Errorf("Error initializing GPG: %v", err)
 	}
 	defer mech.Close()
+
+	if expectedFingerprint == "any" {
+		_, expectedFingerprint, err = mech.Verify(unverifiedSignature)
```
mtrmac, line 123: Absolutely not. The fingerprint must come from a trusted source, not from the potentially-hostile signature. 

(And we should not accept keys of all humans that could possibly be imported into `$HOME/.gnupg`.)

Even if it might work out fine with the code as is, it’s a risk and the conditions that make it work now could be lost in future modifications.

The data is already available from a trusted source for ephemeral mechanisms.

cmd/skopeo/signing.go
```diff
@@ -67,15 +67,21 @@ func (opts *standaloneSignOptions) run(args []string, stdout io.Writer) error {
...
 		Use:   "standalone-verify MANIFEST DOCKER-REFERENCE KEY-FINGERPRINT SIGNATURE",
 		Short: "Verify a signature using local files",
-		RunE:  commandAction(opts.run),
+		Long: `Verify a signature using local files
+
+KEY-FINGERPRINT can be an exact fingerprint, or "any" if you trust all the keys in the trust store.`,
+		RunE: commandAction(opts.run),
 	}
+	flags := cmd.Flags()
+	flags.StringVar(&opts.truststore, "truststore", "", "Trust store of public keys. Defaults to using local gpg keys.")
```
mtrmac, line 84: This parameter is just a file that contains public keys, so `key-file`, `keys` or something like that?

mtrmac (commented):
Thanks for the PR.

Allowing to use `standalone-verify` with a set of keys separate from the global state in $HOMOE does make sense.

> Being able to supply someone with a keystore they can use to verify any of our images with is valuable to us.

That’s worries me quite a bit. `standalone-verify` is a low-level debugging tool, _inherently_ vulnerable to TOCTOU. I think in any kind of production use, almost all users should use `policy.json` set up at the point of consumption (e.g. for CRI-O or `podman pull`).

The idea that some _other_ people, probably not experts, are instructed (even if by an expert) to use `standalone-verify`, seems to make it more likely to me that the TOCTOU risks are going to materialize.

So I’m tempted to push back quite a bit: Why is `pull` + `policy.json` (or, I don’t know, some similar enforcement at the point of image import/approval) not the right tool for those users?

cmd/skopeo/signing.go
Jamstah: I think trust store is more specific, it's saying "these are the keys to use for trusting signatures". Could say "public-keys"?

cmd/skopeo/signing.go
Jamstah: It's really intended to use with the new argument, to say "accept anything from this store".

If the key isn't in the store the verification fails, even with any.

Could make it only work if the argument is set?

cmd/skopeo/signing.go
mtrmac: Works for me.

Jamstah:
> That’s worries me quite a bit. standalone-verify is a low-level debugging tool, inherently vulnerable to TOCTOU. I think in any kind of production use, almost all users should use policy.json set up at the point of consumption (e.g. for CRI-O or podman pull).

I agree completely that verification at time of use is preferable, unfortunately, I don't feel the signing landscape is there yet. I'm looking forward to registry references making it better. Verifying at point of use is not too bad when pulling from the origin registry, but is hard for users who have mirrored the image and have changed the location, which is part of the signature.

There is a TOCTOU problem when using tags, but we're using digests. If you verify a digest is signed then it's safe to use that digest anywhere, at any time, because an image referred to by digest is immutable.

At the moment, the only good solution we have to help people verify images on a registry using the signature extension mechanism seems to be standalone verify. This pr makes that slightly easier, and some customers have asked us how to verify in place.

This is by no means the way I want customers doing this long term, but I don't have a better option for them right now. I'm looking at ways to help them verify representations on disk as part of a mirror operation, or to enable verification on pull from a mirror registry. Any thoughts welcome!

cmd/skopeo/signing.go
Jamstah: Ok. If we get to a place of conceptually agreeing I'll make the change :)

cmd/skopeo/signing.go
mtrmac: Conceptually, sure, if the public keys are explicitly provided, that already specifies the fingerprints, and `policy.json` uses the key data that way already. I just don’t want _this particular implementation approach_, reading the fingerprint from the untrusted signature, to be used; I’ve seen vulnerabilities from somewhat similar uses of untrusted data that are (originally) safe in context.

As for implementing this, `VerifyDockerManifestSignature` currently does not expose that feature to external callers. So we would need a small addition to c/image, which is, in isolation, not a big deal.

… but then again, why is the application not using `policy.json` already?

Jamstah: Would you be more comfortable taking each of the fingerprints in the public key file and attempting verification in turn to see if one of them matches?

I'm not aware of a way to fake a fingerprint in a signature, but if that does exist, testing with a series of explicit keys wouldn't help, would it?

Jamstah: I'll leave the policy.json question to the main thread :)

mtrmac:
> is hard for users who have mirrored the image and have changed the location, which is part of the signature.

Is `remapIdentity` in `policy.json` conceptually not sufficient, or somehow broken for this purpose? Sure, that requires some kind of integration between the software that configures mirrors to access and `policy.json`. Is that the hurdle? Or something else?

> I'm looking at ways to help them verify representations on disk as part of a mirror operation, or to enable verification on pull from a mirror registry.

I’d expect the software that _does_ the mirroring to verify signatures when reading from the upstream source (confirming that it is mirroring the right thing), and should be possible using `policy.json`. Do you mean that you are looking at some kind of after-mirroring sanity check of the data? Or is `policy.json` just not applicable because the software that does the mirroring is a different implementation?

---

> There is a TOCTOU problem when using tags, but we're using digests. If you verify a digest is signed then it's safe to use that digest anywhere, at any time, because an image referred to by digest is immutable.

That’s a good point. Fair enough, I’d still recommend looking for ways to to use `policy.json` instead, but if you continue to think this is the only practical way, let’s do it.

cmd/skopeo/signing.go
mtrmac: I think a possible implementation (which is similar to what `policy.json` does, simplifying things)
- `signature.NewEphemeralGPGSigningMechanism` already returns the fingerprints that can be matched; those can be used instead of being ignored.
-  The core verification logic underlying `VerifyDockerManifestSignature` is already flexible WRT deciding which keys are acceptable; so it’s just a matter of exposing that in c/image, `VerifyImageManifestSignatureMultipleKeys` or something.

---

> I'm not aware of a way to fake a fingerprint in a signature

(BTW the signature doesn’t even contain the full fingerprint… just a trailing part that is used to look up a local key.)

> , but if that does exist, testing with a series of explicit keys wouldn't help, would it?

There’s nothing insecure about the code as it is _in this PR_. I’m just a touch paranoid :)

The thing I’m worried about is that it muddles the trust dependency relationship.

Before: fully trusted user → public keys and a specific fingerprint → cryptographic signature → signature payload → manifest. There’s a clear “source of trust → originally untrusted, newly trusted data” direction, and untrusted data is not used for any purpose until the prerequisites are verified.

After, in _some_ cases: fully trusted user → public keys → cryptographic signature → fingerprint → cryptographic signature → signature payload …

i.e. this line is fine on the code path where it is, but it would be problematic on the other code path, unexpectedly accepting any signature from $HOME/.gnupg ; or it could be copy&pasted into some other software…

mtrmac:
(BTW if you happened to be revisiting the larger mirroring design now, consider using sigstore-formatted signatures instead. It means a new set of keys, a new signature format, and a new signature transport mechanism, but the signatures are stored in ordinary registries, without the huge hassle of separate lookaside servers, or being restricted to registry implementations with x-registry-supports-signatures — and the signatures are subject to ordinary registry access controls.

If you already have a “simple signing” GPG workflow that works, I don’t think there’s any reason to move; but if you were building this entirely from scratch, not having to deal with the lookaside might be a noticeable operational improvement — at the cost of needing an updated mirroring tool that knows how to mirror sigstore signatures.

`{skopeo copy, podman push} --sign-by-{sigstore,sigstore-private-key}` for signing; `sigstoreSigned` in `policy.json` for verification; `use-sigstore-attachments` in `registries.d/*.yaml`.)

Jamstah:
Yes, we have simple signing working now and that's a requirement from above, but sigstore is the strategic direction and I believe we already have access to x509 keys to use so I can look into signing our content with both mechanisms.

We use `oc image mirror` as our mirroring tool, so long term I see these options:
- Putting in a PR to oc image mirror to make it mirror simple signatures (not strategic)
- Putting in a PR to oc image mirror to make it mirror sigstore tags automatically (maybe, but won't be needed after reference support)
- Adding the sigstore tags to our list of images to mirror in our build processes (unlikely to be picked up by all IBM Cloud Pak teams without huge organizational effort)
- Waiting for references support and putting in a PR to oc image mirror to mirror references too (don't know how long that will take).

After signatures are mirrored I can look at how to help customers set up their policy files.

Short term, I'm thinking of making a quick utility to grab the signatures for mirrored content and using skopeo to verify the manifest signatures, and sha256sum to verify the blob content.

I'll make an update so that "any" only works with a supplied public keys file, and rename the arg.

Also, thank you so much for your help and discussion so far, I'm really at the learning stage of image signing and will be using my new found knowledge to pull out my stirring stick and try and improve our approach across the board :)

cmd/skopeo/signing.go
Jamstah: > The core verification logic underlying VerifyDockerManifestSignature is already flexible WRT deciding which keys are acceptable; so it’s just a matter of exposing that in c/image, VerifyImageManifestSignatureMultipleKeys or something.

I was hoping not to have to do the c/image dance if I could avoid it, which is why I went with the approach of verifying we have the key in the public key store, then running the proper verification.

I could just run the verification once per key without having to change c/image, but if you think changing c/image is the only acceptable approach, then I'll go add the new method and suck it up :)

Jamstah: Actually, it would be hard to run it once per key cleanly because the error type is the same for all errors and I can't tell cleanly if its a fingerprint mismatch.

So, its either (a) the approach I have, but restrict it to only working if we specify a key file, or (b) update c/image with a multi-keyprint method.

I'd prefer to do (a), but if you think (b) is the only way, then I'll do that instead.

Jamstah: Have changed the argument name and added a double-check of the fingerprint being found in the public key file. I might pass this around some colleagues so wanted the externals to look about right.

cmd/skopeo/signing.go
```diff
@@ -67,15 +67,21 @@ func (opts *standaloneSignOptions) run(args []string, stdout io.Writer) error {
...
 		Use:   "standalone-verify MANIFEST DOCKER-REFERENCE KEY-FINGERPRINT SIGNATURE",
 		Short: "Verify a signature using local files",
-		RunE:  commandAction(opts.run),
+		Long: `Verify a signature using local files
+
+KEY-FINGERPRINT can be an exact fingerprint, or "any" if you trust all the keys in the public key file.`,
+		RunE: commandAction(opts.run),
 	}
+	flags := cmd.Flags()
+	flags.StringVar(&opts.publickeyfile, "public-key-file", "", `File containing public keys. If not specified, will use local gpg keys. When using a public key file, you can specify "any" as the fingerprint to trust any key in the public key file.`)
```
TomSweeneyRedHat, line 84:
```suggestion
	flags.StringVar(&opts.publickeyfile, "public-key-file", "", `File containing public keys. If not specified, will use local GPG keys. When using a public key file, you can specify "any" as the fingerprint to trust any key in the public key file.`)
```

At least it's supposed to be uppercased to my understanding.

docs/skopeo-standalone-verify.1.md
```diff
@@ -28,6 +28,10 @@ as per containers-policy.json(5).
 
 Print usage statement
 
+**--public-key-file** _public key file_
+
+Trust store of public keys to use when verifying signatures. If this is not specified, keys from gpg home are used.
```
TomSweeneyRedHat, line 33:
```suggestion
The trust store of public keys to use when verifying signatures. If this is not specified, keys from the GPG homedir are used.
```

TomSweeneyRedHat:
@Jamstah thanks for the PR.  You need a rebase and some of the tests are barking.

Jamstah:
> @Jamstah thanks for the PR. You need a rebase and some of the tests are barking.

Thanks for the review, done.

cmd/skopeo/signing.go
```diff
@@ -67,15 +67,21 @@ func (opts *standaloneSignOptions) run(args []string, stdout io.Writer) error {
 }
 
 type standaloneVerifyOptions struct {
+	publickeyfile string
```
mtrmac, line 70: `publicKeyFile` please.

cmd/skopeo/signing.go
```diff
@@ -67,15 +67,21 @@ func (opts *standaloneSignOptions) run(args []string, stdout io.Writer) error {
...
 		Use:   "standalone-verify MANIFEST DOCKER-REFERENCE KEY-FINGERPRINT SIGNATURE",
 		Short: "Verify a signature using local files",
-		RunE:  commandAction(opts.run),
+		Long: `Verify a signature using local files
+
+KEY-FINGERPRINT can be an exact fingerprint, or "any" if you trust all the keys in the public key file.`,
+		RunE: commandAction(opts.run),
 	}
+	flags := cmd.Flags()
+	flags.StringVar(&opts.publickeyfile, "public-key-file", "", `File containing public keys. If not specified, will use local GPG keys. When using a public key file, you can specify "any" as the fingerprint to trust any key in the public key file.`)
```
mtrmac, line 84: Non-blocking: dropping the last sentence might make `--help` look better on not-that-large terminals. The full details can be in the man page.

cmd/skopeo/signing.go
```diff
@@ -98,17 +104,45 @@ func (opts *standaloneVerifyOptions) run(args []string, stdout io.Writer) error
 		return fmt.Errorf("Error reading signature from %s: %v", signaturePath, err)
 	}
 
-	mech, err := signature.NewGPGSigningMechanism()
+	var mech signature.SigningMechanism
+	var fingerprints []string
+	if opts.publickeyfile != "" {
+		publicKeys, err := os.ReadFile(opts.publickeyfile)
+		if err != nil {
+			return fmt.Errorf("Error reading public keys from %s: %v", opts.publickeyfile, err)
```
mtrmac, line 112: (Absolutely non-blocking: Using `%w` would be more idiomatic — but it doesn’t make a difference in practice, and there’s quite a lot of pre-existing `%v`.)

cmd/skopeo/signing.go
```diff
@@ -98,17 +104,45 @@ func (opts *standaloneVerifyOptions) run(args []string, stdout io.Writer) error
...
+		mech, fingerprints, err = signature.NewEphemeralGPGSigningMechanism(publicKeys)
+	} else {
+		mech, err = signature.NewGPGSigningMechanism()
+	}
 	if err != nil {
 		return fmt.Errorf("Error initializing GPG: %v", err)
 	}
 	defer mech.Close()
+
+	if opts.publickeyfile != "" && expectedFingerprint == "any" {
```
mtrmac, line 123: `any` without `--public-key-file` should probably return a readable error instead of trying and failing to verify a signature.

cmd/skopeo/signing.go
```diff
@@ -98,17 +104,45 @@ func (opts *standaloneVerifyOptions) run(args []string, stdout io.Writer) error
...
+		if err != nil {
+			return fmt.Errorf("Could not determine fingerprint from signature: %v", err)
+		}
+		found := false
+		for _, fingerprint := range fingerprints {
+			if expectedFingerprint == fingerprint {
+				found = true
+			}
+		}
+		if !found {
```
mtrmac, line 134: (This can be simplified to `x/exp/slices.Contains`, but it probably shouldn’t remain anyway.)

cmd/skopeo/signing.go
```diff
@@ -98,17 +104,45 @@ func (opts *standaloneVerifyOptions) run(args []string, stdout io.Writer) error
...
+		}
+	}
+
 	sig, err := signature.VerifyDockerManifestSignature(unverifiedSignature, unverifiedManifest, expectedDockerReference, mech, expectedFingerprint)
 	if err != nil {
 		return fmt.Errorf("Error verifying signature: %v", err)
 	}
 
-	fmt.Fprintf(stdout, "Signature verified, digest %s\n", sig.DockerManifestDigest)
+	fmt.Fprintf(stdout, "Signature verified using fingerprint %s, digest %s\n", expectedFingerprint, sig.DockerManifestDigest)
```
mtrmac, line 145: I’m not too sure about this addition… especially I’d like to avoid a policy tool being built on top, parsing this text; human-targeted texts are not a stable API.

Yes, it’s useful for debugging … but for debugging, also available without using `skopeo`.

cmd/skopeo/signing.go
mtrmac: Yes, I’d much prefer a new function in c/image , to make all the enforcement logic centralized … and yes, with a fairly stringent unit-test coverage requirements. Apart from `skopeo sync`, Skopeo is mostly intended to be a _very dumb_ wrapper around c/image.

(We don’t need a c/image release to add the Skopeo feature; Skopeo can just vendor a mainline commit.)

mtrmac:
> Going to assume that ostree-rs-ext is a flake?

Extremely likely; I have now triggered a re-run.

---

(For other reviewers: Note https://github.com/containers/skopeo/pull/1950#discussion_r1152663358 , please don’t merge yet.)

cmd/skopeo/signing.go
Jamstah: Wasn't sure about importing the new package, but yes, it will be going.

cmd/skopeo/signing.go
Jamstah: OK, will do that next :)

The painful part last time was needing to make the skopeo tests pass in the c/Image PR. This time, I expect I don't need to do that bit.

Jamstah: image pr: https://github.com/containers/image/pull/1904

cmd/skopeo/signing.go
mtrmac: >  because the message format just changed and broke them :)

… like our own tests, yes :)

Fair enough, let’s leave it in.

mtrmac:
(The `ostree-rs-ext` failures are unrelated.)

mtrmac:
#1959 updated c/image with the new feature.

Jamstah:
I may have slipped in an extra feature.

mtrmac (approved):
Thanks!

PR #1953, opened 2023-03-27
systemtests: Fix 040-local-registry-auth about XDG_RUNTIME_DIR

sstosh:
Need to restore `XDG_RUNTIME_DIR` when podman removes the registry.

---

The test fails on rootless because it fails to remove the registry container.

<details>
<summary>tests are failed on rootless</summary>

```
ok 1 auth: credentials on command line
not ok 2 auth: credentials via podman login
# (from function `log_and_run' in file systemtest/helpers.bash, line 130,
#  from function `start_registry' in file systemtest/helpers.bash, line 354,
#  from function `setup' in test file systemtest/040-local-registry-auth.bats, line 15)
#   `start_registry --testuser=$testuser --testpassword=$testpassword --enable-delete=true reg' failed with status 125
# grep: /tmp/skopeo_bats.QQTiih/auth/htpasswd: No such file or directory
# $ podman --cgroup-manager=cgroupfs run -d --name reg -v /tmp/skopeo_bats.QQTiih/auth:/auth:Z -p 5000:5000 -e REGISTRY_STORAGE_DELETE_ENABLED=true -e REGISTRY_AUTH=htpasswd -e REGISTRY_AUTH_HTPASSWD_PATH=/auth/htpasswd -e REGISTRY_AUTH_HTPASSWD_REALM=Registry Realm quay.io/libpod/registry:2
# Error: creating container storage: the container name "reg" is already in use by d01c4e29dcd83ff882f15098417e456250760cc1db7a969e1d5a06e24b54aedc. You have to remove that container to be able to reuse that name: that name is already in use
# reg
ok 3 auth: copy with --src-creds and --dest-creds
not ok 4 auth: credentials via authfile
# (from function `log_and_run' in file systemtest/helpers.bash, line 130,
#  from function `start_registry' in file systemtest/helpers.bash, line 354,
#  from function `setup' in test file systemtest/040-local-registry-auth.bats, line 15)
#   `start_registry --testuser=$testuser --testpassword=$testpassword --enable-delete=true reg' failed with status 125
# grep: /tmp/skopeo_bats.PDl4IN/auth/htpasswd: No such file or directory
# $ podman --cgroup-manager=cgroupfs run -d --name reg -v /tmp/skopeo_bats.PDl4IN/auth:/auth:Z -p 5000:5000 -e REGISTRY_STORAGE_DELETE_ENABLED=true -e REGISTRY_AUTH=htpasswd -e REGISTRY_AUTH_HTPASSWD_PATH=/auth/htpasswd -e REGISTRY_AUTH_HTPASSWD_REALM=Registry Realm quay.io/libpod/registry:2
# Error: creating container storage: the container name "reg" is already in use by 66ec0f8a781d1fbc7f0b36153669f06e4e0aec2c9cffe5a404b9ed2fec836ac7. You have to remove that container to be able to reuse that name: that name is already in use
# reg
```
</details>

<details>
<summary>tests are succeeded on rootless after applying this commit</summary>

```
ok 1 auth: credentials on command line
ok 2 auth: credentials via podman login
ok 3 auth: copy with --src-creds and --dest-creds
ok 4 auth: credentials via authfile
```
</details>

### DIFF
--- a/systemtest/040-local-registry-auth.bats
+++ b/systemtest/040-local-registry-auth.bats
@@ -17,6 +17,7 @@ function setup() {
     _cred_dir=$TESTDIR/credentials
     # It is important to change XDG_RUNTIME_DIR only after we start the registry, otherwise it affects the path of $XDG_RUNTIME_DIR/netns maintained by Podman,
     # making it impossible to clean up after ourselves.
+    export XDG_RUNTIME_DIR_OLD=$XDG_RUNTIME_DIR
     export XDG_RUNTIME_DIR=$_cred_dir
     mkdir -p $_cred_dir/containers
     # Remove old/stale cred file
@@ -111,6 +112,9 @@ function setup() {
 }
 
 teardown() {
+    # Need to restore XDG_RUNTIME_DIR.
+    XDG_RUNTIME_DIR=$XDG_RUNTIME_DIR_OLD
+
     podman rm -f reg
 
     if [[ -n $_cred_dir ]]; then


TomSweeneyRedHat:
/lgtm
Thanks @sstosh !

systemtest/040-local-registry-auth.bats
```diff
@@ -17,6 +17,7 @@ function setup() {
     _cred_dir=$TESTDIR/credentials
     # It is important to change XDG_RUNTIME_DIR only after we start the registry, otherwise it affects the path of $XDG_RUNTIME_DIR/netns maintained by Podman,
     # making it impossible to clean up after ourselves.
+    export XDG_RUNTIME_DIR_OLD=$(echo $XDG_RUNTIME_DIR)
```
edsantiago, line 20: Please just use `export X=$Y`, without the `echo` and subshell

systemtest/040-local-registry-auth.bats
```diff
@@ -111,7 +112,8 @@ function setup() {
 }
 
 teardown() {
-    podman rm -f reg
+    # Need to change XDG_RUNTIME_DIR when podman removes the registry.
+    XDG_RUNTIME_DIR=$XDG_RUNTIME_DIR_OLD podman rm -f reg
```
edsantiago, line 116: This does not do what you think it does. This runs the `podman rm` command with the old XDG... but leaves XDG the way it is. Then, on the next `@test`, and next `setup()`, `OLD` will be set to the current `XDG`, which is probably not what you want. 

I don't remember this code, and have not fully read it, but I believe the correct approach would be:
```bash
setup() {
    # temporarily override XDG for (whatever reason) but preserve old setting
    export OLD=$XDG
    export XDG=$newvalue
}
....
teardown() {
   # restore XDG
   XDG=$OLD
```

edsantiago (changes requested):
Thanks for your PR. I haven't done a full review, but found some issues that probably (unless I missed something significant) need addressing.

sstosh:
Thank you for your review. I fixed it.
And I checked the tests passed.

```
ok 1 auth: credentials on command line
ok 2 auth: credentials via podman login
ok 3 auth: copy with --src-creds and --dest-creds
ok 4 auth: credentials via authfile
```

edsantiago (approved):
LGTM except for maybe the condition where `XDG` is undefined, but I'm finding it hard to care about that. I'll let @mtrmac have the final say. Thanks again!

mtrmac:
> LGTM except for maybe the condition where `XDG` is undefined, but I'm finding it hard to care about that. I'll let @mtrmac have the final say. Thanks again!

We get … lucky… in that Go does not have proper optional types, so everything does `Getenv("XDG_RUNTIME_DIR") != ""`, so for Go programs there’s no difference between `XDG_RUNTIME_DIR` unset and set to an empty value. (Ugh… this one time it helps us, but on the whole I’d rather have a good clean programming language.)

mtrmac:
> > LGTM except for maybe the condition where `XDG` is undefined, but I'm finding it hard to care about that. I'll let @mtrmac have the final say. Thanks again!
> 
> We get … lucky… in that Go does not have proper optional types, so everything does `Getenv("XDG_RUNTIME_DIR") != ""`, which means that for Go programs there’s no difference between `XDG_RUNTIME_DIR` unset and set to an empty value. (Ugh… this one time it helps us, but on the whole I’d rather have a good clean programming language.)

mtrmac (approved):
Thanks!

This is a clear improvement, so merging.

Possible cleanups I don’t really think are worth blocking on:
- upper-case shell variable names are, pedantically, reserved by POSIX
- I can imagine, instead of overriding `XDG_RUNTIME_DIR` for all commands in scope, doing that manually for every, single, command, that, needs, that. That would be safer, but also much more visual noise, so I mostly prefer the current approach (and I hate to admit that :) )

edsantiago:
>overriding XDG_RUNTIME_DIR for all commands in scope, doing that manually for every, single, command, that, needs, that

FWIW that was my first inclination also, and I, too, recoiled in horror after looking into it. Sometimes Good Enough is Good Enough.

PR #1965, opened 2023-04-05
Fix error handling of signature.NewEphemeralGPGSigningMechanism

mtrmac:
`signature.NewEphemeralGPGSigningMechanism is called` in an `if` branch where the previous `err :=` introduces a "new" `err` variable, which means the failure isn't visible after the `if`.

So, do the dumb thing and just check on both branches explicitly. (We still need to worry about correctly setting `mech` and `publicKeyfingerprints` to persist after the `if`.)

How I hate Go sometimes. And this shows we really should update the linter.


mtrmac:
@vrothberg @TomSweeneyRedHat PTAL, I’d like to fix this before a release.

TomSweeneyRedHat:
LGTM
looks like the Skopeo tests are being bit by the sha1 issue too.

mtrmac:
The RPM tests are (so far?) intentionally advisory-only, so that’s not an immediate major concern.

PR #1967, opened 2023-04-05
Rework hack/make*, and switch from golint to golangci-lint

### LINKED ISSUE #1892: Replace golint with golangci-lint
https://github.com/golang/lint says `golint` is deprecated and frozen; we are using `golangci-lint` in other repos, so we should move here as well.

Cc: @cevich :
- c/image’s `.cirrus.yml` says “# golangci-lint tool requires heavy cpu/memory or gets OOM'd”. Is that something to explicitly worry about here, or should we just try and see what happens?
- Looking at c/image, we are installing `golangci-lint` in `Makefile` on every test run. Would it pay off to include it in the CI image?

Note to self: After Skopeo migrates from `golint`, we should also remove the rules to install `golint` from c/image.

mtrmac:
Motivated by #1965 , finally fix #1892 .

Also take the opportunity to simplify all the validation scripts, by removing dead code, or no-longer-necessary options, and the like.

(https://golangci-lint.run/usage/install/ recommends using a GitHub action; I went with what I was already familiar with, consistent with what is used during local development, just to make sure that this gets done; we can improve on this later.)

Cc: @cevich @edsantiago , I might well be overlooking something about the CI pipeline or system tests.


mtrmac:
(This is expected to fail without #1965 ; I’d like to see that at least once, then I’ll rebase or something.)

.cirrus.yml
```diff
@@ -91,7 +92,7 @@ osx_task:
         export PATH=$GOPATH/bin:$PATH
         brew update
         brew install gpgme go go-md2man
-        go install golang.org/x/lint/golint@latest
+        curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.52.2
```
cevich, line 95: Consider setting/using an env-var for the version number.  You can do that up under the `Global variables used for all tasks` section.  That should be easier to maintain.

Alternatively, if parity with local development is important, consider doing the install and encoding the version in the `Makefile` instead.

cevich (approved):
Nice work @mtrmac This will reduce the maintenance and enhancement burden quite a bit.  It's also much more pleasing to the eyes :wink:

cevich:
> https://golangci-lint.run/usage/install/ recommends using a GitHub action

BTW, it's good you did this with scripts.  Github-actions aren't portable to local development and coordinating workflow dependencies w/ other CI systems (i.e. cirrus-ci) is a major PITA.

.cirrus.yml
mtrmac: Thanks, not duplicating the version number is a great point.

I went with `make tools`, somewhat similar to what c/image does.

TomSweeneyRedHat:
Looks like a rebase may be needed.  
LGTM

rhatdan:
LGTM

PR #1973, opened 2023-04-13
docs: add MacPorts installation info

herbygillot:
https://ports.macports.org/port/skopeo/

### DIFF
--- a/install.md
+++ b/install.md
@@ -55,11 +55,19 @@ sudo apk add skopeo
 
 [Package Info](https://pkgs.alpinelinux.org/packages?name=skopeo)
 
-### macOS
-
+### macOS - Homebrew
 ```sh
 brew install skopeo
 ```
+
+[Package info](https://formulae.brew.sh/formula/skopeo#default)
+
+### macOS - MacPorts
+```sh
+sudo port install skopeo
+```
+
+[Package info](https://ports.macports.org/port/skopeo/)
 
 ### Nix / NixOS
 ```sh


mtrmac (commented):
Thanks. Adding a MacPorts package would be useful.

Right now, though, at least looking at https://build.macports.org/builders/ports-11_arm64-builder/builds/88010/steps/install-port/logs/files , and comparing with the Homebrew formula, the port seems to only install a small subset of the Skopeo files that `make install` installs.

And at the very least a `policy.json` is necessary for a useful Skopeo setup; I’m not comfortable with recommending a packaging that habituates users to use `--insecure-policy` all the time.

So this seems a bit premature right now.

herbygillot:
Understood - I will make changes to make the MacPorts port contents match what is expected.

herbygillot:
OK, the `skopeo` port has been updated - contents are being installed by `make install`: https://build.macports.org/builders/ports-13_arm64-builder/builds/17582/steps/install-port/logs/files

It will take a while for the builders to update the archives for the other macOS versions, but the port should now be correct.

mtrmac:
Does that work as expected? Compare https://github.com/containers/skopeo/issues/787 .

herbygillot:
Oh that's useful - give me a few to make those modifications.

mtrmac:
As a very minimal litmus test, I think `skopeo inspect` and `skopeo copy docker:… dir:…` needs to work out of the box.

PR #1980, opened 2023-04-24
Improve the static build documentation a tiny bit

### LINKED ISSUE #1700: clean up static binary handling
While I understand it is not officially supported, the documentation and Makefile implementation surrounding the generation of a static binary is rather confusing right now.

install.md says to do this:
```
docker run -v $PWD:/src -w /src -e CGO_ENABLED=0 golang \
make BUILDTAGS=containers_image_openpgp GO_DYN_FLAGS=
```

However, the Makefile makes reference to some non-standard `DISABLE_CGO` environment variable, as well as some additional build tags not mentioned in install.md.
https://github.com/containers/skopeo/blob/1add7a81d769e734d7419e7e9f3f38d1ac746987/Makefile#L112-L114

Instead, it would be a lot more helpful if the files in question used a `//go:build !cgo` (or `// +build !cgo`) build constraint rather than forcing people to also pass these additional build tags. At the very least, the Makefile should reference `go env CGO_ENABLED` instead of defining its own similarly named variable.

Similarly, it is unclear if clearing `GO_DYN_FLAGS` is really necessary, but it is certainly suspicious given that the Makefile itself sets it. As best I can tell this is a non-standard variable that skopeo made up, so probably the Makefile should just be fixed to calculate it correctly. (But first we should confirm whether `-buildmode=pie` is actually a problem with `CGO_ENABLED=0` or not.)
https://github.com/containers/skopeo/blob/1add7a81d769e734d7419e7e9f3f38d1ac746987/Makefile#L55-L59

mtrmac:
- Accurately document the effect of `GO_DYN_FLAGS`
- Fix a typo
- Remove a shell script example; removes a `docker` command, we don't care to maintain it to run correctly, and anyone knowledgeable enough to understand the trade-offs of static linking can almost certainly create that from scratch.
 
Fixes #1700

### DIFF
--- a/install.md
+++ b/install.md
@@ -235,15 +235,8 @@ There have been efforts in the past to produce and maintain static builds, but t
 
 That being said, if you would like to build Skopeo statically, you might be able to do it by combining all the following steps.
 - Export environment variable `CGO_ENABLED=0` (disabling CGO causes Go to prefer native libraries when possible, instead of dynamically linking against system libraries).
-- Set the `BUILDTAGS=containers_image_openpgp` Make variable (this remove the dependency on `libgpgme` and its companion libraries).
-- Clear the `GO_DYN_FLAGS` Make variable (which otherwise seems to force the creation of a dynamic executable).
-
-The following command implements these steps to produce a static binary in the `bin` subdirectory of the repository:
-
-```bash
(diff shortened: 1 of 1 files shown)


vrothberg (approved):
LGTM

rhatdan:
LGTM

Issue #1935, opened 2023-03-07
Merging a single image `dir:` with an `index-only` `dir:`


achilleas-k:
_This is partially related to https://github.com/containers/image/issues/1874._

Our use case requires transferring a container and its manifests (and particularly the manifest list) using a portable format that doesn't affect the image itself.  We also don't want to be downloading all the images for all the architectures just to get the manifest list.  In the end, it should be possible to run the container by referencing its manifest list digest without needing access to the original registry (or even an internet connection).

@mtrmac suggested using `--multi-arch=index-only` to store the manifest list in a `dir:` and then getting the image separately, which means eventually, these would need to be merged.  I've used the following script to accomplish this, but I'd like confirmation that this is the right way.

Using [Fedora 37 on Quay.io as an example](https://quay.io/repository/fedora/fedora/manifest/sha256:176454f0e89d7bda8b8b577bfd855f5cb3854234d781855baef82cb057b0529e)
```bash
skopeo copy --preserve-digests docker://quay.io/fedora/fedora@sha256:176454f0e89d7bda8b8b577bfd855f5cb3854234d781855baef82cb057b0529e dir:container-archive/image

skopeo copy --preserve-digests --multi-arch=index-only docker://quay.io/fedora/fedora@sha256:176454f0e89d7bda8b8b577bfd855f5cb3854234d781855baef82cb057b0529e dir:container-archive/index
cp -a container-archive/image container-archive/merged

digest=$(sha256sum container-archive/image/manifest.json | cut -f1 -d' ')

mv container-archive/merged/manifest.json "container-archive/merged/${digest}.manifest.json"

cp container-archive/index/manifest.json container-archive/merged/manifest.json
```

To verify that it worked:

```bash
sudo skopeo copy --preserve-digests dir:container-archive/merged 'containers-storage:[overlay@/var/lib/containers/storage+/run/containers/storage]quay.io/fedora/fedora@sha256:176454f0e89d7bda8b8b577bfd855f5cb3854234d781855baef82cb057b0529e'
sudo podman run --rm --pull=never quay.io/fedora/fedora@sha256:176454f0e89d7bda8b8b577bfd855f5cb3854234d781855baef82cb057b0529e uname -a
```
runs the container without needing to pull anything.

Given the current `dir:` format, do you see any issues with this way of merging?

Is this something that can be added to the tooling (particularly, skopeo, either as part of the copy command or separately as a "merge" command?

mtrmac:
[...]

PR #1987, opened 2023-05-02
RFC: copy: add --multi-arch=sparse:...

nalind:
Add a "sparse:..." value for the copy --multi-arch option, where the "..." is a comma-separated list of instance digests, platform specifications, or the literal word "system" to signify the local platform.  Along with https://github.com/containers/image/pull/1938, I think this one of the features described at https://github.com/containers/skopeo/issues/1935#issuecomment-1458649836, give or take the presence of "[" and "]" around the list.  If it looks about right, I'll see what sorts of tests I can add to the PR.

### DIFF
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -5,8 +5,10 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"runtime"
 	"strings"
 
+	"github.com/containerd/containerd/platforms"
 	commonFlag "github.com/containers/common/pkg/flag"
 	"github.com/containers/common/pkg/retry"
 	"github.com/containers/image/v5/copy"
@@ -19,6 +21,8 @@ import (
 	"github.com/containers/image/v5/transports/alltransports"
 	encconfig "github.com/containers/ocicrypt/config"
 	enchelpers "github.com/containers/ocicrypt/helpers"
+	"github.com/opencontainers/go-digest"
+	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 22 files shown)


cmd/skopeo/copy.go
```diff
@@ -99,23 +102,40 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
...
+		// There is no CopyNoImages value in copy.ImageListSelection per se, but
+		// we can get the desired effect by using CopySpecificImages.
+		return copy.CopySpecificImages, nil, nil, nil
+	case strings.HasPrefix(multiArch, "sparse:"):
+		sparseArg := strings.TrimPrefix(multiArch, "sparse:")
+		platformSpecs := strings.Split(sparseArg, ",")
+		var platformList []imgspecv1.Platform
+		var instanceList []digest.Digest
+		for _, platformSpec := range platformSpecs {
+			if instanceDigest, err := digest.Parse(platformSpec); err == nil {
```
mtrmac, line 122: Hum… I’m instinctively not a fan of “if it parses like A, it is A, else it is B” formats; OTOH it _can_ be a convenient UI.

(That… kinda applies to the `system` string, but that would be absurd.)

---

Maybe write some kind of an explicit rule that digests contain a `:`, but non-digests don’t contain it? I didn’t read all of the `platforms` package to see whether that would work (OTOH if it didn’t, we… might be in trouble here)

cmd/skopeo/copy.go
```diff
@@ -99,23 +102,40 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
...
+				instanceList = append(instanceList, instanceDigest)
+				continue
+			}
+			if platformSpec == "system" {
+				platformList = append(platformList, imgspecv1.Platform{})
```
mtrmac, line 127: `imgspecv1.Platform{}` does not generally mean “the current platform”.

… it does seem to mean that in the current implementation. To me, that just seems surprising and risky long-term.

(Compare https://github.com/containers/image/pull/1938 — I think _ideally_ I’d like the “turn `system` into a specific choice” logic to live in c/image.)

---

Actually, I don’t think this works as is. `skopeo copy --override-arch amd64 --multi-arch=sparse:system` should, I think, copy the amd64 image, not the image for `runtime.GOARCH` as the two PRs currently do if I’m reading them right. I’ll need to revisit this tomorrow…

cmd/skopeo/copy.go
```diff
@@ -99,23 +102,40 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
...
+			} else {
+				p, err := platforms.Parse(platformSpec)
```
mtrmac, line 129: `platforms.Parse`, if I’m reading it correctly, turns `"linux"` into `linux/$GOARCH`. Or, alternatively, `"amd64"` into `$GOOS/amd64`.

Hum, do we want that? Especially for Mac users?

---

Actually, should the UI be built around the OCI “platform” triplet at all, or should we have OS/arch filters separately? (And what does that mean for Zstd compression? Probably nothing…)

mtrmac (commented):
Code is fine (with the comparatively minor question of `system` semantics)… but then I got stuck on the `platforms.Parse` part below — and I’ll need to leave for the day anyway.

Quite possibly you have already thought about these things and there’s a good reason to interpret the inputs this way.

cmd/skopeo/copy.go
nalind: You're right - we choose the current platform for "system", ignoring `--override-arch`, which doesn't feel right.  Using a Platform built using overrideOS/overrideArch/overrideVariant seems to match the intent better.

cmd/skopeo/copy.go
nalind: > `platforms.Parse`, if I’m reading it correctly, turns `"linux"` into `linux/$GOARCH`. Or, alternatively, `"amd64"` into `$GOOS/amd64`.
> 
> Hum, do we want that? Especially for Mac users?

Hmm, for people on GOOS="darwin" I could actually see defaulting to "linux" as being reasonable, but for people running on GOOS="windows" I think it would be surprising.

mtrmac: One more option is to just provide the full-platform filter now (with a more strict parser requiring a full specification?), and leave a nicer arch-filter-only UI for a future improvement — as long as we can see a path to making that possible, and, uh, `--multi-arch=sparse2:…` is a last-resort option we always have.

“A nice arch-filter-only UI” is not what the primary motivating issue is asking for right now…

nalind:
Took a swing at that sparse syntax.  it's not super-pretty, but I tried to reduce the opportunity for copy-paste errors.

cmd/skopeo/copy.go
```diff
@@ -99,24 +103,122 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
...
+		}); err != nil {
+			return nil, nil, err
+		} else if isDigest {
+			continue
+		}
+		if isArch, err := parseArg("arch=[", "]", func(archSpecList string) (bool, error) {
+			wantedOS := runtime.GOOS
+			if globalOptions.overrideOS != "" {
+				wantedOS = globalOptions.overrideOS
+			}
```
mtrmac, line 190: If Mac users must use `--override-os=linux --multi-arch=sparse:arch=[amd64,arm64]`, that’s not really any simpler than `--multi-arch=sparse:platform=[linux/amd64,linux/arm64]`.

I think if for an `arch`, separate and distinct from `platform`, to be worth adding (extra complexity in implementation and for users to read about), it should be an arch-only filter, not restricting the OS at all.

I also don’t think we need to build that arch-only filter at all for the first version… see the top-level comment.

cmd/skopeo/copy.go
```diff
@@ -99,24 +103,122 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
...
+			return true, nil
+		}); err != nil {
+			return nil, nil, err
+		} else if isArch {
+			continue
+		}
+		if isPlatform, err := parseArg("platform=[", "]", func(platformSpecList string) (bool, error) {
+			for _, platformSpec := range strings.Split(platformSpecList, ",") {
+				p, err := platforms.Parse(platformSpec)
```
mtrmac, line 207: If we are to ever have a separate `arch=` and `platform=`, at most one of them (possibly neither) should default to `runtime.GOOS`.

Requiring the user to always provide os/arch (with an optional variant) here, using a custom parser instead of `platforms.Parse`, would for now provide a clearly deterministic feature, while still giving us the flexibility to addd that defaulting to `GOOS` in the future.

mtrmac (commented):
As a matter of picking a design for `--multi-arch=sparse:`, sure, this works as a direction/plan.

---

WRT the semantics of the UI, the thing I _conceptually_ struggle with here is that “filter for these architectures” is not really “filter for these architectures _and an implied operating system_. But then “filter for architectures, ignore OS” would need to be modeled as an entire new API in `copy.Options`, and we almost certainly don’t need all that implementation complexity of doing both `platform` and `arch` _in the first version_ if we don’t have known high-priority users.

Maybe we can, for now, just implement `digest` and `platform` (certainly allowing _any_ use case), with a custom `platform` parser that requires os/arch (and optionally allows variant), to avoid the defaulting surprises of https://github.com/containers/skopeo/pull/1987#discussion_r1183115878 .

Or, alternatively, implement `digest`+`arch` and not worry about `platform`, for a more convenient UI probably also handling most practical uses (multi-OS images are hard for me to imagine currently…). That would require a new filter implementation in `c/image/copy`, the existing `ChooseInstance` can’t do it.

---

Implementing “system” cleanly might be simplest by adding an extra boolean to c/image, `copy.Options.InstanceIncludeSystem`. [Or, alternatively, we could over-engineer `copy.Options.InstanceSelectors` that allows multiple various criteria, but that seems way more of an effort than necessary.]

cmd/skopeo/copy.go
```diff
@@ -99,24 +103,122 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
...
+	}
+}
+
+func parseMultiArchSparse(globalOptions *globalOptions, sparseArg string) ([]imgspecv1.Platform, []digest.Digest, error) {
+	var platformList []imgspecv1.Platform
+	var instanceList []digest.Digest
+	remainder := "," + sparseArg
+	for remainder != "" {
+		parseArg := func(argStart string, argEnd string, fn func(argVal string) (bool, error)) (bool, error) {
+			if newRemainder, isArg := strings.CutPrefix(remainder, ","+argStart); isArg {
```
mtrmac, line 135: `CutPrefix` was added in Go 1.20; currently we declare Go 1.18.

(It is also very neat, I can’t wait for this to be available…)

We are not necessarily stuck on 1.18, and updating is an option. But looking at https://bodhi.fedoraproject.org/updates/?packages=golang , we might be limited to 1.19.

nalind: Doh, good catch.

mgs255:
@nalind @mtrmac   Is there any chance of this being picked up again?   This feature would certainly be very useful for our use case, where we are only interested in two platforms `linux/amd64` and `linux/arm64`.  Any interface which would allow copying only those two platforms, re-writing the manifest if required would be a fantastic addition.   To me personally, the proposal of `--multi-arch=sparse:platform=linux/amd64,linux/arm64` seems pretty clear and understandable.

Not sure that it has been mentioned already, but there would be an open question as to the behaviour if the source only supported a sub-set of the specified sparse filters.   Here, it would probably make sense to warn on each of the unmatched filters, but error if there were none that matched.

cmd/skopeo/copy.go
mtrmac: #2175 also suggests using `platforms.Parse`, with the same potential surprises.

Both PRs should probably end up with the same decision.

snail2sky:
This RFC is really great, but unfortunately the author has forgotten it, so I decided to fork it to solve my problem.
Thanks  @nalind @mtrmac for the great analysis

nalind:
I should at least clean up the containers/image part.

snail2sky:
> I should at least clean up the containers/image part.

I think your RFC is very good, but the maintainers of skopeo think that they haven't thought about how to make it more elegant, so this function has never been online.
In fact, there is no need to think so much, and this is not a function from zero to one. Docker manifest supports this function, but it is difficult to use.
Referring to the docker manifest, the usage of docker pull --platfrom is actually sufficient.

aguidirh:
@nalind is this work related to the one implemented in #2874 ? 

Could you please have a look if the PR #2874 covers all things related to this PR so we can close #1987?

nalind:
> @nalind is this work related to the one implemented in #2874 ?

It was an attempt at the same sort of thing, yes.
 
> Could you please have a look if the PR #2874 covers all things related to this PR so we can close #1987?

#2874 does not appear to accept instance digests.

PR #1991, opened 2023-05-03
[release-1.11] Cirrus: Add CI self-destruct condition on EOL date

cevich:
This branch will never receive any security-backports when the associated RHEL release reaches EOL.  Add a condition to force CI to break with a helpful message, after this RHEL EOL date.

### DIFF
--- a/contrib/cirrus/runner.sh
+++ b/contrib/cirrus/runner.sh
@@ -6,6 +6,17 @@
 
 set -e
 
+_EOL=20270501
+if [[ $(date +%Y%m%d) -ge $_EOL ]]; then
+    die "As of $_EOL this branch is probably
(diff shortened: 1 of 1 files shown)


mtrmac (commented):
This is assuming that the 9.3/8.9 RHELs will come from a future new branch, do I understand that correctly? Otherwise the last possible EOL seem to be further in the future.

(If that’s the intent, that’s fine; I agree it’s a good habit to have the `_EOL` code on all branches; we can always move the date out.)

mtrmac:
Actually, https://gitlab.com/redhat/centos-stream/rpms/skopeo/-/tree/c9s is now at 1.12, so a future branch will almost certainly be post-1.11, making the (already fairly hypothetical) concern of future 1.11-based branches moot.

PR #1995, opened 2023-05-05
Update c/image with https://github.com/containers/image/pull/1944

mtrmac:
... to update github.com/opencontainers/image-spec to v1.1.0-rc3.

Replaces #1984 .


mtrmac:
The `ostree-rs-ext` test is failing because it is using `golang-1.18.7-1.fc36.x86_64`, but the new code requires Go 1.19.

Meanwhile, Go https://src.fedoraproject.org/rpms/golang lists 1.19.8 as available.

It seems that the base image of the container we use is more fresh, so I’m trying to trigger an update.

rhatdan:
LGTM

vrothberg (approved):
LGTM

PR #2011, opened 2023-06-02
[CI:BUILD] Packit: add jobs for downstream Fedora package builds

lsm5:
Get rid of `skopeo.spec.rpkg` in favour of
`rpm/skopeo.spec` which gets synced with fedora dist-git on every upstream release. The version in the new spec file is set to `0` by default and gets updated by packit automatically on every packit task.

Packit will automatically create a PR on fedora dist-git on every new upstream release. A sample PR will look like:
https://src.fedoraproject.org/rpms/container-selinux/pull-request/10#

A dry run for this can be triggered using:
`$ packit propose-downstream --local-content`

To run this command locally, you would need to have your packit user-configuration-file set.
Ref: https://packit.dev/docs/configuration/#user-configuration-file

along with a fedora api key created at:
https://src.fedoraproject.org/settings#nav-api-tab with sufficient ACLs.

This patch incurs zero additional cost to upstream as the downstream
tasks occur only after upstream release and do not block anything
upstream.

[NO NEW TESTS NEEDED]


lsm5:
@mtrmac @vrothberg @rhatdan @edsantiago @cevich PTAL

lsm5:
Successful copr build at: https://copr.fedorainfracloud.org/coprs/rhcontainerbot/packit-builds/build/6058022/

rpm/update-spec-provides.sh
```diff
@@ -0,0 +1,17 @@
...
+
+set -e
+
+PACKAGE=$(basename $(git rev-parse --show-toplevel))
+# script is run from git root directory
+SPEC_FILE=rpm/$PACKAGE.spec
+
+sed -i '/Provides: bundled(golang.*/d' $SPEC_FILE
+
+GO_IMPORTS=$(golist --imported --package-path github.com/containers/$PACKAGE --skip-self | sort -u | xargs -I{} echo "Provides: bundled(golang({}))")
```
cevich, line 15: I'm not sure in this case, but might confirm that `-I{}` doesn't need to be quoted to protect it from bash gobbling it up.

rpm/skopeo.spec
```diff
@@ -0,0 +1,194 @@
...
+%{_datadir}/zsh/site-functions/_%{name}
+
+%files tests
+%license LICENSE
+%{_datadir}/%{name}/test
+
+%changelog
+%if %{with manual_changelog}
+* Mon May 01 2023 RH Container Bot <rhcontainerbot@fedoraproject.org>
+- Placeholder changelog for envs that are not autochangelog-ready
```
cevich, line 191: Since it's a placeholder it's probably minor, but is `RH Container Bot <rhcontainerbot@fedoraproject.org>` still relevant here (since automation is going to packit)?  Might consider updating this (and the date) if not.

rpm/skopeo.spec
```diff
@@ -0,0 +1,194 @@
...
+
+%ifarch x86_64
+export CGO_CFLAGS="$CGO_CFLAGS -m64 -mtune=generic -fcf-protection=full"
+%endif
+
+export GOPATH=$(pwd)/_build:$(pwd)
+export CGO_CFLAGS=$CFLAGS
+
+%if %{without btrfs}
+export BUILDTAGS="exclude_graphdriver_btrfs btrfs_noversion"
```
cevich, line 142: In `.cirrus.yml` the BUILDTAGS are `'btrfs_noversion libdm_no_deferred_remove containers_image_openpgp`.  Is the value in the spec file correct?

cevich (approved):
LGTM - other than a few total-guesses at things that might need fixing.

rpm/skopeo.spec
lsm5: Changelog entry was copy-pasta from podman spec. This changelog entry is only so that copr builds on centos 8 stream do not complain about no changelog. The copr repos are still owned by `rhcontainerbot` user. 

The changelog entry is not expected to be updated regularly, and might even go away once centos 8 stream gets autochangelog support.

rpm/skopeo.spec
lsm5: hmm, this is what we've had in the fedora spec file all this while. Let me get back on this.

lsm5: I guess I can just include what the upstream makefile includes by default. Looking in the current [skopeo.spec.rpkg](https://github.com/containers/skopeo/blob/main/skopeo.spec.rpkg#L96) on main, it only has libdm_tag.sh apart from the btrfs tags.

lsm5: I can try to use Makefile targets for rpm builds as well and see how those go.

rpm/skopeo.spec
```diff
@@ -0,0 +1,194 @@
...
+%description tests
+%{summary}
+
+This package contains system tests for %{name}
+
+%prep
+%autosetup -Sgit %{name}-%{version}
+sed -i 's/install-binary: bin\/%{name}/install-binary:/' Makefile
+sed -i 's/completions: bin\/%{name}/completions:/' Makefile
+sed -i 's/install-docs: docs/install-docs:/' Makefile
```
edsantiago, line 125: These squick me out a little. Best I can tell, they're removing dependencies. Can you document why? And maybe use `.*` instead of knowing the rest of the command line? And maybe also anchor with `^`?

lsm5: fixed in upcoming..

rpm/skopeo.spec
```diff
@@ -0,0 +1,194 @@
...
+cd ..
+
+mv vendor src
+
+%gobuild -o bin/%{name} %{import_path}/cmd/%{name}
+%{__make} docs
+
+%install
+make \
+    PREFIX=%{buildroot}%{_prefix} \
```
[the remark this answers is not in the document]
lsm5: PREFIX in makefile is set to /usr/local, so that's why setting `PREFIX=%{buildroot}%{_prefix}` takes of DESTDIR too, but anyway, changed this to `DESTDIR=%{buildroot} PREFIX=%{_prefix}` in upcoming..

rpm/skopeo.spec
```diff
@@ -0,0 +1,194 @@
...
+CGO_CFLAGS=$(echo $CGO_CFLAGS | sed 's/-flto=auto//g')
+CGO_CFLAGS=$(echo $CGO_CFLAGS | sed 's/-Wp,D_GLIBCXX_ASSERTIONS//g')
+CGO_CFLAGS=$(echo $CGO_CFLAGS | sed 's/-specs=\/usr\/lib\/rpm\/redhat\/redhat-annobin-cc1//g')
+
+%ifarch x86_64
+export CGO_CFLAGS="$CGO_CFLAGS -m64 -mtune=generic -fcf-protection=full"
+%endif
+
+export GOPATH=$(pwd)/_build:$(pwd)
+export CGO_CFLAGS=$CFLAGS
```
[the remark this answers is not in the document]
lsm5: great catch! fixed in upcoming..

rpm/skopeo.spec
```diff
@@ -0,0 +1,194 @@
...
+
+%if 0%{?rhel}
+%bcond_with btrfs
+%else
+%bcond_without btrfs
+%endif
+
+# go-rpm-macros package and autochangelog exists for fedora and rhel9
+%if 0%{?fedora} || 0%{?rhel} >= 9
+%bcond_without go_rpm_macros
```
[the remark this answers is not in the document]
lsm5: these should be read in the opposite sense actually. See: https://rpm-software-management.github.io/rpm/manual/conditionalbuilds.html  . If you think that's gonna keep confusing people long term, I could probably switch to simply `%bcond` or some other conditional and get rid of the `with and without`. I added a note in the spec along with the link.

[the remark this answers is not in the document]
mtrmac: My mistake, thanks for the pointer.

I think this counts as “expected RPM knowledge”, no need to make the actual spec file mechanism more complicated for ignorant readers like me.

lsm5:
I'm gonna try disabling `.packit.sh` and see how things go. Please hold off further reviews.

rpm/skopeo.spec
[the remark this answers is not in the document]
mtrmac: Yes, right now this does not matter, because the spec file does not run the top-level `install` target.

(That target writes into `$destdir/etc`, so overwriting PREFIX would not work.)

At this point it’s just a matter of precisely matching the semantics to avoid the risk of future surprises.

.packit.sh
```diff
@@ -5,22 +5,28 @@
...
+
+# RPM Version can't take "-"
+RPM_VERSION=$(echo $VERSION | sed -e 's/-/~/')
 
 # Generate source tarball
-git archive --prefix=skopeo-$VERSION/ -o skopeo-$VERSION.tar.gz HEAD
+git archive --prefix=$PACKAGE-$RPM_VERSION/ -o $PACKAGE-$VERSION.tar.gz HEAD
+
+# rpmbuild expects source tarball in the same dir as spec file
+ln -s $PACKAGE-$VERSION.tar.gz rpm/
```
mtrmac:
```suggestion
ln -s ../$PACKAGE-$VERSION.tar.gz rpm/
```
I think

rpm/skopeo.spec
```diff
@@ -0,0 +1,206 @@
...
+# These extra flags present in $CFLAGS have been skipped for now as they break the build
+CGO_CFLAGS=$(echo $CGO_CFLAGS | sed 's/-flto=auto//g')
+CGO_CFLAGS=$(echo $CGO_CFLAGS | sed 's/-Wp,D_GLIBCXX_ASSERTIONS//g')
+CGO_CFLAGS=$(echo $CGO_CFLAGS | sed 's/-specs=\/usr\/lib\/rpm\/redhat\/redhat-annobin-cc1//g')
+
+%ifarch x86_64
+export CGO_CFLAGS="$CGO_CFLAGS -m64 -mtune=generic -fcf-protection=full"
+%endif
+
+export BASEBUILDTAGS="$(hack/libdm_tag.sh) $(hack/libsubid_tag.sh)"
```
mtrmac, line 146: (Non-blocking: If the variable is only consumed immediately below, `export` seems unnecessary.)

rpm/skopeo.spec
```diff
@@ -0,0 +1,194 @@
...
+# unset LDFLAGS earlier set from set_build_flags
+LDFLAGS=''
+
+mkdir _build
+cd _build
+mkdir -p src/%{provider}.%{provider_tld}/%{project}
+ln -s ../../../../ src/%{import_path}
+cd ..
+
+mv vendor src
```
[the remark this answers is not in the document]
lsm5: This is needed for the non-copr official koji builds where the build envs don't have network access.

Without this, I see a build failure like: https://kojipkgs.fedoraproject.org//work/tasks/369/102110369/build.log

rpm/skopeo.spec
```diff
@@ -0,0 +1,194 @@
...
+# These extra flags present in $CFLAGS have been skipped for now as they break the build
+CGO_CFLAGS=$(echo $CGO_CFLAGS | sed 's/-flto=auto//g')
+CGO_CFLAGS=$(echo $CGO_CFLAGS | sed 's/-Wp,D_GLIBCXX_ASSERTIONS//g')
+CGO_CFLAGS=$(echo $CGO_CFLAGS | sed 's/-specs=\/usr\/lib\/rpm\/redhat\/redhat-annobin-cc1//g')
+
+%ifarch x86_64
+export CGO_CFLAGS="$CGO_CFLAGS -m64 -mtune=generic -fcf-protection=full"
+%endif
+
+export GOPATH=$(pwd)/_build:$(pwd)
```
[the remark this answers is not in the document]
lsm5: looks like it still is. See comment: https://github.com/containers/skopeo/pull/2011#discussion_r1228609843

.packit.sh
```diff
@@ -5,22 +5,28 @@
...
+
+# RPM Version can't take "-"
+RPM_VERSION=$(echo $VERSION | sed -e 's/-/~/')
 
 # Generate source tarball
-git archive --prefix=skopeo-$VERSION/ -o skopeo-$VERSION.tar.gz HEAD
+git archive --prefix=$PACKAGE-$RPM_VERSION/ -o $PACKAGE-$VERSION.tar.gz HEAD
+
+# rpmbuild expects source tarball in the same dir as spec file
+mv $PACKAGE-$VERSION.tar.gz rpm/
```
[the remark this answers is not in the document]
lsm5: I'm so happy you made this comment. I see that with the spec file switch: `skopeo.spec.rpkg`  -> `rpm/skopeo.spec` we can just use packit's defaults, so we don't need this script. Got rid of it in the latest version of the PR. Thanks a lot!

.packit.sh
lsm5: this file has been deleted.

rpm/skopeo.spec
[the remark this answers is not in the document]
mtrmac: My immediate thought is to blame it on the `GO111MODULE=off`; `-mod=vendor` is documented to be the default and to not hit the network.

But I didn't actually test that for a long time, I might get to experiment next week.

rpm/skopeo.spec
lsm5: thanks! updated.

rpm/skopeo.spec
lsm5: giving up on Makefile targets in %build section for now. Adjusted buildtags to match upstream Makefile's.

rpm/skopeo.spec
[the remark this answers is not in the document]
lsm5: The `%gobuild` macro provided by Fedora sets `GO111MODULE=off`.

I guess I could try setting %gobuild in the spec file for all environments, but I'm afraid we would miss out on any changes made by the toolchain team.

lsm5:
> alright, I think I have addressed all comments. Added some additional comments in the spec file and also got rid of `.packit.sh`. Packit's defaults are good enough.
> 
> PTAL.

err, once again I spoke too soon. https://copr.fedorainfracloud.org/coprs/rhcontainerbot/packit-builds/package/skopeo/ something seems wrong with version fetching.

lsm5:
> err, once again I spoke too soon. https://copr.fedorainfracloud.org/coprs/rhcontainerbot/packit-builds/package/skopeo/ something seems wrong with version fetching.

So, using packit's defaults means the Version string is fetched using: `git describe --tags --abbrev=0` and not using the string in `version/version.go`. So the current build has version `1.12.0` and not `1.12.1~dev`.  If that's acceptable, then we should be ok. I will upgrade the `Epoch:102` so existing copr users are not broken.

I guess upstream itself doesn't need to concern itself too much with all this, more of a copr usability and breakage thing.

rpm/skopeo.spec
```diff
@@ -0,0 +1,209 @@
...
+Name: %{repo}
+%if %{with copr}
+Epoch: 102
+%else
+Epoch: %{conditional_epoch}
+%endif
+# DO NOT TOUCH the Version string!
+# The TRUE source of this specfile is:
+# https://github.com/containers/skopeo/blob/main/rpm/skopeo.spec
+# If that's what you're reading, Version must be 0, and will be updated by Packit for
```
mtrmac, line 66: The value is not `0` right now, maybe the comment needs updating.

lsm5:
> I will upgrade the `Epoch:102` so existing copr users are not broken.

set `Epoch: 102` to preserve upgrade path for copr users. With that I think we should be good to go.

rpm/skopeo.spec
```diff
@@ -0,0 +1,209 @@
...
+%if %{with btrfs}
+export BUILDTAGS="$BASEBUILDTAGS $(hack/btrfs_tag.sh) $(hack/btrfs_installed_tag.sh)"
+%else
+export BUILDTAGS="$BASEBUILDTAGS btrfs_noversion exclude_graphdriver_btrfs"
+%endif
+
+# unset LDFLAGS earlier set from set_build_flags
+LDFLAGS=''
+
+# These steps are needed for Koji builds
```
mtrmac, line 156: This does not quite help explain the rationale. Admittedly the PR discussion document that now; still, if we have a comment, it seems worth to make that comment useful stand-alone.

Maybe something like
> We want to use the Fedora-provided %%gobuild, which sets `GO111MODULE=off; so we need to set up a
> GOPATH environment to use the vendored dependencies instead of trying to download them from the
> network (which is prohibited by Koji)

?

mtrmac (commented):
> So the current build has version 1.12.0 and not 1.12.1~dev. If that's acceptable, then we should be ok.

The currently-generated version numbers use the snapshot syntax, that’s a clear enough indication of what the sources are.

rpm/skopeo.spec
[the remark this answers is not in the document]
mtrmac: https://pagure.io/go-rpm-macros/issue/28 suggests just overriding that.


Intuitively going back to managing `GOPATH` manually seems like losing progress, but I really don’t have the full context to appreciate which of the two approaches is, overall, more workable or maintainable. I’ll fully leave that up to you.

rpm/skopeo.spec
lsm5: SGTM, will update. Thanks

rpm/skopeo.spec
lsm5: argh, I'll fix that. thanks

rpm/skopeo.spec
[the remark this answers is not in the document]
lsm5: let me give that a quick try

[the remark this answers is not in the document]
lsm5: @mtrmac that worked!! Thanks for bothering to look that up :) . I had stuck around with it for too long to care. Spec file is much simpler now. Other comments addressed as well.

lsm5:
@mtrmac all comments should be addressed now. Marking ready for review. PTAL.

mtrmac (approved):
LGTM. Thanks!

Issue #1648, opened 2022-05-19
Unable to locate package skopeo on Ubuntu 18.04 & 20.04


Hans-J-Klose:
Starting having issues pulling Skopeo into my Ubuntu 18.04 images starting last night.  How can we get `skopeo` to install?

```
# apt-get install -y skopeo
Reading package lists... Done
Building dependency tree       
Reading state information... Done
E: Unable to locate package skopeo
```

felipecrs:
I have the same issue. Also, unable to install `podman`, although the error is a little different:

```console
❯ sudo apt install podman -y
Reading package lists... Done
Building dependency tree
Reading state information... Done
The following additional packages will be installed:
  catatonit conmon containernetworking-plugins containers-common criu crun dns-root-data dnsmasq-base fuse-overlayfs
  libidn11 libnet1 libnl-3-200 libprotobuf-c1 libprotobuf17 libyajl2 podman-machine-cni podman-plugins python3-protobuf
  uidmap
The following NEW packages will be installed:
  catatonit conmon containernetworking-plugins containers-common criu crun dns-root-data dnsmasq-base fuse-overlayfs
  libidn11 libnet1 libnl-3-200 libprotobuf-c1 libprotobuf17 libyajl2 podman podman-machine-cni podman-plugins
  python3-protobuf uidmap
0 upgraded, 20 newly installed, 0 to remove and 0 not upgraded.
Need to get 18.0 MB/29.8 MB of archives.
After this operation, 151 MB of additional disk space will be used.
Get:1 https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_20.04  crun 100:1.2-2 [252 kB]
Err:1 https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_20.04  crun 100:1.2-2
  File has unexpected size (251724 != 251684). Mirror sync in progress? [IP: 190.64.49.124 443]
  Hashes of expected file:
   - SHA256:1c2a3e0d19b27765b7067500bf22bfe0cff2eeefc6a9d2a42150b348b802faba
   - SHA1:c7bab007b17a547540293ba8ad2c63e734d38e6d [weak]
   - MD5Sum:43a01bf8b518728b1981bb9c388144dc [weak]
   - Filesize:251684 [weak]
[...]

PR #2014, opened 2023-06-07
Embed default-policy into binary

felipecrs:
This should make the usage of the skopeo statically linked binary a lot simpler.

- Refs https://github.com/containers/skopeo/issues/1648#issuecomment-1580905503


felipecrs:
I tested it, and it works nice.

felipecrs:
The build errors doesn't seem related:

```
RPM build errors:
    source_date_epoch_from_changelog set but %changelog is missing
    File not found: /builddir/build/BUILDROOT/skopeo-1.12.1~dev-1.20230607224542237766.pr2014.36.gfb180f03.fc39.aarch64/usr/share/man/man1/%{name}*
Finish: rpmbuild skopeo-1.12.1~dev-1.20230607224542237766.pr2014.36.gfb180f03.fc39.src.rpm
```

felipecrs:
How about hide this "feature" behind a compile flag, so that only users building the static binary can set it?

mtrmac:
A static binary having a different security behavior from a dynamic one would be an even more risky footgun.

mtrmac:
Also consider the supply chain security risks of individually distributed static binaries.

Distribution packages really are strongly recommended over that, for many reasons.

Alternatively there is https://github.com/containers/skopeo/tree/main/contrib/skopeoimage .

mtrmac:
My opinion on static binaries is not nearly as strong as my opinion that the security enforcing code needs to be paranoid about both bugs and human operators’ mistakes.

E.g. users could intend to ship a config file by something like Ansible, but make a typo in the path. Or a mount command could fail.

felipecrs:
I understand. That makes sense.

mtrmac:
Of course users are free to patch their own systems, or to wrap Skopeo in a shell script that always adds `--insecure-policy`.

And ignoring signatures may well be the right call for some environments.

I just don’t think it’s the right trade-off for wide distribution.

PR #2026, opened 2023-06-28
[CI:BUILD] RPM: fix ELN builds

lsm5:
For Fedora, we need to ensure ELN builds are successful.

The recent skopeo builds have been failing on copr because of /usr/bin/go-md2man being available via different package names. https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/build/6119200/

This commit changes go-md2man dependency directly on the binary path instead of a pacakge.

ELN scratch builds are now successful:
https://koji.fedoraproject.org/koji/taskinfo?taskID=102706281

### DIFF
--- a/rpm/skopeo.spec
+++ b/rpm/skopeo.spec
@@ -52,7 +52,7 @@ Summary: Inspect container images and repositories on registries
 URL: https://github.com/containers/%{name}
 # Tarball fetched from upstream
 Source0: %{url}/archive/v%{version}.tar.gz
-BuildRequires: go-md2man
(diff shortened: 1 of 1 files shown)


lsm5:
I'll add ELN builds to the CI task list for PRs once everything has been known to be smooth in official Koji builds and `after-commit` copr builds. Mainly because, adding them to the task list and then disabling them for whatever reason can be a lot of trouble especially when dealing with multiple repos.

mtrmac (approved):
Thanks!

PR #2029, opened 2023-06-29
proxy: Policy verification of OCI Image before pulling

RishabhSaini:
Helps address the lack of policy verification in rpm-ostree:
https://github.com/coreos/rpm-ostree/issues/4272

An Image pull triggered by `rpm-ostree rebase` will look to verify the policy by default as a part of `OpenImage` 

The other route taken here adds a new API call
https://github.com/containers/skopeo/issues/1829

### DIFF
--- a/cmd/skopeo/proxy.go
+++ b/cmd/skopeo/proxy.go
@@ -75,6 +75,7 @@ import (
 	"github.com/containers/image/v5/manifest"
 	ocilayout "github.com/containers/image/v5/oci/layout"
 	"github.com/containers/image/v5/pkg/blobinfocache"
+	"github.com/containers/image/v5/signature"
 	"github.com/containers/image/v5/transports"
 	"github.com/containers/image/v5/transports/alltransports"
 	"github.com/containers/image/v5/types"
@@ -95,7 +96,8 @@ import (
 // 0.2.3: Added GetFullConfig
 // 0.2.4: Added OpenImageOptional
 // 0.2.5: Added LayerInfoJSON
-const protocolVersion = "0.2.5"
+// 0.2.6: Policy Verification before pulling OCI
+const protocolVersion = "0.2.6"
 
 // maxMsgSize is the current limit on a packet size.
 // Note that all non-metadata (i.e. payload data) is sent over a pipe.
@@ -264,6 +266,23 @@ func (h *proxyHandler) openImageImpl(args []any, allowNotFound bool) (replyBuf, 
 			return ret, nil
 		}
 		return ret, err
+	}
+
+	unparsedTopLevel := image.UnparsedInstance(imgsrc, nil)
+	policy, err := signature.DefaultPolicy(h.sysctx)
+	if err != nil {
+		return ret, err
+	}
+	policyContext, err := signature.NewPolicyContext(policy)
+	if err != nil {
+		return ret, err
+	}
+	allowed, err := policyContext.IsRunningImageAllowed(context.Background(), unparsedTopLevel)
+	if !allowed || err != nil {
+		return ret, err
+	}
+	if !allowed && err == nil {
+		return ret, fmt.Errorf("policy verification failed unexpectedly")
 	}
 
 	// Note that we never return zero as an imageid; this code doesn't yet


RishabhSaini:
The above change causes the following output if a container image signed by sigstore is given the wrong key:
```
[root@cosa-devsh ~]# rpm-ostree rebase ostree-image-signed:docker://ghcr.io/ahgencer/silverblue:pr-3
Pulling manifest: ostree-image-signed:docker://ghcr.io/ahgencer/silverblue:pr-3
error: Creating importer: Failed to invoke skopeo proxy method OpenImage: remote error: cryptographic signature verification failed: invalid signature when validating ASN.1 encoded signature
```

cgwalters (approved):
Yeah, OK I think this makes sense to just do by default.

Are there cases where we want the equivalent of `skopeo inspect --insecure-policy`?  Hmm...perhaps not.

cgwalters:
As a procedural note I've created [experimental-image-proxy label](https://github.com/containers/skopeo/labels/experimental-image-proxy) to track PRs to this subsystem.

RishabhSaini:
Rebasing to sync the branch

RishabhSaini:
Unsure why the CI is failing "timed out". The tests for ostree-rs-ext passed on the previous run and work locally on my end. Would the owners kindly re-rerun the ostree-rs-ext as I suspect it was a flake

cgwalters:
There's a "Re-run" button next to the check...I clicked it, but so far not seeing a change?
Anyways you should be able to force it with `git commit --amend --allow-empty` or so and re-pushing

mtrmac:
FWIW we do see the `ostree-rs-ext` timing out fairly frequently. Either it takes just a few minutes, or it times out after 30.

cmd/skopeo/proxy.go
```diff
@@ -266,6 +268,20 @@ func (h *proxyHandler) openImageImpl(args []any, allowNotFound bool) (replyBuf,
...
+	if err != nil {
+		return ret, err
+	}
+	policyContext, err := signature.NewPolicyContext(policy)
+	if err != nil {
+		return ret, err
+	}
+	allowed, err := policyContext.IsRunningImageAllowed(context.Background(), unparsedTopLevel)
+	if !allowed || err != nil {
+		return ret, err
```
mtrmac, line 282: (A drive-by comment without fully reviewing the history.)

If `!allowed && err == nil`, this should probably fail (loudly about an unexpected state?), not return a success indication.

cgwalters:
> FWIW we do see the ostree-rs-ext timing out fairly frequently. Either it takes just a few minutes, or it times out after 30.

Ah, dang.  OK, I'm reopening https://github.com/ostreedev/ostree-rs-ext/issues/322#event-9719545551

cmd/skopeo/proxy.go
cgwalters: We're not doing that in the other callees though AFAICS, but agree with the principle for security-relevant code.

But hmm, why does the API here not just return `error`?

RishabhSaini: > (A drive-by comment without fully reviewing the history.)
> 
> If `!allowed && err == nil`, this should probably fail (loudly about an unexpected state?), not return a success indication.

OK updated

RishabhSaini:
Updated

cgwalters (approved):

cmd/skopeo/proxy.go
```diff
@@ -266,6 +268,23 @@ func (h *proxyHandler) openImageImpl(args []any, allowNotFound bool) (replyBuf,
 		return ret, err
 	}
 
+	unparsedTopLevel := image.UnparsedInstance(imgsrc, nil)
+	policy, err := signature.DefaultPolicy(h.sysctx)
```
cgwalters, line 272: While I think this is OK for now, maybe it makes sense to cache this in the future?

RishabhSaini:
> Did you do the whole thing with setting up e.g. cosign and requiring it in the container policy?

Yes, I have tested this system by singing and uploading a signed container image to my registry using cosign and skopeo. Then rebasing from the remote registry behaves appropriately.

cmd/skopeo/proxy.go
```diff
@@ -266,6 +268,23 @@ func (h *proxyHandler) openImageImpl(args []any, allowNotFound bool) (replyBuf,
...
+	if err != nil {
+		return ret, err
+	}
+	policyContext, err := signature.NewPolicyContext(policy)
+	if err != nil {
+		return ret, err
+	}
+	allowed, err := policyContext.IsRunningImageAllowed(context.Background(), unparsedTopLevel)
```
cgwalters, line 280: Here'd what we want is a new API like `IsRunningImageSigned` and it would fail if the signature policy is `insecureAcceptAnything` for https://github.com/ostreedev/ostree-rs-ext/issues/79

cmd/skopeo/proxy.go
mtrmac: > But hmm, why does the API here not just return `error`?

Just excessive paranoia. Looking at call sites like this one, I’m no longer sure this was a good tradeoff at all. But, well, we’re stuck with it.

cmd/skopeo/proxy.go
mtrmac: Ideally, cache the `PolicyContext` (as long as it is used from a single thread).

(Right now that does not make a difference, but a branch caching the imported keys in a native format inside a `PolicyContext` is floating around.)

mtrmac, line 272: Was there a specific reason not to use `opts.global.getPolicyContext` here?

That would also immediately handle `--insecure-policy` per https://github.com/containers/skopeo/pull/2029#pullrequestreview-1506680929 .

cmd/skopeo/proxy.go
```diff
@@ -266,6 +268,23 @@ func (h *proxyHandler) openImageImpl(args []any, allowNotFound bool) (replyBuf,
...
 
+	unparsedTopLevel := image.UnparsedInstance(imgsrc, nil)
+	policy, err := signature.DefaultPolicy(h.sysctx)
+	if err != nil {
+		return ret, err
+	}
+	policyContext, err := signature.NewPolicyContext(policy)
+	if err != nil {
+		return ret, err
+	}
```
mtrmac, line 279: `policyContext.Destroy` is required by the API when done.

cmd/skopeo/proxy.go
```diff
@@ -266,6 +268,23 @@ func (h *proxyHandler) openImageImpl(args []any, allowNotFound bool) (replyBuf,
...
+	allowed, err := policyContext.IsRunningImageAllowed(context.Background(), unparsedTopLevel)
+	if !allowed || err != nil {
+		return ret, err
+	}
+	if !allowed && err == nil {
```
mtrmac, line 284: This is dead code AFAICS. The code has already returned `ret, nil` above.

mtrmac (commented):
A late review I’m afraid.

mtrmac:
> A late review I’m afraid.

Should be resolved by #2048.

PR #2041, opened 2023-07-07
[CI:BUILD] Packit: install golist before updating downstream spec

lsm5:
The default Packit sandbox environment that runs Packit tasks for
downstream Fedora does not have golist installed by default and can't
run superuser tasks.
    
This commit will download and extract the golist binary from the Fedora
rpm and use it to provide golist.
    
The GOPATH mention in `rpm/update-spec-provides.sh` is only required for
golist to generate the gopaths and doesn't affect upstream or the rpm spec.
    
Currently, the only way to reliably test this is on an open github issue by running
`/packit propose-downstream`. This can't be run on an open PR.
The job-specific packit actions can only be tested via the packit
service and not via packit cli.

@mtrmac @vrothberg PTAL

### DIFF
--- a/rpm/update-spec-provides.sh
+++ b/rpm/update-spec-provides.sh
@@ -5,13 +5,30 @@
 # The goimports don't need to be present upstream.
 
 set -eo pipefail
+set -x
 
 PACKAGE=skopeo
 # script is run from git root directory
 SPEC_FILE=rpm/$PACKAGE.spec
 
+# Needed for golist to work
+# This is run in the packit sandbox environment for downstream tasks and
+# doesn't affect upstream repo and CI.
+export GOPATH=~/go
+GOPATHDIR=$GOPATH/src/github.com/containers/
+mkdir -p $GOPATHDIR
+ln -sf $(pwd) $GOPATHDIR/.
+
+# Packit sandbox doesn't allow root
+# Install golist by downloading and extracting rpm
+# We could handle this in packit `sandcastle` upstream itself
+# but that depends on golist existing in epel
+# https://github.com/packit/sandcastle/pull/186
+dnf download golist
+rpm2cpio golist-*.rpm | cpio -idmv
+
 sed -i '/Provides: bundled(golang.*/d' $SPEC_FILE
 
-GO_IMPORTS=$(golist --imported --package-path github.com/containers/$PACKAGE --skip-self | sort -u | xargs "-I{}" echo "Provides: bundled(golang({}))")
+GO_IMPORTS=$(./usr/bin/golist --imported --package-path github.com/containers/$PACKAGE --skip-self | sort -u | xargs "-I{}" echo "Provides: bundled(golang({}))")
 
 awk -v r="$GO_IMPORTS" '/^# vendored libraries/ {print; print r; next} 1' $SPEC_FILE > temp && mv temp $SPEC_FILE


rpm/update-spec-provides.sh
```diff
@@ -4,12 +4,27 @@
...
 PACKAGE=skopeo
 # script is run from git root directory
 SPEC_FILE=rpm/$PACKAGE.spec
 
+# Needed for golist to work
+# This is run in the packit sandbox environment for downstream tasks and
+# doesn't affect upstream repo and CI.
+GOPATHDIR=~/go/src/github.com/containers/
+mkdir -p $GOPATHDIR
+ln -sf $(pwd) $GOPATHDIR/.
```
mtrmac, line 18: I can’t see `golist`, nor the Go standard library, ever referring to `GOPATHDIR`. (Comparing https://pagure.io/go-rpm-macros/blob/master/f/rpm/macros.d/macros.go-rpm#_187 .)

rpm/update-spec-provides.sh
```diff
@@ -4,12 +4,27 @@
...
+
+# Packit sandbox doesn't allow root
+# Install golist by downloading and extracting rpm
+# We could handle this in packit `sandcastle` upstream itself
+# but that depends on golist existing in epel
+# https://github.com/packit/sandcastle/pull/186
+dnf download golist
+rpm2cpio golist-*.rpm | cpio -idmv
```
mtrmac, line 26: I would expect this to extract to `./usr/bin/golist`, but that’s not in `PATH`, so how does the command get executed?

rpm/update-spec-provides.sh
```diff
@@ -4,12 +4,27 @@
...
+
 sed -i '/Provides: bundled(golang.*/d' $SPEC_FILE
 
 GO_IMPORTS=$(golist --imported --package-path github.com/containers/$PACKAGE --skip-self | sort -u | xargs "-I{}" echo "Provides: bundled(golang({}))")
```
mtrmac, line 30: (BTW looking at what Golist does, it seems that it should be possible to use the standard Go command, something like `go list -deps -f '{{range .Imports}}{{.}}\n{{end}}'  ./...`. I fully support using the same command that `go-rpm-macros` uses, and I know nothing about the environment — quite possibly installing all of Go is even worse than installing `golist` — I just wanted to note the theoretical possibility, now that I took some time to learn what `golist` is.)

mtrmac (commented):
Thanks very much for looking at that failure!

Is there a test run somewhere? I’m afraid know nothing about the environment, still, some of the additions are a bit surprising. It’s very likely that’s just my ignorance.

Or is the best way to deal with this to merge and let the next triggered job prove that the script works, because other testing is not all that convenient? That’s fine with me as well :)

rpm/update-spec-provides.sh
lsm5: I need to switch to `./usr/bin/golist`. Thanks for pointing that out. I'll check if I can verify this in the packit local reproducible environment as well.

rpm/update-spec-provides.sh
lsm5: ack. I haven't checked with packit upstream about this, but golang might be a heavy dependency to include in their sandbox. So, I'll keep golist for now.

lsm5:
> Or is the best way to deal with this to merge and let the next triggered job prove that the script works, because other testing is not all that convenient? That’s fine with me as well :)

There's a local test environment, but I haven't yet fully digested it. I'll work on that first to avoid future surprises. Thanks

mtrmac:
I really know nothing about this; if you as the expert think merging & testing live is the most productive approach, I’m totally fine with it. I just wanted to understand what the plan is.

rpm/update-spec-provides.sh
```diff
@@ -4,14 +4,30 @@
...
+# We could handle this in packit `sandcastle` upstream itself
+# but that depends on golist existing in epel
+# https://github.com/packit/sandcastle/pull/186
+dnf download golist
+rpm2cpio golist-*.rpm | cpio -idmv
+
 sed -i '/Provides: bundled(golang.*/d' $SPEC_FILE
 
-GO_IMPORTS=$(golist --imported --package-path github.com/containers/$PACKAGE --skip-self | sort -u | xargs "-I{}" echo "Provides: bundled(golang({}))")
+GO_IMPORTS=$(./usr/bin/golist --imported --package-path github.com/containers/$PACKAGE --skip-self | sort -u | xargs "-I{}" echo "Provides: bundled(golang({}))")
```
mtrmac, line 31: Maybe
```suggestion
GO_IMPORTS=$(GOPATH=$GOPATH ./usr/bin/golist --imported --package-path github.com/containers/$PACKAGE --skip-self | sort -u | xargs "-I{}" echo "Provides: bundled(golang({}))")
```
? `GOPATH` isn’t exported.

mtrmac (approved):
Thanks!

LGTM, please merge whenever you think is a good time.

lsm5:
@mtrmac Thank you, I'll keep this open for now and merge it once I have a better understanding of the local testing env.

lsm5:
I checked with packit folks, the only way so far to test this change would be to run `slash-packit propose-downstream` on a github issue (can't be done on an open PR). And given I have greenlight from @mtrmac and that it doesn't affect upstream, I'll merge this once tests pass and try running the propose-downstream task on the github issue. I'll also unlink #2036 so it remains open after merging this PR.

PR #2042, opened 2023-07-11
[release-1.13 backport] [CI:BUILD] Packit: install golist before updating downstream spec

lsm5:
The default Packit sandbox environment that runs Packit tasks for downstream Fedora does not have golist installed by default and can't run superuser tasks.

This commit will download and extract the golist binary from the Fedora rpm and use it to provide golist.

The GOPATH mention in `rpm/update-spec-provides.sh` is only required for golist to generate the gopaths and doesn't affect upstream or the rpm spec.

Currently, the only way to reliably test this is on an open github issue by running `/packit propose-downstream`. This can't be run on an open PR. The job-specific packit actions can only be tested via the packit service and not via packit cli.

(cherry picked from commit a39972ca35e334b0f04ac999f57faf1ec32cdc79)

### DIFF
--- a/rpm/update-spec-provides.sh
+++ b/rpm/update-spec-provides.sh
@@ -5,13 +5,30 @@
 # The goimports don't need to be present upstream.
 
 set -eo pipefail
+set -x
 
 PACKAGE=skopeo
 # script is run from git root directory
 SPEC_FILE=rpm/$PACKAGE.spec
 
+# Needed for golist to work
+# This is run in the packit sandbox environment for downstream tasks and
+# doesn't affect upstream repo and CI.
+export GOPATH=~/go
+GOPATHDIR=$GOPATH/src/github.com/containers/
+mkdir -p $GOPATHDIR
+ln -sf $(pwd) $GOPATHDIR/.
+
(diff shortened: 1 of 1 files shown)


lsm5:
i'll merge this once tests pass based on https://github.com/containers/skopeo/pull/2041#issuecomment-1630835207

mtrmac:
ACK. (#2036 shows a failure, but it also clearly is not testing the updated code.)

lsm5:
@mtrmac yes, i'm actually going by https://dashboard.packit.dev/results/propose-downstream/2892 which created the expected PR at https://src.fedoraproject.org/rpms/aardvark-dns/pull-request/5#request_diff .  In that case, the script was correct and the cargo dependency was included in packit's sandbox. Just that earlier it wasn't producing the correct output in the fedora PR.

So, in skopeo's case the script should get properly triggered for the next upstream skopeo release and assuming golist has been fetched correctly, everything should be smooth for our next release. Sorry for the PITA, just that currently there's no direct cheap way to verify this until we cut a new release.

mtrmac:
@lsm5 No worries, it’s great to be able to trust your expertise.

PR #2045, opened 2023-07-13
Add Gosec GitHub Action

candrews:
GoSec inspects source code for security problems by scanning the Go AST.

Currently, gosec reports a number of findings that skopeo may want to address. At least with gosec as part of GitHub Actions, the skopeo project can be aware of the current state of skopeo as well the impact changes make upon it.

See: https://github.com/securego/gosec

### DIFF
--- a/.github/workflows/gosec.yml
+++ b/.github/workflows/gosec.yml
@@ -1,0 +1,25 @@
+name: Run Gosec
+on:
+  push:
+jobs:
+  gosec:
+    permissions:
+      contents: read # for actions/checkout to fetch code
+      security-events: write # for github/codeql-action/upload-sarif to upload SARIF results
+      actions: read # only required for a private repository by github/codeql-action/upload-sarif to get the Action run status
+    runs-on: ubuntu-latest
+    env:
+      GO111MODULE: on
+    steps:
+      - name: Checkout Source
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3
+      - name: Run Gosec Security Scanner
+        uses: securego/gosec@3a6fd99e54aca25365f6873883ede35d4c00d917 # master
+        with:
+          # we let the report trigger content trigger a failure using the GitHub Security features.
+          args: '-no-fail -fmt sarif -out results.sarif ./...'
+      - name: Upload SARIF file
+        uses: github/codeql-action/upload-sarif@46ed16ded91731b2df79a2893d3aea8e9f03b5c4 # v2
+        with:
+          # Path to SARIF file relative to the root of the repository
+          sarif_file: results.sarif


lsm5:
@candrews thanks for the PR submission. Just checking if the github action will show up as an additional task in the list? I don't see it in the checks list.

@cevich @mtrmac PTAL.

mtrmac:
In general, I’m skeptical of “bottom-up” expansions of the CI pipeline, or adding integrations / badges just because. Let’s start with the benefit and needs of to the project, not with “this tool exists”.

Looking at `golangci-lint run --enable=gosec ./...`, this reports 5 problems, all in tests, and none of them are a security issue (even in the context of potentially exposing the user running the tests). So that’s not immediately attractive; at best we can update `WriteFile` to use `0600` instead of `0644`, for the hypothetical case of someone looking for the right way to invoke tools in our tests, and not noticing that these calls are safe in the test context.

(c/image has a similar score: 13 false positives, at best some `WriteFile` permissions in tests.

---

If we should do this, `gosec` is a vendored component of `golangci-lint`, so I think we could just enable that instead of adding a new action. (But then again, it is a linter disabled by default, as a data point…)

cevich:
I too am against CI extensions outside of the existing cirrus-based toolchain.  Not as a tool fitment judgment, rather that it introduces "automation non-sequiturs".  For example, Renovate may suggest updates that are only tested/reflected by the cirrus jobs (or vis-versa with github-actions).  Also, github actions are a PITA to run locally.

Suggestions:
1. Can use of this tool be added as a (phony) Makefile target so developers can use it?
2. Can it be told to ignore (or skip) checking tests or test-support code?
3. Can/Should the existing (cirrus) validation check incorporate this tool, or can/should it be made a separate check (maybe even [a manually-triggered one](https://cirrus-ci.org/guide/writing-tasks/#manual-tasks))?

Note: If Cirrus-CI is/becomes a sticking point, I'm happy to help assist with resolving any issues.  Just ping me.

mtrmac:
@cevich AFAICT we can get an equivalent set of validations by enabling a `golangci-lint` option.

---

(Overall I’m not seeing any benefit to enabling this linter, so far. That’s not to say that it could never be useful — but I am not _nearly_ enthusiastic enough to prioritize enabling it and cleaning up the code.)

cevich:
ACK.  I could be wrong, but IIRC doesn't our current Makefile/CI usage of `golangci-lint` ignore tests?  If so, this may be a transparent addition that could prove helpful later (I dunno, I didn't look closely at tool)?

mtrmac:
AFAIK `golangci-lint` covers everything (apart from `vendor` I think). I have confirmed that by seeing that adding a
```go
	_ = strings.Replace("a", "b", "c", 0)
```
both to `cmd/skopeo/main.go` and to `integration/sync_test.go`; `make validate-local` found that in both cases.

PR #2046, opened 2023-07-14
Packit: remove pre-sync action

lsm5:
The pre-sync action constantly breaks and is currently not possible to reliably test until the subsequent upstream release due to limitations in packit.

The lines being added by the action script to the downstream Fedora spec were only meant to keep Fedora happy. But given that they provide no tangible benefit as github notifies us of security issues in libraries mentioned in go.mod and go.sum, along with redhat prodsec's own magic for creating security alerts, there's absolutely no point to having the pre-sync action run and add a layer of uncertainty.

This commit removes the pre-sync action and
`rpm/update-spec-provides.sh`.

Ref: https://github.com/containers/podman/issues/19232


lsm5:
@mtrmac it broke again for podman v4.6.0-rc2. I've finally had it :smile: .  I'll merge it once green unless you object.

There should be no more troubles with this unless fedora packaging committee nags me  :sweat_smile:

mtrmac (approved):
LGTM.

Technically I don’t have a preference, and I’ll trust your expertise WRT the Fedora packaging relationship.

PR #2047, opened 2023-07-18
copy: add support for `--format zstd`

flouthoc:
`skopeo copy --format zstd` copy contents of manifest list from source to destination and verifying the destination is converted to `OCI` manifest list and each image converted and compression with `zstd` compression algorithm.

Needs: https://github.com/containers/image/pull/2047

### DIFF
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -14,6 +14,7 @@ import (
 	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/pkg/cli"
 	"github.com/containers/image/v5/pkg/cli/sigstore"
+	"github.com/containers/image/v5/pkg/compression"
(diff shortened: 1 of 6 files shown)


mtrmac:
> Why mix the manifest format and compression algorithm options? Was it perhaps some example where I said `--format zstd`? If so, I apologize.

I did, in https://github.com/containers/image/pull/1987 . I apologize again.

flouthoc:
No Issues, I will amend this to incorporate `--dest-compress-format`

mtrmac:
`--dest-compress-format` already exists, although full text search might not find the implementation: https://github.com/containers/skopeo/blob/38650252d51d626ef16f3b1c30fb50760fd0d499/cmd/skopeo/utils.go#L269 .

flouthoc:
Not needed.

PR #2064, opened 2023-08-02
Update c/image for golang.org/x/exp

mtrmac:
```
go get github.com/containers/image/v5@main
go mod tidy && go mod vendor
```

This updates c/image with a new version of x/exp. That package has changed API in an incompatible way, so just bumping x/exp (as in https://github.com/containers/skopeo/pull/2060 ) would break Skopeo builds.

This updates both c/image and x/exp in lockstep (and nothing needs updating in Skopeo itself for the x/exp breakage).


vrothberg (approved):
LGTM

PR #2069, opened 2023-08-05
Update to Go 1.19

mtrmac:
We already require it, because docker/credential-helpers uses Go 1.19 os/exec.Cmd.Environ(). So make that official.

> go mod tidy -go=1.19


vrothberg (approved):
LGTM

PR #2073, opened 2023-08-08
packit: Build PRs into default packit COPRs

martinpitt:
Building all PRs of all container projects into the same COPR does not properly isolate PRs from each other.

To avoid that, change the copr_build configuration to use the packit default COPRs, which are specific to the particular PR, and disappear after a few weeks. Depending projects should only run against what landed in skopeo/main i.e. the podman-next COPR.

---

This is exactly the same as https://github.com/containers/crun/pull/1260 ; please see the discussion there, this change needs to be applied to all container projects, and then all land together. Let's keep all the relevant discussions and tracking in  the crun PR please.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -6,29 +6,36 @@
 # supported Fedora and CentOS Stream arches.
 # They do not block the current Cirrus-based workflow.
 
-# Build targets can be found at:
-# https://copr.fedorainfracloud.org/coprs/rhcontainerbot/packit-builds/
-# and
-# https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/
-
 specfile_path: rpm/skopeo.spec
 upstream_tag_template: v{version}
 
-jobs:
-  - &copr
-    job: copr_build
-    trigger: pull_request
-    owner: rhcontainerbot
-    project: packit-builds
-    enable_net: true
-    srpm_build_deps:
-      - make
+srpm_build_deps:
+  - make
 
-  - <<: *copr
-    # Run on commit to main branch
+jobs:
+  - job: copr_build
+    trigger: pull_request
+    enable_net: true
+    # keep in sync with https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next
+    targets:
+      - fedora-all-x86_64
+      - fedora-all-aarch64
+      - fedora-eln-x86_64
+      - fedora-all-aarch64
+      - centos-stream+epel-next-8-x86_64
+      - centos-stream+epel-next-8-aarch64
+      - centos-stream+epel-next-9-x86_64
+      - centos-stream+epel-next-9-aarch64
+    additional_repos:
+      - "copr://rhcontainerbot/podman-next"
+
+  # Run on commit to main branch
+  - job: copr_build
     trigger: commit
     branch: main
+    owner: rhcontainerbot
     project: podman-next
+    enable_net: true
 
   - job: propose_downstream
     trigger: release


martinpitt:
That's the last PR in that series, and of course it's putting up a fight. [srpm build](https://download.copr.fedorainfracloud.org/results/packit/containers-skopeo-2073/srpm-builds/06257545/backend.log.gz) fails with

> [ ERROR][PID:592895] Build failed: Failed to create key-pair for user: packit, project:containers-skopeo-2073 with error: Attempt to talk to server timeouted (we gave it 5 attempts)

That could be transient, as other PRs are building COPRs just fine. If it fails again after a retry, I'll hail the COPR cavalry.

But also, what's that [Java exception](https://github.com/containers/skopeo/pull/2073/checks?check_run_id=15714103322) now?

martinpitt:
[ELN build failed](https://download.copr.fedorainfracloud.org/results/packit/containers-skopeo-2073/fedora-eln-x86_64/06257653-skopeo/builder-live.log.gz) with

> No matching package to install: '/usr/bin/go-md2man'

The problem is that the [podman-next COPR](https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/build/4868782/) does not have a build of golang-github-cpuguy83-md2man for ELN.

@lsm5 Is that something that you can fix on the COPR side, or should I take out the ELN builds here for the time being? Note that I most probably won't be able to finish this as my PTO starts pretty much "now". But you can force-push over my branch/PR of course.

martinpitt:
[ELN build failed](https://download.copr.fedorainfracloud.org/results/packit/containers-skopeo-2073/fedora-eln-x86_64/06257653-skopeo/builder-live.log.gz) with

> No matching package to install: '/usr/bin/go-md2man'

The problem is that the [podman-next COPR](https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/build/4868782/) does not have a build of golang-github-cpuguy83-md2man for ELN.

@lsm5 Is that something that you can fix on the COPR side, or should I take out the ELN builds here for the time being? Note that I most probably won't be able to finish this as my PTO starts pretty much "now". But you can force-push over my branch/PR of course.

lsm5:
> [ELN build failed](https://download.copr.fedorainfracloud.org/results/packit/containers-skopeo-2073/fedora-eln-x86_64/06257653-skopeo/builder-live.log.gz) with
> 
> > No matching package to install: '/usr/bin/go-md2man'
> 
> The problem is that the [podman-next COPR](https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/build/4868782/) does not have a build of golang-github-cpuguy83-md2man for ELN.
> 
> @lsm5 Is that something that you can fix on the COPR side, or should I take out the ELN builds here for the time being? Note that I most probably won't be able to finish this as my PTO starts pretty much "now". But you can force-push over my branch/PR of course.

Sure, no worries. Enjoy PTO. I'll try to force-push or create a new PR and list you as co-author.

Podman recently switched to using a vendored go-md2man because of this issue, so that's why we didn't see this in podman's copr switch. skopeo doesn't seem to vendor go-md2man yet, so I could either look at vendoring it, or switch to pandoc to build manpages, based on what @mtrmac and @vrothberg are ok with.

mtrmac:
I really don’t have an opinion.

Aesthetically, on every Skopeo build having to build a tool that is available as a pre-built RPM seems wasteful to me, OTOH if it is _not_ available, that argument falls apart. (Or would we continue to use an externally-built one in upstream and CI, but a vendored one in RPM?)

As for go-md2man vs. Pandoc, I have no opinion at all. go-md2man was used from the very start of the project, and I never had a reason to look for alternatives, so I have no idea how the two compare.

---

IIRC go-md2man has always _somehow_ been available in EL as an RPM, and that includes RHEL 8. I wouldn’t think a RPM would be dropped without replacement during a single major version of the product. Is it possible that this could be resolved just by enabling some repo? (I didn’t check.)

martinpitt:
> Is it possible that this could be resolved just by enabling some repo? (I didn’t check.)

You need to enable a Fedora ELN build for [the  golang-github-cpuguy83-md2man  package in the podman-next COPR](https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/build/4868782/). ELN is already enabled in podman-next in general, but either that was done after that build, or the build restricts its architectures to exclude Fedora ELN somehow.

Or of course, not build skopeo in ELN, that's also an option.

lsm5:
closing in favour of https://github.com/containers/skopeo/pull/2084 . There was an additional commit added to fix an eln target.

PR #2074, opened 2023-08-08
[CI:BUILD] DO NOT MERGE: debugging multiarch image builds

cevich:
Ref: https://github.com/containers/automation/pull/150


cevich:
Confirmed, the next release of new CI VM images will give us more debugging details as to what's happening when builds fail.

PR #2081, opened 2023-08-15
Improve the docs with cross-compilation info

michalbiesek:
~~The rationale behind this pull request is that it simplifies the process of building and testing by utilizing the RISCV-64 SBC.~~

Add information about cross-compilation in install.md

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,6 +117,7 @@ help:
 	@echo " * 'install' - Install binaries and documents to system locations"
 	@echo " * 'binary' - Build skopeo with a container"
 	@echo " * 'bin/skopeo' - Build skopeo locally"
+	@echo " * 'bin/skopeo.OS.ARCH' - Build skopeo for specific OS and ARCH"
 	@echo " * 'test-unit' - Execute unit tests"
 	@echo " * 'test-integration' - Execute integration tests"
 	@echo " * 'validate' - Verify whether there is no conflict and all Go source files have been formatted, linted and vetted"
--- a/install.md
+++ b/install.md
@@ -123,7 +123,7 @@ podman run docker://quay.io/skopeo/stable:latest copy --help
 
 Otherwise, read on for building and installing it from source:
 
-To build the `skopeo` binary you need at least Go 1.12.
+To build the `skopeo` binary you need at least Go 1.19.
 
 There are two ways to build skopeo: in a container, or locally without a
 container. Choose the one which better matches your needs and environment.
--- a/install.md
+++ b/install.md
@@ -174,6 +174,16 @@ document generation can be skipped by passing `DISABLE_DOCS=1`:
 DISABLE_DOCS=1 make
 ```
 
+### Cross-compilation
+
+For cross-building skopeo, use the command `make bin/skopeo.OS.ARCH`, where OS represents
+the target operating system and ARCH stands for the desired architecture. For instance,
+to build skopeo for RISC-V 64-bit Linux, execute:
+
+```bash
+make bin/skopeo.linux.riscv64
+```
+
 ### Building documentation
 
 To build the manual you will need go-md2man.


edsantiago:
I'd like to weigh in here, but I can't make it past this first sentence:

>The rationale behind this pull request is that it simplifies the process of building and testing

@michalbiesek could I ask you to expand on this a little? What exactly gets simplified?

My inclination is always toward better test coverage -- however, I agree with @mtrmac's concerns. What happens when `riscv64` build fails with a non-obvious reason?

michalbiesek:
Hi, thank you for the response. I apologize for any confusion. To put it simply, I utilize the cross-compilation option to construct the binary on an x86_64 host, which I then transfer to my RISC-V64 SBC computer. On this SBC computer, I execute the `skopeo` tool. What makes this process convenient for me is that I'm not required to build the binary directly on the SBC computer itself.

edsantiago:
Thank you, that does clarify. So IIUC your process is currently:
```console
$ make local-cross bin/skopeo.linux.riscv64
```
...and you would like to shorten that to just `make local-cross` to type less? Is that accurate?

mtrmac:
It’s just `make local-cross bin/skopeo.linux.riscv64` right now. This allows to type `make local-cross` to achieve the same thing, at the cost of building 5 more binaries (i.e. multiplying the total time by 6).

---

So there was actually no intent at all to have this tested in CI?

michalbiesek:
You are correct. I can use `make bin/skopeo.linux.riscv64` to achieve my goal.

Please let me know if the change in this Pull Request is not desirable to the upstream team I will close it.

edsantiago:
I'm starting to understand better, after reading https://github.com/containers/podman/pull/19626. The difference here is that in the podman PR, that does not affect CI because `make cross` is not run anywhere by default. In skopeo CI, it is https://github.com/containers/skopeo/blob/c19118d46fc0e4fa2b02f541a1f12b923b44320e/contrib/cirrus/runner.sh#L103-L105

I'm really torn, but it's mostly because I don't understand the purpose of `make local-cross`. I see it as a testing tool, or perhaps even a packaging tool. As an end user I would find it useless because I'm likely to only have one (maybe two) targets, and will not want to waste cycles+time on everything. But this is *my* ignorance speaking: I was not involved in adding this target, neither to podman nor skopeo. @mtrmac do you have a sense for what purpose this target serves? @michalbiesek in your real-world workflow are you OK with `make local-cross` and waiting for every single build?

michalbiesek:
To be honest, when I initially performed cross-compilation, I used "make local-cross" because the target completion for make is supported there. It was more convenient for me than creating the command without completion :)

mtrmac:
@edsantiago `git log -G local-cross` points at it being added in https://github.com/containers/skopeo/pull/1055, just to be run in CI.

Also Podman, AFAICT, is in a similar boat: Users can also just run `make bin/podman.cross.linux.riscv64` instead of building for all the targets. (OTOH that PR does contain a real, valuable, code fix.)

---

AFAICT there is no reason to run `local-cross` instead of `make bin/…` if only one architecture is desired. Admittedly both Make targets are currently undocumented… which suggests that either one is not currently being very frequently used. (We probably wouldn’t want to document the CI-exclusive targets in end-user installation documentation.)

mtrmac:
@michalbiesek I think the above is fairly conclusive in that `local-cross` is not the recommended way to cross-compile, so this PR as proposed is unnecessary for your purpose.

But it also shows that we don’t currently document any way how to do that; maybe documenting that (in `install.md` and/or `make help`) could be useful. Would you care to work on that instead?

cevich:
Coming in late to the review but...I believe we ***DO*** run a massive cross-compile check as part of podman CI (runs in a VM, so we don't pay for cirrus-minutes).  It's called `Alt Arch. Cross` and [runs these steps](https://github.com/containers/podman/blob/9cd42869223ab8acdd5be3d9bb2a84d988d8d73f/contrib/cirrus/runner.sh#L281-L295) (as an example).

mtrmac:
Yes — that’s unaffected by the linked Podman PR.

TomSweeneyRedHat:
Changes LGTM, I'll let @cevich and @edsantiago determine if this is the right approach.

cevich:
I'm okay with this in principal.  The PR description would be nice to clean up.  It would also be nice to have it included in CI, but I totally understand if that is asking too much.  If this gets merged w/o inclusion in CI, I'm happy to take an RFE issue.

michalbiesek:
> @michalbiesek I think the above is fairly conclusive in that `local-cross` is not the recommended way to cross-compile, so this PR as proposed is unnecessary for your purpose.
> 
> But it also shows that we don’t currently document any way how to do that; maybe documenting that (in `install.md` and/or `make help`) could be useful. Would you care to work on that instead?

Completed. (I have adjusted the title and incorporated the changes in this pull request.)

mtrmac (approved):
Thanks! LGTM.

PR #2086, opened 2023-08-21
rpm: spdx compatible license field

lsm5:
The lowercase `and` in the License field isn't compatible with spdx license format.

This commit replaces all `and` with `AND` in the License field in spec.


lsm5:
doesn't affect upstream. I'll take the liberty to merge once green.

mtrmac (approved):
Thanks!

PR #2093, opened 2023-08-28
Cirrus: Remove multi-arch skopeo image builds

cevich:
These jobs have been failing since early August due to technical/scripting problems.  Disable/remove entirely since a fix is unlikely to be implemented anytime soon.

If anybody ever cares, I gave up trying to debug the script problem in https://github.com/containers/podman/pull/19720.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -77,14 +77,13 @@ doccheck_task:
       "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" doccheck
 
 osx_task:
-    # Don't run for docs-only or multi-arch image builds.
+    # Don't run for docs-only builds.
     # Also don't run on release-branches or their PRs,
     # since base container-image is not version-constrained.
     only_if: &not_docs_or_release_branch >-
         ($CIRRUS_BASE_BRANCH == $CIRRUS_DEFAULT_BRANCH ||
          $CIRRUS_BRANCH == $CIRRUS_DEFAULT_BRANCH ) &&
-        $CIRRUS_CHANGE_TITLE !=~ '.*CI:DOCS.*' &&
-        $CIRRUS_CRON != 'multiarch'
+        $CIRRUS_CHANGE_TITLE !=~ '.*CI:DOCS.*'
     depends_on:
         - validate
     macos_instance:
@@ -106,8 +105,7 @@ osx_task:
 cross_task:
     alias: cross
     only_if: >-
-        $CIRRUS_CHANGE_TITLE !=~ '.*CI:DOCS.*' &&
-        $CIRRUS_CRON != 'multiarch'
+        $CIRRUS_CHANGE_TITLE !=~ '.*CI:DOCS.*'
     depends_on:
         - validate
     gce_instance: &standardvm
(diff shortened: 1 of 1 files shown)


cevich:
[Test `[CI:DOCS]` works as expected.](https://cirrus-ci.com/task/5447456364691456)

mtrmac:
I’m afraid I haven’t been really paying attention to these `jq failures` — is resolving this a matter of days/weeks, and is this PR just to save on CI in the meantime? Or is this failure something I should have dealt with? Or is it “probably broken indefinitely”?

If it’s the last one, `install.md` and `README.md` (and other places?) will need updating. And we might need a deprecation notice… somewhere.

cevich:
> I’m afraid I haven’t been really paying attention to these jq failures — is resolving this a matter of days/weeks, and is this PR just to save on CI in the meantime?

No worries, I added a link in the PR description.  Yes, I spent weeks debugging and likely weeks more are needed.  It's entirely possible that future build automation will use a new script or no script at all.  I haven't decided yet.  In all cases, the builds can't happen in cirrus for other reasons, so might as well clobber it.

> If it’s the last one

The same (probably?) jq problem was happening in all podman, buildah and skopeo build automation.  So I don't think the image source or docs needs any changes at the moment.

mtrmac:
> In all cases, the builds can't happen in cirrus for other reasons, so might as well clobber it.

ACK.

>  So I don't think the image source or docs needs any changes at the moment.

The other two repos don’t point at `quay.io` as official ways to install, AFAICS (and that makes sense, running them containerized would be much harder).

For Skopeo, _eventually_ we need to decide if we are going to continue to provide container images (somehow), or not — and if not, the documentation needs to stop advertising them. _Eventually_; this is not urgent, new releases only happen every few months; but if we, after a time. ended up pointing users at year-old releases, that would not be ideal.

cevich:
I agree.  My intention is to replace the build automation.  I think that's most helpful for the users.  Might be a good idea to sign the images as well, but that may be a further update later on.

cevich:
Force-push: Rebased.  This should be ready to go.

mtrmac:
`README.md` and `install.md` point at the `quay.io/containers/skopeo` and `quay.io/skopeo` container images.

I’m mildly worried about us stopping to build them, and continuing to point users at old unmaintained software. Can we leave a traceable trail, at least? Is there a WIP of the new replacement somewhere? If not, I can open a Skopeo issue to track the situation and we can link to it here.

cevich:
Unless I'm misunderstanding something, those FQINs are all still valid.  Recent images were built and pushed there manually (unless I forgot? I haven't checked).  This PR just disables/removes the (broken since August) built automation.  Re-implementing that is number-2 on my hot-list right behind adding a "make sure credentials can't leak" check.

mtrmac:
(I think 1.13.3 is missing (but the difference between 1.13.2 and 1.13.3 is irrelevant). Dan has also been tagging releases this week…)

Either way, I have filed #2110 , the stale bot will eventually remind me to revisit.

mtrmac (approved):

cevich:
Good idea.  Thanks.

PR #2114, opened 2023-09-15
proxy: Make `OpenImageOptional` work with `oci-archive:`

cgwalters:
By adding the standard ENOENT to our list of errors.

I hit this while working on
https://github.com/coreos/rpm-ostree/pull/4598
which is a tool that builds base images and wants to query if the image exists beforehand.


cmd/skopeo/proxy.go
```diff
@@ -235,7 +235,8 @@ func isDockerManifestUnknownError(err error) bool {
 // TODO drive this into containers/image properly
 func isNotFoundImageError(err error) bool {
 	return isDockerManifestUnknownError(err) ||
-		errors.Is(err, ocilayout.ImageNotFoundError{})
+		errors.Is(err, ocilayout.ImageNotFoundError{}) ||
+		errors.Is(err, os.ErrNotExist)
```
cgwalters, line 239: OK self-review...perhaps the real fix is in c/image to map enoent there to `ocilayout.ImageNotFoundError` ?

mtrmac (commented):
My first impression is to wonder why this needs special handling.

Obviously “an image not found on the registry” is something that can only be detected by talking to the registry with full credentials, i.e. inside Skopeo. Similarly “a tag not found inside an OCI archive” requires an OCI archive implementation.

But if the archive just doesn’t exist…  it might make sense for Skopeo to treat that uniformly, for callers that provide an user-specified _transport_`:`_image_ syntax without understanding the details of that… and then we get to discuss semantics of what “uniformly” means.

But `rpm-ostree compose image` seems to have its own `format=oci` option, so the _transport_`:`_image_ value is not user-provided.


---

Maybe a different way to ask this: Would we be expected to similarly return 0 for `docker://nxdomain.invalid/something`?) If not, what’s the difference?

Right now the documentation of `OpenImageOptional` says “If the image does not exist, zero
is returned.”.

What is the criteria for classifying errors as “image does not exist” vs. “other error opening an image”? Does the documentation of `OpenImageOptional` need clarifying, or updating?

---

\*shrug\* `rpm-ostree` is the only known user, so pragmatically speaking it gets to shape the proxy API however it is useful for `rpm-ostree`; but I’d like the Skopeo-repo documentation of that API to be reasonably clear about what are the properties that need maintaining. (“The `ostree-rs-ext` tests continue to pass” is one way to document those properties, I guess. I find that inelegant but I can live with that.)

cgwalters:
How about this variant instead?  (requires a patch to c/image)

```
From 63f4e4665b74e2d0d7b58e4fffa7fe665b85590f Mon Sep 17 00:00:00 2001
From: Colin Walters <walters@verbum.org>
Date: Fri, 15 Sep 2023 08:53:45 -0400
Subject: [PATCH] proxy: Make `OpenImageOptional` work with `oci-archive:`

By adding the standard ENOENT to our list of errors.

I hit this while working on
https://github.com/coreos/rpm-ostree/pull/4598
which is a tool that builds base images and wants to query
if the image exists beforehand.

---
 cmd/skopeo/proxy.go                                  | 12 +++++++++++-
 integration/proxy_test.go                            |  7 +++++++
 .../containers/image/v5/oci/archive/oci_src.go       |  7 ++++++-
 3 files changed, 24 insertions(+), 2 deletions(-)

diff --git a/cmd/skopeo/proxy.go b/cmd/skopeo/proxy.go
index aab85365..40c8e85f 100644
--- a/cmd/skopeo/proxy.go
+++ b/cmd/skopeo/proxy.go
@@ -73,6 +73,7 @@ import (
 
 	"github.com/containers/image/v5/image"
 	"github.com/containers/image/v5/manifest"
+	ociarchive "github.com/containers/image/v5/oci/archive"
 	ocilayout "github.com/containers/image/v5/oci/layout"
 	"github.com/containers/image/v5/pkg/blobinfocache"
 	"github.com/containers/image/v5/transports"
@@ -229,13 +230,22 @@ func isDockerManifestUnknownError(err error) bool {
 	return ec.ErrorCode() == dockerdistributionapi.ErrorCodeManifestUnknown
 }
 
+func isOciImageNotFound(err error) bool {
+	var lerr ocilayout.ImageNotFoundError
+	if errors.As(err, &lerr) {
+		return true
+	}
+	var aerr ociarchive.ImageNotFoundError
+	return errors.As(err, &aerr)
+}
+
 // isNotFoundImageError heuristically attempts to determine whether an error
 // is saying the remote source couldn't find the image (as opposed to an
 // authentication error, an I/O error etc.)
 // TODO drive this into containers/image properly
 func isNotFoundImageError(err error) bool {
 	return isDockerManifestUnknownError(err) ||
-		errors.Is(err, ocilayout.ImageNotFoundError{})
+		isOciImageNotFound(err)
 }
 
 func (h *proxyHandler) openImageImpl(args []any, allowNotFound bool) (retReplyBuf replyBuf, retErr error) {
diff --git a/integration/proxy_test.go b/integration/proxy_test.go
index 826128da..28949eff 100644
--- a/integration/proxy_test.go
+++ b/integration/proxy_test.go
@@ -354,4 +354,11 @@ func (s *proxySuite) TestProxy() {
 		err = fmt.Errorf("Testing optional image %s: %v", knownNotExtantImage, err)
 	}
 	assert.NoError(t, err)
+
+	nonExistentArchive := "oci-archive:/enoent"
+	err = runTestOpenImageOptionalNotFound(p, nonExistentArchive)
+	if err != nil {
+		err = fmt.Errorf("Testing optional image %s: %v", nonExistentArchive, err)
+	}
+	assert.NoError(t, err)
 }
diff --git a/vendor/github.com/containers/image/v5/oci/archive/oci_src.go b/vendor/github.com/containers/image/v5/oci/archive/oci_src.go
index 6c9ee334..819659e7 100644
--- a/vendor/github.com/containers/image/v5/oci/archive/oci_src.go
+++ b/vendor/github.com/containers/image/v5/oci/archive/oci_src.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 
 	"github.com/containers/image/v5/internal/imagesource"
 	"github.com/containers/image/v5/internal/imagesource/impl"
@@ -41,7 +42,11 @@ type ociArchiveImageSource struct {
 func newImageSource(ctx context.Context, sys *types.SystemContext, ref ociArchiveReference) (private.ImageSource, error) {
 	tempDirRef, err := createUntarTempDir(sys, ref)
 	if err != nil {
-		return nil, fmt.Errorf("creating temp directory: %w", err)
+		if errors.Is(err, fs.ErrNotExist) {
+			return nil, ImageNotFoundError{ref: ref}
+		} else {
+			return nil, fmt.Errorf("creating temp directory: %w", err)
+		}
 	}
 
 	unpackedSrc, err := tempDirRef.ociRefExtracted.NewImageSource(ctx, sys)
-- 
2.40.1

```

cgwalters:
(not going to lie, it took me *at least* 20 minutes of confusion to realize that there's *two* copies of `ImageNotFoundError` in c/image and that the previous code was always buggy and also needed to use `errors.As`)

cgwalters:
Tangential to this of course is my very strong opinion that downcasting and parsing errors should be an *extremely rare* event in code, and in e.g. languages like Rust the right thing to do is exactly how we [model this API in the Rust bindings](https://github.com/containers/containers-image-proxy-rs/blob/d641f5074cbb2c17a92c6a490bb22148a033ba74/src/imageproxy.rs#L404): 

> fn open_image_optional(&self, imgref: &str) -> Result<Option<OpenedImage>> 

The common "not found" case is the `Option` - it's not an error!   More on this in e.g. https://users.rust-lang.org/t/why-i-use-anyhow-error-even-in-libraries/68592

cgwalters:
> But rpm-ostree compose image seems to have its own format=oci option, so the transport:image value is not user-provided.

Yes, they are distinct options *but* they are very much intended to exactly map to the c/image transports and work the same way from a user PoV.  (I just take the liberty to add a nicer `registry` alias instead of `docker://` which just looks weird/legacy IMO)

mtrmac:
> How about this variant instead? (requires a patch to c/image)

- Still https://github.com/containers/skopeo/pull/2114#discussion_r1327313108 . Introducing a _new_ error type would maintain the existing API. (And introducing a new precisely-targeted error is, I think the cleanest solution.)

> ```patch
> @@ -41,7 +42,11 @@ type ociArchiveImageSource struct {
>  func newImageSource(ctx context.Context, sys *types.SystemContext, ref ociArchiveReference) (private.ImageSource, error) {
>  	tempDirRef, err := createUntarTempDir(sys, ref)
>  	if err != nil {
> -		return nil, fmt.Errorf("creating temp directory: %w", err)
> +		if errors.Is(err, fs.ErrNotExist) {
> +			return nil, ImageNotFoundError{ref: ref}
> +		} else {
> +			return nil, fmt.Errorf("creating temp directory: %w", err)
> +		}
> ```

This still matches the `/var/tmp` case; pushing the detection code down inside `createUntarTempDir` would allow precisely targeting the right `os.Open`.

cgwalters:
>  Introducing a new error type would maintain the existing API. (And introducing a new precisely-targeted error is, I think the cleanest solution.)

Hmm.  So with `oci:` transports, we can distinguish between "oci directory exists, has `index.json` etc." and "oci directory doesn't exist".  But with `oci-archive` there's no such thing...either the file exists, or it doesn't - and ISTM that for `oci-archive`, this does semantically map to `ImageNotFoundError`.  Right?

mtrmac:
> The common "not found" case is the `Option` - it's not an error!

(Many things in Go are awkward and unsafe, with no sum types, no optionality… The complete lack of compiler support for errors, and things like `errors.Is` vs. `errors.As`, with no compile-time checking, is also just _sad_.)

I think that works for a frequent expected cause of a specific operation; the c/image “transport” abstraction makes this harder (in a sense, that’s exactly what this PR is hitting — after abstracting to the `NewImageSource` interface method we can no longer discuss what exactly failed / how exactly it failed, without talking about concepts specific to a particular transport).

c/image can just about manage to provide a common `types.ImageReference` / `transport.ImageName` concept, but I think providing a common “image container” that can deal with registries, c/storage, `oci-archive` (a filesystem of `dir:` subdirectories?) is just not feasible. And “image not found in a container” vs “container not found at all” is that kind of a situation.

Ultimately, API stability does not allow us to change `NewImageSource` return values anyway…

mtrmac:
> Hmm. So with `oci:` transports, we can distinguish between "oci directory exists, has `index.json` etc." and "oci directory doesn't exist". But with `oci-archive` there's no such thing...either the file exists, or it doesn't

No, `oci-archive` can also contain multiple images. So, `oci-archive:existing-path:non-matching-reference` is the current `ImageNotFoundError`, and `oci-archive:/dev/null/this/does/not/exist:whatever` is a different failure mode.

cgwalters:
> No, oci-archive can also contain multiple images. So, oci-archive:existing-path:non-matching-reference is the current ImageNotFoundError, and oci-archive:/dev/null/this/does/not/exist:whatever is a different failure mode.

Thanks, I didn't know that.  But...I guess them I'm arguing that at least for all use cases I can think of, I really do want to treat "oci-archive file doesn't exist" here the same as `ImageNotFoundError`.

Unlike `oci` directories which *require* initialization, to me kind of the point of `oci-archive` is that it's just a single file that you can just...manage as a file.  Requiring users of `rpm-ostree compose image` to create an "empty" oci-archive file would just be weird and defeat the purpose of the ergonomics here.

cgwalters:
I've updated the main PR here so far to:

In c/image:

- Move the error check further down
- Necessarily reorder the open invocation first

In skopeo:

- Add a test case for the "parent directory doesn't exist" correctly erroring

cgwalters:
Moved the c/image changes to https://github.com/containers/image/pull/2123

mtrmac:
> Requiring users of `rpm-ostree compose image` to create an "empty" oci-archive file would just be weird and defeat the purpose of the ergonomics here.

(The archives are not editable that way — they can either be read or written. But I do think that when reading a multi-image archive, differentiating the two failures should be possible.)

mtrmac:
@cgwalters the c/image error type PR has merged; do you want to update this?

mtrmac:
@cgwalters ping

mtrmac:
#2613 should make `OpenImageOptional` recognize missing tags in `oci-archive:` (and, now, also for `oci:`).

For `oci-archive:`, it only recognizes `ImageNotFoundError`, not `ArchiveFileNotFoundError`. Given this stale PR, I assume it doesn’t matter either way.

Issue #1545, opened 2022-01-25
Prebuilt and dynamically linked binaries released somewhere as artifacts


thesayyn:
You are probably sick of seeing this coming up over and over again and I apologize for bringing this up yet again. I am just curious about whether the proposed approach below would work for everyone.

Before I begin, I want to express that I am well aware of the conversation that went on #1478 #755 #932 #670

# Use case
Recently, I started working on https://github.com/thesayyn/rules_container which is a bazel rule for building container images. It utilizes skopeo to fetch OCI and DockerV2 images and lay them out in the bazel-out tree as OCI layout. Then we use umoci to manipulate the OCI layout (adding cross-platform layers etc.) that resides in the bazel-out tree that we fetched previously. All is well and works perfectly. 

# Problem
By the nature of bazel, everything is hermetic and reproducible. but what does this mean? well, it means that bazel has to know everything that goes into the pot (aka: targets) to determine whether a compilation unit is affected. this includes tools that are part of the compilation; which is skopeo in this case. hope all is well to this point because things are gonna complicated. 

there are a few ways to make skopeo part of the compilation;

1) We could rely on users to get skopeo despite being non-deterministic and impractical way.
This is the only way forward we can go currently as skopeo is only distributed as container images and proprietary packages (deb etc). 
This method has a few drawbacks. 1) it is non-hermetic as two distinct people can have different versions of skopeo installed on their machine thus leading to different outputs and hashes. breaks the determinism and this is big. 2) users have to fetch skopeo for every building environment. 

2) Fetching the skopeo repository under bazel and building it via prefetched go toolchain.
This is the most feasible way in terms of determinism and hermeticity as we will be fetching a pinned version of the repository and building it via a deterministic go toolchain. everything might sound all right when looked at from the maintainers of skopeo maintainers, unfortunately, this way is also sub-optimal as it has a great overhead of fetching the toolchain and building the source each time. also, we will have our own version of BUILD files (similar to Makefiles) and they can break at any given upgrade. 

3) Fetching prebuilt binaries from Github and comparing checksums
This is the most favorable and optimal way to fetch external tools within the bazel ecosystem. Basically, we tell the bazel where to look for platform-specific binaries and their checksums. then bazel fetches them for us and checks if the binary was tempered with (addresses the concern https://github.com/containers/skopeo/issues/1478#issuecomment-958106748) as we had the checksums declared beforehand. No overhead of fetching go toolchain and building the source. no risk of being disconnected from the upstream repository.

[...]

PR #2124, opened 2023-10-08
Update install documentation & add Build pipeline

hurzelpurzel:
Refer to the Windows build of user "passcod".

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -50,7 +50,7 @@ jobs:
       - name: Install dependencies
         run: |
           sudo apt-get update
-          sudo apt-get install -y mingw-w64 curl golang make devmapper gpgme
+          sudo apt-get install -y mingw-w64 curl golang make libdevmapper-dev gpgme
 
       - name: Fetch source
         run: curl -L https://github.com/containers/skopeo/archive/refs/tags/v${{ env.VERSION }}.tar.gz -o - | tar xzv
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -59,6 +59,10 @@ jobs:
         working-directory: skopeo-${{ env.VERSION }}
         run: make bin/skopeo.windows.amd64.exe
 
+      - name: Build skopeo.linux.amd64
+        working-directory: skopeo-${{ env.VERSION }}
+        run: make bin/skopeo.linux.amd64
+
       - name: Create release 
         uses: actions/github-script@v6
         with:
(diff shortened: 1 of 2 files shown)


passcod:
I would much rather this project start producing builds, even copying my config to do so if needed, than pointing to my repo.

hurzelpurzel:
Know I added the Pipeline from @passcod  and provided the linuxs build as well.

mtrmac:
Thanks for the PR.

First of all, we can’t even look at a PR where the DCO is missing.

---

It’s also, still, the case, that I don’t think we can provide useful stand-alone Linux binaries. Compare #1545 and earlier discussion. I’m happy if they work for you, and it’s perfectly fine to publish them, but I don’t think we want to knowingly recommend that path in the official repo.

---

WRT macOS / Windows binaries, I have a much less strong opinion. (I am rather worried that the Windows binary is probably not a good Windows ecosystem citizen, but, well, maybe we got to start somewhere.)

The binaries for those operating systems could be useful, but it’s also something where I know very little. It also needs to align with how the GitHub.com/containers organization has build systems and clouds set up, and what we can commit to watching / maintaining. @cevich , WDYT, at least for an initial triage?

TomSweeneyRedHat:
@hurzelpurzel thanks for the PR, as @mtrmac noted, please sign it.

@cevich, PTAL when you get a chance.

cevich:
So in general producing test-builds is a perfectly reasonable thing to do, however they must be very clearly labeled as testing-only builds (if they wreck your universe, we don't care).  If this PR wants to go in that direction for the initial implementation, great!  I think all that's needed is stripping out the release components (and could drop the version query as well).

"Official" builds must come from a trusted build pipeline and include some form of trust-verification.  Both parts are critical and non-trivial to pull off.  What happens in podman today is maybe less than ideal, but it works and could potentially be duplicated here:

1. Builds (all platforms all arches) are done manually in a clean-room environment and uploaded to a github pre-release.
2. A dry-run is manually initiated for a github-actions workflow which downloads (from pre-release page) and signs the windows and mac installer binaries.  Making them available for manual inspection/testing.
3. Assuming everything is "good", the pre-release is published which triggers the same signing workflow to perform the official signing.

The critical "good-citizen" thing here WRT Windows (and maybe Mac) is the signing can only happen once per version/release.  This is because a UUID is generated in the installer that the OS tracks for both uninstall and upgrade purposes.  Having duplicate UUIDs per version completely screws this up for end-users.

Hence why there is always a human in the release publishing and signing loop.

mtrmac:
> Hence why there is always a human in the release publishing and signing loop.

At this point, to me that is pretty much a show-stopper. Combined with the lack of sign-off…

PR #2136, opened 2023-10-24
Fix container image reference in install docs

robmoore-i:
See: https://quay.io/repository/containers/skopeo

### DIFF
--- a/install.md
+++ b/install.md
@@ -109,11 +109,11 @@ always welcome.
 
 ## Container Images
 
-Skopeo container images are available at `quay.io/skopeo/stable:latest`.
+Skopeo container images are available at `quay.io/containers/skopeo/stable:latest`.
 For example,
 
 ```bash
-podman run docker://quay.io/skopeo/stable:latest copy --help
+podman run docker://quay.io/containers/skopeo/stable:latest copy --help
 ```
 
 [Read more](./contrib/skopeoimage/README.md).


mtrmac (changes requested):
Thanks for the PR.

The `quay.io/containers/skopeo/stable` repository referenced in this PR doesn’t exist at all, so I don’t understand what this is supposed to fix.

robmoore-i:
Not only does it exist, it is the repository which works if you actually attempt to run the command listed in the docs.

Have a go at running the command in the existing docs - does it work? It didn't for me. That repository does not seem to exist.

Now have a go at running the command I have updated it to. Does that work? It does for me.

mtrmac:
```sh
$ skopeo inspect --override-os linux docker://quay.io/skopeo/stable:latest
… data …
```

>  does it work? 

Yes.

> It didn't for me. 

How does it fail, then?

---

vs.
```sh
% bin/skopeo inspect --override-os linux docker://quay.io/containers/skopeo/stable:latest
FATA[0001] Error parsing image name "docker://quay.io/containers/skopeo/stable:latest": reading manifest latest in quay.io/containers/skopeo/stable: unauthorized: access to the requested resource is not authorized 
```

That repo isn’t intended to exist, and, in fact, doesn’t. If Podman can run a command referencing an image in that repo, that would be a bug in Podman.

robmoore-i:
I'm using docker rather than podman. Nevertheless it's really odd that we get different behaviour. I'll get back to you in a bit with more details.

TomSweeneyRedHat:
OK, I'm a bit confused here.  quay.io/containers/skopeo is supposed to be a repo containing tags like :latest, :v1.9.0, etc.  

So a pull of docker://quay.io/skopeo/stable:latest and then a pull of docker://quay.io/containers/skopeo:latest is supposed to return different images containing the same makeup.  

@cevich the quay.io/containers/{project} repos all look to have been turned into pointers to the quay.io/{project}/stable repos.  Was that intentional or was that an artifact of the fun security drill we had with the images a month or so ago?

robmoore-i:
Here is some behaviour I got just now using `quay.io/containers/skopeo`:

```
rmoore@MacBook-Pro-2 ge % docker run quay.io/containers/skopeo:v1.5.0 --version
skopeo version 1.5.0
rmoore@MacBook-Pro-2 ge % docker run quay.io/containers/skopeo:latest --version
Unable to find image 'quay.io/containers/skopeo:latest' locally
latest: Pulling from containers/skopeo
44fae78d9571: Pull complete 
1eb50d17de7e: Pull complete 
982c05207b36: Pull complete 
2b9681efa0ee: Pull complete 
555e2f5844f6: Pull complete 
Digest: sha256:8998057b898aedebf3393177df0b1680101264c37daddc42da6f38d0806a692c
Status: Downloaded newer image for quay.io/containers/skopeo:latest
skopeo version 1.13.3
rmoore@MacBook-Pro-2 ge % 
```

It seems like `quay.io/skopeo/stable` is now working 🎉 Thanks for looking into that @mtrmac  / @TomSweeneyRedHat.

I found the repository `quay.io/containers/skopeo` because after experiencing that the repository listed in the docs didn't seem to exist, I went to https://quay.io and did a search for "skopeo" where I found https://quay.io/repository/containers/skopeo as the top search result. It looked to me as though the docs were simply out of date. It seems that's not the case, and this was a temporary issue caused by some recent infra work. I'll close this PR now that it seems to be fixed 👍

cevich:
Oh goodie, thanks for the followup.  Glad things are working as documented.  

> So a pull of docker://quay.io/skopeo/stable:latest and then a pull of docker://quay.io/containers/skopeo:latest is supposed to return different images containing the same makeup.

Indeed there should be (effectively) a 1:1 match across version-tags and `latest` between `quay.io/<repo>/stable:<tag>` and `quay.io/containers/<repo>:<tag>`.  One image is built, then it's companion build should re-use the same layer-cache.  So even the timestamps should be w/in a minute or two of each other.

We monitor the builds daily, but it's always possible something sneaks under the radar.  So please feel free to let us know if you notice something broken.

PR #2137, opened 2023-10-24
Remove unused environment variables in Cirrus

mtrmac:
Some other containers/* repos use these values in test names; we don't, so remove them so that we don't have to worry about keeping them up to date.

Previously #2131. Cc: @cevich


cevich (approved):
LGTM.  Thanks!

vrothberg (approved):
LGTM

PR #2138, opened 2023-10-24
Fix ENTRYPOINT documentation, drop others.

### LINKED ISSUE #2134: misleading ENTRYPOINT info in README
I just stumbled on the [README](https://github.com/containers/skopeo/blob/main/contrib/skopeoimage/README.md), where it states:

> Also, the ENTRYPOINT and the WORKDIR variables are not set within these container images, as such they default to /.

While https://github.com/containers/skopeo/blob/ca00d96b6bae830b9b19b222597871ae1727dbe8/contrib/skopeoimage/stable/Containerfile#L47

explicitly defines one, which is very confusing when someone tries to pull the container image and want's to test it out.

Therefore either the README or the Containerfile should be updated that the information is correct again.

mtrmac:
`ENTRYPOINT` was incorrectly documented to be set to `/` (which doesn't even make sense).

Stop mentioning `PATH` and `WORKDIR` in the top-level `README`, typical users of the container shouldn't need to care, and it's already somewhat implied by "built using the latest Fedora".

Fixes #2134.

We should also update the READMEs in some https://quay.io/organization/skopeo repos and https://quay.io/repository/containers/skopeo ; @TomSweeneyRedHat do you know who has write access to that?

### DIFF
--- a/contrib/skopeoimage/README.md
+++ b/contrib/skopeoimage/README.md
@@ -25,9 +25,7 @@ the images live are public and can be pulled without credentials.  These contain
 resulting containers can run safely with privileges within the container.
 
 The container images are built using the latest Fedora and then Skopeo is installed into them.
-The PATH in the container images is set to the default PATH provided by Fedora.  Also, the
-ENTRYPOINT and the WORKDIR variables are not set within these container images, as such they
-default to `/`.
+The ENTRYPOINT of the container is set to execute the `skopeo` binary.
 
 The container images are:
 


vrothberg (approved):
LGTM

contrib/skopeoimage/README.md
```diff
@@ -25,9 +25,7 @@ the images live are public and can be pulled without credentials.  These contain
 resulting containers can run safely with privileges within the container.
 
 The container images are built using the latest Fedora and then Skopeo is installed into them.
-The PATH in the container images is set to the default PATH provided by Fedora.  Also, the
-ENTRYPOINT and the WORKDIR variables are not set within these container images, as such they
-default to `/`.
+ENTRYPOINT of the container is set to execute the `skopeo` binary.
```
TomSweeneyRedHat, line 28:
```suggestion
The ENTRYPOINT of the container is set to execute the `skopeo` binary.
```
Or, and I'm on the fence about noting the default values or not noting them here as they can be found elsewhere

```suggestion
The PATH in the container images is set to the default PATH provided by Fedora, and as it is not set, the WORKDIR variable defaults to `/`.  The ENTRYPOINT of the container is set to execute the `skopeo` binary.
```

Your call on which to use.  I'd go with the lengthier one, but it's a coin flip to me.

mtrmac: Thanks, fixed the first part.

As for `PATH`/`WORKDIR`, the way I think about this (and the way the README documents usage), users are guided to say something like `podman run …/skopeo…:latest copy`, and not actually do anything _inside_ the container. So `PATH`/`WORKDIR` should not matter for such usage. Of course users can execute the container with `--entrypoint=/bin/sh`… and the `Containerfile`s do make some effort to make `container-storage:` work within the container itself.

So it seems to me the question is whether this document is a short introduction to using the image, or whether it is documenting our “API” commitment to have `PATH` and `WORKDIR` set that way. I could see an argument for either, I very weakly prefer brevity.

TomSweeneyRedHat (commented):
LGTM
@mtrmac pushed me on the brevity side of the fence!  :smiley:

PR #2150, opened 2023-11-07
Fix for CVE-2023-39325

lsm5:
Updated golang.org/x/net to v0.17.0

Ref: https://github.com/advisories/GHSA-4374-p667-p6c8


lsm5:
@mtrmac PTAL

go.mod
```diff
@@ -130,3 +130,5 @@ require (
 	gopkg.in/square/go-jose.v2 v2.6.0 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
+
+replace golang.org/x/net v0.11.0 => golang.org/x/net v0.17.0
```
mtrmac, line 134: This is not the best way to do that; future updates of the `x/net` version number above (around line 119), even to a version > 0.17.0, would still be replaced by 0.17.0.

IIRC the right way is to `go get golang.org/x/net/@v0.17.0`.

mtrmac (commented):
FWIW, Skopeo, AFAIK, never creates a http/2 _server_, so this vulnerability is irrelevant. I can very well accept that the fairly large changes are worth doing anyway, just to silence security scanners.

(The “request changes” is just about the `replace` approach, not the principle of the thing.)

TomSweeneyRedHat:
@lsm5 on the off-chance that you have done a go vendoring before, checkout https://github.com/containers/buildah/blob/main/install.md#vendoring

lsm5:
Done. PTAL.

mtrmac (approved):
Thanks!

Issue #560, opened 2018-10-13
copy: add null transport format to support signature verification


alonbl:
Hi,

Maybe I am missing something, but there is no way to perform verify only of an image signature with the policy configuration. The only option I found is to use the copy the image to other image.

If that's true, it would be great if you add null transport format to allow something like:

```
$ skopeo copy docker://registry1.com/image:tag null:
Writing manifest to image destination
Storing signatures
```

This will execute the cycle of verifying the source while not download/uploading the entire content, the null transport should also allow writing signatures (to null...) to avoid explicit `--remove-signatures`.

Any better option to verify only?

Thanks!

mtrmac:
Thanks for your report.

Signatures are intended provide end-to-end nonrepudiability (proof of who signed the image, verified directly by the final consumer), even if the intervening registries / mirrors / HTTP proxies / other transport mechanisms are unreliable or outright malicious.

With these goals, “verify a remote image” is a very dangerous operation to provide: the verification results are only valid for that particular registry reply, nothing prevents the registry from providing a maliciously modified image on subsequent requests. So, such a “verify image on a registry” operation would provide a fundamentally unusable—and dangerous to rely on!—result.

So, if you are designing your workflow to have such a “verify signature” step with a not-100%-trusted registry, please reconsider.

If your registry—and the backing storage!—_is_ 100% trusted, and you want the “verify signature” operation maybe for a server-side audit run that all images are correctly signed, or something similar, that can of course be built — as you say, with a `skopeo copy` to a temporary destination (like a `dir:` which is immediately deleted), or with `skopeo standalone-verify`, or ultimately with the underlying GitHub.com/containers/image Go package. But as described above, I’d very strongly prefer to not provide “verify remote image” operation (quite a few users ask for it when they first hear of signatures, without considering the implications).

Is there any important use case I am missing?

alonbl:
Hi,

Thank you for the long description! I fully agree with anything you write. However, the use cases I had in mind are different... Mostly verify integrity of the repository and configuration not to actually execute the image.

Using standalone-verify is what I begun with, until I found out that it does not use the policy nor the repositories.d so it cannot verify the entire settings. The copy, however, does the work, so I currently download it into a docker-archive to enjoy the benefit, however, I think that downloading the actual data to verify the signature configuration is a bit of an overkill.

Examples:

[...]

PR #2153, opened 2023-11-10
Add --check-policy option to skopeo inspect

AlbanBedel:
When handling images with skopeo it can be useful to be able to check that the image is valid before processing any of its metadata. With the new `--check-policy` flag `skopeo inspect` will error out if the image doesn't pass the policy check.

### DIFF
--- a/cmd/skopeo/inspect.go
+++ b/cmd/skopeo/inspect.go
@@ -29,6 +29,7 @@ type inspectOptions struct {
 	raw           bool // Output the raw manifest instead of parsing information about the image
 	config        bool // Output the raw config blob instead of parsing information about the image
 	doNotListTags bool // Do not list all tags available in the same repository
+	checkPolicy   bool // Error out if image is not allowed to run
 }
 
 func inspectCmd(global *globalOptions) *cobra.Command {
@@ -61,6 +62,7 @@ skopeo inspect --format "Name: {{.Name}} Digest: {{.Digest}}" docker://registry.
 	flags.BoolVar(&opts.config, "config", false, "output configuration")
 	flags.StringVarP(&opts.format, "format", "f", "", "Format the output to a Go template")
 	flags.BoolVarP(&opts.doNotListTags, "no-tags", "n", false, "Do not list the available tags from the repository in the output")
+	flags.BoolVar(&opts.checkPolicy, "check-policy", false, "Verify that the image pass the policy check")
 	flags.AddFlagSet(&sharedFlags)
 	flags.AddFlagSet(&imageFlags)
 	flags.AddFlagSet(&retryFlags)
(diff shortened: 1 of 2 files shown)


cmd/skopeo/inspect.go
```diff
@@ -106,6 +108,24 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
...
+			return fmt.Errorf("Error loading trust policy: %v", err)
+		}
+		defer func() {
+			if err := policyContext.Destroy(); err != nil {
+				retErr = noteCloseFailure(retErr, "tearing down policy context", err)
+			}
+		}()
+
+		// Be paranoid and fail if either return value indicates so.
+		if allowed, err := policyContext.IsRunningImageAllowed(ctx, unparsedImage); !allowed || err != nil {
```
mtrmac, line 124: This should also be inside `retry.IfNecessary` because it contacts remote servers to fetch the signatures.

mtrmac (commented):
Thanks for the PR.

Um… There are definitely some valid use cases, but frequently users ask for things like this to build _incorrect_ verification workflows like https://github.com/containers/skopeo/issues/560#issuecomment-429939281 (there’s much more in there). The balance of utility vs. risk is not just in the existence of the feature, but also in the prevalence of one vs. the other user base.

I don’t really know what to here…

Maybe name this `--dangerous-{check,use,enforce}-policy` and strongly warn in both in the man page and `--help` that this command succeeding gives _no guarantees at all_ about the inspected image being correctly signed at any other time or for any other purpose? That would be certainly not elegant, but it would be at least _some_ way to give naive users pause.

cmd/skopeo/inspect.go
mtrmac, line 124: Also, being security-critical, this should have at least minimal tests that the option works and policy is enforced (not that all possible kinds of policy are enforced), maybe somewhere around `copySuite.TestCopySignatures` which already creates correctly/incorrectly signed images.

AlbanBedel:
> Thanks for the PR.
> 
> Um… There are definitely some valid use cases, but frequently users ask for things like this to build _incorrect_ verification workflows like [#560 (comment)](https://github.com/containers/skopeo/issues/560#issuecomment-429939281) (there’s much more in there). The balance of utility vs. risk is not just in the existence of the feature, but also in the prevalence of one vs. the other user base.

First of all thanks for the review. Before continuing I would like to be sure that I understand correctly which risk you are talking about. In the issue you pointed the risk was about having an image referenced by tag being later replaced by another image. But with `skopeo inspect` you get the digest of the image, so if from then on the image is only referenced using the returned digest this attack is not possible, or am I missing something?

My use case is to create some k8s Jobs using some parameters derived from the image metadata, so even without verification I have to reference the images by digest to ensure that the image and metadata are matching.

mtrmac:
> But with `skopeo inspect` you get the digest of the image, so if from then on the image is only referenced using the returned digest this attack is not possible

Yes. **If**. I’ve encountered enough users who don’t immediately, unprompted, see the need for this (and even users who are hard to convince this is necessary) to make me worried.

PR #2166, opened 2023-11-29
[release-1.6] Stop trying to install sif prerequisite RPMs

mtrmac:
Fedora 36 repos no longer exist, so these installations always fail: https://cirrus-ci.com/build/5827089207656448 .

The SIF test in systemtest/020-copy.bats is already conditionalized on the presence of the tools, so it should now be skipped.

### DIFF
--- a/contrib/cirrus/runner.sh
+++ b/contrib/cirrus/runner.sh
@@ -66,9 +66,6 @@ _run_setup() {
     # VM's come with the distro. skopeo package pre-installed
     dnf erase -y skopeo
 
-    # Required for testing the SIF transport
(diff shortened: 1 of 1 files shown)


mtrmac:
Cc: @cevich

rhatdan:
LGTM

cevich (approved):
LGTM. Thanks for taking care of this.

FYI (I don't know if it applies to these old images), the current set of CI VM images already have these packages bundled in.

Ref: https://github.com/containers/automation_images/blob/main/cache_images/fedora_packaging.sh

PR #2172, opened 2023-12-01
[Makefile] Don't compile during install

rahilarious:
Usually while building from source, `make` and `make install` is a common phenomenon. Former only compiles and latter only installs.

Standard procedure of `make && make install` compiles same package twice wasting CPU cycles and producing heat which causes climate change. So trying to fix that. #MakeEarthGreenAgain

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -92,7 +92,7 @@ osx_task:
         export PATH=$GOPATH/bin:$PATH
         go version
         go env
-        make validate-local test-unit-local bin/skopeo
+        make validate-local test-unit-local bin/skopeo docs completions
         sudo make install
         /usr/local/bin/skopeo -v
 
--- a/Makefile
+++ b/Makefile
@@ -104,7 +104,7 @@ endif
 #     Note: Uses the -N -l go compiler options to disable compiler optimizations
 #           and inlining. Using these build options allows you to subsequently
 #           use source debugging tools like delve.
-all: bin/skopeo docs
+all: bin/skopeo docs completions
 
 codespell:
 	codespell -S Makefile,build,buildah,buildah.spec,imgtype,copy,AUTHORS,bin,vendor,.git,go.sum,CHANGELOG.md,changelog.txt,seccomp.json,.cirrus.yml,"*.xz,*.gz,*.tar,*.tgz,*ico,*.png,*.1,*.5,*.orig,*.rej" -L fpr,uint,iff,od,ERRO -w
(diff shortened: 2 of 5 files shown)


mtrmac (commented):
I think, as a general principle, computers should work to remove tedious work. So I’m unhappy with removing `make` dependencies and forcing users to manually understand and deal with the dependencies (as this PR shows that several callers need to be manually extended).

Also the origin of the `./configure && make && make install` convention, https://www.gnu.org/prep/standards/html_node/Standard-Targets.html#Standard-Targets , specifies
> _Compile the program and_ copy …

---

Rather, I think several of of the `make` targets involved (from this PR, `completions` and `bin/skopeo` at least) should, instead, be turned into “proper” `make` targets which _correctly_ specify their inputs and outputs, without using `.PHONY`, so that `make` can correctly understand the targets don’t need to be rebuilt. (Something of that kind does exist e.g. in Podman’s `bin/podman` target.)

And then `install` can depend on the binary artifact for users convenience, while the usual `make && {,sudo} make install` users will only trigger one compilation.

rahilarious:
> I think, as a general principle, computers should work to remove tedious work. 

I don't think just typing extra `make` prior to `make install` is tedious work. 

 > And then `install` can depend on the binary artifact for users convenience, while the usual `make && {,sudo} make install` users will only trigger one compilation.

*recompilation

User already sacrifices own "User convenience" when one decides to compile from source instead of using package manager :) 

Restricting options and diverting from standards in the name of avg joe's convenience is Apple way. Personally, I hate apples, love open raspberries.

mtrmac:
> > I think, as a general principle, computers should work to remove tedious work.
> 
> I don't think just typing extra `make` prior to `make install` is tedious work.

I’m open to that argument, given GNU coding standards. Typing extra `make docs` or `make completions`, _and having to figure that out_, does count as a tedious work to me. Nobody wakes up with “I want to run `make completions` today”.

> > And then `install` can depend on the binary artifact for users convenience, while the usual `make && {,sudo} make install` users will only trigger one compilation.
> 
> *recompilation

No, my point is that with properly specified dependencies `make install` _does the right thing_: it compiles the binary if the user didn’t run `make`, and it doesn’t compile again if `make` was run without modifying the sources.

rahilarious:
> it compiles the binary if the user didn’t run `make`, and it doesn’t compile again if `make` was run without modifying the sources.

But it still checks if it was previously compiled or not. This step is considered as "building". The reason I insist that `make install` should _JUST_ install (nothing else) is most package managers consider **building** in **install** phase as a red flag and not best practice.

mtrmac:
I don’t know how to agree on what is a “best practice”; opinions will vary.

Pragmatically, any Linux package manager has certainly had to install with GNU coding standards-targeting software; so it either has a known technique to reconcile these two approaches, or a long-standing practice of ignoring the difference between them.

rahilarious:
> I don’t know how to agree on what is a “best practice”; opinions will vary.
> 
> Pragmatically, any Linux package manager has certainly had to install with GNU coding standards-targeting software; so it either has a known technique to reconcile these two approaches, or a long-standing practice of ignoring the difference between them.

I respect your decision. Will close the PR.

mtrmac:
@lsm5 Would it be valuable to add the precise dependencies, so that a second `make` doesn’t recompile? Or would you be patching that code out either way?

Issue #2170, opened 2023-12-01
Are default-policy.json & default.yaml in wrong place? Shouldn't they be in containers/image?


rahilarious:
According to Fedora's source [spec](https://src.fedoraproject.org/rpms/containers-common/blob/rawhide/f/containers-common.spec), various "common" files are pulled from containers/{common,image,storage,shortnames} which makes sense as they're all modules. But these 2 common files are pulled from skopeo which already depends on containers/common, so it doesn't make sense to have both in skopeo.

All major distros (Fedora, debian, archlinux, gentoo, alpine, opensuse) are placing these 2 files in containers-common package.
Thoughts on moving them to containers/image?

lsm5:
The 2 files in question are now in c/image.

mtrmac:
@lsm5 the Skopeo repo still contains copies. Can we remove them?

lsm5:
> @lsm5 the Skopeo repo still contains copies. Can we remove them?

The containers-common rpm [fetches them from c/image](https://github.com/containers/common/blob/main/rpm/containers-common.spec#L66-L67) . So it should be safe to remove unless other tests need it.

PR #2173, opened 2023-12-01
Moving default-policy.json and default.yaml to containers/image

rahilarious:
All major distros (Fedora, debian, archlinux, gentoo, alpine, opensuse) are placing these 2 files in containers-common package. Why not fix it upstream?

1st step towards addressing https://github.com/containers/skopeo/issues/2170


mtrmac (commented):
I have no opinion on the location of these files, I’ll fully defer to our packaging experts (and this _must_ be coordinated with them).

@lsm5 @jnovy PTAL.

---

I do have one opinion: Skopeo’s tests must continue to work, and this causes a failure.

.cirrus.yml
```diff
@@ -147,6 +147,7 @@ ostree-rs-ext_task:
         - dnf builddep -y skopeo
         - make
         - make install
+        - echo '{"default":[{"type":"insecureAcceptAnything"}],"transports":{"docker-daemon":{"":[{"type":"insecureAcceptAnything"}]}}}' > /etc/containers/policy.json
```
mtrmac, line 150: - I’m afraid this is failing
- I’m tempted to say that this should do the same thing we are asking other users to do. I.e. if users are expected to install the files from the c/image repo, this should also go through the troubler of installing the files from the c/image repo.

mtrmac:
For the record, the files, and the Make rules to install them, were moved to https://github.com/containers/image/pull/2215 .

.cirrus.yml
rahilarious: >     * I’m tempted to say that this should do the same thing we are asking other users to do. I.e. if users are expected to install the files from the c/image repo, this should also go through the troubler of installing the files from the c/image repo.

I agree. I just don't know how to handle cirrus stuff. It's up to you to properly fix it.

mtrmac: It is… just a shell script?

---

… Uh. The one thing we _might_ need to provide some guidance on is to figure out which c/image version to use. ~Luckily the data is available in `go.mod`, but turning that into a `git` command is a bit cumbersome.

@lsm5 @jnovy @rhatdan What do we do for Podman users who build from source, and need `containers.conf` from c/common? It seems to be either “nothing and let them figure it out” or “install a previous packaged of `containers-common` and hope the files match”, is there something I’m missing?

rhatdan: Nothing but let them figure it out.  Luckily Podman runs fine with no containers.conf and we encourage distributions to comment out containers.conf to be used information purposes only, IE Document the default in the system file, and allow admin and users to customize individual fields.

mtrmac: I’m afraid `policy.json` is mandatory. Oh well… I’m always happy to recommend using packaged versions :)

rahilarious:
> @mtrmac I guess we can revise this PR to handle any outstanding cleanup issues if @rahilarious is willing to continue this.

I'm unfamiliar with cirrus stuff, any help would be appreciated

rahilarious:
as per https://github.com/podman-container-tools/container-libs/blob/main/common/rpm/containers-common.spec , these files are being installed from `container-libs/image` repo so this PR should be merged now. 

PTAL @lsm5

lsm5:
@rahilarious there's a bash completion installation issue that's causing most (all?) tests to fail. PTAL.

rahilarious:
> @rahilarious there's a bash completion installation issue that's causing most (all?) tests to fail. PTAL.

Being a Gentoo guy, I'd prefer not to fiddle with rpm spec files. But still, fixed. It's all green now

lsm5:
> Being a Gentoo guy, I'd prefer not to fiddle with rpm spec files. But still, fixed. It's all green now

Sorry, I should've been clear. I mainly meant the github actions jobs which will also need the completions target. Consider Packit / testing-farm as secondary CI, and thanks for the spec file change :slightly_smiling_face: .

The testing-farm ELN job failures can be ignored here. I'll probably just remove them if they keep failing.

rahilarious:
> > Being a Gentoo guy, I'd prefer not to fiddle with rpm spec files. But still, fixed. It's all green now
> 
> Sorry, I should've been clear. I mainly meant the github actions jobs which will also need the completions target. Consider Packit / testing-farm as secondary CI, and thanks for the spec file change 🙂 .
> 
> The testing-farm ELN job failures can be ignored here. I'll probably just remove them if they keep failing.

Let me know if anything else is needed from my side

lsm5:
> Let me know if anything else is needed from my side

Yes, see previous comment about the failing github actions jobs that also need the `completions` target added in `doccheck` and `test_skopeo` jobs. These do not reuse the spec file. You'll see them in `.github/workflows/ci.yml`

lsm5:
@podman-container-tools/skopeo-maintainers @podman-container-tools/skopeo-reviewers PTAL as well.

If there are concerns about breaking current Makefile behaviour, maybe we could probably consider the `nobuild` approach: https://github.com/podman-container-tools/buildah/pull/6984

Makefile
```diff
@@ -159,13 +152,8 @@ clean:
...
 
 install: install-binary install-docs install-completions
-	install -d -m 755 ${DESTDIR}${LOOKASIDEDIR}
-	install -d -m 755 ${DESTDIR}${CONTAINERSCONFDIR}
-	install -m 644 default-policy.json ${DESTDIR}${CONTAINERSCONFDIR}/policy.json
-	install -d -m 755 ${DESTDIR}${REGISTRIESDDIR}
-	install -m 644 default.yaml ${DESTDIR}${REGISTRIESDDIR}/default.yaml
 
-install-binary: bin/skopeo
+install-binary:
```
mtrmac, line 156: The closest thing there is to a standard for this, https://www.gnu.org/prep/standards/html_node/Standard-Targets.html , specifies that `install` should build if necessary. Admittedly our use of phony targets for the things to be installed is very suboptimal.

Either way I don’t know why the changes to `install-binary` or `install-completions` or `skopeo.spec` should be a part of this PR at all.

.cirrus.yml
```diff
@@ -160,6 +160,7 @@ ostree-rs-ext_task:
...
```
mtrmac: > as per https://github.com/podman-container-tools/container-libs/blob/main/common/rpm/containers-common.spec , these files are being installed from `container-libs/image` repo so this PR should be merged now.

Well then the CI stanza that adds a hard-coded policy shouldn’t be needed, and shouldn’t be part of the PR?!

(In the meantime, we have migrated the CI from `.cirrus.yml` to GitHub actions; that makes the `.cirrus.yml` change ineffective, but either way the PR should not be adding that line.)

mtrmac (changes requested):
https://github.com/podman-container-tools/skopeo/blob/main/install.md documents how users can use `make install`, and that would now be _insufficient_ to obtain a working installation; so that must be extended to also add the other required steps … and that gets us back to https://github.com/podman-container-tools/skopeo/pull/2173#discussion_r1443011122 .

rahilarious:
> https://github.com/podman-container-tools/skopeo/blob/main/install.md documents how users can use `make install`, and that would now be _insufficient_ to obtain a working installation

According to install.md user would 1st run `make` (also build completions now) then `sudo make install` which would only install completions. Nobody should be running `sudo make install` directly anyways.

Makefile
rahilarious: > Either way I don’t know why the changes to `install-binary` or `install-completions` or `skopeo.spec` should be a part of this PR at all

You're right, at least by PR title they shouldn't be there. Should I create new PR for those or change title mentioning both changes?

mtrmac:
_Almost nothing_ works if `policy.json` does not exist! So if `make install` from this project does not install it, the existing instructions are not sufficient.

rahilarious:
> Yes, see previous comment about the failing github actions jobs that also need the `completions` target added in `doccheck` and `test_skopeo` jobs. These do not reuse the spec file. You'll see them in `.github/workflows/ci.yml`

done

rahilarious:
> _Almost nothing_ works if `policy.json` does not exist! So if `make install` from this project does not install it, the existing instructions are not sufficient.

Skopeo depends on container-libs/image. policy.json should be installed by c/image. Should I add the line in install.md that c/image should be installed first?

.cirrus.yml
rahilarious: done. Removed that line

rahilarious:
ping @mtrmac as https://github.com/podman-container-tools/buildah/pull/6984 also depends on this

mtrmac:
> The closest thing there is to a standard for this, https://www.gnu.org/prep/standards/html_node/Standard-Targets.html , specifies that `install` should build if necessary.

---

And we’ve had this discussion several years ago: https://github.com/podman-container-tools/skopeo/pull/2172#pullrequestreview-1762472451

mtrmac (changes requested):
https://github.com/podman-container-tools/skopeo/pull/2173#pullrequestreview-4765932069 still applies AFAICT.

rahilarious:
Ok. I guess I'll just 

> > The closest thing there is to a standard for this, https://www.gnu.org/prep/standards/html_node/Standard-Targets.html , specifies that `install` should build if necessary.
> 
> And we’ve had this discussion several years ago: [#2172 (review)](https://github.com/podman-container-tools/skopeo/pull/2172#pullrequestreview-1762472451)

Good memory :) I'll close this one and maintain patch downstream.

PR #2175, opened 2023-12-06
Add support for --override-platform flag

### LINKED ISSUE #2118: support override via platform specifier
Currently skopeo accepts three distinct flags for overriding the default (host) platform selection: `--override-os`, `--override-arch`, and `--override-variant`.

When working with other tools such as buildkit, the platform is instead specified as a single slash-delimited string, such as `linux/arm64` or `linux/arm/v7`. This format is explained in further detail [here](https://github.com/containerd/containerd/blob/8686ededfc90076914c5238eb96c883ea093a8ba/platforms/platforms.go#L63).

It would be nice if skopeo accepted a new flag such as `--override-platform` that accepted such a platform specifier. And probably the three existing override flags should be deprecated.

rittneje:
Closes #2118.

### DIFF
--- a/cmd/skopeo/main.go
+++ b/cmd/skopeo/main.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/containerd/containerd/platforms"
 	commonFlag "github.com/containers/common/pkg/flag"
 	"github.com/containers/image/v5/signature"
 	"github.com/containers/image/v5/types"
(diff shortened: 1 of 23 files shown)


mtrmac (commented):
Thanks!

My immediate reaction is that I don’t know how I feel about the way `platforms.Parse` defaults values; compare https://github.com/containers/skopeo/pull/1987#discussion_r1183115878 . The trade-off is consistency with `containerd`, which is certainly _somewhat_ valuable.

Either way, this and #1987 should probably end up with a consistent answer to this trade-off.

rhatdan:
@nalind thoughts?

nalind:
This doesn't break anything, and it gets us closer to not being in the business of figuring out for ourselves which "variant" value to use on ARM boxes, so I'm fine with it.  If that means reworking #1987, or scrapping it entirely, I'm not going to be particularly upset.

PR #2182, opened 2023-12-19
Accept URI for Sigstore Signed Images

lukewarmtemp:
Fixes https://github.com/coreos/rpm-ostree/issues/4272#issuecomment-1689160021

Allows for container images signatures with a URI instead of a email
address as the SAN. This is needed in certain cases such as using a github workflow to sign a container using the github actions OIDC token.

The URI field was added to the `fulcioTrustRoot` struct and associated functions were modifed/created. Logic to handle either having a URI or email address as the SAN was also implemented.

**Testing:**
1. Copy and store `fulcio_v1.crt.pem` to `/etc/pki/containers/fulcio.sigstore.dev.pub` and `rekor.pub` to `/etc/pki/containers/rekor.sigstore.dev.pub`, both of which can be found at: https://github.com/sigstore/root-signing/blob/main/README.md

2. Configure `/etc/containers/policy.json`
```
{
  "default": [
    {
      "type": "reject"
    }
  ],
  "transports": {
    "docker": {
      "registry.access.redhat.com": [
        {
          "type": "signedBy",
          "keyType": "GPGKeys",
          "keyPath": "/etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release"
        }
      ],
      "registry.redhat.io": [
        {
          "type": "signedBy",
          "keyType": "GPGKeys",
          "keyPath": "/etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release"
        }
      ],
      "ghcr.io/jmpolom": [
        {
          "type": "sigstoreSigned",
          "fulcio": {
            "caPath": "/etc/pki/containers/fulcio.sigstore.dev.pub",
            "oidcIssuer": "https://token.actions.githubusercontent.com",
            "URI": "https://github.com/jmpolom/fedora-ostree-ws-container/.github/workflows/build.yml@refs/heads/main"
          },
          "rekorPublicKeyPath": "/etc/pki/containers/rekor.sigstore.dev.pub",
          "signedIdentity": {
            "type": "matchRepository"
          }
        }
      ]
    },
    "docker-daemon": {
      "": [
        {
          "type": "insecureAcceptAnything"
        }
      ]
    }
  }
}
```

3. Create `/etc/containers/registries.d/ghcr.io.yaml`
```
docker:
     ghcr.io/jmpolom:
         use-sigstore-attachments: true
```

4. Build the skopeo binary and run the following
```
$ ./bin/skopeo copy docker://ghcr.io/jmpolom/fedora-silverblue-ws:38-main dir:/some/local/directory
```

### DIFF
--- a/vendor/github.com/containers/image/v5/signature/fulcio_cert.go
+++ b/vendor/github.com/containers/image/v5/signature/fulcio_cert.go
@@ -7,6 +7,7 @@ import (
 	"encoding/asn1"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/containers/image/v5/signature/internal"
@@ -21,14 +22,15 @@ type fulcioTrustRoot struct {
 	caCertificates *x509.CertPool
 	oidcIssuer     string
 	subjectEmail   string
+	URI            string
 }
 
 func (f *fulcioTrustRoot) validate() error {
 	if f.oidcIssuer == "" {
 		return errors.New("Internal inconsistency: Fulcio use set up without OIDC issuer")
 	}
-	if f.subjectEmail == "" {
-		return errors.New("Internal inconsistency: Fulcio use set up without subject email")
+	if f.subjectEmail == "" && f.URI == "" {
+		return errors.New("Internal inconsistency: Fulcio use set up without subject email or URI")
 	}
 	return nil
 }
@@ -174,10 +176,17 @@ func (f *fulcioTrustRoot) verifyFulcioCertificateAtTime(relevantTime time.Time, 
 	}
 
 	// == Validate the OIDC subject
-	if !slices.Contains(untrustedCertificate.EmailAddresses, f.subjectEmail) {
-		return nil, internal.NewInvalidSignatureError(fmt.Sprintf("Required email %s not found (got %#v)",
-			f.subjectEmail,
-			untrustedCertificate.EmailAddresses))
+	if !slices.Contains(untrustedCertificate.EmailAddresses, f.subjectEmail) && !strings.Contains(untrustedCertificate.URIs[0].String(), f.URI) {
+		if len(untrustedCertificate.EmailAddresses) > 0 {
+			return nil, internal.NewInvalidSignatureError(fmt.Sprintf("Required email %s not found (got %#v)",
+				f.subjectEmail,
+				untrustedCertificate.EmailAddresses))
+		}
+		if len(untrustedCertificate.URIs) > 0 {
+			return nil, internal.NewInvalidSignatureError(fmt.Sprintf("Required URI %s not found (got %#v)",
+				f.URI,
+				untrustedCertificate.URIs))
+		}
 	}
 	// FIXME: Match more subject types? Cosign does:
 	// - .DNSNames (can’t be issued by Fulcio)
--- a/vendor/github.com/containers/image/v5/signature/policy_config_sigstore.go
+++ b/vendor/github.com/containers/image/v5/signature/policy_config_sigstore.go
@@ -261,6 +261,17 @@ func PRSigstoreSignedFulcioWithSubjectEmail(subjectEmail string) PRSigstoreSigne
 	}
 }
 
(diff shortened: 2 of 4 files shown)


lukewarmtemp:
Hello @mtrmac, this is a small patch I've made to the `sigstore` verification process in order to address the issue linked in the PR message. I was wondering if you could both review my PR and offer some insight on some next steps/ways to generalize the patch so that we can catch more bugs related to `sigstore` from coming up down the line.

cgwalters:
Thanks for working on this!

Without digging into the details, I think (unless I'm misunderstanding something here) this should be a PR against https://github.com/containers/image - any changes to the vendored sources here would be lost when updating.

BTW a common thing when developing Go is to use [the replace directive](https://go.dev/ref/mod#go-mod-file-replace) so that you can build a `skopeo` binary with your changes to the vendored library for integration testing.

vendor/github.com/containers/image/v5/signature/fulcio_cert.go
```diff
@@ -174,10 +176,17 @@ func (f *fulcioTrustRoot) verifyFulcioCertificateAtTime(relevantTime time.Time,
...
+		if len(untrustedCertificate.EmailAddresses) > 0 {
+			return nil, internal.NewInvalidSignatureError(fmt.Sprintf("Required email %s not found (got %#v)",
+				f.subjectEmail,
+				untrustedCertificate.EmailAddresses))
+		}
+		if len(untrustedCertificate.URIs) > 0 {
+			return nil, internal.NewInvalidSignatureError(fmt.Sprintf("Required URI %s not found (got %#v)",
+				f.URI,
+				untrustedCertificate.URIs))
+		}
```
mtrmac, line 189: - ⚠️ If the policy requires _both_ an email address and an URI, this must not accept signatures which only have one of them
- This is invoking `slices.Contains(…, "")` if the field is not set. That’s at the very least unclean.
- I suspect the error reporting logic should also be tightened. It must primarily decide based on the _policy_, not on attacker-set `untrusted*` fields.

vendor/github.com/containers/image/v5/signature/policy_config_sigstore.go
```diff
@@ -329,9 +343,12 @@ func (f *prSigstoreSignedFulcio) UnmarshalJSON(data []byte) error {
 	if gotOIDCIssuer {
 		opts = append(opts, PRSigstoreSignedFulcioWithOIDCIssuer(tmp.OIDCIssuer))
 	}
-	if gotSubjectEmail {
+	if gotSubjectEmail && !gotURI {
 		opts = append(opts, PRSigstoreSignedFulcioWithSubjectEmail(tmp.SubjectEmail))
 	}
+	if !gotSubjectEmail && gotURI {
+		opts = append(opts, PRSigstoreSignedFulcioWithURI(tmp.URI))
+	}
```
mtrmac, line 351: This is silently ignoring the options if both are specified?!

mtrmac (commented):
Thanks!
- Yes, this needs to be a PR against containers/image, not against Skopeo . Please file one and we’ll continue there.
- I’m not immediately sure the URI field is the one we should target. Can you post the full `openssl x509 -text` of the generated certificate, please? https://github.com/sigstore/fulcio/blob/main/docs/oid-info.md#directory documents various more specific fields, maybe this should actually target one (or more?) of them. OTOH just matching on SubjectAltName URIs might be more general. I don’t have a strong opinion so far.
- For c/image/signature, it would also need unit tests with close-to-perfect coverage (both of the policy parsing and implementation/enforcement).

lukewarmtemp:
Moved to https://github.com/containers/image/pull/2235

vendor/github.com/containers/image/v5/signature/fulcio_cert.go
jmpolom: > If the policy requires both an email address and an URI, this must not accept signatures which only have one of them

I believe this would be an invalid policy with cosign. One or the other.

mtrmac: Then such an invalid policy must be rejected at parsing/initialization time, not processed in an unexpected way.

PR #2189, opened 2024-01-04
Adding support for semver filter in sync command 

husseinferas:
Adding Kubasobon's changes in the [PR](https://github.com/containers/skopeo/pull/1918) and rebase it with main to move it forward 

more changes added:
- using single semver constraint
```yaml
docker.io:
  images-by-semver:
    node: ">= 20.0.0"
```


mtrmac (commented):
Thanks for taking this on!

Outstanding review comments: https://github.com/containers/skopeo/pull/1918#pullrequestreview-1324143881

(Also, eventually please squash the “formate sync.go file” commit, so that the code passes tests on every commit.)

husseinferas:
> error: RPC failed; HTTP 500 curl 22 The requested URL returned error: 500
@mtrmac 
Regarding the CI failure, I can't retry the step on cirrus-ci.

husseinferas:
> error: RPC failed; HTTP 500 curl 22 The requested URL returned error: 500
@mtrmac 
Regarding the CI failure, I can't retry the step on cirrus-ci, I think it's only a network gulch

cmd/skopeo/sync.go
```diff
@@ -439,60 +440,60 @@ func filterSourceReferences(sys *types.SystemContext, registryName string, colle
 			continue
 		}
 
-		sourceReferences = filter(logger, sourceReferences)
-		if len(sourceReferences) == 0 {
+		var filteredSourceReferences []types.ImageReference
+		for _, ref := range sourceReferences {
+			if filter(logger, ref) {
+				filteredSourceReferences = append(filteredSourceReferences, ref)
```
mtrmac, line 446: `break` here, so that multiple filters don’t cause the same image to be copied twice.

cmd/skopeo/sync.go
```diff
@@ -368,61 +378,154 @@ func imagesToCopyFromRegistry(registryName string, cfg registrySyncConfig, sourc
...
+			tagVersion, err := semver.NewVersion(tagged.Tag())
+			if err != nil {
+				logger.Tracef("Tag %q cannot be parsed as semver, skipping", tagged.Tag())
+				return false
+			}
+			for _, constraint := range constraints {
+				if !constraint.Check(tagVersion) {
+					return false
+				}
+			}
```
mtrmac, line 521: The previous version accepted a tag if any of the constraints matched; this one only accepts a tag if all of them match. That doesn’t match the documentation, AFAICS for the
```
            - "3.12 - 3.13"
            - ">= 3.17"
```
inputs no tag could ever match.

mtrmac (commented):
Thanks!

(Eventually a rebase&squash would be preferred over intermediate commits and merges, but that’s non-blocking)

> RPC failed; HTTP 500 curl 22 The requested URL returned error: 500

Let’s see if a future version passes; the macOS failures at that stage are certainly unrelated.

cmd/skopeo/sync.go
husseinferas: honestly I changed that on purpose I think the relation between array of tags should be "AND" instead of "OR"
let's take this example
```
docker.io:
  images-by-semver:
    node: 
      - "18 - 20"
      - ">= 18.19"
      - "<= 20.0"
```
In the previous code, the selected images were all of these constraints. It even selected old versions from `8` to the newer versions like `21`. In my view, this is confusing because what is the point of having multiple constraints then? If we can't limit the filter more by adding more constraints, otherwise we just need to provide one constraint instead of a list. 
the current setup will reflect the meaning of filter and each constraint in the list will limit the output more 
I'm happy to update the docs to match the current behaviour if we agree on it.

mtrmac: Most importantly the code must match the documentation.

---

https://github.com/Masterminds/semver#basic-comparisons shows that there already is a way to AND / OR constraints; there is only one level of nesting, but that should be sufficient to express any policy. (`a AND (b OR c)` == `(a AND b) OR (a AND c)`.

So, in that sense, it doesn’t matter whether we OR or AND the top-level entries — we might not even need multiple lines, just one constraint per repo, and let the user do the boolean math to form it.

I’d argue for the previous (OR) semantics:
- It is consistent with the way `cfg.Images` works; those are ORed
- I don’t expect constraints to be that complicated? The three constraints in your example seem just to be redundant, equivalent to `18.19 - 20.0`; users would presumably just write that one. I would _guess_ that a typical rule to sync an image can well fit on one line, using the `Masterminds/semver` AND/OR language. I don’t have extensive practical experience, that might be wrong.
- OTOH it does seem to me that it would be _fairly frequent_ to sync several narrowly-specified stable versions: “we are deploying foo v1.2, v1.4, and v1.8 throughout the company; grab all 1.X.Y versions of those”. The OR semantics makes it easy to express that.

So I’d prefer OR but that’s not a blocker; the code matching the documentation is a blocker.

husseinferas: I think having a single constraint is enough to use and clear from a user perspective 
if the users need more advanced filtering, they can use regex. The point of this feature is to provide an easy way to reference tags 
and single constraint will match the behaviour of regex 

```
docker.io:
  images-by-semver:
    alpine: ">= 3.12"
```

cmd/skopeo/sync.go
husseinferas: adding the `break` here made the program stop after finding the first tag that match regex or semver

husseinferas: I think you meant `continue` instead of `break`? it worked with `continue`

husseinferas:
I now have problem: when I run the sync command, it prints "Start filtering using the regular expression:" log message multiple times, it is definitely being called in a loop but it is not obvious to me where @mtrmac

MShekow:
Looking forward to getting this PR merged.

Do @husseinferas or @mtrmac have any opinions on what I posted in https://github.com/containers/skopeo/pull/1918#issuecomment-1875414488 ? 

From long-term experience I can conclude that:
- using only regular expressions is not a good solution, because the regexp either matches _too many_ tags, or you need to build a very complicated regex that is hard to understand and maintain
- using only semver comparisons is not possible in practice, because many container images do _not_ use "proper" semver tags
- The only solution I can think of is to _mix_ both approaches, as I discussed in https://github.com/containers/skopeo/pull/1918#issuecomment-1875414488

My gut feeling is that _now_ would be the best time to solve that problem, as otherwise breaking changes might be inevitable.

husseinferas:
Thank you @MShekow 
I really like your proposal but I would say this is an improvement that we can make later after introducing `images-by-semver` first. 
I'd prefer to implement this in different PR

MShekow:
> Thank you @MShekow I really like your proposal but I would say this is an improvement that we can make later after introducing `images-by-semver` first. I'd prefer to implement this in different PR

I fully agree with you to keep PRs (which _implement_ things) small. What I'm suggesting is merely that we should already brainstorm whether any potential _syntax_ changes (in the input yaml file) would be good to introduce _now_.

cmd/skopeo/sync.go
mtrmac: Regex and version comparisons seem rather different to me, one is hardly a replacement for the other.

\*shrug\* there is `||` to express OR, so a single expression should work.

cmd/skopeo/sync.go
mtrmac: I think I was just wrong — for some reason I thought this is iterating over filters, not over tags.

`continue` does nothing at the ~end of a loop.

mtrmac:
@MShekow Looking at your proposal, it does seem to be compatible with the proposed structure, right now. Is there anything that you think needs changing?

MShekow:
> @MShekow Looking at your proposal, it does seem to be compatible with the proposed structure, right now. Is there anything that you think needs changing?

To reiterate, the currently proposed structure (that is also implemented in this PR) is something like this:

```yaml
images-by-semver:
  alpine:
    - "3.12 - 3.13"
```
I can't think of a "good" solution for adding the prefilter-regex (that contains the named capture group) that I suggested in https://github.com/containers/skopeo/pull/1918#issuecomment-1875414488. You would have to use funky syntax, mangling the prefilter-regex into the version string somehow, e.g. by replacing the version string (here: `"3.12 - 3.13"`) with something like `"3.12 - 3.13|||stable-(?P<semver>\\d+\\.\\d+\\.\\d)"` (using an arbitrary delimiter like `|||`).

That was the entire reason why I proposed to add an additional YAML "layer" in https://github.com/containers/skopeo/pull/1918#issuecomment-1875414488 that separates the list of semver-strings from the list of prefilter-regexes.

husseinferas:
good, so our current structure will work just fine in future with regex layer

```
images-by-semver:
  alpine: "3.12 - 3.13"
```
with regex, in next implementation 

```
images-by-semver:
  alpine: "3.12 - 3.13 || stable-(?P<semver>\\d+\\.\\d+\\.\\d)"
```

cmd/skopeo/sync.go
husseinferas: @mtrmac I think the code is now ready for your review

TomSweeneyRedHat:
Code LGTM

TomSweeneyRedHat:
@husseinferas, the branch needs updating. I've just kicked that off.  Once completed, I'd like @mtrmac to do the final approval to ensure his comments were addressed.

mtrmac:
> That was the entire reason why I proposed to add an additional YAML "layer" in https://github.com/containers/skopeo/pull/1918#issuecomment-1875414488 that separates the list of semver-strings from the list of prefilter-regexes.

Oh… now I get what you mean.

I definitely agree that a separate field is better than trying to stuff things inside the server filter. OTOH it’s a bit tedious to force every user to add one more level of nesting for a hopefully fairly rare case.

Instead of
```yaml
registry.example.com:
    images-by-semver:
        alpine:
            regex-filters: "…"
            versions: ">= 3.12.0"
```
we can do
```yaml
registry.example.com:
    images-by-semver:
        alpine: ">= 3.12.0"
    semver-regex-filters:
        alpine: "…"
```
and only the repos which need that filter would pay the price.

(Really long-term we should probably end up with
```yaml
sync:
   -  source: registry.source.com/source-repo
       destination: registry.source.com/dest-repo
       tags: …
       regex: …
       semver: …
       semver-regex-filter: …
```
but that’s a long-term future direction.)

cmd/skopeo/sync.go
```diff
@@ -71,6 +72,7 @@ type tlsVerifyConfig struct {
 type registrySyncConfig struct {
 	Images           map[string][]string    // Images map images name to slices with the images' references (tags, digests)
 	ImagesByTagRegex map[string]string      `yaml:"images-by-tag-regex"` // Images map images name to regular expression with the images' tags
+	ImagesBySemver   map[string]string      `yaml:"images-by-semver"`    // ImagesBySemver maps images name with a list of semver constraints (e.g. '>=3.14') to match images' tags to
```
mtrmac, line 75:
```suggestion
	ImagesBySemver   map[string]string      `yaml:"images-by-semver"`    // ImagesBySemver maps a repository to a semver constraint (e.g. '>=3.14') to match images' tags to
```

mtrmac (commented):
Thanks!

LGTM; please squash&rebase into a single commit.

mtrmac:
(For transparency, I have rebased the work, to avoid WIP commits; per https://github.com/containers/skopeo/compare/db8f93e33ea3b99b157ef0c74226c922d7ecd658..177d4adb205dffa0e0e50d0709ee27b953ee0c00 the code is identical.)

MShekow:
I'm super happy that this PR has come to a successful conclusion :).

@husseinferas are you interested in implementing the _regex filter_ I mentioned (using the proposal made in https://github.com/containers/skopeo/pull/2189#issuecomment-1898993006), to increase the usefulness of this PR's semver filter? If not, I would have a go at it, it might take a while, however.

husseinferas:
Thanks @MShekow @mtrmac 
@mtrmac, should we wait for the next version release to start using Semver?

I think we need to update the sync YAML file structure for the next iteration because the current sync file is not scalable and has another problem that we need to fix, which is the destination repository mentioned [here](https://github.com/containers/skopeo/pull/1531).

```yaml
sync:
   -  source: registry.source.com/source-repo
       destination: registry.source.com/dest-repo
       tags: …
       regex: …
       semver: …
       semver-regex-filter: … 
```

I am willing to work on this feature, @mtrmac   How does the team communicate? Is there a Slack channel or chat group somewhere? I had many questions about this update before I started working on it.

MShekow:
Congrats on the release in v1.15.0 🎉

> Thanks @MShekow @mtrmac @mtrmac, should we wait for the next version release to start using Semver?
> 
> I think we need to update the sync YAML file structure for the next iteration because the current sync file is not scalable and has another problem that we need to fix, which is the destination repository mentioned [here](https://github.com/containers/skopeo/pull/1531).
> 
> ```yaml
> sync:
>    -  source: registry.source.com/source-repo
>        destination: registry.source.com/dest-repo
>        tags: …
>        regex: …
>        semver: …
>        semver-regex-filter: … 
> ```
> 
> I am willing to work on this feature, @mtrmac How does the team communicate? Is there a Slack channel or chat group somewhere? I had many questions regarding this update.

Really looking forward to the next iteration. I suppose that we're waiting for input from @mtrmac ?

mtrmac:
In general there’s a long-term need to redesign the new YAML format; the GitHub repo contains quite a few open issues directly or indirectly touching on the needs and design requirements. I’m afraid it’s also not a feature I’ve not been able to invest significant time in.

These GitHub issues are the primary collaboration mechanism of the project.

PR #2213, opened 2024-02-02
Update after https://github.com/containers/image/pull/2273

mtrmac:
This is necessary so that c/image tests can pass.


mtrmac:
(I will merge without a review, to keep c/image CI working. Hopefully someone looked at this as a part of https://github.com/containers/image/pull/2273 .)

PR #2232, opened 2024-02-16
Add function to check if REGISTRY_AUTH_FILE is valid json

JoshPasek:
if REGISTRY_AUTH_FILE is set, then the script fails if the file is empty. Wanting to add a simple script so that if the file is empty then  it adds valid json.


mtrmac (changes requested):
This
- says it adds a function, and doesn’t
- Does not compile
- reads a file in a public `/tmp` without having any reason to expect other users haven’t used it for their own purposes
- deals with a file content _at all_, when this is a unit test for option parsing and the file is never read in the code under test.

What are you actually trying to do?

PR #2252, opened 2024-03-09
Update to Go 1.20

mtrmac:
After https://github.com/containers/image/pull/2337 (notably https://github.com/go-openapi/swag/compare/v0.22.10...v0.23.0 ), c/image will require Go 1.20 to build.

This does not yet update to that version of c/image, but that will happen soon enough.

So, update to Go 1.20, and take advantage of new standard library features.


felipecrs:
Agreed. v1.15.0 fails to compile with go 1.19.

TomSweeneyRedHat:
LGTM
@cevich PTAL

rhatdan:
LGTM

cevich:
golang 1.20 is fine, it's 1.21 we need to be careful with.

PR #2262, opened 2024-03-18
[skip-ci] rpm: use macro supported vendoring

lsm5:
This removes the need for any `Provides: bundled()` we needed in spec files.

The updated Provides will be visible in the build logs and can also be verified with `rpm -q --provides $RPM_FILE`.


mtrmac (approved):
Thanks! This is nice.

PR #2273, opened 2024-03-27
[CI:DOCS] Add golang 1.21 update warning

cevich:
This is needed on the off-chance that some tool or a human suggests updating the minimum version to 1.21 or later. Since doing so would cause Fedora and Debian to start behaving differently WRT builds.

Ref: https://github.com/containers/podman/pull/22073


mtrmac (approved):
Thanks!

PR #2277, opened 2024-03-29
feat: add `--image-parallel-copies` flag

Atry:
It would be useful when there are many layers while most of layers are unchanged between versions.

See discussion at https://github.com/nlewo/nix2container/pull/125

### DIFF
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -44,6 +44,7 @@ type copyOptions struct {
 	encryptLayer             []int                     // The list of layers to encrypt
 	encryptionKeys           []string                  // Keys needed to encrypt the image
 	decryptionKeys           []string                  // Keys needed to decrypt the image
+	imageParallelCopies      uint                      // Maximum number of parallel requests when copying images
 }
 
 func copyCmd(global *globalOptions) *cobra.Command {
@@ -95,6 +96,7 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
 	flags.StringSliceVar(&opts.encryptionKeys, "encryption-key", []string{}, "*Experimental* key with the encryption protocol to use needed to encrypt the image (e.g. jwe:/path/to/key.pem)")
 	flags.IntSliceVar(&opts.encryptLayer, "encrypt-layer", []int{}, "*Experimental* the 0-indexed layer indices, with support for negative indexing (e.g. 0 is the first layer, -1 is the last layer)")
 	flags.StringSliceVar(&opts.decryptionKeys, "decryption-key", []string{}, "*Experimental* key needed to decrypt the image")
+	flags.UintVar(&opts.imageParallelCopies, "image-parallel-copies", 0, "Maximum number of image layers to be copied (pulled/pushed) simultaneously. Not setting this field will fall back to containers/image defaults.")
 	return cmd
 }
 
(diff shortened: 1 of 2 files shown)


cmd/skopeo/copy.go
```diff
@@ -95,6 +96,7 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
 	flags.StringSliceVar(&opts.encryptionKeys, "encryption-key", []string{}, "*Experimental* key with the encryption protocol to use needed to encrypt the image (e.g. jwe:/path/to/key.pem)")
 	flags.IntSliceVar(&opts.encryptLayer, "encrypt-layer", []int{}, "*Experimental* the 0-indexed layer indices, with support for negative indexing (e.g. 0 is the first layer, -1 is the last layer)")
 	flags.StringSliceVar(&opts.decryptionKeys, "decryption-key", []string{}, "*Experimental* key needed to decrypt the image")
+	flags.UintVar(&opts.maxParallelDownloads, "max-parallel-downloads", 6, "Maximum number of parallel downloads")
```
mtrmac, line 99: This should not hard-code a value; by default, this should be set to 0 so that the c/image default applies.

mtrmac, line 99: The “downloads” name is not really accurate; `skopeo copy` could have well been used for a “push” or a registry--to-registry-copy.

(Yes, the `copy.Options` field uses that naming. That was a mistake.)

`containers.conf` uses `image_parallel_copies`; that could probably be also improved upon, but at least it’s better in this respect, and the consistency seems attractive to me.

mtrmac (commented):
Thanks!

Compare https://github.com/containers/image/pull/1968 ; that might, or might not, also have a noticeable effect for that use case.

(Hypothetically even to the point of making this option unnecessary? I think that’s unlikely, but worth measuring.)

cmd/skopeo/copy.go
```diff
@@ -95,6 +96,7 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
 	flags.StringSliceVar(&opts.encryptionKeys, "encryption-key", []string{}, "*Experimental* key with the encryption protocol to use needed to encrypt the image (e.g. jwe:/path/to/key.pem)")
 	flags.IntSliceVar(&opts.encryptLayer, "encrypt-layer", []int{}, "*Experimental* the 0-indexed layer indices, with support for negative indexing (e.g. 0 is the first layer, -1 is the last layer)")
 	flags.StringSliceVar(&opts.decryptionKeys, "decryption-key", []string{}, "*Experimental* key needed to decrypt the image")
+	flags.UintVar(&opts.imageParallelCopies, "image-parallel-copies", 0, "Maximum number of image layers to be copied (pulled/pushed) simultaneously. Not setting this field will fall back to containers/image defaults. (6)")
```
mtrmac, line 99: I don’t think this should be making any promises about the default parallelism, in either place.

mtrmac (commented):
Does https://github.com/containers/image/pull/1968 change the need for this flag?

cmd/skopeo/copy.go
Atry: Fixed!

Atry:
> Does [containers/image#1968](https://github.com/containers/image/pull/1968) change the need for this flag?

Both PRs can speed up `skopeo copy`. It's better to have both.

TomSweeneyRedHat:
@Atry Looks like this needs a rebase.
Changes LGTM on a quick scan.

Atry:
Rebased

rhatdan:
LGTM
@mtrmac PTAL

mtrmac (approved):
Thanks again!

PR #2161, opened 2023-11-24
[hack/libdm_tag.sh] Fix failed build in absence of libdevmapper.h

rahilarious:

PR #2283, opened 2024-04-05
Hard-code the device-mapper graph driver to disabled

mtrmac:
This follows similar changes to Podman in
https://github.com/containers/podman/commit/407354198169f33a1ebebca3ae47d5d0c34d9b42 and Buildah in
https://github.com/containers/buildah/pull/4832

Compare earlier #2161 .

(I didn’t actually _test_ any of the updated installation instructions.)

Cc: @giuseppe


rpm/skopeo.spec
```diff
@@ -67,7 +67,6 @@ BuildRequires: go-rpm-macros
 %endif
 BuildRequires: gpgme-devel
 BuildRequires: libassuan-devel
-BuildRequires: pkgconfig(devmapper)
```
mtrmac, line 70: Cc: @lsm5 

Also @cevich — maybe the dependency can be removed from some of the `automation_images` as well.

cevich: Thanks for thinking of the CI VMs.  We do install `device-mapper-devel` in them, but the images are used across many repos.  W/o trying or checking every single one of them, I can't think of an easy way to confirm :thinking:  I think I'm more comfortable leaving it there, and up to some future CI VM package-pruning effort.  Though please let me know if you object for any reason.

TomSweeneyRedHat (commented):
LGTM
but would like a head nod from @giuseppe

giuseppe (approved):
LGTM

lsm5 (approved):
LGTM

PR #2280, opened 2024-04-03
Freeze the fedora-minimal image reference at Fedora 38

mtrmac:
... because the tests are assuming a v2s2 image, but as of Fedora 39, the image uses the OCI format.

See e.g. https://api.cirrus-ci.com/v1/task/6504550402621440/logs/integration.log for the test failures this is trying to fix.

PR #2285, opened 2024-04-08
[release-1.13] Backport #2280

mtrmac:
This branch is used by the c/image `release-5.26` branch.

### DIFF
--- a/integration/copy_test.go
+++ b/integration/copy_test.go
@@ -30,7 +30,8 @@ const (
 	v2DockerRegistryURL   = "localhost:5555" // Update also policy.json
 	v2s1DockerRegistryURL = "localhost:5556"
 	knownWindowsOnlyImage = "docker://mcr.microsoft.com/windows/nanoserver:1909"
-	knownListImage        = "docker://registry.fedoraproject.org/fedora-minimal" // could have either ":latest" or "@sha256:..." appended
+	knownListImageRepo    = "docker://registry.fedoraproject.org/fedora-minimal"
+	knownListImage        = knownListImageRepo + ":38"
 )
 
 func TestCopy(t *testing.T) {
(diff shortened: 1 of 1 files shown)


mtrmac:
Merging without review to remove noise from daily CI runs.

lsm5:
@mtrmac checked with the packit folks. This occurred because the branch was changed from `main` to `release-1.13` after PR creation. I've filed an issue with them which has been linked in this PR too. So, I'd say we leave things as they are for now.

mtrmac:
@lsm5 Thanks!

lsm5:
@mtrmac i'm unable to reproduce this issue on https://github.com/containers/skopeo/pull/2298 . So, I wonder if this is a corner case.
-  Do you remember if you changed the target branch from `main` to `release-1.13` right after you created the PR and before the builds completed (regardless of completion status).
- Did you try rerunning the packit jobs after you changed the branch?

mtrmac:
@lsm5 the tooltips on “last week” for filing the PR and changing the branch are both 7:51, so that must have been an ~immediate edit (and that matches what I would expect to happen, as soon as I noticed the large diff I would retarget).

I really can’t remember whether I triggered a restart of the Packit jobs, I’m afraid.

lsm5:
> @lsm5 the tooltips on “last week” for filing the PR and changing the branch are both …:51, so that must have been an ~immediate edit (and that matches what I would expect to happen, as soon as I noticed the large diff I would retarget).

ah yes, I should've checked tooltips before asking :| , anyway thanks for confirming.
> 
> I really can’t remember whether I triggered a restart of the Packit jobs, I’m afraid.

No worries at all. Thanks for the info. I'll look further with the packit team, but hopefully this issue won't be regular occurrence.

PR #1813, opened 2022-11-22
[CI:BUILD] Cirrus: Migrate OSX task to M1

ashley-cui:
Migrate our OSX build to a M1 instance, since Cirrus is sunsetting Intel-based macOS instances.

Signed-off-by: Ashley Cui <acui@redhat.com>

PR #2293, opened 2024-04-11
[release-1.9] Bump ocicrypt and go-jose CVE-2024-28180

TomSweeneyRedHat:
Bump github.com/go-jose/go-jose to v3.0.0 and
github.com/containers/ocicrypt to v1.1.10

Addresses: CVE-2024-28180
https://issues.redhat.com/browse/OCPBUGS-30788


mtrmac (commented):
Thanks!
- Shouldn’t this also include an update to `gopkg.in/go-jose/go-jose.v2`?
- I don’t _mind_ the `ocicrypt` update, but it seems not to be directly related to the vulnerability. (One major effect it _does_ have is updating from go-jose v2 to v3. That might be a good enough reason to include it.) The downside of this is that it rather extends the scope of the update, so I just wanted to double-check that it is intentional.

Feel free to merge as is if all of this is intentional.

mtrmac:
At a first glance, the test failure probably requires back porting also https://github.com/containers/skopeo/pull/2286 .

TomSweeneyRedHat:
I've reworked this.  I skipped the ocicrypt bump, and just went with the sigstore bump like we decided upon in the release-1.11 branch #2292

mtrmac:
- (That’s a lot of updates for an old branch. I assume we are fine with that level of risk, and that it builds correctly on the relevant old target platforms — I haven’t checked.)
- Does this _intentionally_ avoid updating the `square/go-jose` version? I’d like to clearly differentiate between “we have checked and know that the vulnerability is unreachable” and “we didn’t check”. (Per earlier analysis, I’m fine with the `letsencrypt/boulder` caller calling an old version. I didn’t check `ocicrypt` in detail.)
- The test failures require at least #2286 , and doing something about the macOS test runner.

TomSweeneyRedHat:
I've done a cherry pick and repushed.  We'll see where that gets it.

mtrmac:
Linux looks good.

I think the macOS builds can just be disabled: They don’t matter for the backports. (Alternatively, https://github.com/containers/skopeo/pull/1813 , but we would be building with a recent Go, and that’s probably eventually going to break anyway.)

TomSweeneyRedHat:
I repushed with the .cirrus.yaml change. @mtrmac, I added them to the cherry pick without thinking. If you want me to move it to the first commit or a new commit, let me know.

@cevich, could you eyeball the .cirrus.yaml change, please?  My first official change to a CI yaml file.

.cirrus.yml
```diff
@@ -74,33 +74,39 @@ doccheck_task:
...
+#        brew install gpgme go@1.18 go-md2man
+#        go install golang.org/x/lint/golint@latest
+#    test_script: |
+#        export PATH=$GOPATH/bin:/usr/local/opt/go@1.18/bin:$PATH
+#        go version
+#        go env
+#        make validate-local test-unit-local bin/skopeo
+#        sudo make install
+#        /usr/local/bin/skopeo -v
+#######
```
cevich, line 105: It's not a big deal, but just a gentle reminder: This is version control, you can simply delete the lines and they will live on in the history :grin: 
(Though it helps the archaeologists if this is done in a dedicated commit)

.cirrus.yml
```diff
@@ -74,33 +74,39 @@ doccheck_task:
...
 
 cross_task:
     alias: cross
-    only_if: *not_docs_multiarch
+#    only_if: *not_docs_multiarch
```
cevich, line 109: Hrmm, this jumped out at me.   I checked and see it has been removed from `main` as well.  As above, I'd suggest just deleting the line instead of commenting it out.

.cirrus.yml
```diff
@@ -241,7 +247,7 @@ success_task:
     depends_on:
         - validate
         - doccheck
-        - osx
+#        - osx
```
cevich, line 250: If you clobber the commented-out task, make sure to remove this line also.

cevich (approved):
The `.cirrus.yml` changes mostly LGTM.  All my comments are non-blocking, just suggestions.  I'm happy to take a peek @TomSweeneyRedHat you did a fine job :smile:

.cirrus.yml
TomSweeneyRedHat: I was on the fence for the clobber, I'll go do so and will move the yaml change to a new commit.

mtrmac (approved):
Thanks again!

Merging as is.

@TomSweeneyRedHat just to make sure:

The vulnerability is in `JSONWebEncryption.Decrypt` and `JSONWebEncryption.DecryptMulti`.

`ocicrypt` does actually call `DecryptMulti`, and on this brach, it is calling the `square` version of the library which remains at unpatched 2.6.0. OTOH the call is in `jweKeyWrapper.UnwrapKey`, processing a private key of the invoking user. So there should be no privilege escalation.

I’m comfortable with the technical decision to leave the unfixed `square`…2.6.0 around, and invoked; I just want to highlight this, in case that were insufficient.

mtrmac:
IIRC a small (not full) bump of ocicrypt could do the trick. (A full bump would go all the way to v3, which is not used on this branch yet.)

TomSweeneyRedHat:
ocicrypt is dragging in the square version, which disappeared in ocicrypt v1.1.8. However, that's when ocicrypt started pulling in go-jose:v3, and there are a lot of files that are added. Decisions, decisions.

mtrmac:
So there isn’t an intermediate version of ocicrypt, I’m sorry.

- Even an update to go-jose/v3 _removes_ the square/v2 one. It’s not a net addition; even git mostly shows that as renamed files.
- The difference between square/v2.6.0 and go-jose/v3.0.0 is very small (mostly the package name change, in errors), and we would migrating along with `ocicrypt`, the _only_ user, and removing an unmaintained dependency. I rather like that part.
- The update of ocicrypt to 1.1.8 is mostly lint updates and removal of `pkg/errors`, perhaps unwanted churn but mostly manageable.
- The difference between JOSE v3.0.0 and v3.0.3 does update a lot of the golang.org/x dependencies, that’s more uncomfortable.

I am fine with either decision.

TomSweeneyRedHat:
@mtrmac I just went through the code and concur with your analysis, the only spot we have that's a problem is the one you spotted in the keywrapper_Jwe.go file.  But not being a person that wins bets often, I think we better move forward with #2314 just to be ultra safe.

PR #2297, opened 2024-04-15
Update to Go 1.21

mtrmac:
…  primarily because some dependencies started to require it: if we ever needed to quickly update a dependency for a vulnerability fix, we might have to update to Go 1.21 on a short notice, or fork the dependency.

Cc: @jnovy @lsm5 @cevich @TomSweeneyRedHat


go.mod
```diff
@@ -1,7 +1,7 @@
 module github.com/containers/skopeo
 
 // Minimum required golang version
-go 1.20 // *****  ATTENTION  WARNING  CAUTION  DANGER  ******
+go 1.21 // *****  ATTENTION  WARNING  CAUTION  DANGER  ******
```
cevich, line 4: My understanding is that along with this, we also need `toolchain 1.21` as well?  Please double-check, and also please remove the big warning comment since it no longer applies.

cevich:
After this and https://github.com/containers/image/pull/2377 merge, do you have plans for buildah and podman?  Reason is, we're currently blocking all updates that require anything past `1.20` with https://github.com/containers/automation/pull/187/files#diff-4d5df5c9d8fa729092ddd28691af9c033a34c9fba7434c06501ae5f875e7b1a6R115-R132

Meaning, that will either need to be removed, or relocated to the buildah and podman Renovate config overrides.

mtrmac:
(Marking as draft to make sure we have a decision the Renovate configuration first; I understand Renovate would immediately start failing otherwise.)

go.mod
cevich: As I understand the situation (could easily be wrong), it's a complex interaction between Renovate proposed package update PRs and our CI systems.

Without the `toolchain` directive, and without any go version restrictions to Renovate, on Non-Fedora platforms (Debian, Mac, Windows), go will "auto-update" to the in-use  version of go for the package update in question (as proposed by Renovate).  Meaning for Renovate PRs, builds in CI might not be uniform across all platforms.  In other words, Windows might use golang 5.4.3 for updated package FOO 0.2 while Fedora uses golang 1.21 for FOO 0.2   I'm assuming this is a bad/unexpected thing.

OTOH, setting a `toolchain` forces the maximum go version, and both go and Renovate will honor that for non-Fedora platforms.  Meaning builds across all platforms will be exactly the same (for all shared packages).

Note: The only thing making Fedora special in these matters is they set some magic envar by default which disables the auto-update behavior.

@Luap99 please double-check my understanding, when you have a chance.

cevich:
>  understand Renovate would immediately start failing otherwise.)

I don't believe it will fail, it will simply (and mostly silently) refuse to open any module update PRs which also require a bump of `go 1.12`.

mtrmac:
IIRC from the F38 situation, the Go 1.20 toolchain refuses to work on `go 1.21` repos. I could well be misremembering.

Either way, “Renovate silently stops filing PRs” is the worst kind of failure mode for me.

go.mod
mtrmac: The way I read the paragraph linked above, `toolchain` is the _minimum_ Go version; it triggers unwanted upgrades (but not downgrades).

---

… and IIRC the effect _with Renovate included_ is that at the times where Renovate adds a `toolchain` directive, it adds one that matches Renovate’s default toolchain.

… and _that_ behavior we don’t want, you’re right.

I’ll add the `toolchain` directives tomorrow. (The renovate hard-coded version will, I think, orthogonally, need bumping to 1.21.)

Luap99:
> IIRC from the F38 situation, the Go 1.20 toolchain refuses to work on go 1.21 repos. I could well be misremembering.

No it will just fail to compile if the repo actually uses newer features, only starting with 1.21 go will enforce the minimum go version stated in go.mod. So we should never set a newer go version than our supported distros use.
And f38 was updated to go 1.21 so we should be good even on repos where we still test on f38.

go.mod
Luap99: yes I think we should have the toolchain there so the normal go commands can be used without adding the toolchain later, especially so renovate can be used properly.

I do however think that we need to make sure the toolchain is only ever set to the same value as the go line so we do not force update users unnecessary.

Luap99: and yes toolchain is the minimum version, not maximum.

mtrmac: I have added `toolchain 1.21.0`, and updated the comment a bit.

I’m _very_ open to suggestions about the comment.

mtrmac:
> Ephemeral COPR build failed. @containers/packit-build please check.

```
vendor/github.com/containers/libtrust/ec_key_openssl.go:23:15: undefined: ecdsa.HashSign
```
The build log does contain mentions of /usr/lib/golang/src/vendor/github.com/golang-fips/openssl/v2 , so the FIPS patch does seem to be present.

Best I can trace this, https://github.com/golang-fips/go/commit/56ac3dbbca083eaa6ce0b75893115a797f6682c9 has removed the `ecdsa.HashSign` function from the FIPS-patched Go standard library.

@derekparker, is the above correct, and if so, what is the recommended migration path? (I seem to vaguely remember that the upstream `ecsa.Sign` did not work in FIPS mode. Is that no longer the case, and should we now always call it?) And is there a single code implementation that now works in all Go versions, or is that Go version-dependent, so that we need to continue using `HashSign` for older Go versions?

derekparker:
Hey,

Yes it is correct, that particular symbol is removed in the `openssl@v2` version. The upgrade path is to use this equivalent symbol instead: https://github.com/golang-fips/openssl/blob/9dbc52b7d686ce75e707d18bd997964a3570f554/ecdsa.go#L112. So the new symbol name with be `HashSignECDSA`.

go.mod
cevich: I think the comment just needs to say:

``Warning: Ensure the `go` and `toolchain` versions match exactly to prevent unwanted auto-updates`

cevich:
@mtrmac @Luap99 Just to be sure I'm keeping-up:  Before this (and the c/image) PR merges, I need to open PRs for Buildah and Podman that disable Renovate updates past go 1.20.  After those merge, I need another PR to disable the global Renovate go 1.20 limitation.  Then this (and the c/image) PRs are safe to merge.

Somehow that sounds overcomplicated :disappointed:

Luap99:
> @mtrmac @Luap99 Just to be sure I'm keeping-up: Before this (and the c/image) PR merges, I need to open PRs for Buildah and Podman that disable Renovate updates past go 1.20. After those merge, I need another PR to disable the global Renovate go 1.20 limitation. Then this (and the c/image) PRs are safe to merge.
> 
> Somehow that sounds overcomplicated 😞

I see no reason why this couldn't be merged before updating the renovate config. Sure vendor PRs that need a newer golang will be broken but so are they today so there is really no difference regardless of merge order. Ideally these changes are merged sooner than later given the vast amount of deps that need go 1.21 so we can unlock many updates.

mtrmac:
> @mtrmac @Luap99 Just to be sure I'm keeping-up: Before this (and the c/image) PR merges, I need to open PRs for Buildah and Podman that disable Renovate updates past go 1.20. After those merge, I need another PR to disable the global Renovate go 1.20 limitation.

I think that once this merges, Buildah and Podman need to be structurally ready to accept 1.21 (because they will get it on any c/image bump).

So I was thinking just bump Renovate globally to 1.21 and that’s all. Probably pay close attention to the `toolchain` lines if they were added to Renovate.

Alternatively, we could line up PRs to bump to 1.21 with a `toolchain` we like for all projects, and merge them all immediately followed by a Renovate config update.

go.mod
mtrmac: Done.

Luap99:
> So I was thinking just bump Renovate globally to 1.21 and that’s all. Probably pay close attention to the toolchain lines if they were added to by Renovate.

Yeah that should happen, and even if someone somewhere merges a PR with a to new toolchain directive we can always manually revert it back to an older version so no big problem.

cevich:
Great, this sounds like a problem that can be solved easily by a short e-mail :grin:

cevich:
Ref: https://github.com/containers/automation/pull/189

cevich:
PR merged and e-mail sent.

mtrmac:
> > Ephemeral COPR build failed. @containers/packit-build please check.
> 
> ```
> vendor/github.com/containers/libtrust/ec_key_openssl.go:23:15: undefined: ecdsa.HashSign
> ```

Dealing with this continues in https://github.com/containers/skopeo/pull/2305 .

Luap99 (approved):
LGTM

TomSweeneyRedHat:
LGTM
once the tests are happy

mtrmac:
(I think those test failures are orthogonal, but it does make a lot of sense to fix those failures before introducing more changes.)

PR #2308, opened 2024-04-24
CI VMs: bump to new versions with tmpfs /tmp

edsantiago:
For the last long time, Fedora CI VMs have had a disk /tmp. Real-world setups typically have tmpfs /tmp. This switches to CI VMs that reflect the real world.

See https://github.com/containers/automation_images/pull/340


rhatdan:
LGTM

PR #2312, opened 2024-05-02
Packit c10s

lsm5:
This commit will enable downstream syncing to CentOS Stream 10. The
centos maintainer will need to manually run `packit propose-downstream`
and `centpkg build` until better centos integration is in place.
    
This commit also builds both rhel9 and centos9 copr rpms so we can check
for things like differences in golang compiler.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -24,8 +24,6 @@ jobs:
       - fedora-all-aarch64
       - fedora-eln-x86_64
       - fedora-eln-aarch64
-      - centos-stream+epel-next-8-x86_64
-      - centos-stream+epel-next-8-aarch64
       - centos-stream+epel-next-9-x86_64
       - centos-stream+epel-next-9-aarch64
     additional_repos:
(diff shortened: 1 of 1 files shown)


lsm5:
@mtrmac PTAL. the centos-stream builds are failing on `ecdsa.HashSign` while the epel builds (use RHEL + EPEL) passed because of an older golang compiler. Having both epel and centos-stream would be useful for such cases.

The change in the PR itself should be good to go.

mtrmac (approved):
Thanks!

mtrmac:
The EPEL-9 builds are reported as successful in COPR, but still pending in Packit.

I’ll try rebasing this PR to re-trigger builds…

lsm5:
> The EPEL-9 builds are reported as successful in COPR, but still pending in Packit.
> 
> I’ll try rebasing this PR to re-trigger builds…

oh actually that may have been packit getting confused because of the centos-stream and epel jobs having same parameters with different targets. Let me push an additional change to include a distinct `packages` key.

mtrmac:
@lsm5 Thanks. All but centos-stream is now passing. If this looks good, please merge.

lsm5:
> @lsm5 Thanks. All but centos-stream is now passing. If this looks good, please merge.

Thanks a lot @mtrmac

PR #2315, opened 2024-05-03
fix summaries for `standalone-sign` and `standalone-verify`

ktdreyer:
Correct the summary for `standalone-sign`. The old summary appears to be a bad copy-and-paste from `skopeo copy`.

Clarify that both of these tools are debugging tools, only operating on local files.


mtrmac (approved):
Thanks!

PR #2317, opened 2024-05-04
Add info on Skopeo image to README.md

TomSweeneyRedHat:
In the Buildah repo, we had a request for a ptr to the container image location in the docs.  We've added similar to Podman and Buildah.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,8 @@ Skopeo works with API V2 container image registries such as [docker.io](https://
 For a detailed description how to install or build skopeo, see
 [install.md](./install.md).
 
+Skopeo is also available as a Container Image on [quay.io](https://quay.io/skopeo/stable).  For more information, see the [Skopeo Image](https://github.com/containers/image_build/blob/main/skopeo/README.md) page.
+
 ## Inspecting a repository
 `skopeo` is able to _inspect_ a repository on a container registry and fetch images layers.
 The _inspect_ command fetches the repository's manifest and it is able to show you a `docker inspect`-like


TomSweeneyRedHat:
@cevich PTAL

README.md
```diff
@@ -47,6 +47,8 @@ Skopeo works with API V2 container image registries such as [docker.io](https://
 For a detailed description how to install or build skopeo, see
 [install.md](./install.md).
 
+Skopeo is also available as a Container Image on [quay.io](https://quay.io/skopeo/stable).  For more information, see the [Skopeo Image](https://github.com/containers/image_build/blob/main/skopeo/README.md) page.
+
```
mtrmac, line 51: The `install.md` linked just above already contains a similar pointer.

I think we’d typically recommend using a non-containerized distribution package rather than a container (e.g. because maintaining login state, and dealing with credential helpers, in containerized software is harder), so having the top-level readme point at _only_ the container image might not be optimal.

cevich: I think Tom is acting based on the 20-40,000 hits per day on the container image (according to quay).  Surely there must be another 50k people that would use the image if only they knew about it from the readme :rofl: 

> using a non-containerized distribution package

This is a valid point, though I'm not sure it's in-scope for this PR.  Reason being, I think it might be a good thing to mention both here and in the "Sample Usage" portion of the container image docs.  Would you be okay if we took these on as followup items?

cevich (approved):

mtrmac:
@cgwalters Several times recently we’ve seen `ostree-rs-ext` tests failing with
```
    1: skopeo failed: time="2024-05-04T00:19:54Z" level=fatal msg="initializing source oci:/proc/self/fd/3:: open /proc/self/fd/3/index.json: no such file or directory"
````
e.g. https://cirrus-ci.com/task/5588090834649088 . Could there be some race in that code?

README.md
mtrmac: I’m certainly not going to argue the current README structure is optimal… More prominently advertising the non-preferred alternative, _and_ adding redundancy to the documentation, does not quite satisfy my perfectionism.

But then, enough other things are broken that I’ll accept being outvoted and hit the green button.

cevich: Opened https://github.com/containers/image_build/pull/29

PR #2322, opened 2024-05-08
hack: Support picking cc and cpp via environment variables.

### LINKED ISSUE #2278: The checks in hack/ directory hardcode `cc'
The scripts in the hack/ directory assume that `cc' binary is present in the build environment.  That might not always be the case (the C compiler might be present as `gcc' for example).  Podman project does https://github.com/containers/podman/blob/main/hack/btrfs_tag.sh#L2 , would be nice if skopeo started to do the same.

graywolf:
Until now `cc' was hard-coded as the only compiler used.  Supporting selecting the compiler and preprocessor to be used via environment variables makes life easier for distributors, so this commit mimics how podman does it in its hack/* scripts.

Fixes #2278.

### DIFF
--- a/hack/btrfs_installed_tag.sh
+++ b/hack/btrfs_installed_tag.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-cc -E - > /dev/null 2> /dev/null << EOF
+${CPP:-${CC:-cc} -E} ${CPPFLAGS} - > /dev/null 2> /dev/null << EOF
 #include <btrfs/ioctl.h>
 EOF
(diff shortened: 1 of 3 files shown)


lsm5 (commented):
Thanks for the PR @graywolf 

> Ephemeral COPR build failed. @containers/packit-build please check.

ignore unrelated `rpm-build:centos-stream` failures.

LGTM looking at the same files in podman. FWIW, `hack/libsubid_tag.sh` in podman doesn't have the change yet.

@mtrmac PTAL.

mtrmac (approved):
Thanks!

PR #2327, opened 2024-05-13
[skip-ci] Packit: `packages: [skopeo-fedora]` for podman-next jobs

lsm5:
Without this change, packit will trigger duplicate jobs on podman-next for `skopeo-fedora`, `skopeo-centos` etc, which essentially point to the same spec file.

1 triggered job builds packages for all environments enabled on the COPR, so simply `skopeo-fedora` should suffice.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -63,6 +63,7 @@ jobs:
   # Run on commit to main branch
   - job: copr_build
     trigger: commit
+    packages: [skopeo-fedora]
     notifications:
       failure_comment:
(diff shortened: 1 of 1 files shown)


mtrmac (approved):

mtrmac:
Makes sense to me with the little I know, please merge when appropriate.

lsm5:
centos-stream failures are dependent on a newer golang build. Ignoring for now and merging. Thanks @mtrmac

Issue #1992, opened 2023-05-04
help: skopeo sync How to keep mirroring to project names


jinjianming:
For example mine : sync.yml 

```
 registry.js.design:
      images:
          justwatch/elasticsearch_exporter:
              - "1.1.0"
          elasticsearch/elasticsearch:
              - "7.8.1"
```
My Sync Command
```
skopeo sync --tls-verify=false --src yaml --dest docker sync.yml js-registry:5000/
```

My expected effect
```
js-registry:5000/justwatch/elasticsearch_exporter:1.1.0
js-registry:5000/elasticsearch/elasticsearch:7.8.1
```
Actual effect
```
js-registry:5000/elasticsearch_exporter:1.1.0
js-registry:5000/elasticsearch:7.8.1
```

mtrmac:
There is `--scoped`, but that might also not do exactly what you need.

jinjianming:
> 有`--scoped`，但这也可能无法完全满足您的需要。

I tried it and brought over the domain name of the source warehouse. I only want to bring the following path

```
[root@js-staging-env-177 ~]# skopeo sync  --scoped --src docker --dest docker --src-tls-verify=false --dest-tls-verify=false registry.js.design/bitnami/redis-exporter:1.24.0-debian-10-r9 js-registry:5000/
INFO[0000] Tag presence check                            imagename="registry.js.design/bitnami/redis-exporter:1.24.0-debian-10-r9" tagged=true
INFO[0000] Copying image ref 1/1                         from="docker://registry.js.design/bitnami/redis-exporter:1.24.0-debian-10-r9" to="docker://js-registry:5000/registry.js.design/bitnami/redis-exporter:1.24.0-debian-10-r9"
```
reality

```
js-registry:5000/registry.js.design/bitnami/redis-exporter:1.24.0-debian-10-r9
```
expect
```
js-registry:5000/bitnami/redis-exporter:1.24.0-debian-10-r9
```

mtrmac:
Yes; I think the only way right now is to split up the YAML file and run `skopeo sync` several times with different destinations.

jinjianming:
> 是的; 我认为现在唯一的方法是拆分 YAML 文件并`skopeo sync`在不同的目的地运行多次。

I implemented it through the following script

```
function sync_docker(){
    REGISTRY="js-registry:5000"
    local yml_file=$2
# Get image names and tags from sync.yml and iterate through them
[...]

PR #2330, opened 2024-05-15
feat(sync): preserve subpath in repositories when registry is defined

TechIsCool:
This PR enhances the Skopeo `sync` command by ensuring that image paths are preserved. This adds functionality while also addressing the issue where paths were previously flattened with `path.Base(destSuffix)`. 
 
  - Add `RegistryName` field to `repoDescriptor` for better registry handling.
  - Introduce `getSubPath` function to cleanly handle subpath extraction.
  - Update functions `imagesToCopyFromRegistry` and `filterSourceReferences` to include `RegistryName`.
  - Modify `syncOptions.run` to use `getSubPath`.
  - Add unit tests for `getSubPath`.
  
  Resolves
  - https://github.com/containers/skopeo/issues/1992


mtrmac (changes requested):
Thanks for the PR.

We can’t just change what existing commands do, and break many existing scripts.

(And, as detailed throughout the GitHub issues, I’m uninterested in adding more mapping features. We should just let the user specify (source repo + destination repo) pairs; unfortunately that requires quite a noticeable redesign of the YAML format.)

PR #2337, opened 2024-05-23
[release-1.14] CVE-2024-3727 

TomSweeneyRedHat:
Add the fix to the release-14 branch for CVE-2024-3727.  This will fix the issue in OCP v4.16 and RHEL 8.10/9.4

Currently addresses: https://issues.redhat.com/browse/OCPBUGS-33267 until the RHEL 8.10  and 9.4 Jira cards are created.

PR #2341, opened 2024-05-29
Packit: update packit targets

lsm5:
We only care about buildability on epel 9 (RHEL). It won't be shipped to centos stream or to fedora.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -20,40 +20,5 @@ jobs:
         message: "Ephemeral COPR build failed. @containers/packit-build please check."
     enable_net: true
     targets:
-      - fedora-all-x86_64
-      - fedora-all-aarch64
-      - fedora-eln-x86_64
-      - fedora-eln-aarch64
-      - centos-stream+epel-next-8-x86_64
-      - centos-stream+epel-next-8-aarch64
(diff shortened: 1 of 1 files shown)


lsm5:
failing on
```
# github.com/containers/libtrust
vendor/github.com/containers/libtrust/ec_key_openssl.go:23:9: not enough return values
	have ([]byte, error)
	want (*big.Int, *big.Int, error)
```

lsm5:
I think these should just be epel targets. Updating..

mtrmac:
> ```
> vendor/github.com/containers/libtrust/ec_key_openssl.go:23:9: not enough return values
> ```
For the record, the `ecdsa.HashSign` function (which has missing for a time) was recently restored with a different prototype. That was later fixed in https://github.com/golang-fips/go/pull/199 .

mtrmac (approved):
Thanks!

mtrmac:
To double check, ideally this should be merged before #2337, is that correct?

lsm5:
> To double check, ideally this should be merged before #2337, is that correct?

@mtrmac That's right

lsm5:
I set these to epel because we don't care about shipping this to centos-stream and fedora AFAIK

PR #2356, opened 2024-06-17
Refer to registry.k8s.io instead of k8s.gcr.io

mtrmac:
... per https://kubernetes.io/blog/2023/02/06/k8s-gcr-io-freeze-announcement/ .

We are seeing intermittent failures (sufficient to reliably cause a test suite failure) pulling from k8s.gcr.io, let's see if using the newer one improves things.


cevich:
Something very similar was done in buildah, for ref.: https://github.com/containers/buildah/pull/4210

Luap99 (commented):
LGTM

PR #2362, opened 2024-06-20
[release-1.15] Refer to registry.k8s.io instead of k8s.gcr.io

mtrmac:
... per https://kubernetes.io/blog/2023/02/06/k8s-gcr-io-freeze-announcement/ .

We are seeing intermittent failures (sufficient to reliably cause a test suite failure) pulling from k8s.gcr.io, let's see if using the newer one improves things.

This is a backport of #2356.

### DIFF
--- a/integration/copy_test.go
+++ b/integration/copy_test.go
@@ -479,7 +479,7 @@ func (s *copySuite) TestCopySimple() {
 
 	// FIXME: It would be nice to use one of the local Docker registries instead of needing an Internet connection.
 	// "pull": docker: → dir:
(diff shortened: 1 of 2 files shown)


lsm5:
hmm, the packit config seems stale here.  @TomSweeneyRedHat is this branch shipped anywhere on RHEL?

TomSweeneyRedHat:
@lsm5 it's currently on RHEL 9.5 and 10.0 Beta, but will be replaced by v1.16.  So it's only on the test build systems at the moment, and Skopeo 1.15 won't be part of a RHEL release.  It is on Fedora and I'm assuming other distros though.

PR #2375, opened 2024-07-08
[skip-ci] RPM: spec file cleanup

lsm5:
All our supported distros except CentOS Stream 9 support autochangelog and we currently build rpms for CentOS Stream 9 only on the podman-next copr, where we don't really care much about changelog.

So, there's no need for this conditional to exist.

Podman Ref: https://github.com/containers/podman/pull/23110


mtrmac (approved):
Thanks!

Issue #658, opened 2019-05-20
Please provide a basic progress / transfer speed output ("CI compatible") on copy


frittentheke:
First of all, thank you massively for Skopeo and it's ability to efficiently copy images between registries. Especially when having to use proxies and other special setups Skopeo is gold!

Just like the @baracoder the issue https://github.com/containers/skopeo/issues/597 we are using Skopeo within Gitlab CI. 
While having Skopeo not print and update the otherwise nice progress indicators during a CI run is an improvement - not having any indication of the current transfer speed is an issue for our use case though. With copy being silent until it either succeeds or fails makes it hard to spot if there is a (speed) issue with the transfer.

In our case having some feedback on the transfer speed / likely success is crucial. May I kindly suggest to introduce a setting (or by checking for a tty to be present) to simply report the current speed / progress along with a timestamp in front about once a minute or every 30 seconds as a single line? This should not clutter ones output in CI and with the --quite option still available can still be switched off. This would also be nice for cron jobs to have some sort of "logging" of the progress maybe?

mtrmac:
Thanks for your report.

Reporting the recent transfer speed should be reasonably possible using `c/image/copy.Options.Progress`; reporting overall progress is not currently possible through that interface (it does not report about all the involved blobs, and their sizes, in advance).  Enhancing the progress reporting interfaces of `c/image/copy` to make this possible (or, ideally, to implement the current / WIP progress bar code in c/image/copy on top of that generic interface) would be nice, of course.

frittentheke:
Thanks for the quick response. The issue I am having is to see the current progress (transfer speed per second) rather than the total process. In short: I simply want to see how quickly things are progressing currently.

mtrmac:
This got notably more complex in the meantime, because up to 6 blobs are now copied simultaneously; so, the concept of “current progress” does not make sense any more, without some sort of aggregated view.

frittentheke:
@mtrmac  urgh, too bad :-( 
Thanks you keeping me / this issue updated.

As cool as it is when Skopeo does sync / copy multiple gigabytes of Docker layers in seconds on a fast local network, when you are dealing with potentially slow registries which are accessed via the internet across half the globe any indication if things are still moving or (almost) stalled would help.

rhatdan:
@vrothberg Can we use the progress bars that we use in Podman for this?

vrothberg:
> @vrothberg Can we use the progress bars that we use in Podman for this?

[...]

PR #2400, opened 2024-08-15
 proxy: Verify *either* toplevel or target

cgwalters:
proxy: Move policycontext into global state

I am not aware of a reason not to just cache this for the life
of the proxy, like we do other global state.

Prep for further changes.

---

proxy: Verify *either* toplevel or target

An issue was raised in that a current Red Hat internal build system
was performing a signature just on the per-architecture manifest,
but the current proxy code is expecting a signature on the manifest
list.

To quote Miloslav from that bug:

> Podman signs both the per-platform items and the top level,
> and enforces signatures only on per-platform items. cosign,
> by default, signs only the top level (and has an option to
> sign everything), I’m not sure which one it enforces.
> I don’t immediately recall other platforms.

We believe the current proxy code is secure since
we always require signatures (if configured) on the manifest
list, and the manifest list covers the individual manifests
via sha256 digest.

However, we want to support signatures only being present
on the per-arch manifest too in order to be flexible.

Yet, we can't hard switch to requiring signatures on the
per-arch manifests as that would immediately break anyone
relying on the current behavior of validating the toplevel.

Change the logic here to:

- Verify signature on the manifest list, and cache the error (if any)
- Fetch the manifest
- Verify signature on the manifest
- Allow if *either* signature was accepted; conversely, only error
  if signature validation failed on *both* the manifest list and
  manifest

This also switches things to cache the manifest upfront instead
of doing it lazily on `GetManifest/GetConfig`; in practice
callers were always immediately requesting those anyways.
The use case of just fetching blobs exists (e.g. to resume
an interrupted fetch), but is relatively obscure and
in general I think it's good to re-verify signatures on
each operation.

---

### DIFF
--- a/cmd/skopeo/proxy.go
+++ b/cmd/skopeo/proxy.go
@@ -153,9 +153,9 @@ type activePipe struct {
 // openImage is an opened image reference
 type openImage struct {
 	// id is an opaque integer handle
-	id        uint64
-	src       types.ImageSource
-	cachedimg types.Image
+	id  uint64
+	src types.ImageSource
+	img types.Image
 }
 
 // proxyHandler is the state associated with our socket.
@@ -225,6 +225,7 @@ func (h *proxyHandler) OpenImage(args []any) (replyBuf, error) {
 }
 
 func (h *proxyHandler) openImageImpl(args []any, allowNotFound bool) (retReplyBuf replyBuf, retErr error) {
+	ctx := context.Background()
 	h.lock.Lock()
 	defer h.lock.Unlock()
 	var ret replyBuf
@@ -254,12 +255,53 @@ func (h *proxyHandler) openImageImpl(args []any, allowNotFound bool) (retReplyBu
 	}
 
 	unparsedTopLevel := image.UnparsedInstance(imgsrc, nil)
-	allowed, err := h.policyctx.IsRunningImageAllowed(context.Background(), unparsedTopLevel)
+	// Check the signature on the toplevel (possibly multi-arch) manifest, but we don't
+	// yet propagate the error here.
+	allowed, toplevelVerificationErr := h.policyctx.IsRunningImageAllowed(context.Background(), unparsedTopLevel)
+	if toplevelVerificationErr == nil && !allowed {
+		return ret, fmt.Errorf("internal inconsistency: policy verification failed without returning an error")
+	}
+
+	mfest, manifestType, err := unparsedTopLevel.Manifest(ctx)
 	if err != nil {
 		return ret, err
 	}
-	if !allowed {
-		return ret, fmt.Errorf("internal inconsistency: policy verification failed without returning an error")
+	var target *image.UnparsedImage
+	if manifest.MIMETypeIsMultiImage(manifestType) {
+		manifestList, err := manifest.ListFromBlob(mfest, manifestType)
+		if err != nil {
+			return ret, err
+		}
+		instanceDigest, err := manifestList.ChooseInstance(h.sysctx)
+		if err != nil {
+			return ret, err
+		}
+		target = image.UnparsedInstance(imgsrc, &instanceDigest)
+
+		allowed, targetVerificationErr := h.policyctx.IsRunningImageAllowed(context.Background(), target)
+		if targetVerificationErr == nil && !allowed {
+			return ret, fmt.Errorf("internal inconsistency: policy verification failed without returning an error")
+		}
+
+		// Now, we only error if *both* the toplevel and target verification failed.
+		// If either succeeded, that's OK. We want to support a case where the manifest
+		// list is signed, but the target is not (because we previously supported that behavior),
+		// and we want to support the case where only the target is signed (consistent with what c/image enforces).
+		if targetVerificationErr != nil && toplevelVerificationErr != nil {
+			return ret, targetVerificationErr
+		}
+	} else {
+		target = unparsedTopLevel
+
+		// We're not using a manifest list, so require verification of the single arch manifest.
+		if toplevelVerificationErr != nil {
+			return ret, toplevelVerificationErr
+		}
+	}
+
+	img, err := image.FromUnparsedImage(ctx, h.sysctx, target)
+	if err != nil {
+		return ret, err
 	}
 
 	// Note that we never return zero as an imageid; this code doesn't yet
@@ -268,6 +310,7 @@ func (h *proxyHandler) openImageImpl(args []any, allowNotFound bool) (retReplyBu
 	openimg := &openImage{
 		id:  h.imageSerial,
 		src: imgsrc,
+		img: img,
 	}
 	h.images[openimg.id] = openimg
 	ret.value = openimg.id
@@ -367,44 +410,6 @@ func (h *proxyHandler) returnBytes(retval any, buf []byte) (replyBuf, error) {
 	return ret, nil
 }
 
-// cacheTargetManifest is invoked when GetManifest or GetConfig is invoked
-// the first time for a given image.  If the requested image is a manifest
-// list, this function resolves it to the image matching the calling process'
-// operating system and architecture.
-//
-// TODO: Add GetRawManifest or so that exposes manifest lists
-func (h *proxyHandler) cacheTargetManifest(img *openImage) error {
-	ctx := context.Background()
-	if img.cachedimg != nil {
-		return nil
-	}
-	unparsedToplevel := image.UnparsedInstance(img.src, nil)
-	mfest, manifestType, err := unparsedToplevel.Manifest(ctx)
-	if err != nil {
-		return err
-	}
-	var target *image.UnparsedImage
-	if manifest.MIMETypeIsMultiImage(manifestType) {
-		manifestList, err := manifest.ListFromBlob(mfest, manifestType)
-		if err != nil {
-			return err
-		}
-		instanceDigest, err := manifestList.ChooseInstance(h.sysctx)
-		if err != nil {
-			return err
-		}
-		target = image.UnparsedInstance(img.src, &instanceDigest)
-	} else {
-		target = unparsedToplevel
-	}
-	cachedimg, err := image.FromUnparsedImage(ctx, h.sysctx, target)
-	if err != nil {
-		return err
-	}
-	img.cachedimg = cachedimg
-	return nil
-}
-
 // GetManifest returns a copy of the manifest, converted to OCI format, along with the original digest.
 // Manifest lists are resolved to the current operating system and architecture.
 func (h *proxyHandler) GetManifest(args []any) (replyBuf, error) {
@@ -424,14 +429,8 @@ func (h *proxyHandler) GetManifest(args []any) (replyBuf, error) {
 		return ret, err
 	}
 
-	err = h.cacheTargetManifest(imgref)
-	if err != nil {
-		return ret, err
-	}
-	img := imgref.cachedimg
-
 	ctx := context.Background()
-	rawManifest, manifestType, err := img.Manifest(ctx)
+	rawManifest, manifestType, err := imgref.img.Manifest(ctx)
 	if err != nil {
 		return ret, err
 	}
@@ -460,7 +459,7 @@ func (h *proxyHandler) GetManifest(args []any) (replyBuf, error) {
 	// docker schema and MIME types.
 	if manifestType != imgspecv1.MediaTypeImageManifest {
 		manifestUpdates := types.ManifestUpdateOptions{ManifestMIMEType: imgspecv1.MediaTypeImageManifest}
-		ociImage, err := img.UpdatedImage(ctx, manifestUpdates)
+		ociImage, err := imgref.img.UpdatedImage(ctx, manifestUpdates)
 		if err != nil {
 			return ret, err
 		}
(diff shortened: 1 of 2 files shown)


cmd/skopeo/proxy.go
```diff
@@ -245,31 +254,64 @@ func (h *proxyHandler) openImageImpl(args []any, allowNotFound bool) (retReplyBu
...
+		// and we want to support the case where only the target is signed (as some signing
+		// systems are known to do this).
+		if targetVerificationErr != nil && toplevelVerificationErr != nil {
+			return ret, toplevelVerificationErr
+		}
+	} else {
+		target = unparsedTopLevel
+
+		// We're not using a manifest list, so require verification of the single arch manifest.
+		if toplevelVerificationErr != nil {
```
cgwalters, line 298: @mtrmac pointed out that this was missing - I've updated it, good catch!

cmd/skopeo/proxy.go
```diff
@@ -245,31 +254,64 @@ func (h *proxyHandler) openImageImpl(args []any, allowNotFound bool) (retReplyBu
...
+		allowed, targetVerificationErr := h.policyctx.IsRunningImageAllowed(context.Background(), target)
+		if targetVerificationErr == nil && !allowed {
+			return ret, fmt.Errorf("internal inconsistency: policy verification failed without returning an error")
+		}
+
+		// Now, we only error if *both* the toplevel and target verification failed.
+		// If either succeeded, that's OK. We want to support a case where the manifest
+		// list is signed, but the target is not (because we previously supported that behavior),
+		// and we want to support the case where only the target is signed (as some signing
+		// systems are known to do this).
```
mtrmac, line 290: Something like
```suggestion
		// and we want to support the case where only the target is signed (consistent with what c/image enforces).
```
? Or maybe both that this is the behavior consistent with c/image, and that we’ve seen images with only these signatures in the wild.

cmd/skopeo/proxy.go
```diff
@@ -245,31 +254,64 @@ func (h *proxyHandler) openImageImpl(args []any, allowNotFound bool) (retReplyBu
...
 
 	// Note that we never return zero as an imageid; this code doesn't yet
 	// handle overflow though.
 	h.imageSerial++
 	openimg := &openImage{
-		id:  h.imageSerial,
-		src: imgsrc,
+		id:        h.imageSerial,
+		src:       imgsrc,
+		cachedimg: cachedimg,
```
mtrmac, line 314: This value is now always present, so no longer a cache — maybe `parsedImg` or just `image`?

cmd/skopeo/proxy.go
```diff
@@ -426,10 +430,6 @@ func (h *proxyHandler) GetManifest(args []any) (replyBuf, error) {
 		return ret, err
 	}
 
-	err = h.cacheTargetManifest(imgref)
-	if err != nil {
-		return ret, err
-	}
 	img := imgref.cachedimg
```
mtrmac, line 433: Applies to all users: This variable often has a single user, or at most two — I think inlining `imgref.[cachedImg]` would be a tiny bit easier to follow.

mtrmac (commented):
Thanks!

Implementation LGTM, just some naming nits.

cmd/skopeo/proxy.go
```diff
@@ -204,6 +206,12 @@ func (h *proxyHandler) Initialize(args []any) (replyBuf, error) {
 	h.sysctx = sysctx
 	h.cache = blobinfocache.DefaultCache(sysctx)
 
+	policyContext, err := h.opts.global.getPolicyContext()
```
mtrmac, line 209: For the record, caching it for the whole was already agreed upon as a possible improvement in https://github.com/containers/skopeo/pull/2029#discussion_r1252325538 .

The downside is that for a very long-running proxy, policy file changes would not affect the proxy.

cgwalters: The proxy is not intended to run as a long-running separate daemon but be lifecycle bound to an invoking process (which typically should not also be long lived; rpm-ostree runs as a daemon but auto-exits on idle), so I think we're OK here.

mtrmac (approved):
LGTM. Please mark “ready for review” when appropriate.

cmd/skopeo/proxy.go
```diff
@@ -245,31 +254,63 @@ func (h *proxyHandler) openImageImpl(args []any, allowNotFound bool) (retReplyBu
...
+		if targetVerificationErr == nil && !allowed {
+			return ret, fmt.Errorf("internal inconsistency: policy verification failed without returning an error")
+		}
+
+		// Now, we only error if *both* the toplevel and target verification failed.
+		// If either succeeded, that's OK. We want to support a case where the manifest
+		// list is signed, but the target is not (because we previously supported that behavior),
+		// and we want to support the case where only the target is signed (consistent with what c/image enforces).
+		if targetVerificationErr != nil && toplevelVerificationErr != nil {
+			return ret, toplevelVerificationErr
```
mtrmac, line 291: … I think preferring `targetVerificationError` would be better; or report both failures.

cgwalters:
OK I went to add tests and got burned by the fact that actually `quay.io/coreos/11bot` seemed to become a single-arch manifest list at some point...

But, using my own test image, I did at least verify that 
```
git diff
diff --git a/cmd/skopeo/proxy.go b/cmd/skopeo/proxy.go
index a4a90898..c3fdabed 100644
--- a/cmd/skopeo/proxy.go
+++ b/cmd/skopeo/proxy.go
@@ -294,9 +294,9 @@ func (h *proxyHandler) openImageImpl(args []any, allowNotFound bool) (retReplyBu
                target = unparsedTopLevel
 
                // We're not using a manifest list, so require verification of the single arch manifest.
-               if toplevelVerificationErr != nil {
-                       return ret, toplevelVerificationErr
-               }
+               // if toplevelVerificationErr != nil {
+               //      return ret, toplevelVerificationErr
+               // }
        }
 
        img, err := image.FromUnparsedImage(ctx, h.sysctx, target)
```

(i.e. my original bug) will correctly cause the tests to fail.

It should be trivial to get someone with creds to upload a non-manifest-list fixture to quay.io/libpod - do you know who to ping? Chris?

cgwalters:
It'd also be useful to have sigstore-signed fixtures there of course.

mtrmac:
@edsantiago probably has a good idea of what images exist / are appropriate to use from `quay.io/libpod`.

If we have no requirements for how the subject images {are, are not} signed: in the worst case, given a multi-platform image, we can read the index, choose an instance, and use a digest reference as a single-platform image.

integration/proxy_test.go
```diff
@@ -355,3 +369,30 @@ func (s *proxySuite) TestProxy() {
...
+
+// Verify that a policy that denies all images is correctly rejected
+// for both manifest listed and direct per-arch images.
+func (s *proxySuite) TestProxyPolicyRejectAll() {
+	t := s.T()
+	tempd := t.TempDir()
+	config := proxyConfig{
+		policyFilePath: tempd + "/policy.json",
+	}
+	err := os.WriteFile(config.policyFilePath, []byte("{ \"default\": [ { \"type\":\"reject\"} ] }"), 0o644)
```
mtrmac, line 381: (This might be a tiny bit easier to read with raw string literals per https://go.dev/ref/spec#String_literals.)

integration/proxy_test.go
```diff
@@ -355,3 +369,30 @@ func (s *proxySuite) TestProxy() {
...
+
+	err = runTestGetManifestAndConfig(p, knownListImage)
+	assert.ErrorContains(t, err, "is rejected by policy")
+
+	// This one should continue to be fine.
+	err = runTestOpenImageOptionalNotFound(p, knownNotExtantImage)
+	if err != nil {
+		err = fmt.Errorf("Testing optional image %s: %v", knownNotExtantImage, err)
+	}
+	assert.NoError(t, err)
```
mtrmac, line 397: `assert.NoError(t, err, "testing image %s", knownNotExtantImage)`

cmd/skopeo/proxy.go
```diff
@@ -701,9 +682,13 @@ func (h *proxyHandler) close() {
 		err := image.src.Close()
 		if err != nil {
 			// This shouldn't be fatal
-			logrus.Warnf("Failed to close image %s: %v", transports.ImageName(image.cachedimg.Reference()), err)
+			logrus.Warnf("Failed to close image %s: %v", transports.ImageName(image.img.Reference()), err)
```
mtrmac, line 685: (Pre-existing, non-blocking: using `image.src.Reference()` here would more obviously correct — but it ends up being the same value in the end.)

mtrmac (commented):
(These are all non-blocking aesthetics.)

edsantiago:
>what images exist / are appropriate to use from quay.io/libpod.

Short answer: anything you want, it's under our control with multiple people having access.

Longer answer: CI VM images now include a local cache registry running on port 60333, with copies of all images required by buildah and podman. I'd like to extend that to skopeo too. Purpose is to reduce our exposure to quay flakes. Can signed images be served by a simple local registry process? If not, is there a way to make it possible?

mtrmac:
sigstore signatures are ordinary objects on a registry, and can be served that way.

simple signing signatures are served using plain HTTP. It should not be very hard (set up a directory as a staging destination in `registries.d`, mirror the images, then publish contents of that directory as HTTP), but a bit of extra work.

mtrmac:
… alternatively, we _could_ also design a way to store simple-signing signatures on the registry (perhaps not using exactly the sigstore tag convention, but using OCI referrers as intended). We’ve been converging on sigstore as the future, so it’s unclear that it is necessary, and in some sense, having this convenient feature sigstore-only might motivate users to migrate, so maybe we should not be adding that.

cgwalters:
Tangential to this PR landing but...per your comment on the bug:

> (It would be easy enough to accept top-level-only signatures in c/image, but then users would lose the ability to create sparse mirrors with a subset of platforms; and it seems that there is a class of users which does care about that feature. Requiring pre-platform signatures, and preserving this feature for users, seems valuable.)

The next problem is, if we start trying to move bootc itself over to more like what podman/skopeo does (which includes hopefully [unified storage](https://github.com/containers/bootc/issues/20)) it will also necessarily imply that we'd need to ensure we replicate this "accept top-level-only" on at least the pull code path we use for bootc. I don't know what that would look like exactly...maybe some sort of new `accept-toplevel-signatures` flag in `containers-registries.conf`? Or a CLI flag, or...maybe we actually do just continue to use the proxy but expose a much higher level `Copy` API that can write to c/storage too.

(When we do that type of stuff though we'll definitely want a basic progress bar... https://github.com/containers/skopeo/issues/658 is related in that I think a lot of these "script skopeo on the CLI" cases would really more nicely generalize to "language-independent c/image RPC API")

mtrmac:
Ugh, I’d prefer not to have that option, recognizing that it really might be unavoidable. That makes me more motivated to argue that this PR should just do what c/image does in this PR, and we would see if anything breaks.

---

The implementation would probably be a new field in `c/image/copy.Options`; there is no obvious config file wired to that. (It would probably be `containers.conf`, except that currently Skopeo does not use it at all.)

cgwalters:
> If we have no requirements for how the subject images {are, are not} signed: in the worst case, given a multi-platform image, we can read the index, choose an instance, and use a digest reference as a single-platform image.

I looked at this, but it's a bit logistically annoying to do in the tests at the moment.

> Short answer: anything you want, it's under our control with multiple people having access.

Ideally there'd be a git repository defining the images in quay.io/libpod - in this case a periodic sync like e.g. `skopeo copy --multi-arch system docker://quay.io/centos/centos:stream9 docker://quay.io/libpod/centos-non-manifest-list` or so. But, a one-time invocation of that is also fine.

cgwalters:
Hmm. Maybe what would be helpful here is for the proxy to expose back to the client the signature state - especially the case when we detect only the manifest list was signed. I could then add something to bootc to loudly warn if this is the case...the problem is that there are going to be users that aren't paying attention to the journal/process-stdout for the often unattended systems that are a primary target for bootc. But it may help.

mtrmac:
> Maybe what would be helpful here is for the proxy to expose back to the client the signature state 

Or have the proxy warn directly, and direct proxy’s stderr to a log?

cgwalters:
> Short answer: anything you want, it's under our control with multiple people having access.

@edsantiago ok but can you link me to the procedure or code or process for adding images there?

edsantiago:
I've lost the thread here, so this is the question I am answering:
* how does a given image X get pushed to `quay.io/libpod` ?
If that is not the question you intended, please ignore this, and clarify your question.

Answer: someone with credentials (me, @baude, not sure who else) pushes it.

If it is me, I'm neurotic about generating reproducible, documented, clearly tagged multiarch images. I will be glad to help craft something then push it. Please let me know how/where to begin with the image building process.

cgwalters:
>If it is me, I'm neurotic about generating reproducible, documented, clearly tagged multiarch images. I will be glad to help craft something then push it. 

This sounds good to me but to achieve reproducible documented process I'd expect a a git repository to contain that, that's what I was looking for.

At this point I'm basically just repeating https://github.com/containers/skopeo/pull/2400#issuecomment-2293394273

edsantiago:
IIUC you are referencing this part of the comment:

>Ideally there'd be a git repository defining the images in quay.io/libpod - in this case a periodic sync like e.g. skopeo copy --multi-arch system docker://quay.io/centos/centos:stream9 docker://quay.io/libpod/centos-non-manifest-list or so. But, a one-time invocation of that is also fine.

That's not something I can ever imagine myself doing. I can envision myself doing a one-time `skopeo copy ... /libpod/centos-non-manifest-list:20240821`, preferably with a corresponding commit to the skopeo repo, in the section(s) that use this image, describing how/when/who that image was copied and what is so special about the source centos image.

You mentioned `periodic`. What would be the purpose of that?

cgwalters:
Coming back to this one, I believe the resolution for this was the build system was changed to sign the manifest list too.

So....dunno. There's definitely a lingering compatibility issue here that'd be fixed by this, but there's also no immediate strong driver.

mtrmac:
The code has moved to container-libs; https://github.com/containers/container-libs/pull/765 is a slightly updated implementation in that repo.

Thanks!

PR #2401, opened 2024-08-16
Replace egrep with grep -E

lsm5:
This should silence any `egrep is obsolescent; using grep -E` warnings.


lsm5:
@mtrmac @edsantiago @cevich PTAL

cevich:
Good catch / thanks for fixing this.

LGTM

cevich (approved):

mtrmac (approved):
Thanks!

PR #2402, opened 2024-08-19
Add support for `--digestfile` for skopeo `sync`

ankitforcode:
Added `--digestfile` for skopeo `sync` option
* This will create the digest file if not exists. 
* Will append the sha and image reference for each image that is copied separated by space. 
* This option would not be available in `dry-run` flag is set.

### DIFF
--- a/cmd/skopeo/sync.go
+++ b/cmd/skopeo/sync.go
@@ -20,6 +20,7 @@ import (
 	"github.com/containers/image/v5/directory"
 	"github.com/containers/image/v5/docker"
 	"github.com/containers/image/v5/docker/reference"
+	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/pkg/cli"
 	"github.com/containers/image/v5/pkg/cli/sigstore"
 	"github.com/containers/image/v5/signature/signer"
@@ -46,6 +47,7 @@ type syncOptions struct {
 	format                   commonFlag.OptionalString // Force conversion of the image to a specified format
 	source                   string                    // Source repository name
 	destination              string                    // Destination registry name
+	digestFile               string                    // Write digest to this file
 	scoped                   bool                      // When true, namespace copied images at destination using the source repository name
 	all                      bool                      // Copy all of the images if an image in the source is a list
 	dryRun                   bool                      // Don't actually copy anything, just output what it would have done
@@ -121,6 +123,7 @@ See skopeo-sync(1) for details.
 	flags.StringVarP(&opts.destination, "dest", "d", "", "DESTINATION transport type")
 	flags.BoolVar(&opts.scoped, "scoped", false, "Images at DESTINATION are prefix using the full source image path as scope")
 	flags.StringVar(&opts.appendSuffix, "append-suffix", "", "String to append to DESTINATION tags")
+	flags.StringVar(&opts.digestFile, "digestfile", "", "Write the digests and Image References of the resulting images to the specified file, separated by newlines")
 	flags.BoolVarP(&opts.all, "all", "a", false, "Copy all images if SOURCE-IMAGE is a list")
 	flags.BoolVar(&opts.dryRun, "dry-run", false, "Run without actually copying data")
 	flags.BoolVar(&opts.preserveDigests, "preserve-digests", false, "Preserve digests of images and lists")
@@ -723,10 +726,24 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) (retErr error) {
 		logrus.Warn("Running in dry-run mode")
 	}
 
+	var digestFile *os.File
+	if opts.digestFile != "" && !opts.dryRun {
+		digestFile, err = os.OpenFile(opts.digestFile, os.O_TRUNC|os.O_CREATE|os.O_WRONLY, 0644)
+		if err != nil {
+			return fmt.Errorf("Error creating digest file: %w", err)
+		}
+		defer func() {
+			if err := digestFile.Close(); err != nil {
+				retErr = noteCloseFailure(retErr, "closing digest file", err)
+			}
+		}()
+	}
+
 	for _, srcRepo := range srcRepoList {
 		options.SourceCtx = srcRepo.Context
 		for counter, ref := range srcRepo.ImageRefs {
 			var destSuffix string
+			var manifestBytes []byte
 			switch ref.Transport() {
 			case docker.Transport:
 				// docker -> dir or docker -> docker
@@ -758,7 +775,7 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) (retErr error) {
 			} else {
 				logrus.WithFields(fromToFields).Infof("Copying image ref %d/%d", counter+1, len(srcRepo.ImageRefs))
 				if err = retry.IfNecessary(ctx, func() error {
-					_, err = copy.Image(ctx, policyContext, destRef, ref, &options)
+					manifestBytes, err = copy.Image(ctx, policyContext, destRef, ref, &options)
 					return err
 				}, opts.retryOpts); err != nil {
 					if !opts.keepGoing {
@@ -770,7 +787,19 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) (retErr error) {
 					logrus.WithError(err).Errorf("Error copying ref %q", transports.ImageName(ref))
 					continue
 				}
+				// Ensure that we log the manifest digest to a file only if the copy operation was successful
+				if opts.digestFile != "" {
+					manifestDigest, err := manifest.Digest(manifestBytes)
+					if err != nil {
+						return err
+					}
+					outputStr := fmt.Sprintf("%s %s", manifestDigest.String(), transports.ImageName(destRef))
+					if _, err = digestFile.WriteString(outputStr + "\n"); err != nil {
+						return fmt.Errorf("Failed to write digest to file %q: %w", opts.digestFile, err)
+					}
+				}
 			}
+
 			imagesNumber++
 		}
 	}
--- a/docs/skopeo-sync.1.md
+++ b/docs/skopeo-sync.1.md
@@ -68,6 +68,16 @@ Print usage statement.
 
 **--append-suffix** _tag-suffix_ String to append to destination tags.
 
+**--digestfile** _path_
+
+After copying the images from source, write the digest of the resulting images along with Image Reference.
+
+```
+sha256:bf91f90823248017a4f920fb541727fa8368dc6cf377a7debbd271cf6a31c8a7 docker://myhost.com/alpine:edge
+sha256:31603596830fc7e56753139f9c2c6bd3759e48a850659506ebfb885d1cf3aef5 docker://myhost.com/postgres:14.3
+
+```
+
 **--preserve-digests**
 
 Preserve the digests during copying. Fail if the digest cannot be preserved.


mtrmac (changes requested):
Thanks!

Did you try running the PR? AFAICS this would keep overwriting the same file with digests, and in effect only record the last digest of them. I don’t see how that’s useful.

ankitforcode:
> Thanks!
> 
> Did you try running the PR? AFAICS this would keep overwriting the same file with digests, and in effect only record the last digest of them. I don’t see how that’s useful.

hey @mtrmac 
Thanks for reviewing, I have fixed the code in a way that each sha for the image that is copied would be appended to the file once copied. This option would also not be available in case the `dry-run` flag is set.

ankitforcode:
> Thanks!
> 
> Did you try running the PR? AFAICS this would keep overwriting the same file with digests, and in effect only record the last digest of them. I don’t see how that’s useful.

Done.

TomSweeneyRedHat:
LGTM

cmd/skopeo/sync.go
```diff
@@ -723,6 +726,15 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) (retErr error) {
 		logrus.Warn("Running in dry-run mode")
 	}
 
+	var digestFile *os.File
+	if opts.digestFile != "" && !opts.dryRun {
+		digestFile, err = os.OpenFile(opts.digestFile, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
```
mtrmac, line 731: Why `O_APPEND` here? It’s not the typical behavior when providing a destination path to a program.

It’s not _completely unheard of_ but it seems unusual to me.

cmd/skopeo/sync.go
```diff
@@ -723,6 +726,15 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) (retErr error) {
...
+		if err != nil {
+			return fmt.Errorf("Error creating digest file: %w", err)
+		}
+		defer digestFile.Close()
```
mtrmac, line 735: For writes, I think we want to also check for errors here (so that e.g. failures updating the directory or the inode with the final data are reported).

cmd/skopeo/sync.go
```diff
@@ -758,7 +770,17 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) (retErr error) {
...
 				logrus.WithFields(fromToFields).Infof("Copying image ref %d/%d", counter+1, len(srcRepo.ImageRefs))
 				if err = retry.IfNecessary(ctx, func() error {
-					_, err = copy.Image(ctx, policyContext, destRef, ref, &options)
+					manifestBytes, err := copy.Image(ctx, policyContext, destRef, ref, &options)
+					if opts.digestFile != "" {
+						manifestDigest, err := manifest.Digest(manifestBytes)
+						if err != nil {
+							return err
+						}
+						_, err = digestFile.WriteString(ref.DockerReference().String() + " " + manifestDigest.String() + "\n")
```
mtrmac: `ref.DockerReference()` can be `nil`, then this would crash.

I think `transports.ImageName` is the only fully general format we can use here.

mtrmac: Using `ref` (the source) here is inconsistent: the image’s manifest can change during the copy (via `ImageListSelection` or `ForceManifestMIMEType`, for example), so this would report the source reference + the destination manifest, a combination that might not exist anywhere.

Using the destination reference would make more sense to me.

cmd/skopeo/sync.go
```diff
@@ -771,6 +793,7 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) (retErr error) {
 					continue
 				}
 			}
+
```
mtrmac, line 796: (Non-blocking: I think keeping the counter increment immediately after the action being counted is a bit cleaner than adding the blank line here.)

docs/skopeo-sync.1.md
```diff
@@ -68,6 +68,10 @@ Print usage statement.
 
 **--append-suffix** _tag-suffix_ String to append to destination tags.
 
+**--digestfile** _path_
+
+After copying the image, write the digest of the resulting image to the file.
```
mtrmac, line 73: The description needs updating a bit, there is no single “the resulting image”.

---

This should at least briefly document the format of the output, whatever the format ends up being.

cmd/skopeo/sync.go
```diff
@@ -121,6 +123,7 @@ See skopeo-sync(1) for details.
 	flags.StringVarP(&opts.destination, "dest", "d", "", "DESTINATION transport type")
 	flags.BoolVar(&opts.scoped, "scoped", false, "Images at DESTINATION are prefix using the full source image path as scope")
 	flags.StringVar(&opts.appendSuffix, "append-suffix", "", "String to append to DESTINATION tags")
+	flags.StringVar(&opts.digestFile, "digestfile", "", "Write the digest of the pushed image to the specified file")
```
mtrmac, line 126: The description needs updating a bit, there is no single “the resulting image”.

mtrmac, line 126: [I was thinking about naming the option `digest-file`, this is just _ugly_.

But `skopeo copy`, as well as Buildah and Podman, all use the dash-less variant. And we have `--authfile`.

So I guess `digestfile` makes the most sense; still, highlighting this as a possible point of discussion.]

cmd/skopeo/sync.go
mtrmac: Still outstanding. Why was this marked as resolved?

docs/skopeo-sync.1.md
```diff
@@ -68,6 +68,15 @@ Print usage statement.
 
 **--append-suffix** _tag-suffix_ String to append to destination tags.
 
+**--digestfile** _path_
+
+After copying the images from source, write the digest of the resulting images along with Image Reference.
+
+```
+bf91f90823248017a4f920fb541727fa8368dc6cf377a7debbd271cf6a31c8a7 <dest-repository>/alpine:edge
+31603596830fc7e56753139f9c2c6bd3759e48a850659506ebfb885d1cf3aef5 <dest-repository>/postgres:14.3
```
mtrmac, line 77: `<dest-repository>` is not really accurate — `docker://…` for registry locations, and there is no “repository” for directory destinations.

cmd/skopeo/sync.go
```diff
@@ -723,6 +726,19 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) (retErr error) {
 		logrus.Warn("Running in dry-run mode")
 	}
 
+	var digestFile *os.File
+	if opts.digestFile != "" && !opts.dryRun {
+		digestFile, err = os.OpenFile(opts.digestFile, os.O_CREATE|os.O_WRONLY, 0644)
+		if err != nil {
+			if os.IsNotExist(err) {
+				return fmt.Errorf("File system is likely out of inodes or has other issues.: %w", err)
```
mtrmac, line 734: Why should “IsNotExist” map to “out of inodes???” That’s by far not the most common cause (almost certainly be a typo in a parent directory name).

Why does this need a special case for `IsNotExist` anyway?

cmd/skopeo/sync.go
```diff
@@ -758,7 +774,20 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) (retErr error) {
...
 				if err = retry.IfNecessary(ctx, func() error {
-					_, err = copy.Image(ctx, policyContext, destRef, ref, &options)
+					manifestBytes, err := copy.Image(ctx, policyContext, destRef, ref, &options)
+					if opts.digestFile != "" {
+						manifestDigest, err := manifest.Digest(manifestBytes)
+						if err != nil {
+							return err
+						}
+						outputStr := fmt.Sprintf("%s %s", manifestDigest.String(), transports.ImageName(destRef))
+						if counter < len(srcRepo.ImageRefs)-1 {
```
mtrmac, line 784: Why is this now conditional?

Traditionally in UNIX, “a line” is some characters _and a terminating newline_. Not having the terminating newline makes the file inconvenient to use, e.g. a trivial `cat` has the final line mixed with the next shell prompt (or, worse, hidden by the prompt).

cmd/skopeo/sync.go
```diff
@@ -723,6 +726,19 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) (retErr error) {
 		logrus.Warn("Running in dry-run mode")
 	}
 
+	var digestFile *os.File
+	if opts.digestFile != "" && !opts.dryRun {
+		digestFile, err = os.OpenFile(opts.digestFile, os.O_CREATE|os.O_WRONLY, 0644)
```
mtrmac, line 731: Without `O_APPEND`, this now needs `O_TRUNC`, otherwise the file will contain data from a previous run.

mtrmac (changes requested):

cmd/skopeo/sync.go
ankitforcode: not sure if understood the ask here @mtrmac

mtrmac: If `digestFile.Close` fails, that should be detected and treated as a failure.

ankitforcode:
> Thanks!
> 
> Did you try running the PR? AFAICS this would keep overwriting the same file with digests, and in effect only record the last digest of them. I don’t see how that’s useful.

Ran the changes on local and validated the output for manifest file.

cmd/skopeo/sync.go
ankitforcode: Done, now handled correctly.

cmd/skopeo/sync.go
```diff
@@ -723,6 +726,19 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) (retErr error) {
...
 	}
 
+	var digestFile *os.File
+	if opts.digestFile != "" && !opts.dryRun {
+		digestFile, err = os.OpenFile(opts.digestFile, os.O_TRUNC|os.O_CREATE|os.O_WRONLY, 0644)
+		if err != nil {
+			return fmt.Errorf("Error creating digest file: %w", err)
+		}
+		defer func() {
+			if errs := digestFile.Close(); errs != nil {
```
mtrmac, line 736: If there were some other earlier error, that error is more likely to be a root cause, so it should be preferred.

… I forgot that we have a helper for this already: `noteCloseFailure`. Compare how it is used elsewhere.

cmd/skopeo/sync.go
```diff
@@ -758,7 +774,18 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) (retErr error) {
...
+					if opts.digestFile != "" && err == nil {
+						manifestDigest, err := manifest.Digest(manifestBytes)
+						if err != nil {
+							return err
+						}
+						outputStr := fmt.Sprintf("%s %s", manifestDigest.String(), transports.ImageName(destRef))
+						if _, err = digestFile.WriteString(outputStr + "\n"); err != nil {
+							return fmt.Errorf("Failed to write digest to file %q: %w", opts.digestFile, err)
+						}
+					}
```
mtrmac, line 788: Actually, all of this would be better outside of the `retry.IfNecessary` block; e.g. if the write to `digestFile` failed, we should terminate immediately, and definitely not restart the image copy.

cmd/skopeo/sync.go
```diff
@@ -780,6 +808,10 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) (retErr error) {
 	} else {
 		logrus.Infof("Synced %d images from %d sources", imagesNumber, len(srcRepoList))
 	}
+	// If there were errors during the sync process, return an error
+	if err != nil {
```
mtrmac, line 812: What does this do? AFAICS `err` is always `nil` on this path (and if it isn’t, that should be detected locally where that happens).

Guessing, this might have been intended to work with the added `defer` handler of `Close` but that’s not how `defer` works; it is not yet called at this point.

cmd/skopeo/sync.go
```diff
@@ -723,6 +726,19 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) (retErr error) {
...
 
+	var digestFile *os.File
+	if opts.digestFile != "" && !opts.dryRun {
+		digestFile, err = os.OpenFile(opts.digestFile, os.O_TRUNC|os.O_CREATE|os.O_WRONLY, 0644)
+		if err != nil {
+			return fmt.Errorf("Error creating digest file: %w", err)
+		}
+		defer func() {
+			if errs := digestFile.Close(); errs != nil {
+				err = fmt.Errorf("Error closing digest file: %w", err)
```
mtrmac, line 737: Changing `err` in the error affects nothing AFAICS. We have `retErr` for this purpose.

ankitforcode: Done.

cmd/skopeo/sync.go
ankitforcode: Done.

cmd/skopeo/sync.go
ankitforcode: Fixed

cmd/skopeo/sync.go
```diff
@@ -770,7 +787,19 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) (retErr error) {
 					logrus.WithError(err).Errorf("Error copying ref %q", transports.ImageName(ref))
 					continue
 				}
+				// Ensure that we log the manifest digest to a file only if the copy operation was successful
+				if opts.digestFile != "" && err == nil {
```
mtrmac, line 791: `err != nil` can’t happen here AFAICS

mtrmac (commented):
Thanks!

One last thing, and please squash the commits into one.

mtrmac (approved):
Thanks!

PR #2406, opened 2024-08-21
[release-1.16] Bump Skopeo to v1.16.1, c/common to v0.60.2, c/image to v5.32.2

TomSweeneyRedHat:
Bumping the c/* projects in Skopeo to the same versions that are in Buildah and Podman, which are headed to RHEL 9.5/10.0 Beta.  Then bump Skopeo itself to v1.16.1.


mtrmac (approved):
Thanks!

PR #2432, opened 2024-10-08
Packit: constrain downstream koji job to fedora package

lsm5:
This constraint helps to avoid duplicate Packit jobs. The effect will only be seen at the time of Fedora koji build and not on upstream. This doesn't affect upstream.


mtrmac (approved):
Thanks!

PR #2439, opened 2024-10-11
proxy: Add various debug logging

cgwalters:
To aid debugging https://github.com/ostreedev/ostree-rs-ext/issues/657

### DIFF
--- a/cmd/skopeo/proxy.go
+++ b/cmd/skopeo/proxy.go
@@ -687,6 +687,7 @@ func (h *proxyHandler) FinishPipe(args []any) (replyBuf, error) {
 
 	// Wait for the goroutine to complete
 	f.wg.Wait()
+	logrus.Debug("Completed pipe goroutine")
 	// And only now do we close the write half; this forces the client to call this API
 	f.w.Close()
 	// Propagate any errors from the goroutine worker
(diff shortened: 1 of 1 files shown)


cmd/skopeo/proxy.go
```diff
@@ -600,6 +600,7 @@ func (h *proxyHandler) GetBlob(args []any) (replyBuf, error) {
 		tr := io.TeeReader(blobr, verifier)
 		n, err := io.Copy(f.w, tr)
 		if err != nil {
+			logrus.Debugf("")
```
mtrmac, line 603: Do you want to add some distinct text here?

mtrmac (commented):
Sure.

Feel free to merge as is this is sufficient.

cmd/skopeo/proxy.go
cgwalters: Forgot to amend the commit after vacillating on whether a debug log is needed there; I don't think it is, so deleted.

cmd/skopeo/proxy.go
```diff
@@ -782,6 +784,8 @@ func (h *proxyHandler) processRequest(readBytes []byte) (rb replyBuf, terminate
 		err = fmt.Errorf("invalid request: %v", err)
 		return
 	}
+	logrus.Debugf("Executing method %s", req.Method)
```
mtrmac, line 787: (For very broken or adversarial values, `%q` throughout would be better, but that’s not really relevant here.)

mtrmac (approved):
Thanks!

cmd/skopeo/proxy.go
cgwalters: Yeah...I guess I need to just internalize "use %q by default" in general.

mtrmac:
@cgwalters please try rebasing on top of `main`; that will include an updated `golangci-lint`, and there’s a _chance_ that might affect the `osx` test timing out. At least, we would be working off a shared baseline.

PR #2440, opened 2024-10-14
Fix format string inconsistency causing a build failure

mtrmac:
After a manually-pushed version of #2400 .

Cc: @cgwalters .

(There are some indications the tests are going to fail for an unrelated purpose later; if so, I’ll include a fix for that in this PR.)


cgwalters (approved):
Argh sorry about that!

PR #2443, opened 2024-10-15
Build Windows binary

mattn:


Makefile
```diff
@@ -21,16 +21,16 @@ FISHINSTALLDIR=${PREFIX}/share/fish/vendor_completions.d
...
+GOARCH ?= $(shell $(GO) env GOARCH)
 
 # N/B: This value is managed by Renovate, manual changes are
 # possible, as long as they don't disturb the formatting
 # (i.e. DO NOT ADD A 'v' prefix!)
 GOLANGCI_LINT_VERSION := 1.61.0
 
 ifeq ($(GOBIN),)
-GOBIN := $(GOPATH)/bin
+  GOBIN := $(subst \,/,$(shell $(GO) env GOPATH))/bin
```
mtrmac, line 33: This is surprising to me. `make` doesn’t like native-style paths? Which implementation is this?

Makefile
```diff
@@ -99,11 +99,16 @@ ifeq ($(DISABLE_CGO), 1)
 	override BUILDTAGS = exclude_graphdriver_btrfs containers_image_openpgp
 endif
 
+BIN=bin/skopeo
```
mtrmac, line 102: Naming this something a bit more explicit would help. `SKOPEO_BIN`? `SKOPEO_TARGET`?

---

There are quite a few other instances of `bin/skopeo` in the Makefile, why don’t they need updating?

mtrmac (commented):
@mattn IIRC this PR has never passed tests. Are you still working on it?

Makefile
mattn: On Windows, GOPATH include backslash not slash.

PR #2445, opened 2024-10-28
support copy multi arch instance

cleverhu:
support copy multi arch instance, this PR must work with [https://github.com/containers/image/pull/2612](https://github.com/containers/image/pull/2612)

### DIFF
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -53,7 +53,8 @@ func copyCmd(global *globalOptions) *cobra.Command {
 	srcFlags, srcOpts := imageFlags(global, sharedOpts, deprecatedTLSVerifyOpt, "src-", "screds")
 	destFlags, destOpts := imageDestFlags(global, sharedOpts, deprecatedTLSVerifyOpt, "dest-", "dcreds")
 	retryFlags, retryOpts := retryFlags()
-	opts := copyOptions{global: global,
+	opts := copyOptions{
+		global:              global,
 		deprecatedTLSVerify: deprecatedTLSVerifyOpt,
 		srcImage:            srcOpts,
 		destImage:           destOpts,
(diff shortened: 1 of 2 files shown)


cmd/skopeo/copy.go
```diff
@@ -102,22 +103,36 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
...
 	case "system":
-		return copy.CopySystemImage, nil
+		return copy.CopySystemImage, nil, nil
 	case "all":
-		return copy.CopyAllImages, nil
+		return copy.CopyAllImages, nil, nil
 	// There is no CopyNoImages value in copy.ImageListSelection, but because we
 	// don't provide an option to select a set of images to copy, we can use
 	// CopySpecificImages.
-	case "index-only":
```
mtrmac, line 114: We must not break existing options.

cmd/skopeo/copy.go
```diff
@@ -102,22 +103,36 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
...
+			platforms = append(platforms, manifest.Schema2PlatformSpec{OS: os, Architecture: arch})
+		}
+		return copy.CopySpecificImages, platforms, nil
+		// case "index-only":
+		// 	return copy.CopySpecificImages, nil
+		// // We don't expose CopySpecificImages other than index-only above, because
+		// // we currently don't provide an option to choose the images to copy. That
+		// // could be added in the future.
+		// default:
+		// 	return copy.CopySystemImage, fmt.Errorf("unknown multi-arch option %q. Choose one of the supported options: 'system', 'all', or 'index-only'", multiArch)
```
mtrmac, line 135: Ultimately, code should be removed, unless it is relevant for future readers. Git preserves the history anyway.

mtrmac (changes requested):
Thanks.

Similarly to c/image, all existing options must keep working.

Also compare previous #1987, that probably discusses relevant concerns; I didn’t now go back to re-read that.

mtrmac (commented):
Eventually, a substantially equivalent feature was added in #2874 . Thanks for this work!

PR #2459, opened 2024-11-12
update CI images to f41

Luap99:
Also fix a problem with a dnf command that no longer works on f41


mtrmac (commented):
ACK; I suppose this should go only after we tag the release, shouldn’t it?

Luap99:
Doesn't matter really, given the CI is not passing on the release PR I see no harm merging this and then rebase there.

mtrmac (approved):
Fair enough, we have already paid for the cost of this CI run, so let’s take advantage of it.

PR #2458, opened 2024-11-12
Bump Skopeo to v1.17.0, c/common v0.61.0, c/image v5.33.0, c/storage v1.56.0

TomSweeneyRedHat:
As the title says.  In preparation of Skopeo v1.17.0 to go
out with Podman v5.3, also bump:

containers/storage:  v1.56.0
containers/image:    v5.33.0
containers/common:   v0.61.0

PR #2460, opened 2024-11-12
DO NOT MERGE: Trigger an ostree-rs-ext test

mtrmac:
Testing for #2458 .


Luap99:
the error included the current go version 1.22.5:
```
go: go.mod requires go >= 1.22.6 (running go 1.22.5; GOTOOLCHAIN=local)
```
https://github.com/containers/skopeo/pull/2458/checks?check_run_id=32876317732

We must trigger a rebuild of the image, I don't see how this would be do it

mtrmac:
I’m sorry, I’m being confusing with the multiple avenues I’ve been pursuing concurrently. This PR exists for the now-added “[Determine GO versions](https://github.com/containers/skopeo/pull/2460/commits/c7333cc6b89a782537791e689479155e97210665)” commit, pursuing the fallback option of doing a RPM update (and this shows that a RPM update would bring us to 1.22.7).

Concurrently, I have been looking at the `quay.io/coreos-assembler/fcos-buildroot:testing-devel` base image — it does seem to now contain Go 1.23.2.

mtrmac:
This looks good; I’ll trigger the container image rebuild in https://github.com/containers/skopeo/pull/2453 .

PR #2469, opened 2024-12-02
Fix handling of errorShouldDisplayUsage

### LINKED ISSUE #2468: Skopeo commands should return non-0 exit status if called incorrectly
If a `skopeo` command is called with invalid flags, e.g. `skopeo copy --dest-tls-verify false docker-archive://... docker://...`, then `skopeo` fails and prints a Usage help message to STDOUT. However, the exit status of the command is `0`, so there is no easy way to programatically detect whether the command has failed.

In this case, the `=` was missed in `--dest-tls-verify=false` and so `false` is seen as the Source for the `copy` command, which is invalid. But the issue is that my script couldn't then detect that the `skopeo` command had failed, and instead reported that the `copy` was successful (which it wasn't, and only became obvious once checking logs).

mtrmac:
- Exit with a non-zero exit code
- Actually report the error we detected
- Write usage to stderr

Fixes #2468.

### DIFF
--- a/cmd/skopeo/utils.go
+++ b/cmd/skopeo/utils.go
@@ -26,7 +26,7 @@ import (
 	"golang.org/x/term"
 )
 
-// errorShouldDisplayUsage is a subtype of error used by command handlers to indicate that cli.ShowSubcommandHelp should be called.
+// errorShouldDisplayUsage is a subtype of error used by command handlers to indicate that the command’s help should be included.
 type errorShouldDisplayUsage struct {
 	error
 }
@@ -62,7 +62,8 @@ func commandAction(handler func(args []string, stdout io.Writer) error) func(cmd
 		err := handler(args, c.OutOrStdout())
 		var shouldDisplayUsage errorShouldDisplayUsage
 		if errors.As(err, &shouldDisplayUsage) {
-			return c.Help()
+			c.SetOut(c.ErrOrStderr()) // This mutates c, but we are failing anyway.
+			_ = c.Help()              // Even if this failed, we prefer to report the original error
 		}
 		return err
 	}
--- a/cmd/skopeo/copy_test.go
+++ b/cmd/skopeo/copy_test.go
@@ -1,0 +1,18 @@
+package main
+
+import "testing"
+
+func TestCopy(t *testing.T) {
+	// Invalid command-line arguments
+	for _, args := range [][]string{
+		{},
+		{"a1"},
+		{"a1", "a2", "a3"},
+	} {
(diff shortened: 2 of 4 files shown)


cmd/skopeo/utils.go
```diff
@@ -26,7 +26,7 @@ import (
 	"golang.org/x/term"
 )
 
-// errorShouldDisplayUsage is a subtype of error used by command handlers to indicate that cli.ShowSubcommandHelp should be called.
+// errorShouldDisplayUsage is a subtype of error used by command handlers to indicate that the command’s help should be included.
 type errorShouldDisplayUsage struct {
```
cgwalters, line 30: Is this a cobra idiom? Is there equivalent to this in other projects we maintain?

Basically just wondering if this could/should be abstracted/reused in multiple places to make bugs like this less likely.

cgwalters (approved):

cmd/skopeo/utils.go
mtrmac: > Is this a cobra idiom?

Not… particularly? cobra has a default behavior of printing usage on _every_ error. That seems excessive, but anyway Skopeo opts out (via `SilenceUsage`), and so do the other projects.

Cobra’s idiomatic way to handle these particular checks is via `command.Args` (e.g. `Args: cobra.ExactArgs(2)`). That doesn’t do anything special about printing, or not printing, the usage message.

---

Skopeo isn’t even consistent internally in usage this `errorShouldDisplayUsage`; it happened to be created when migrating to Cobra, and it was sporadically copy&pasted into some of the other commands.

Printing the usage message here comes from way back, https://github.com/containers/skopeo/commit/f398c9c035fb4aae0df3f3338f02a131979efaf3 , with no documented rationale or reference to any specific convention (e.g. particular the GNU coding standards are silent on this).


I don’t feel particularly strongly about this behavior; if I were told to just report the error and drop all this usage logic, I wouldn’t object. As for consistently printing the usage documentation in other subcommands, I’m not opposed but I’m also not going to work on that now, given other priorities — it’s just the current behavior, returning a success exit code, that I think is unacceptable with high priority.

---

WRT reuse in other projects, not easily: the `commandAction` wrapper exists (due to some … lessons learned :) ) to make the Cobra API _unavailable_ in the command handlers, to force everything to go through the option structs and never look up flags by string names. I do think that’s a better design, but using cobra flag lookups by name is extremely widespread throughout Buildah/Podman (in some cases it’s the only the flags get shared between the Buildah and Podman codebases), so `commandAction` can’t be practically adopted by Buildah/Podman right now.

PR #2493, opened 2025-01-10
systemtest: update quay.io registry image

Luap99:
The "2" tag is very old and not a multi arch manifest. As such testing on aarch64 failed because it pulled and x86_64 image instead. This was found in downstream RHEL testing.

The "2.8.2" is multi arch and used in podman testing were we successfully run aarch64 based testing.


ddarrah (approved):
lgtm

mtrmac (commented):
This _looks_ good to me, but none of the usual Cirrus tests ran for this PR. I’d prefer not to merge this completely untested.

I can’t see why the tests didn’t run — they are supposed to be skipped only for some CI:… titles.

Luap99:
Force pushed, sometimes cirrus doesn't seem to trigger any jobs. That might happen when there are cirrus, github or google cloud outages or some other weird issues. If you notice this more often let me know so we can report this to cirrus.

PR #2500, opened 2025-01-16
feat: Add `--retry-delay` Option

### LINKED ISSUE #2476: feature request: add delay between --retry-times
This is my calling sequence:
```
ssh -4 <user>@<ip> -O ControlMaster=auto -fNL localhost:13340:localhost:20048     # localhost:20048 is the remote registry
skopeo copy --retry-times 10 --dest-tls-verify=false docker://<image> docker://localhost:13340/<image>
ssh -4 <user>@<ip> -O cancel -L localhost:13340:localhost:20048
```
As the remote registry is not directly accessible, I have to use ssh port forwarding as a bridge. 

This works fine if there is no error during skopeo copy. If there is any error in the first try, then any subsequent retries will fail as well. This is because the remote registry is not immediately responding to subsequent connections, until after a time delay. 

This following sequence works fine in the case of retries:
```
ssh -4 <user>@<ip> -O ControlMaster=auto -fNL localhost:13340:localhost:20048     # localhost:20048 is the remote registry
for retries in { 1..10 }; do
    skopeo copy --dest-tls-verify=false docker://<image> docker://<image> && break
    sleep 5s
done
ssh -4 <user>@<ip> -O cancel -L localhost:13340:localhost:20048
``` 
The request is to add time delay between retries.

adambkaplan:
- Add option to set a fixed delay between retries.
- Clarify that if delay is unset, skopeo uses an exponential delay for retries.

Fixes #2476

### DIFF
--- a/cmd/skopeo/utils.go
+++ b/cmd/skopeo/utils.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"strings"
+	"time"
 
 	commonFlag "github.com/containers/common/pkg/flag"
 	"github.com/containers/common/pkg/retry"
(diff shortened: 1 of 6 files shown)


mtrmac (approved):
Thanks!

PR #2501, opened 2025-01-23
RPM: cleanup gobuild macro for CentOS Stream

lsm5:
The default gobuild macro on CentOS Stream now accounts for `BUILDTAGS`, so we don't need to redefine the macro in rpm spec.

### DIFF
--- a/rpm/skopeo.spec
+++ b/rpm/skopeo.spec
@@ -145,6 +145,10 @@ cp -pav systemtest/* %{buildroot}/%{_datadir}/%{name}/test/system/
 #define license tag if not already defined
 %{!?_licensedir:%global license %doc}
 
+# Include this to silence rpmlint.
+# Especially annoying if you use syntastic vim plugin.
+%check
+
 %files
 %license LICENSE
 %doc README.md
--- a/rpm/skopeo.spec
+++ b/rpm/skopeo.spec
@@ -7,20 +7,15 @@
 %global debug_package %{nil}
 %endif
 
-# RHEL's default %%gobuild macro doesn't account for the BUILDTAGS variable, so we
-# set it separately here and do not depend on RHEL's go-[s]rpm-macros package
-# until that's fixed.
-# c9s bz: https://bugzilla.redhat.com/show_bug.cgi?id=2227328
-# c8s bz: https://bugzilla.redhat.com/show_bug.cgi?id=2227331
-%if %{defined rhel}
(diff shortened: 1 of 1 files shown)


lsm5:
The `builder-live.log.gz` files in the [copr builds](https://copr.fedorainfracloud.org/coprs/packit/containers-skopeo-2501/build/8562681/) show BUILDTAGS set in spec are used by gobuild.

lsm5:
@mtrmac PTAL.

rpm/skopeo.spec
```diff
@@ -7,15 +7,6 @@
 %global debug_package %{nil}
 %endif
 
-# RHEL's default %%gobuild macro doesn't account for the BUILDTAGS variable, so we
-# set it separately here and do not depend on RHEL's go-[s]rpm-macros package
-# until that's fixed.
-# c9s bz: https://bugzilla.redhat.com/show_bug.cgi?id=2227328
-# c8s bz: https://bugzilla.redhat.com/show_bug.cgi?id=2227331
-%if %{defined rhel}
-%define gobuild(o:) go build -buildmode pie -compiler gc -tags="rpm_crashtraceback libtrust_openssl ${BUILDTAGS:-}" -ldflags "-linkmode=external -compressdwarf=false ${LDFLAGS:-} -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \\n') -extldflags '%__global_ldflags'" -a -v -x %{?**};
```
mtrmac, line 16: `libtrust_openssl` is not a global RHEL tag, it is specific to our `github.com/containers/libtrust` . Should we really just _drop it_?

I see that the c9s builds still include the tag (apparently it was hard-coded in `go-srpm-macros` in some versions/patches?) but the c10s builds don’t.

Should we add that tag back into our `BUILDTAGS` in `skopeo.spec`? Or is it truly OK to drop it?

mtrmac (commented):
Thanks! I love the cleanup, but there’s one part I’m unsure about.

rpm/skopeo.spec
lsm5: good catch. You're right and my bad for missing that. I'll fix that in spec in this PR itself. Putting back on draft.

lsm5: Alright, the latest build does show `libtrust_openssl` tag for centos stream builds: https://copr.fedorainfracloud.org/coprs/packit/containers-skopeo-2501/build/8566716/ . It will be set for rhel too.

I also included another commit to silence rpmlint.

rpm/skopeo.spec
```diff
@@ -146,6 +145,10 @@ cp -pav systemtest/* %{buildroot}/%{_datadir}/%{name}/test/system/
 #define license tag if not already defined
 %{!?_licensedir:%global license %doc}
 
+# Include this to silence rpmlint.
+# Especially annoying if you use syntastic vim plugin.
+%check
```
mtrmac, line 150: Would we consider at least a smoke-test, like `skopeo --version`? Anyway, that’s not this PR.

mtrmac (approved):
Thanks!

Issue #2497, opened 2025-01-16
skopeo delete - multiarch image oci v1 not fully deleted (artifactory)


ggjulio:
`skopeo delete` doesn't fully delete oci v1 multiarch images.  (at least on artifactory )

It just deletes the "index" `list.manifest.json` but won't delete any of the referenced manifests corresponding to each architecture & attestation.  
Thus the associated layers will never be deleted because they are still referenced by an existing manifest.  
In other terms it (almost) don't free any disk space. (a warning in the man page would be nice)

Maybe this issue should be at container/image, let me know.
Any inputs welcomed.

<details>

<summary>Example</summary>

`list.manifest.json` of tag `latest` :

<details>

```json
{
    "schemaVersion": 2,
    "mediaType": "application/vnd.oci.image.index.v1+json",
    "manifests": [{
            "mediaType": "application/vnd.oci.image.manifest.v1+json",
            "digest": "sha256:bafedd5a97eac06124376a79f857e9506d451c9f8f074f79535b986a8fd5a0a7",
            "size": 42,
            "platform": {
                "architecture": "amd64",
                "os": "linux"
            }
        },
        {
            "mediaType": "application/vnd.oci.image.manifest.v1+json",
            "digest": "sha256:dca88a204d74dc532b0a005068c578d46f0de98dc15921c22d16e5f92769a13e",
            "size": 42,
            "annotations": {
                "vnd.docker.reference.digest": "sha256:bafedd5a97eac06124376a79f857e9506d451c9f8f074f79535b986a8fd5a0a7",
                "vnd.docker.reference.type": "attestation-manifest"
            },
            "platform": {
                "architecture": "unknown",
                "os": "unknown"
            }
        }
    ]
}
```

</details>

Manifest.json of `sha256:bafedd5...`

<details>

```json
{
    "schemaVersion": 2,
    "mediaType": "application/vnd.oci.image.manifest.v1+json",
    "config": {
        "mediaType": "application/vnd.oci.image.config.v1+json",
        "digest": "sha256:b629ae61a4cc1ec62d644dc80c00021d106e87422dc1a54758f8e7c4bec5651c",
        "size": 42
    },
    "layers": [{
            "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip",
            "digest": "sha256:6414378b647780fee8fd903ddb9541d134a1947ce092d08bdeb23a54cb3684ac",
            "size": 42
        },
        {
[...]

PR #2504, opened 2025-01-28
docs: Update skopeo delete man page #2497

ggjulio:
#2497

### DIFF
--- a/docs/skopeo-delete.1.md
+++ b/docs/skopeo-delete.1.md
@@ -15,6 +15,7 @@ The effect of this is registry-specific; many registries don’t support this op
 
 **WARNING**: If _image-name_ contains a tag (but not a digest), in the current version of Skopeo this resolves the tag into a digest, and then deletes the manifest by digest, as described above (possibly deleting all tags pointing to that manifest, not just the provided tag). This behavior may change in the future.
 
+**WARNING**: If _image-name_ is a list manifest (ex: multi-platform images), only the `list.manifest.json` is deleted which may lead to storage issue.  Only the registry server is able to find dandling manifests as one may be referenced by many `list.manifest.json`. Some registries like artifactory don't have any garbage collector built in.
 
 When using the github.com/distribution/distribution registry server:
 To release the allocated disk space, you must login to the container registry server and execute the container registry garbage collector. E.g.,


docs/skopeo-delete.1.md
```diff
@@ -15,6 +15,7 @@ The effect of this is registry-specific; many registries don’t support this op
 
 **WARNING**: If _image-name_ contains a tag (but not a digest), in the current version of Skopeo this resolves the tag into a digest, and then deletes the manifest by digest, as described above (possibly deleting all tags pointing to that manifest, not just the provided tag). This behavior may change in the future.
 
+**WARNING**: If _image-name_ is a list manifest (ex: multi-platform images), only the `list.manifest.json` is deleted which may lead to storage issue.  Only the registry server is able to find dandling manifests as one may be referenced by many `list.manifest.json`. Some registries like artifactory don't have any garbage collector built in.
```
mtrmac, line 18: I don’t think this needs as much emphasis as the warning above, which is about data loss.

mtrmac, line 18: “storage issue” does not say what the problem is.

mtrmac, line 18: `s/find dandling manifests/correctly determine which manifests are dangling/` , or better “unreferenced” or “unreachable”

mtrmac, line 18: `list.manifest.json` is not a widely-used terminology, to my knowledge. “A multi-platform manifest list” maybe.

mtrmac (commented):
Thanks for the PR.

I can’t verify the factual accuracy; https://jfrog.com/help/r/jfrog-artifactory-documentation/delete-multi-architecture-docker-tags suggests that there is _some_ support for deleting everything in one operation, and it seems a bit surprising to me that using the Docker API would not trigger the same logic. But I wouldn’t know.

With that disclaimer: I think this caveat makes sense to document, but we need to structure the man page in a way that is reasonably easy to follow, both for users who are need to do something registry-specific and for users who just care about the CLI options.

Maybe, at that point in the man page, introduce a new “registry server-side concerns” section, talking both (first?) about the distribution/distribution server, and about Artifactory.

PR #2524, opened 2025-02-17
WIP: Transport completions

yedayak:
Another thing I would like to add is completion of `containers-storage:`, but I'm not sure how to get the list of available images.


cmd/skopeo/completions.go
```diff
@@ -4,10 +4,47 @@ import (
...
+	"path/filepath"
+	"strings"
 )
 
-// autocompleteSupportedTransports list all supported transports with the colon suffix.
-func autocompleteSupportedTransports(cmd *cobra.Command, args []string, toComplete string) ([]cobra.Completion, cobra.ShellCompDirective) {
+func autocompleteTransports(cmd *cobra.Command, args []string, toComplete string) ([]cobra.Completion, cobra.ShellCompDirective) {
+	directive := cobra.ShellCompDirectiveNoSpace
+	// We still don't have a transport
+	if !strings.Contains(toComplete, ":") {
```
Luap99, line 16: Overall this would likely be much better written with `strings.Cut()`

That way you can easily check if a colon was set and the you have the clear prefix and can match that without the colon below.

cmd/skopeo/completions.go
```diff
@@ -4,10 +4,47 @@ import (
...
+		transports := supportedTransportSuggestions()
+		return transports, directive | cobra.ShellCompDirectiveNoFileComp
+	}
+	if strings.HasPrefix(toComplete, "oci-archive:") || strings.HasPrefix(toComplete, "docker-archive:") || strings.HasPrefix(toComplete, "sif:") || strings.HasPrefix(toComplete, "oci:") {
+		return nil, directive
+	}
+	if toComplete == "docker:" {
+		return []cobra.Completion{"docker://"}, directive | cobra.ShellCompDirectiveNoFileComp
+	}
+	if strings.HasPrefix(toComplete, "dir:") {
```
Luap99, line 26: here you then should match the transpart names as defined by the code and not hard code the strings
`docker.Transport.Name()`, `directory.Transport.Name()`,etc...

cmd/skopeo/completions.go
```diff
@@ -4,10 +4,47 @@ import (
...
+			// Fallback to whatever the shell gives us
+			return nil, cobra.ShellCompDirectiveFilterDirs | cobra.ShellCompDirectiveNoFileComp
+		}
+		suggestions := make([]cobra.Completion, 0, len(entries))
+		for _, e := range entries {
+			if e.IsDir() {
+				suggestions = append(suggestions, "dir:"+path.Join(curDir, e.Name())+"/")
+			}
+		}
+		return suggestions, cobra.ShellCompDirectiveNoFileComp | cobra.ShellCompDirectiveNoSpace
```
Luap99, line 41: Is that actually worth it? Returning all paths as done by the shell seems more than fine, compared of having to read all files here.

Luap99 (commented):
Overall this looks like a nice improvement, just some comments on the implementation. (Note I am not an official maintainer here but I have done most of the cobra completion work in podman and skopeo)

cmd/skopeo/completions.go
yedayak: Yeah, this works much better. Updated to do this

cmd/skopeo/completions.go
yedayak: I agree its a bit overkill, this just makes the completions more accurate, showing only directories. I'm fine with not doing it.

lsm5:
@Luap99 do we test completions in CI anywhere yet?

Luap99:
> @Luap99 do we test completions in CI anywhere yet?

In podman we have number of rather complex tests (test/system/600-completion.bats). But that is not something I would recommend for skopeo, most of the changes here are trivial to test manually (and it is not like they will get changed often) and there is basically no risk if the completion is not working fully.

cmd/skopeo/completions.go
```diff
@@ -1,13 +1,57 @@
...
+		return []cobra.Completion{docker.Transport.Name() + "://"}, directive | cobra.ShellCompDirectiveNoFileComp
+	}
+	if transport == directory.Transport.Name() {
+		// just ShellCompDirectiveFilterDirs is more correct, but doesn't work here in bash, see https://github.com/spf13/cobra/issues/2242. Instead we get the directories ourselves.
+		curDir := filepath.Dir(details)
+		entries, err := os.ReadDir(curDir)
+		if err != nil {
+			cobra.CompErrorln("Failed ReadDir at " + curDir)
+			// Fallback to whatever the shell gives us
+			return nil, cobra.ShellCompDirectiveFilterDirs | cobra.ShellCompDirectiveNoFileComp
```
mtrmac, line 40: The way I read the documentation, `NoFileComp` should make `FilterDirs` entirely irrelevant. (In the implementation, that seems not to matter.)

cmd/skopeo/completions.go
```diff
@@ -1,13 +1,57 @@
...
+	transport, details, haveTransport := strings.Cut(toComplete, ":")
+	if !haveTransport {
+		transports := supportedTransportSuggestions()
+		return transports, directive | cobra.ShellCompDirectiveNoFileComp
+	}
+	if transport == ociArchive.Transport.Name() || transport == dockerArchive.Transport.Name() || transport == sif.Transport.Name() || transport == oci.Transport.Name() {
+		return nil, directive
+	}
+	if transport == docker.Transport.Name() && details == "" {
+		return []cobra.Completion{docker.Transport.Name() + "://"}, directive | cobra.ShellCompDirectiveNoFileComp
```
mtrmac, line 31: `transport+"://"` please: we’ve had a proposal to support other names for this transport, so let’s not hard-code more than necessary.

cmd/skopeo/completions.go
```diff
@@ -1,13 +1,57 @@
...
+		entries, err := os.ReadDir(curDir)
+		if err != nil {
+			cobra.CompErrorln("Failed ReadDir at " + curDir)
+			// Fallback to whatever the shell gives us
+			return nil, cobra.ShellCompDirectiveFilterDirs | cobra.ShellCompDirectiveNoFileComp
+		}
+		suggestions := make([]cobra.Completion, 0, len(entries))
+		for _, e := range entries {
+			if e.IsDir() {
+				suggestions = append(suggestions, directory.Transport.Name()+":"+path.Join(curDir, e.Name())+"/")
```
mtrmac, line 45: `transport+":"…` please

cmd/skopeo/completions.go
```diff
@@ -1,13 +1,57 @@
...
 	"github.com/spf13/cobra"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
 )
 
-// autocompleteSupportedTransports list all supported transports with the colon suffix.
-func autocompleteSupportedTransports(cmd *cobra.Command, args []string, toComplete string) ([]cobra.Completion, cobra.ShellCompDirective) {
+func autocompleteTransports(cmd *cobra.Command, args []string, toComplete string) ([]cobra.Completion, cobra.ShellCompDirective) {
```
mtrmac, line 19: This is now  `autocompleteImageNames`, now that it is trying to complete the full arguments.

cmd/skopeo/completions.go
```diff
@@ -1,13 +1,57 @@
...
+	directive := cobra.ShellCompDirectiveNoSpace
+	// We still don't have a transport
```
mtrmac, line 21: This comment belongs inside the next `if` (or on the `if !haveTransport` line), it is incorrect for other paths through this function.

mtrmac (commented):
Thanks!

mtrmac:
> > @Luap99 do we test completions in CI anywhere yet?
> 
> In podman we have number of rather complex tests (test/system/600-completion.bats).

Here, I think it would be easier, and we would not give up on ~any test coverage compared to what Podman does, to use Go unit tests for the individual parts (and, for that purpose, to split the `dir:` logic into a separate function).

> and there is basically no risk if the completion is not working fully.

Yes. I don’t think missing tests are a blocker.

cmd/skopeo/completions.go
```diff
@@ -1,13 +1,57 @@
...
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
 )
 
-// autocompleteSupportedTransports list all supported transports with the colon suffix.
-func autocompleteSupportedTransports(cmd *cobra.Command, args []string, toComplete string) ([]cobra.Completion, cobra.ShellCompDirective) {
+func autocompleteTransports(cmd *cobra.Command, args []string, toComplete string) ([]cobra.Completion, cobra.ShellCompDirective) {
+	directive := cobra.ShellCompDirectiveNoSpace
```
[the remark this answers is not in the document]
yedayak: Yeah, I over complicated it :) removed

cmd/skopeo/completions.go
yedayak: Yeah ShellCompDirectiveFilterDirs by itself should just complete directories (though in bash it currently doesn't). fixed

cmd/skopeo/completions.go
yedayak: done

cmd/skopeo/completions.go
yedayak: Removed the comment since it's now obvious from the variable name

cmd/skopeo/completions.go
```diff
@@ -1,13 +1,55 @@
...
-// autocompleteSupportedTransports list all supported transports with the colon suffix.
-func autocompleteSupportedTransports(cmd *cobra.Command, args []string, toComplete string) ([]cobra.Completion, cobra.ShellCompDirective) {
+func autocompleteImageNames(cmd *cobra.Command, args []string, toComplete string) ([]cobra.Completion, cobra.ShellCompDirective) {
+	transport, details, haveTransport := strings.Cut(toComplete, ":")
+	if !haveTransport {
+		transports := supportedTransportSuggestions()
+		return transports, cobra.ShellCompDirectiveNoSpace | cobra.ShellCompDirectiveNoFileComp
+	}
+	switch transport {
+	case ociArchive.Transport.Name(), dockerArchive.Transport.Name(), sif.Transport.Name(), oci.Transport.Name():
```
Luap99, line 26: Isn't oci also a directory so it should be in the case below with dir:

cmd/skopeo/completions.go
```diff
@@ -1,13 +1,55 @@
...
+		return nil, cobra.ShellCompDirectiveNoSpace
```
Luap99, line 27: This should return the Default and not NoSpace, no space means the default shell path completion never adds the final space meaning users have to add it manually which is not how most shell completion works.

cmd/skopeo/completions.go
```diff
@@ -1,13 +1,55 @@
...
+		entries, err := os.ReadDir(curDir)
+		if err != nil {
+			cobra.CompErrorln("Failed ReadDir at " + curDir)
+			// Fallback to whatever the shell gives us
+			return nil, cobra.ShellCompDirectiveFilterDirs
+		}
+		suggestions := make([]cobra.Completion, 0, len(entries))
+		for _, e := range entries {
+			if e.IsDir() {
+				suggestions = append(suggestions, transport+":"+path.Join(curDir, e.Name())+"/")
```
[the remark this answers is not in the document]
Luap99: yes, BUT this is not actually working correctly in all scenarios, it fails short when using relative paths that start with `./` or paths that have `//` and many more, the reason is that Join() calls Clean() which may modify the input path curDir. That causes shell completion to ignore the resuls as the string no longer match

i.e. this returns the wrong result as ./ is stipped
```
$ bin/skopeo __complete inspect dir:./
dir:.fmf/
dir:.git/
dir:.github/
dir:bin/
dir:cmd/
dir:completions/
dir:contrib/
dir:docs/
dir:hack/
dir:integration/
dir:plans/
dir:rpm/
dir:systemtest/
dir:vendor/
dir:version/
:6
Completion ended with directive: ShellCompDirectiveNoSpace, ShellCompDirectiveNoFileComp
```

In podman I worked around that with this function: https://github.com/containers/podman/blob/ebec104391a1f49cb6c91d7dd7a7a04900b18f61/cmd/podman/common/completion.go#L418-L431

also when using filepath it should be using `+string(os.PathSeparator)` not `/`

Overall I am not convinced that the extra complications are worth it here, I rather just return the Default/FilterDirs and let the shell handle it and figure out a way to fix these things in the completion script in cobra if possible as you reported in https://github.com/spf13/cobra/issues/2242

cmd/skopeo/completions.go
mtrmac: `oci`/`oci-archive`/`docker-archive` also accept a reference after the directory/regular file/regular file path.

`sif:` accepts only a path, so adding a space there seems better.

cmd/skopeo/completions.go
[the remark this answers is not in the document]
yedayak: I removed the code to get the directories for now, just went with the default completions since that seems more useful how it is now

cmd/skopeo/completions.go
```diff
@@ -1,13 +1,55 @@
...
+			}
+		}
+		return suggestions, cobra.ShellCompDirectiveNoFileComp | cobra.ShellCompDirectiveNoSpace
+	}
+	if transport == docker.Transport.Name() && details == "" {
```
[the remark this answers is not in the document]
yedayak: done

cmd/skopeo/completions.go
yedayak: Put a comment about this

cmd/skopeo/completions.go
yedayak: Put oci in a separate case since it can have a ref

mtrmac (approved):
Thanks! This is a great improvement.

PR #2527, opened 2025-02-28
Enforce manifest and blob digests in (skopeo inspect) and (skopeo layers)

mtrmac:
Nobody should be using `skopeo layers`.

For `skopeo inspect`, I could see an argument that it should return the data even if it does not match the digest, so that users can diagnose that. On balance, I think matching the expectation that digests are enforced is more valuable.

### DIFF
--- a/cmd/skopeo/inspect.go
+++ b/cmd/skopeo/inspect.go
@@ -106,8 +106,9 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error) 
 		}
 	}()
 
+	unparsedInstance := image.UnparsedInstance(src, nil)
 	if err := retry.IfNecessary(ctx, func() error {
-		rawManifest, _, err = src.GetManifest(ctx, nil)
+		rawManifest, _, err = unparsedInstance.Manifest(ctx)
 		return err
 	}, opts.retryOpts); err != nil {
 		return fmt.Errorf("Error retrieving manifest for image: %w", err)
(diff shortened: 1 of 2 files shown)


cmd/skopeo/inspect.go
```diff
@@ -106,8 +106,9 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
 		}
 	}()
 
+	unparsedInstance := image.UnparsedInstance(src, nil)
 	if err := retry.IfNecessary(ctx, func() error {
-		rawManifest, _, err = src.GetManifest(ctx, nil)
```
cgwalters, line 110: I have a dim memory we talked about this before but is there a reason why `GetManifest` does not verify by default? We're just talking about some small JSON so the overhead of sha256 on it has to be pretty small.

mtrmac: There are ~11 implementations in c/image + several outside, added over time; trusting that all of them are going to do the right thing is hard to enforce.

The idea originally was that the transports only deal with “I/O” and centralized code in c/image, primarily c/image/copy, provides a _single_ location where the enforcement provably happens.

That doesn’t really work with the various external ImageSource users… compare https://github.com/containers/image/pull/2742 .

We might provide some other interface, but we can’t prevent direct calls to `GetManifest` without breaking the stable API.

cgwalters: > The idea originally was that the transports only deal with “I/O” and centralized code in c/image, primarily c/image/copy, provides a single location where the enforcement provably happens.

Maybe the way to think of this is that we should offer a "copy metadata to *memory*" operation?

TomSweeneyRedHat:
/approve
LGTM

PR #2539, opened 2025-03-07
.github: remove cirrus rerun action

Luap99:
As pointed out in buildah[1] the action is broken in bad ways where it can trigger 1000+ rerun wasting our cloud resources.

Get rid of it for now until we find something better or can properly identify and fix the root cause.


mtrmac (approved):
Thanks!

Here’s hoping for a real fix, eventually…

PR #2554, opened 2025-03-21
proxy: Add GetLayerInfoPiped

cgwalters:
I was experimenting with images with lots of layers (> 200) and this invocation was incorrectly adding the entire response into what was intended as the metadata plane.

`GetManifest` and `GetConfig` (even those are relatively small) still always return their data over a pipe, same as blobs.

Add a new `GetLayerInfoPiped` that does the same so we can easily get this information for images with a lot of layers.


cgwalters:
Consumer in https://github.com/containers/containers-image-proxy-rs/pull/76

integration/proxy_test.go
```diff
@@ -291,6 +292,20 @@ func runTestGetManifestAndConfig(p *proxy, img string) error {
 		return fmt.Errorf("No CMD or ENTRYPOINT set")
 	}
 
+	// Also test this legacy interface
```
mtrmac, line 295: Is this comment appropriate?

integration/proxy_test.go
```diff
@@ -229,6 +229,7 @@ type byteFetch struct {
 	err     error
 }
 
+// This exercises all the metadata fetching APIs.
 func runTestGetManifestAndConfig(p *proxy, img string) error {
```
mtrmac, line 233: Then maybe rename this instead of adding a comment that the name doesn’t match?

cmd/skopeo/proxy.go
```diff
@@ -619,9 +620,8 @@ func (h *proxyHandler) GetBlob(args []any) (replyBuf, error) {
 
 // GetLayerInfo returns data about the layers of an image, useful for reading the layer contents.
 //
-// This needs to be called since the data returned by GetManifest() does not allow to correctly
-// calling GetBlob() for the containers-storage: transport (which doesn’t store the original compressed
-// representations referenced in the manifest).
+// This is the same as GetLayerInfoPiped, but returns its contents inline. This is subject to
+// failure for large images and is only retained for backwards compatibility.
```
mtrmac, line 624: I don’t think there is a Skopeo-side failure – “hard to handle in callers which want to impose a limit on the total size of the initial `reply`” maybe?

mtrmac (commented):
Thanks!

The code is fair enough, but, just to be clear, this changes ~nothing Skopeo-side: the manifest, and the layer info array, all exist in memory. AFAICS this makes a difference only because the caller has a https://github.com/containers/containers-image-proxy-rs/blob/63f5f96d4196ffb31a9167f6421ceb0df07ec111/src/imageproxy.rs#L382 .

That’s a reasonable design — a caller looking for a success/failure indicator has a fair need to get that value without having to read an unbounded value into memory first — I just want to make sure there is no misunderstanding.

cgwalters:
> AFAICS this makes a difference only because the caller has a https://github.com/containers/containers-image-proxy-rs/blob/63f5f96d4196ffb31a9167f6421ceb0df07ec111/src/imageproxy.rs#L382

No, that issue is inherently because we use SOCK_SEQPACKET and the kernel doesn't support queuing arbitrary message sizes there. 

There's an equivalent constant here https://github.com/containers/skopeo/blob/dd71592115d48548b5527eea696f5ef3b572ace8/cmd/skopeo/proxy.go#L98

cmd/skopeo/proxy.go
cgwalters: I clarified this in the comment

mtrmac (approved):
Thanks! LGTM.

cgwalters:
Just to recap briefly here the rationale for the protocol as designed is that it clearly splits metadata and data channels, allowing us to pipeline operations.

We previously discussed alternative IPC protocols in e.g. https://github.com/containers/skopeo/pull/1844#issuecomment-1378777187

cgwalters:
Hmm...it looks like the testing farm jobs succeeded, but didn't report that success to GH?
/packit retest

mtrmac:
You’re all good, the “release” jobs don’t run on the `main` branch, but, sadly, show up in the list.

@lsm5 this is https://github.com/packit/packit-service/issues/2678 , isn’t it?

cgwalters:
Hmm, but all I can do here is "Enable auto-merge" but that I think is waiting for those jobs?

mtrmac:
Not on those. I’m confused about that button as well — my current theory is that it is waiting on the PR to be rebased (in some other PR, “Update Branch / Update with merge commit” didn’t help IIRC), but I didn’t properly test that theory yet.

cgwalters:
Hmm, does this repo have a config enabled that requires each PR to be against current main? That seems to be an unnecessary hit to ergonomics

cgwalters:
Ah yes 
image
- intentional?

mtrmac:
I’m not at all sure about the full history. ISTR requiring PRs to be fresh is intentional, but also admins had the option to override it and force-merge — maybe _that_ has been disabled. I don’t know when or by who.

TomSweeneyRedHat:
LGTM
I'll leave the merge honors to @mtrmac

lsm5:
> @lsm5 this is [packit/packit-service#2678](https://github.com/packit/packit-service/issues/2678) , isn’t it?

yes. Actually I think it's best to remove those additional jobs until I have the right github actions figured out. Will submit a PR.

cgwalters:
Minor thing but I'd definitely like this in RHEL 10.1, do I need to file a JIRA or I presume skopeo will just be rev'd at some point for usual reasons anyways?

PR #2558, opened 2025-03-25
[skip-ci] TMT: keep PR-label independent tests

lsm5:
Switch to keeping TMT tests independent of PR labels for now.

In order to keep PR-label dependent tests, Packit UI would need improvement making it clear that some tests are not meant to run, perhaps also changing the status to `ignore` or `neutral`.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -102,27 +102,11 @@ jobs:
       failure_comment:
         message: "Tests failed. @containers/packit-build please check."
     targets: *fedora_copr_targets
-    require: &dev_branch
-      label:
-        absent:
-          - release
     tf_extra_params:
       environments:
         - artifacts:
           - type: repository-file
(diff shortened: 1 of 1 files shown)


mtrmac (approved):
Thank you! This is a great improvement for drive-by contributors.

(On release branches, it’s mostly us doing the backports, so I’m not concerned about a bit confusing UI. But, to leave a breadcrumb, on _future_ release branches we will need to re-enable the release-branch testing.)

lsm5:
> (On release branches, it’s mostly us doing the backports, so I’m not concerned about a bit confusing UI. But, to leave a breadcrumb, on _future_ release branches we will need to re-enable the release-branch testing.)

A dirty-but-works way would be to remove the entire `tf-extra-params` from the packit job. That would ensure we only test with official distro packages.

mtrmac:
If we expect new branches to be cut in 3-6 months, maybe we don’t need a dirty-but-works hack for now. Hopefully TMT will improve this in the meantime — and if not, we have these comments just to record what is going on, and we can add hacks later.

PR #2564, opened 2025-04-02
proxy: call maybeReexec() up front

### LINKED ISSUE #2563: `experimental-image-proxy` broken with `containers-storage:`
```
skopeo version 1.18.0
skopeo-1.18.0-1.fc42.x86_64
```

Using `containers-image-proxy-rs`, if I attempt to pull an image from a `containers-storage:` URL I get this:

```
Error: Opening image

Caused by:
    failed to invoke method OpenImage: failed to invoke method OpenImage: mount /var/home/lis/.local/share/containers/storage/overlay:/var/home/lis/.local/share/containers/storage/overlay, flags: 0x1000: operation not permitted
```

It seems to be caused by a bug in `skopeo` itself: normally during a `skopeo copy` command trying to access a `containers-storage:` source, `skopeo` does the equivalent of `podman unshare` in order to gain access to the containers storage.  The `skopeo experimental-image-proxy` seems to "forget" to do the same and gets the permissions error.  In particular, the process that skopeo is running in does:

```
mount("/var/home/lis/.local/share/containers/storage/overlay", "/var/home/lis/.local/share/containers/storage/overlay", 0xc000013c88, MS_BIND, NULL) = -1 EPERM (Operation not permitted)
```

which fails because it didn't call `unshare()` anywhere before that.

You can resolve the issue by adding a wrapper around `skopeo` in your `$PATH` which calls `podman unshare /usr/bin/skopeo` and then it will work.

allisonkarlitskaya:
Commands like `skopeo copy` do this first depending on the image names involved in the operation, but we don't have the luxury of having access to that information up-front.  Always call maybeReexec(), just in case it's required later on.

Fixes #2563

### DIFF
--- a/cmd/skopeo/proxy.go
+++ b/cmd/skopeo/proxy.go
@@ -876,6 +876,12 @@ func (h *proxyHandler) processRequest(readBytes []byte) (rb replyBuf, terminate 
 
 // Implementation of podman experimental-image-proxy
 func (opts *proxyOptions) run(args []string, stdout io.Writer) error {
+	// We may get a request to pull from a containers-storage: URL, so do
+	// this up front.
+	if err := maybeReexec(); err != nil {
+		return err
+	}
+
 	handler := &proxyHandler{
 		opts:        opts,
 		images:      make(map[uint64]*openImage),
--- a/cmd/skopeo/unshare.go
+++ b/cmd/skopeo/unshare.go
@@ -2,6 +2,10 @@
 
 package main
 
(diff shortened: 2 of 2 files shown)


mtrmac (changes requested):
Thanks!

Let’s discuss the full impact in #2563 first, to make sure this doesn’t break important users.

allisonkarlitskaya:
I think we decided not to do this.  Let's leave #2563 open anyway.

PR #2565, opened 2025-04-04
Add golangci-lint run --tests=false, fix found issues

kolyshkin:
See individual commits for detals.


hack/validate-lint.sh
```diff
@@ -1,16 +0,0 @@
...
-errors=$($GOBIN/golangci-lint run --build-tags "${BUILDTAGS}" 2>&1)
-
-if [ "$?" -eq 0 ]; then
-	echo 'Congratulations!  All Go source files have been linted.'
-else
-	{
-		echo "Errors from golangci-lint:"
-		echo "$errors"
-		echo
-		echo 'Please fix the above errors. You can test via "golangci-lint" and commit the result.'
```
mtrmac, line 12: \*shrug\* IIRC things like this were originally added because external contributors were confused about some of the CI failures. But that must have happened with some earlier linter, and it’s not strong reason to keep this.

Makefile
```diff
@@ -237,7 +237,8 @@ test-all-local: validate-local validate-docs test-unit-local
 validate-local:
 	hack/validate-git-marks.sh
 	hack/validate-gofmt.sh
-	GOBIN=$(GOBIN) hack/validate-lint.sh
+	$(GOBIN)/golangci-lint run --build-tags "${BUILDTAGS}"
+	$(GOBIN)/golangci-lint run --build-tags "${BUILDTAGS}" --tests=false
```
mtrmac, line 241: \*shrug\* I don’t mind moving the files to make them test-only, but, also, there’s ultimately not that much point to this? If we are not going to add this kind of extra run on every project, singling out Skopeo here is a bit surprising.

So is the _need_ for an extra run like this at the first place — I rather think that all of this is a missing feature of the `unused` linter, and it should conceptually happen inside an ordinary run.

mtrmac (commented):
Thanks!

The added build tags are clearly correct. The file renames, why not. Adding an extra linter run… feels weird, but I’ll let others weigh in.

hack/validate-lint.sh
kolyshkin: Yeah, I think that golangci-lint-action is verbose these days, clearly showing the errors in the annotated diffs.

Makefile
kolyshkin: Conceptually, yes. In the real world, though, there are some rough edges around linting wrt build tags and test files. Practically, golangci-lint can't (yet?) lint with/without various build tags and/or tests in a single run. Maybe this will be fixed, but now it's not the case.

As to why this is needed -- the biggest thing I was able to find using `golangci-lint run --tests=false` is here: https://github.com/containers/buildah/pull/6101 (some no-longer-used code was kept because it had a unit test). I think it's worth it, given that the second golangci-lint run is much faster thanks to go and golangci-lint caching.

kolyshkin: Having said that, I can drop the last patch from this PR if that's preferred.

mtrmac: I’m generally unenthusiastic about maintaining lint configs… but that Buildah PR _does look_ pretty attractive for the price. Removing unused code is always beneficial.

Maybe add a comment here, something like
```
# --tests=false allow detecting code unused outside of tests; ideally the linter should be able to find this automatically.
# Since everything is already cached, this additional run doesn't take much time.
```

justifying the extra run as a ~special case (contrary to the extra cost of https://github.com/containers/image/pull/2815 ), promoting the commit message to better visibility for future maintainers, and then leave it in?

mtrmac (approved):
Thank you!

lsm5 (approved):
LGTM

PR #2566, opened 2025-04-04
.golangci.yml: simplify

kolyshkin:
1. Remove some unused exclusions, document the used ones.
2. ~~Add `errorlint`, fix found warnings ("Use `%w` to format errors").~~ (dropped)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -2,10 +2,12 @@ version: "2"
 linters:
   settings:
     staticcheck:
-      checks: ["all", "-ST1000", "-ST1003", "-ST1016", "-ST1020", "-ST1021", "-ST1022", "-ST1005"] # golangci-lint's default, we add -ST1005
+      checks:
+        # Compared to golangci-lint v2.0.2 defaults, we don’t exclude
+        # ST1003, ST1016, ST1020, ST1021, ST1022 as we don't hit those.
+        - all
+        - -ST1000 # Incorrect or missing package comment.
+        - -ST1005 # Incorrectly formatted error string.
   exclusions:
     presets:
-      - comments
-      - common-false-positives
-      - legacy
       - std-error-handling


.golangci.yml
```diff
@@ -1,11 +1,13 @@
...
 linters:
+  enable:
+    - errorlint
   settings:
     staticcheck:
-      checks: ["all", "-ST1000", "-ST1003", "-ST1016", "-ST1020", "-ST1021", "-ST1022", "-ST1005"] # golangci-lint's default, we add -ST1005
+      checks:
+        - all
+        - -ST1000 # Incorrect or missing package comment.
+        - -ST1005 # Incorrectly formatted error string.
```
mtrmac, line 10: This is not the same thing as the previous one. I don’t really want to be in the business of micromanaging this.

(Before `version: "2"`, the configuration used to start with `inherit`, with us adding only the exceptions, but that no longer works for some reason.)

.golangci.yml
```diff
@@ -1,11 +1,13 @@
 version: "2"
 linters:
+  enable:
+    - errorlint
```
mtrmac, line 4: Compare https://github.com/containers/image/pull/2512 . As in there, adding this linter only adds unnecessary work. (`%#v` vs. `%#w` makes no difference at the CLI level.)

.golangci.yml
```diff
@@ -1,11 +1,13 @@
...
-      checks: ["all", "-ST1000", "-ST1003", "-ST1016", "-ST1020", "-ST1021", "-ST1022", "-ST1005"] # golangci-lint's default, we add -ST1005
+      checks:
+        - all
+        - -ST1000 # Incorrect or missing package comment.
+        - -ST1005 # Incorrectly formatted error string.
   exclusions:
     presets:
-      - comments
-      - common-false-positives
-      - legacy
```
mtrmac, line 10: The linter’s default is `[]`, so, this is a move in the preferred direction.

mtrmac (changes requested):
Thanks!

I’m generally very skeptical of detailed linter configuration given our limited capacity (https://github.com/containers/image/pull/2244 ), because that’s committing us to a long-term maintenance effort.

I’d prefer not to deviate from defaults if at all possible.

.golangci.yml
kolyshkin: > This is not the same thing as the previous one.

Right, it's not. As per commit message, this patch removes some unused exclusions.

My thinking is, it makes no sense to disable warnings which we never hit. If and when we hit those warnings, we can decide on a case by case basis whether we prefer to exclude or fix those.
 
As to "let's stick to golangci-lint defaults" mantra, I generally agree, but in this case we clearly already don't follow it. Since we deviate, it's probably best to minimize the linter configuration.

.golangci.yml
kolyshkin: Removed the second patch. I'm slightly concerned about not being able to lint for missing `%w` in other parts of the code, but perhaps this is not a big deal.

kolyshkin:
Second commit removed.

.golangci.yml
mtrmac: > If and when we hit those warnings, we can decide on a case by case basis whether we prefer to exclude or fix those.

= committing to spend more time on this in the future…

… Fine. Would documenting `all # Compared to Golangci-lint 2.0.2 defaults, we don’t exclude  "-ST1003", "-ST1016", "-ST1020", "-ST1021", "-ST1022" because we don’t currently hit those` work?

kolyshkin: Updated.

lsm5:
The Packit rawhide failures should go away once the copr buildroot gets a new libgpg-error.

mtrmac (approved):
Thanks!

(And thanks @lsm5 for keeping an eye on the build failures.)

PR #2574, opened 2025-04-11
.github: check_cirrus_cron work around github bug

Luap99:
So I wondered why our email workflow only reported things for podman...

It seems the secrets: inherit is broken and no longer working, I see all jobs on all repos failing with:

Error when evaluating 'secrets'. .github/workflows/check_cirrus_cron.yml (Line: 19, Col: 11): Secret SECRET_CIRRUS_API_KEY is required, but not provided while calling.

This makes no sense to me I doubled checked the names, nothing changed on our side and it is consistent for all projects. Interestingly this same thing passed on March 10 and 11 (on all repos) but failed before and after this as well.

Per[1] we are not alone, anyway let's try to get this working again even if it means more duplication.

[1] actions/runner#2709


mtrmac (approved):
Thanks!

mheon:
LGTM

PR #2596, opened 2025-05-07
[CI:DOCS] README.md: Add badges

lsm5:
Add badges for License, Latest Release and Go Report Card. Copied and modified from podman.


mtrmac (commented):
Thanks!

I’m generally skeptical of badges — I think users mostly don’t care; maintainers are certainly not going to visit the repository top page to _poll_ for status (we have `podman-monitor` for that); and to the extent the badges do anything, their primary effect is to advertise the link destination, at cost to users/maintainers.

Those are ~trivial concerns — it’s far more important for the projects to be consistent — so I’m fine in principle with adding these things; I did want to write the skepticism down explicitly, so that I’m not interpreted to be enthusiastically supporting the spread of the badges.

lsm5:
Ack. This PR in particular is not really required by anyone AFAIK, just that it was there in podman and I basically thought `why not here`. I guess if the cost of advertizing turns out to be too high, we could remove it later. I'll defer to you.

mtrmac (approved):
\*shrug\* At least 2 people think the badges are a good idea, I should not be making everyone spend more time on this.

Issue #1933, opened 2023-03-06
Skopeo inspect outright fails with OCI artifacts


mpagot:
With Skopeo 1.5.0 I was able to get

```
% podman run -it skopeo/stable:v1.5.0 inspect docker://registry.opensuse.org/devel/sap/trento/factory/charts/trento/trento-server:1.2.2-dev2

{
    "Name": "registry.opensuse.org/devel/sap/trento/factory/charts/trento/trento-server",
...
    "Env": null
}
```

With the latest Skopeo version I have an error

```
% podman run -it skopeo/stable:v1.11.1 inspect docker://registry.opensuse.org/devel/sap/trento/factory/charts/trento/trento-server:1.2.2-dev2

FATA[0000] unsupported image-specific operation on artifact with type "application/vnd.cncf.helm.config.v1+json"
```

mtrmac:
Thanks for your report.

The change to refuse to provide image-specific data for non-image artifacts is intentional, although we didn’t anticipate that completely breaking `skopeo inspect`.

What do you actually use `skopeo inspect` for in this case? The only data that is relevant for the artifact is name (already known to the user), other tags in the repo (available via `list-tags`), and data about the blob digests/MIME types/sizes.

So what is the necessary output? Is it essentially an image presence check? Is the goal to list data about the blobs (and if so, for what purpose)? Something else?

mpagot:
Last version working for me is 1.8.0. First version with this FATA is 1.9.0
Looking at the code I'd guess it has been introduced with https://github.com/containers/skopeo/commit/62158a58bcef9240e821d92253484420b952ddbe

mpagot:
I agree with you that `skopeo inspect` is not needed for helm chart

Feel free to close this issue

mtrmac:
I think this report is valuable — and I meant the question about what is the purpose and expected output.

---

On the one hand, the current fields of `skopeo inspect` are mostly inapplicable to artifacts, and there are no reasonable substitute values that could be used instead (e.g. for `architecture`), so just failing does seem appropriate right now.

OTOH, there are probably _some_ uses of `skopeo inspect` that make sense; at the very least “what is that thing at all?” is a question that should be possible to answer using `skopeo inspect` in some reasonably convenient way.

Compare #1283 . At this point it’s not clear to me what a good UI would look like, so I’m starting by looking for use cases.

mpagot:
[...]

Issue #1283, opened 2021-05-13
Skopeo inspect: provide --top-level about included architectures


pombredanne:
I would expect that `inspect` could return data from images even though they do not match my current OS/arch and without me first knowing about the image OS/Arch.
To me the whole purpose of inspect is that It should not require to know anything about the image OS or arch in the first place. 

I have integrated skopeo here to fetch images https://github.com/nexB/scancode.io/blob/62911d1fe692327a23af85fa9668344deb5b9ebd/scanpipe/pipes/fetch.py#L132 but I cannot get inspect to get me data on arbitrary docker:// references unless I use `inspect --raw` first to find out about the architecture and OS.

Using the latest 1.2.3 version from Linux, I cannot get inspect to work when running this:
```
$ skopeo inspect docker://mcr.microsoft.com/windows:20H2
FATA[0000] Error parsing manifest for image: Error choosing image instance: 
no image found in manifest list for architecture amd64, variant "", OS linux
```

Yet things work with `--override-os=windows`:
```
$ skopeo inspect  --override-os=windows docker://mcr.microsoft.com/windows:20H2 
{
    "Name": "mcr.microsoft.com/windows",
    "Digest": "sha256:c52e2549de5fbd32a83b6f4f43e471b79a111432ec3fcd2f605f2a1c0653bd29",
    "RepoTags": [
        "10.0.17763.1039",
        "10.0.17763.1039-amd64",
        "10.0.17763.1040",
        "10.0.17763.1040-amd64",
        "10.0.17763.1098",
        "10.0.17763.1098-amd64",
        "10.0.17763.1158",
```

and this:
```
$ skopeo inspect  --raw docker://mcr.microsoft.com/windows:20H2 
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 865,
         "digest": "sha256:01eaaa20c1ea39b30c5cbf0e24a528d02d7a23d0de316b0fc462cb6af029297f",
         "platform": {
            "architecture": "amd64",
            "os": "windows",
            "os.version": "10.0.19042.985"
         }
      }
   ]
}
```

vrothberg:
Thanks for reaching out!

Some background on the behavior:
[...]

PR #2600, opened 2025-05-12
[do-not-merge] Print as much as we can in 'inspect' even for non-image artifacts.

jankaluza:
This commit removes the check for MediaTypeImageConfig MediType from Inspect function. Instead, it lets the existing code to parse the OCI Config blob and print the ImageInspectInfo.

For non-image artifacts, this info can be incomplete, but still valid. If invalid config is found, the Unmarshal returns an error.

Resolves: #1933

### DIFF
--- a/hack/tree_status.sh
+++ b/hack/tree_status.sh
@@ -9,5 +9,6 @@ else
 	echo "tree is dirty, please commit all changes and sync the vendor.conf"
 	echo ""
 	echo "$STATUS"
-	exit 1
+	# TEMPORARY DISABLE FOR TESTING.
+	exit 0
 fi
--- a/vendor/github.com/containers/image/v5/manifest/oci.go
+++ b/vendor/github.com/containers/image/v5/manifest/oci.go
@@ -198,14 +198,6 @@ func (m *OCI1) Serialize() ([]byte, error) {
 
 // Inspect returns various information for (skopeo inspect) parsed from the manifest and configuration.
 func (m *OCI1) Inspect(configGetter func(types.BlobInfo) ([]byte, error)) (*types.ImageInspectInfo, error) {
-	if m.Config.MediaType != imgspecv1.MediaTypeImageConfig {
-		// We could return at least the layers, but that’s already available in a better format via types.Image.LayerInfos.
-		// Most software calling this without human intervention is going to expect the values to be realistic and relevant,
-		// and is probably better served by failing; we can always re-visit that later if we fail now, but
-		// if we started returning some data for OCI artifacts now, we couldn’t start failing in this function later.
-		return nil, manifest.NewNonImageArtifactError(&m.Manifest)
-	}
-
 	config, err := configGetter(m.ConfigInfo())
 	if err != nil {
 		return nil, err


jankaluza:
I will have to do this change in c/image. Just testing it here.

mtrmac:
Thanks!

> For non-image artifacts, this info can be incomplete, but still valid.

I don’t think that’s very likely; the fields are fairly image-specific.

And, fundamentally, it’s just not _valid_ to randomly read JSON fields from an entirely unknown data format and assume their meaning is consistent with images. I think such code is a forward-compatibility risk.

As is, out of the 12 items in current output, there are about 3 that are relevant to OCI artifacts _in general_, without specific support for a specific artifact MIME type.

---

The larger context of handling OCI artifacts with `skopeo inspect` is that the current output format is a historical accident and IMHO a dead end; if we are adding OCI artifacts, it should be in some other UI and output format. Consider:
- The current output, by default, includes a tag list. That’s entirely unrelated to the single image being inspected, it can be rather costly (for build systems which assign a tag to every build, many thousands of items and dozens of round-trips), and it’s very unlikely that users would want _both_ information about a single image and a tag list at exactly the same time.
- The current output mixes data about the whole possibly multi-platform image (e.g. the digest, and repo name [the repo name is pointless anyway]) and a single per-platform instance.
- Users probably need a way to ask “what is this? Multi-platform image / Single-platform image / artifact / (other???)”. There is no way to answer that now.
- The _very first_ question about an OCI artifact needs to be “what is the artifact’s MIME type?”
- There’s a long-standing need to list the platforms included in a multi-platform image; currently users do `--raw` and parse the OCI index because `inspect` does not provide the necessary format abstraction.

Hence I’m thinking about something like https://github.com/containers/skopeo/issues/1283 . I don’t have much of a design in mind, but I feel fairly confident there’s ~no value in adding support for OCI artifacts to the current format.

mtrmac:
> * Add support for inspecting whole Image Index without resolving it to specific Image Manifest. Still print the output in user-friendly format (one can use --raw to get the raw data).

(In general, I think of `--raw` as a last-resort escape mechanism. `skopeo inspect`, or some future variant of it, should expose the information that makes sense in a single stable format, to make it easy to write scripts, and without users having to understand the details of the underlying format (… much; the design assumptions always leak through abstractions).)

> * In case of unknown manifest.artifactType / manifes.config.mediaType, only show the artifactType / mediaType and suggest using the --raw?

Ultimately, users can either write their own consumer, or script something using `skopeo copy $src dir:…`, and then they can do whatever they want with the artifact. _Consuming_ the artifact is not a job of `skopeo inspect`.

> * Add a way to easily add new mediaType parsers for different artifacts in code.

Maybe? Generally Skopeo has been a _very thin_ wrapper over c/image, and Skopeo-specific features tend to suffer. What would the artifact-specific parser do?

(Well, there is the current Podman mechanism of `application/vnd.unknown.artifact.v1` where we could list the file names… which are otherwise visible in `LayersData`.)

> * Not proactively try to list the repoTags, since this is expensive call and might not be supported by all the registries (at least it seems like that from code). New skopeo command could be added to support that in the future?

`skopeo list-tags` already exists.

> I think something like that would definitely make the skopeo more predictable to use. I'm not sure how aligned is this plan with RUN-2712, but I like it more than changing the current "inspect" in the backward-incompatible way. What do you think?

I’d start with (some variant of / name of `inspect --top-level`) which reports _something sensible_ and forward-looking for all of (artifact, index, manifest), so that we can add features and fields to it later. We don’t need to report details of specific artifact types, or deep details of indices, in a first version, we just need to make it possible to add in the future.

For many artifacts, the only things we can reasonably report without building specific parsers is the MIME type, and config/layer blob details, and I think that’s perfectly fine. Anyone _consuming_ a specific MIME type will probably need a tool to do that, and that tool can also provide a specific “inspect” feature; that doesn’t strictly need to exist in Skopeo.

Skopeo does need to provide a way to answer “I have this repo:tag; what _is this thing_?”.

jankaluza:
> For many artifacts, the only things we can reasonably report without building specific parsers is the MIME type, and config/layer blob details, and I think that’s perfectly fine. Anyone consuming a specific MIME type will probably need a tool to do that, and that tool can also provide a specific “inspect” feature; that doesn’t strictly need to exist in Skopeo.

I'm working on that and submit it as a new PR. I'm closing this WIP PR without merging.

PR #2601, opened 2025-05-13
proxy: Add GetRawBlob

cgwalters:
The original model the idea here is the proxy centralizes verification of things like digest. However in practice, this causes reading to be seriously awkward; ref
https://github.com/containers/containers-image-proxy-rs/issues/79

Also, I have a project to implement a registry frontend to `containers-storage:` and a core problem with `GetBlob` right now is it *requires* the blob size up front even though the underlying Go logic doesn't.

Moving to a "raw" interface solves that.

### DIFF
--- a/cmd/skopeo/proxy.go
+++ b/cmd/skopeo/proxy.go
@@ -70,6 +70,7 @@ import (
 	"sync"
 	"syscall"
 
+	"github.com/containers/common/pkg/retry"
 	"github.com/containers/image/v5/image"
 	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/pkg/blobinfocache"
@@ -93,7 +94,8 @@ import (
 // 0.2.5: Added LayerInfoJSON
 // 0.2.6: Policy Verification before pulling OCI
 // 0.2.7: Added GetLayerInfoPiped
-const protocolVersion = "0.2.7"
+// 0.2.8: Added GetRawBlob and reply.error_code
+const protocolVersion = "0.2.8"
 
 // maxMsgSize is the current limit on a packet size.
 // Note that all non-metadata (i.e. payload data) is sent over a pipe.
@@ -116,6 +118,23 @@ type request struct {
 	Args []any `json:"args"`
 }
 
+type proxyErrorCode string
+
+const (
+	// proxyErrPipe means we got EPIPE writing to a pipe owned by the client
+	proxyErrPipe proxyErrorCode = "EPIPE"
+	// proxyErrRetryable can be used by clients to automatically retry operations
+	proxyErrRetryable proxyErrorCode = "retryable"
+	// All other errors
+	proxyErrOther proxyErrorCode = "other"
+)
+
+// proxyError is serialized over the errfd channel for GetRawBlob
+type proxyError struct {
+	Code    proxyErrorCode `json:"code"`
+	Message string         `json:"message"`
+}
+
 // reply is serialized to JSON as the return value from a function call.
 type reply struct {
 	// Success is true if and only if the call succeeded.
@@ -124,6 +143,8 @@ type reply struct {
 	Value any `json:"value"`
 	// PipeID is an index into open pipes, and should be passed to FinishPipe
 	PipeID uint32 `json:"pipeid"`
+	// ErrorCode will be non-empty if error is set (new in 0.2.8)
+	ErrorCode proxyErrorCode `json:"error_code"`
 	// Error should be non-empty if Success == false
 	Error string `json:"error"`
 }
@@ -132,8 +153,11 @@ type reply struct {
 type replyBuf struct {
 	// value will be converted to a reply Value
 	value any
-	// fd is the read half of a pipe, passed back to the client
+	// fd is the read half of a pipe, passed back to the client for additional data
 	fd *os.File
+	// errfd will be a serialization of error state. This is optional and is currently
+	// only used by GetRawBlob.
+	errfd *os.File
 	// pipeid will be provided to the client as PipeID, an index into our open pipes
 	pipeid uint32
 }
@@ -180,6 +204,30 @@ type convertedLayerInfo struct {
 	Digest    digest.Digest `json:"digest"`
 	Size      int64         `json:"size"`
 	MediaType string        `json:"media_type"`
+}
+
+// mapProxyErrorCode turns an error into a known string value.
+func mapProxyErrorCode(err error) proxyErrorCode {
+	switch {
+	case err == nil:
+		return ""
+	case errors.Is(err, syscall.EPIPE):
+		return proxyErrPipe
+	case retry.IsErrorRetryable(err):
+		return proxyErrRetryable
+	default:
+		return proxyErrOther
+	}
+}
+
+// newProxyError creates a serializable structure for
+// the client containing a mapped error code based
+// on the error type, plus its value as a string.
+func newProxyError(err error) proxyError {
+	return proxyError{
+		Code:    mapProxyErrorCode(err),
+		Message: fmt.Sprintf("%v", err),
+	}
 }
 
 // Initialize performs one-time initialization, and returns the protocol version
@@ -617,6 +665,91 @@ func (h *proxyHandler) GetBlob(args []any) (replyBuf, error) {
 	return ret, nil
 }
 
+// GetRawBlob can be viewed as a more general purpose successor
+// to GetBlob. First, it does not verify the digest, which in
+// some cases is unnecessary as the client would prefer to do it.
+//
+// It also does not use the "FinishPipe" API call, but instead
+// returns *two* file descriptors, one for errors and one for data.
+//
+// On (initial) success, the return value provided to the client is the size of the blob.
+func (h *proxyHandler) GetRawBlob(args []any) (replyBuf, error) {
+	h.lock.Lock()
+	defer h.lock.Unlock()
+
+	var ret replyBuf
+
+	if h.sysctx == nil {
+		return ret, fmt.Errorf("client error: must invoke Initialize")
+	}
+	if len(args) != 2 {
+		return ret, fmt.Errorf("found %d args, expecting (imgid, digest)", len(args))
+	}
+	imgref, err := h.parseImageFromID(args[0])
+	if err != nil {
+		return ret, err
+	}
+	digestStr, ok := args[1].(string)
+	if !ok {
+		return ret, fmt.Errorf("expecting string blobid")
+	}
+
+	ctx := context.TODO()
+	d, err := digest.Parse(digestStr)
+	if err != nil {
+		return ret, err
+	}
+	blobr, blobSize, err := imgref.src.GetBlob(ctx, types.BlobInfo{Digest: d, Size: int64(-1)}, h.cache)
+	if err != nil {
+		return ret, err
+	}
+
+	// Note this doesn't call allocPipe; we're not using the FinishPipe infrastructure.
+	piper, pipew, err := os.Pipe()
+	if err != nil {
+		blobr.Close()
+		return ret, err
+	}
+	errpipeR, errpipeW, err := os.Pipe()
+	if err != nil {
+		piper.Close()
+		pipew.Close()
+		blobr.Close()
+		return ret, err
+	}
+	// Asynchronous worker doing a copy
+	go func() {
+		// We own the read from registry, and write pipe objects
+		defer blobr.Close()
+		defer pipew.Close()
+		defer errpipeW.Close()
+		logrus.Debugf("Copying blob to client: %d bytes", blobSize)
+		_, err := io.Copy(pipew, blobr)
+		// Handle errors here by serializing a JSON error back over
+		// the error channel. In either case, both file descriptors
+		// will be closed, signaling the completion of the operation.
+		if err != nil {
+			logrus.Debugf("Sending error to client: %v", err)
+			serializedErr := newProxyError(err)
+			buf, err := json.Marshal(serializedErr)
+			if err != nil {
+				// Should never happen
+				panic(err)
+			}
+			_, writeErr := errpipeW.Write(buf)
+			if writeErr != nil && !errors.Is(err, syscall.EPIPE) {
+				logrus.Debugf("Writing to client: %v", err)
+			}
+		}
+		logrus.Debugf("Completed GetRawBlob operation")
+	}()
+
+	ret.value = blobSize
+	ret.fd = piper
+	ret.errfd = errpipeR
+	return ret, nil
+}
+
 // GetLayerInfo returns data about the layers of an image, useful for reading the layer contents.
 //
 // This is the same as GetLayerInfoPiped, but returns its contents inline. This is subject to
@@ -763,29 +896,36 @@ func (h *proxyHandler) close() {
 
 // send writes a reply buffer to the socket
 func (buf replyBuf) send(conn *net.UnixConn, err error) error {
-	logrus.Debugf("Sending reply: err=%v value=%v pipeid=%v", err, buf.value, buf.pipeid)
+	logrus.Debugf("Sending reply: err=%v value=%v pipeid=%v datafd=%v errfd=%v", err, buf.value, buf.pipeid, buf.fd, buf.errfd)
 	replyToSerialize := reply{
 		Success: err == nil,
 		Value:   buf.value,
 		PipeID:  buf.pipeid,
 	}
 	if err != nil {
+		replyToSerialize.ErrorCode = mapProxyErrorCode(err)
 		replyToSerialize.Error = err.Error()
 	}
 	serializedReply, err := json.Marshal(&replyToSerialize)
 	if err != nil {
 		return err
 	}
-	// We took ownership of the FD - close it when we're done.
+	// We took ownership of these FDs, so close when we're done sending them or on error
 	defer func() {
 		if buf.fd != nil {
 			buf.fd.Close()
 		}
+		if buf.errfd != nil {
+			buf.errfd.Close()
+		}
 	}()
-	// Copy the FD number to the socket ancillary buffer
+	// Copy the FD number(s) to the socket ancillary buffer
 	fds := make([]int, 0)
 	if buf.fd != nil {
 		fds = append(fds, int(buf.fd.Fd()))
+	}
+	if buf.errfd != nil {
+		fds = append(fds, int(buf.errfd.Fd()))
 	}
 	oob := syscall.UnixRights(fds...)
 	n, oobn, err := conn.WriteMsgUnix(serializedReply, oob, nil)
@@ -858,6 +998,8 @@ func (h *proxyHandler) processRequest(readBytes []byte) (rb replyBuf, terminate 
 		rb, err = h.GetFullConfig(req.Args)
 	case "GetBlob":
 		rb, err = h.GetBlob(req.Args)
+	case "GetRawBlob":
+		rb, err = h.GetRawBlob(req.Args)
 	case "GetLayerInfo":
 		rb, err = h.GetLayerInfo(req.Args)
 	case "GetLayerInfoPiped":
--- a/cmd/skopeo/proxy.go
+++ b/cmd/skopeo/proxy.go
@@ -897,6 +897,15 @@ func (h *proxyHandler) close() {
 // send writes a reply buffer to the socket
 func (buf replyBuf) send(conn *net.UnixConn, err error) error {
 	logrus.Debugf("Sending reply: err=%v value=%v pipeid=%v datafd=%v errfd=%v", err, buf.value, buf.pipeid, buf.fd, buf.errfd)
+	// We took ownership of these FDs, so close when we're done sending them or on error
+	defer func() {
+		if buf.fd != nil {
+			buf.fd.Close()
+		}
+		if buf.errfd != nil {
+			buf.errfd.Close()
+		}
+	}()
 	replyToSerialize := reply{
 		Success: err == nil,
 		Value:   buf.value,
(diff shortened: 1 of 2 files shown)


cgwalters:
Draft since this works well for me, but needs tests

cgwalters:
Forgot to mention in this PR but this is a dependency of https://github.com/cgwalters/cstor-dist which you may also be interested in

cgwalters:
> It seems to me that the Rust client side can trivially call FinishPipe at the place where it would do the digest verification, 

Sorry the issue with more real discussion is https://github.com/containers/containers-image-proxy-rs/issues/80 which TL;DR is that it's a problem that `FinishPipe` blocks the control channel.

Which we could definitely also fix without `GetBlobRaw` indeed.

> If nothing else, after removing the error reporting mechanism a consumer won’t be able to differentiate between a (retryable) network error and a digest validation error.

Yes it'd be really nice to handle that indeed. I guess a possible design here would be `GetBlobFramed` which returns
```
#[derive(Serialize)]
enum Message {
  Data(&[u8])
  Error(...)
}
```
or so? (Framed in something like [bincode](https://github.com/bincode-org/bincode) since json is pretty suboptimal for binary data)

I guess this whole thing quickly goes back to less-ad-hoc IPC frameworks.

mtrmac:
> [containers/containers-image-proxy-rs#80](https://github.com/containers/containers-image-proxy-rs/issues/80) which TL;DR is that it's a problem that `FinishPipe` blocks the control channel.

Do I understand it correctly that there is thread A consuming the blob, and thread B calling `FinishPipe` _before_ thread A is done reading? Yes, that would stall for a potentially very long time, but, also, _if_ that is the design, why doesn’t A call `FinishPipe` itself? Having B receive the digest verification / network error and then somehow communicate it back to A looks needlessly complicated to me.

After A finishes consuming data, `FinishPipe` should be a metadata operation. Sure, there is some context switching and it’s probably a few $something thousand instructions, but it should be trivial compared to any network or disk cost.

(Even with a design where the proxy is all “owned” by a single proxy client thread, making it undesirable for thread A to make requests directly, having A communicate to the proxy client and block on a reply might not be that much worse? It does serialize everything at the proxy client, but that would be ~equivalent in effect to the serialization that happens in the proxy server anyway.)

---

There is a potentially real issue here in that `FinishPipe` serializes against some other thread C’s `GetBlob` _initiating_ a download, i.e. network round-trips required to set up TLS and do a `GET` request, before the body starts arriving. That would be _some_ delay but, on the whole, I would also expect it to not be significant — this is not interactive UI with a 50 ms deadline, and the critical path will probably include this thread C with its slow `GET` as well as consuming all of that blob afterwards: before C gets to `FinishPipe`, A will have ample opportunity to perform its own `FinishPipe`, so A is not the slowest part.
 
---

> Yes it'd be really nice to handle that indeed. I guess a possible design here would be `GetBlobFramed`

Yes.

> I guess this whole thing quickly goes back to less-ad-hoc IPC frameworks.

And that too.

cgwalters:
> It seems to me that the Rust client side can trivially call FinishPipe at the place where it would do the digest verification, what’s the difficulty?! Some kind of Rust-related reference counting making it hard to hold on to a proxy handle at that time?

Yeah I think you're right. We could probably just automatically handle this in the rust client code. So it's not a strong argument for `GetBlobRaw` in and of itself.

> And if FinishPipe remained, I don’t see much of a reason to prefer the digest verification happening at one side or the other, 

Note for the use case of cstor-dist we're going:
`containers-storage:` -> `cstor-dist` (registry) -> `containers-storage:|ostree|composefs-rs`

And it'd be silly to have cstor-dist also do checksum verification because we know the client is going to do it. So probably regardless IMO some sort of `GetBlobRaw` makes sense right?

> except perhaps that sha512 digests are coming, and that will be transparently handled with the Go code as is.

I think any nontrivial consumer of the proxy is going to be parsing the manifest digests and doing things with them. The ostree and composefs-rs projects definitely are and so there's no magic adaptation.

cgwalters:
Anyways though, the other option is to just cut the size requirement for `GetBlob`, that would be a pretty straightforward patch too. I may look at that.

mtrmac:
> > And it'd be silly to have cstor-dist also do checksum verification because we know the client is going to do it. So probably regardless IMO some sort of `GetBlobRaw` makes sense right?
> 
> Yes, that’s a good enough reason.

… to allow the client to opt out of digest verification, _somehow_. I’m not too convinced about removing the error handling entirely.

cgwalters:
> I’m not too convinced about removing the error handling entirely.

But in the raw case the error handling is literally just whether or not the proxy successfully wrote to the pipe, right? Which, most often is actually an `EPIPE` because the client closed the pipe because it found some other error, but that then boomerangs back to the client! We have some hacks here https://github.com/bootc-dev/bootc/blob/c1d67aa07c3408694b46b74deb7be10587b70fe7/ostree-ext/src/container/unencapsulate.rs#L151

Which, the only way to fix that sanely I think is to have a new API creatively called `GetBlob2`. So putting several things together:

- Add GetBlob2 DIGEST [verify: bool]

It returns...umm, I am vacillating on what the format should be. I tried hard so far to only make this thing about JSON and `pipe()` but it's tricky to keep doing it. I guess one option is that it returns *two* pipe read fds, one which returns JSON only in the case of error, and the other which is the raw data stream? The other option is BSON or again going into a bigger RPC framework like capn'proto or whatever but for now I think my preference is continuing to avoid that.

So what gets sent over the error fd would be at most one of:

```
struct Error {
  code: String,
  message: String
}
```

But we'd need to bikeshed how `code` appears, I am actually unfamiliar with the best practices here. Error mapping cross-language IPC is in my experience a hard problem. One possibility is to squash everything into one of say four cases:

- `network`
- `EPIPE`
- `checksum`
- `other`

(And yes we should probably emit an error if we get `EPIPE`, but the client can know to safely ignore it)

mtrmac:
Just a note on this part for now:

> > I’m not too convinced about removing the error handling entirely.
> 
> But in the raw case the error handling is literally just whether or not the proxy successfully wrote to the pipe, right?

On an error reading from the registry, it’s the full information the Go server side has: “connection reset / unexpected EOF / `Transfer-Encoding: Chunked` framing invalid / timed out / completely unexpected EPERM / whatever”. _Maybe_ the ~expected network errors should be handled / retried automatically. I think it would be very valuable for the 100% unexpected errors to be propagated to the Rust client and ultimately to a human. For that, we need, at the very least, to transfer an error string.

`EPIPE` is an interesting case that needs special handling, sure, but I don’t think it’s the only one.

---

> It returns...umm, I am vacillating on what the format should be. I tried hard so far to only make this thing about JSON and `pipe()` but it's tricky to keep doing it. I guess one option is that it returns _two_ pipe read fds, one which returns JSON only in the case of error, and the other which is the raw data stream?

Note to self: Return here. We’ve had `GetBlobAt` returning a separate data stream and an error stream, and it was very inconvenient to consume — but that might not be applicable here.

---

> ```
> struct Error {
>   code: String,
>   message: String
> }
> ```
> 
> But we'd need to bikeshed how `code` appears, I am actually unfamiliar with the best practices here.

> Error mapping cross-language IPC is in my experience a hard problem.

It’s hard, in the corner cases, even within a single-language process :)

There’s ample space to over-engineer. I think the above is a good start, and we can always add `code`-specific fields (turning this into an ~tagged union).

The extreme is to adopt something like https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties , where the error can have an open-ended list of _non-exclusive_ bools (“network-related”? “retryable”? “timeout”? All 3 could be true simultaneously). I think that’s overkill.

---

I think it would be convenient to use the same error mechanism for the current `reply`. Not blocking, but it might simplify consumers.

mtrmac:
> > It returns...umm, I am vacillating on what the format should be. I tried hard so far to only make this thing about JSON and `pipe()` but it's tricky to keep doing it. I guess one option is that it returns _two_ pipe read fds, one which returns JSON only in the case of error, and the other which is the raw data stream?
> 
> Note to self: Return here. We’ve had `GetBlobAt` returning a separate data stream and an error stream, and it was very inconvenient to consume — but that might not be applicable here.

Ref. https://github.com/containers/storage/pull/2162 : If there is a `dataFD` and `errorFD`, and the producer writes an error to `errorFD` and closes both, the consumer must, after seeing EOF on `dataFD`, _still_ read `errorFD`, because the kernel does not guarantee a `poll` or similar will see the error incoming on `errorFD` before the EOF on `dataFD` (and user-space poll / event loop abstractions probably don’t guarantee ordering either).

A single pipe is easier to consume and reason about.

cgwalters:
> Ref. https://github.com/containers/storage/pull/2162 : If there is a dataFD and errorFD, and the producer writes an error to errorFD and closes both, the consumer must, after seeing EOF on dataFD, still read errorFD, because the kernel does not guarantee a poll or similar will see the error incoming on errorFD before the EOF on dataFD (and user-space poll / event loop abstractions probably don’t guarantee ordering either).

Right, but that's easy? I mean at least in Rust I'd map this as two futures which both need to be joined, not a `select` as happened in that c/storage code originally. This is only a little tangential but probably interesting nonetheless; `tokio::select!` in Rust has a [giant trap](https://tomaka.medium.com/a-look-back-at-asynchronous-rust-d54d63934a1c) and always needs careful auditing and people talk about "cancel safety" because of this. (TL;DR Go uses stackful coroutines, Rust's futures are stackless and precisely sized, but the lack of stack means you can get burned by cancellation).

> One possible way to think about this is that consuming two pipes requires the caller to always read the error pipe until EOF; that’s probably not any easier than, in the current API, calling FinishPipe after reaching EOF on the data stream.

Mmmmm...except the problem again with today's API is that the client *doesn't get EOF* until it calls `FinishPipe`. The rationale for that is it forces callers to invoke `FinishPipe` (to check for errors), and the workaround we did for this is basically "read only up until the known blob size" which is viable but definitely not ergonomic or obvious at all for callers.

mtrmac:
> Mmmmm...except the problem again with today's API is that the client _doesn't get EOF_ until it calls `FinishPipe`.

Oops, I have _completely missed_ that when discussing this PR. That makes many of the things I suggested above, thinking that they are simple, rather less practical.

> the workaround we did for this is basically "read only up until the known blob size" which is viable 

[`GetBlob` is allowed to return -1 as blob size, and that happens sometimes.]

… does error reporting work _at all_ if there is a network error in the middle of a blob? The goroutine in `GetBlob` will see an error, and terminate, but the write end of the pipe will be kept open. The client will be stuck in `read` on a pipe, never see an EOF, and it will never have a reason to call `FinishPipe`.

Am I missing something again? If it does fail this way, then yes, we definitely need a new `GetBlob` (and everything else that uses pipes??).

cgwalters:
> [GetBlob is allowed to return -1 as blob size, and that happens sometimes.]

That doesn't matter though because we always know the expected size from the manifest descriptor and that's what clients should use, right?

> … does error reporting work at all if there is a network error in the middle of a blob? The goroutine in GetBlob will see an error, and terminate, but the write end of the pipe will be kept open. The client will be stuck in read on a pipe, never see an EOF, and it will never have a reason to call FinishPipe.

Well, the way this is working right now is that the calling process in Rust creates two futures, one doing the read/processing and one calling `FinishPipe`. So in the case of network error we'll get an error from the latter. But that brings the concurrency problem that Allison was talking about where any `FinishPipe` blocks metadata stream and so...yeah it's a huge pain in the butt to in practice use this to do concurrent fetches, even though it was designed to do so.

It'd be really quite ugly but I suspect a viable hack for the status quo would be to only queue the FinishPipe on a timeout if we stop getting data from the pipe or so (and that happens before the expected size)

mtrmac:
> we always know the expected size from the manifest descriptor 

schema1 manifests don’t have sizes. (OTOH at least Docker has been working for years to disable schema1 manifests, e.g. `distribution/distribution` only accepts them with an opt-in.)

cgwalters:
> schema1 manifests don’t have sizes. 

[We hard error on that case](https://github.com/containers/skopeo/blob/a4770636501c2dfe76a8ec503511260f79979261/cmd/skopeo/proxy.go#L441) today already.

I honestly would actually like to add a `Open` method that hard requires OCI even - for the use cases I care about in bootc, there's simply no reason to continue to support v2s2.

cgwalters:
OK I reworked this per discussion and also updated the example consumer in https://github.com/containers/containers-image-proxy-rs/pull/83 - just got as far as verifying the example client there works. Could still use tests here etc.

cmd/skopeo/proxy.go
```diff
@@ -93,7 +94,8 @@ import (
 // 0.2.5: Added LayerInfoJSON
 // 0.2.6: Policy Verification before pulling OCI
 // 0.2.7: Added GetLayerInfoPiped
-const protocolVersion = "0.2.7"
+// 0.2.8: Added GetRawBlob
```
mtrmac, line 97: … and `reply.ErrorCode`

cmd/skopeo/proxy.go
```diff
@@ -116,6 +118,21 @@ type request struct {
...
 }
 
+// Set if we caught EPIPE writing to the client
+const PROXY_ERR_PIPE = "EPIPE"
+
+// Errors that the client believes are retryable (e.g. network flakes)
+const PROXY_ERR_RETRYABLE = "retryable"
+
+// Any other error
+const PROXY_ERR_UNKNOWN = "unknown"
```
mtrmac, line 128:
```go
type proxyErrorCode string
const (
   proxyErrPipe proxyErrorCode = "EPIPE"
   proxyErrRetryable = "retryable"
   …
)
```
and then using a named type in `struct` fields / return values etc. would add a bit of type safety.

cmd/skopeo/proxy.go
```diff
@@ -182,6 +204,30 @@ type convertedLayerInfo struct {
 	MediaType string        `json:"media_type"`
 }
 
+// mapProxyErrorCode turns an error into a known string value.
+func mapProxyErrorCode(err error) string {
+	if err == nil {
```
mtrmac, line 209: (Absolutely non-blocking: Using a Go `switch` would be a tiny bit shorter. This works just fine.)

cmd/skopeo/proxy.go
```diff
@@ -617,6 +663,86 @@ func (h *proxyHandler) GetBlob(args []any) (replyBuf, error) {
...
+
+	// Note this doesn't call allocPipe; we're not using the FinishPipe infrastructure.
+	piper, pipew, err := os.Pipe()
+	if err != nil {
+		blobr.Close()
+		return ret, err
+	}
+	errpipeR, errpipeW, err := os.Pipe()
+	if err != nil {
+		blobr.Close()
```
mtrmac, line 713: Close `pipe*` here.

(Alternatively,`defer func() { if !succeeded { /* close things */ } }` and set `succeeded` after the goroutine starts.)

cmd/skopeo/proxy.go
```diff
@@ -617,6 +663,86 @@ func (h *proxyHandler) GetBlob(args []any) (replyBuf, error) {
...
+		return ret, err
+	}
+
+	// Note this doesn't call allocPipe; we're not using the FinishPipe infrastructure.
+	piper, pipew, err := os.Pipe()
+	if err != nil {
+		blobr.Close()
+		return ret, err
+	}
+	errpipeR, errpipeW, err := os.Pipe()
```
mtrmac, line 711: (`pipeR`, `pipeW`, `errPipeR`, `errPipeW` would be a tiny bit more idiomatic in Go. OTOH all-lowercase is widespread in this file.)

cmd/skopeo/proxy.go
```diff
@@ -617,6 +663,86 @@ func (h *proxyHandler) GetBlob(args []any) (replyBuf, error) {
...
+		_, err := io.Copy(pipew, blobr)
+		// Handle errors here by serializing a JSON error back over
+		// the error channel. In either case, both file descriptors
+		// will be closed, signaling the completion of the operation.
+		if err != nil {
+			serializedErr := newProxyError(err)
+			buf, err := json.Marshal(serializedErr)
+			if err != nil {
+				// Should never happen
+				panic(err)
```
mtrmac, line 731: … \*shrug\*

cmd/skopeo/proxy.go
```diff
@@ -770,23 +896,31 @@ func (buf replyBuf) send(conn *net.UnixConn, err error) error {
 		PipeID:  buf.pipeid,
 	}
 	if err != nil {
+		replyToSerialize.ErrorCode = mapProxyErrorCode(err)
 		replyToSerialize.Error = err.Error()
+
```
mtrmac, line 901: Non-blocking: Please drop this empty line.

(Also, setting `Error` and `ErrorCode` in the declaration order would be a tiny bit easier to check. I don’t care whether the declaration or this code is adjusted.)

cmd/skopeo/proxy.go
```diff
@@ -770,23 +896,31 @@ func (buf replyBuf) send(conn *net.UnixConn, err error) error {
...
 	}
 	serializedReply, err := json.Marshal(&replyToSerialize)
 	if err != nil {
 		return err
 	}
-	// We took ownership of the FD - close it when we're done.
+	// We took ownership of these FDs, so close when we're done sending them or on error
 	defer func() {
```
mtrmac, line 908: (Non-blocking, this is pre-existing: this `defer` should happen before we exit on the failure above.)

mtrmac (commented):
Thanks!

ACK overall. (Note also the lint failures.)

cmd/skopeo/proxy.go
cgwalters: Thanks, I will fix that in a separate PR for clarity

cgwalters: Eh that'd be a conflict, so added another commit here

cmd/skopeo/proxy.go
```diff
@@ -124,6 +143,8 @@ type reply struct {
 	Value any `json:"value"`
 	// PipeID is an index into open pipes, and should be passed to FinishPipe
 	PipeID uint32 `json:"pipeid"`
+	// ErrorCode will be non-empty if error is set (new in 0.2.8)
+	ErrorCode string `json:"error_code"`
```
mtrmac, line 147:
```suggestion
	ErrorCode proxyErrorCode `json:"error_code"`
```

cmd/skopeo/proxy.go
```diff
@@ -182,6 +206,30 @@ type convertedLayerInfo struct {
 	MediaType string        `json:"media_type"`
 }
 
+// mapProxyErrorCode turns an error into a known string value.
+func mapProxyErrorCode(err error) string {
```
mtrmac, line 210:
```suggestion
func mapProxyErrorCode(err error) proxyErrorCode {
```

cmd/skopeo/proxy.go
```diff
@@ -182,6 +206,30 @@ type convertedLayerInfo struct {
...
+	switch {
+	case err == nil:
+		return ""
+	case errors.Is(err, syscall.EPIPE):
+		return string(proxyErrPipe)
+	case retry.IsErrorRetryable(err):
+		return proxyErrRetryable
```
mtrmac, line 217: The difference here suggests that the `proxyErrorCode` type needs to be explicitly used for all of the enum values.

cmd/skopeo/proxy.go
```diff
@@ -116,6 +118,23 @@ type request struct {
...
+	proxyErrPipe proxyErrorCode = "EPIPE"
+	// proxyErrRetryable can be used by clients to automatically retry operations
+	proxyErrRetryable = "retryable"
+	// All other errors
+	proxyErrOther = "other"
+)
+
+// proxyError is serialized over the errfd channel for GetRawBlob
+type proxyError struct {
+	Code    string `json:"code"`
```
mtrmac, line 134:
```suggestion
	Code    proxyErrorCode `json:"code"`
```

cmd/skopeo/proxy.go
```diff
@@ -617,6 +665,92 @@ func (h *proxyHandler) GetBlob(args []any) (replyBuf, error) {
...
+		// will be closed, signaling the completion of the operation.
+		if err != nil {
+			logrus.Debugf("Sending error to client: %v", err)
+			serializedErr := newProxyError(err)
+			buf, err := json.Marshal(serializedErr)
+			if err != nil {
+				// Should never happen
+				panic(err)
+			}
+			// write all of buf to errpipew
```
mtrmac, line 739: (A single `.Write` should actually be sufficient
>  Write must return a non-nil error if it returns n < len(p)

the current version is also fine.)

cgwalters: Thanks. In Rust the `write()` trait is the same as the classic write syscall ([and it's the source of numerous bugs](https://internals.rust-lang.org/t/could-we-ever-fix-naming-of-io-write/22226)); here I had a different bug that caused me to think there were short writes going on here.

cmd/skopeo/proxy.go
cgwalters: Yes, thanks for that!

cgwalters:
(It'd sure be nice if github had a "commit as fixup" button)

cmd/skopeo/proxy.go
```diff
@@ -116,6 +118,23 @@ type request struct {
...
+type proxyErrorCode string
+
+const (
+	// proxyErrPipe means we got EPIPE writing to a pipe owned by the client
+	proxyErrPipe proxyErrorCode = "EPIPE"
+	// proxyErrRetryable can be used by clients to automatically retry operations
+	proxyErrRetryable = "retryable"
+	// All other errors
+	proxyErrOther = "other"
+)
```
mtrmac, line 130:
```suggestion
const (
	// proxyErrPipe means we got EPIPE writing to a pipe owned by the client
	proxyErrPipe proxyErrorCode = "EPIPE"
	// proxyErrRetryable can be used by clients to automatically retry operations
	proxyErrRetryable proxyErrorCode = "retryable"
	// All other errors
	proxyErrOther proxyErrorCode = "other"
)
```

cgwalters: Ah sorry I'd missed that. Thanks, squashed in!

mtrmac (approved):
Thanks!

PR #2604, opened 2025-05-15
Bump to c/common 0.63.0, c/image 5.35.0, Skopeo to v1.19.0, v1.20.0-dev

TomSweeneyRedHat:
This bumps c/common and c/image to v0.63.0 and v5.35.0, respectively.  It then bumps Skopeo to v1.19.0, and then to the next dev version, v1.20.0-dev.

c/storage has already been bumped to v1.58.0, with this PR, we will line up with Podman v5.5.0

There are way more updates from the c/common and c/image bumps than I anticipated.....

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -1,4 +1,4 @@
 package version
 
 // Version is the version of the build.
-const Version = "1.19.0-dev"
+const Version = "1.19.0"
--- a/version/version.go
+++ b/version/version.go
@@ -1,4 +1,4 @@
 package version
 
 // Version is the version of the build.
-const Version = "1.19.0"
+const Version = "1.20.0-dev"


go.mod
```diff
@@ -1,14 +1,16 @@
 module github.com/containers/skopeo
 
 // Minimum required golang version
-go 1.23.0
+go 1.23.3
+
+toolchain go1.23.9
```
mtrmac, line 6: I think downstreams would prefer this line to be dropped.

(Even the 1.23.3 requirement causes build failures, I assume that has also happened for Podman and we have a plan for that.)

TomSweeneyRedHat: good thought, let me go tweak that.

TomSweeneyRedHat:
c/common is requiring a bump to Go 1.20.3.  I wasn't able to get around that, but was able to remove the toolchain line which is generally problematic.

TomSweeneyRedHat:
I tried bumping the go.mod to just have go 1.23.3 as Podman does, but the go vendoring  insisted on a toolchain statement being added for 1.23.1 which messed things up in here.  I tried removing it, and now adding the toolching at 1.23.3, and still no go. 

@Luap99 or @lsm5 any thoughts?

Luap99:
You likely should apply the toochain none workaround we have in podman https://github.com/containers/podman/commit/0acd192b5936e7161944d1c91f081148758fba10.

Functionally the toolchain line must >= go version. So it can only be used to require an even newer go than actually required per go to build. As such this toolchain line has no purpose for us because we never want to arbitrarily bump the go version to exactly prevent these situations where distros cannot build because of it.
You can always manually remove the the toolchain line (after you updated go.mod) as go will only add it when it upgrades the go version. I think they removed that behavior recently, not sure if it was in go 1.24 already or 1.25. So hopefully that should not cause us so much pain in the future.

Now the real issue is that we have no direct control of what the go version is as it uses the newest version of all dependencies in this case go 1.23.3 is now required. Why anyone would depend on these patch versions specifically is way beyond me and I have been ranting about this IMO anti-feature ever since go added that ability.

Luap99:
(I am not sure if I didn't listen but I never heard anyone mention a new skopeo release)

Packaging wise it would be nice to have this released with or before podman. We use sidetags in fedora to bundle the common/buildah/podman/skopeo updates into one big bodhi update. So right now buildah had to wait almost a month since we were waiting on the final podman release to get this out. We have the podman updates set up for f42/41 yesterday that then need 3 karma or 14 days until they can shipped in stable. Adding the skopeo release later will always reset karma/stable days so it will delay the releases to fedora even further.
And given our container images build of stable fedora packages this always delays them as well.

mtrmac:
> You can always manually remove the the toolchain line (after you updated go.mod)

Yes; for me, `go mod tidy` does that if `toolchain` == `go`, and the CI seems to want that to be done.

---

> Why anyone would depend on these patch versions specifically

These situations are months-old, e.g. https://github.com/sigstore/fulcio/issues/1899 . The upstreams learned about the issue and got a bit better in the meantime (never mind that recent Fulcio is on 1.24…)

mtrmac:
Note to self: Trying the `go.mod` / `toolchain` handling, and CI impact, in https://github.com/containers/skopeo/pull/2607 .

mtrmac:
@TomSweeneyRedHat the dependency updates / `toolchain`/ CI updates are done on `main`, so we need just the version bump.

Do you want to rebase this PR, or would you prefer me to prepare it in a new one?

TomSweeneyRedHat:
Thanks for the heads up.  I'd like to put this one to bed and I'll go rebase in a few.

TomSweeneyRedHat:
@mtrmac repushed, fingers crossed for happy tests.

mtrmac (approved):
Thanks!

PR #2606, opened 2025-05-19
[skip-ci] Packit: set fedora-all after F40 EOL

lsm5:
F40 is now EOL and all current Fedora releases have Go 1.23. So, we're safe to re-enable fedora-all.


mtrmac (approved):
Thank you!

PR #2609, opened 2025-05-26
Add --retry to test operations that tend to fail

mtrmac:
... to decrease impact of flakes.

---

We now see a network error at least once in most PRs in this repo, it’s becoming too much of a hassle.

I intend to keep expanding the scope of this PR as long as CI keeps showing new classes of avoidable failures.

### DIFF
--- a/contrib/cirrus/runner.sh
+++ b/contrib/cirrus/runner.sh
@@ -64,7 +64,7 @@ _run_setup() {
     # things directly on the host VM.  Fortunately they're all
     # located in the container under /usr/local/bin
     msg "Accessing contents of $SKOPEO_CIDEV_CONTAINER_FQIN"
-    podman pull --quiet $SKOPEO_CIDEV_CONTAINER_FQIN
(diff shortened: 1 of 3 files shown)


mtrmac:
The Skopeo tests also serve as the primary tests for the c/image/docker code, so exercising that
- without mirrors set up
- potentially, to contact some real-world registries (although we don’t test `docker.io`, despite its important position in the ecosystem)

is valuable.

I agree that _most_ tests in the repo could run against a local server, but that would require a more detailed analysis. This is just a dumb simple band-aid to immediately improve the developer experience.

cgwalters (approved):

Luap99 (commented):
LGTM, it certainly helps and doing the local registry stuff is more complicated to get right and I see you have tmt tests there it would not help you either.

IMO flakes are just one problem the other is that the network resource can chnage any time and then break the tests, then you get to update main + all active branches, etc...
If we only have a set of local images than will not change this cannot happen.

mtrmac:
GitHub branch protection requires a formal approval. Thanks!

PR #2612, opened 2025-05-28
[release-1.9] Support CI testing on non-x86_64

cevich:
Previously, internal CI gating tests sometimes fail because the required registry container image only supports x86_64.  Update to the `2.8.2` image tag with support for all primary architectures.


cevich:
For the record, this matches main’s https://github.com/containers/skopeo/commit/b5a13bccfddd7eecdb4d046d3abc7519cf0ade0c .

PR #2570, opened 2025-04-09
Add MAINTAINERS.md and Governance, update OWNERS

mheon:
Many of the people in OWNERS no longer contribute to the project, so clean the file and restrict to those who are still active. Alongside this, add our core maintainers who have merge authority on all repos.

Governance is a simple link to the Podman governance model, and MAINTAINERS.md mirrors the new OWNERS.

PR #2613, opened 2025-05-28
Fix recognizing "tag not found" in oci:, and add that for oci-archive:

mtrmac:
It turns out the `oci:` “tag not found” handling in #1757 never worked (it used a value equality check instead of a type conformance check). Fix that, and add a unit test.

Also, implement the same semantics for `oci-archive:`, per #2114. Only manually tested.

Unlike #2114, this does not add “exit status 2” for “the whole archive is missing”. I don’t think that needs special handling, and it seems no caller currently cares.


cgwalters:
BTW it looks like I lost my "official reviewer" status on this repo, do you want to readd it?

mtrmac:
> Out of curiosity how did you find this?

I added `oci-archive:` using the same code pattern, and in manual testing it turned out not to work.

---

> BTW it looks like I lost my "official reviewer" status on this repo, do you want to readd it?

That might have happened while formalizing the rules (for CNCF, around #2570) due to a disconnect between GitHub permissions and text records… and now there’s a bit of a process involved. I’ll ask around.

lsm5:
@mtrmac do we need any systemtest updates here?

mtrmac:
@lsm5 Specifically _system_ tests? This is upstream-motivated, not tracked in a downstream product, so I don’t think there’s a reason to have a system test and not an integration tests (but there could well be something I’m missing).

Arguably the `oci-archive:` code path should have an automated test, it doesn’t have one now.

PR #2618, opened 2025-05-30
Don't BuildRequires: ostree-devel

mtrmac:
We are not opting into the ostree backend, and it doesn't build: https://github.com/containers/image/pull/2821 . So, stop referencing the dependency.

Should not change behavior.

@lsm5 @jnovy PTAL.

### DIFF
--- a/rpm/skopeo.spec
+++ b/rpm/skopeo.spec
@@ -63,7 +63,6 @@ BuildRequires: go-rpm-macros
 %endif
 BuildRequires: gpgme-devel
 BuildRequires: libassuan-devel
-BuildRequires: ostree-devel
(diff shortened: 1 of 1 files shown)


cgwalters (approved):

rpm/skopeo.spec
```diff
@@ -63,7 +63,6 @@ BuildRequires: go-rpm-macros
 %endif
 BuildRequires: gpgme-devel
 BuildRequires: libassuan-devel
-BuildRequires: ostree-devel
```
jnovy, line 66: LGTM

lsm5 (approved):
LGTM

PR #2619, opened 2025-06-03
Move RHEL branches to TMT

lsm5:
**Commit 1:**
Advantages:
    - Test RHEL branches on on actual RHEL
    - Reduce our GCE/AWS usage
    - Easier to switch git host (RE: CNCF)
    
The spec file is fetched from CentOS Stream as it's publicly available and the RHEL branches are most often close enough, if not identical.

**Commit 2:**
Cirrus: remove config
    
This commit removes the cirrus config file, thus all cirrus jobs.
    
It also hardcodes the Makefile var SKOPEO_CIDEV_CONTAINER_FQIN to get rid of the cirrus.yml requirement.
    
TODO:
1. It would be nice to have targets in the main Makefile independent of any CI system.
    
2. Fix test-integration-local target to handle the `registry-v2` binary issue. Currently, it's not buildable and Makefile doesn't account for it either.

This does NOT enable any workflow from upstream to downstream. It won't automatically create PRs to downstream dist-git.


plans/no-rpm.fmf
```diff
@@ -0,0 +1,28 @@
...
+      how: install
+      package: [ golang, podman-docker ]
+    - name: enable-linger
+      how: shell
+      script: loginctl enable-linger $(whoami)
+    - name: golangci
+      how: shell
+      script: |
+        # https://golangci-lint.run/welcome/install/
+        curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh | sh -s -- -b /bin v2.1.6
```
mtrmac: Should this read the version from `Makefile`? On this branch it’s a different major version, with a different config file format.

plans/no-rpm.fmf
```diff
@@ -0,0 +1,28 @@
...
+
+/validate:
+    execute+:
+        script: |
+            make validate
+            make validate-docs
+
+/integration:
+    execute+:
+        script: make test-integration
```
mtrmac: This would try to run everything in a container (and it’s a barely-maintained test path not otherwise exercised in CI, AFAIK). Is this extra layer Intentional? It might not be necessary if everything were running in a dedicated VM. (I have ~no opinion.)

mtrmac (commented):
Thanks!

I know ~nothing about TMT and Packit.

@jnovy PTAL.

plans/no-rpm.fmf
lsm5: it's running in a vm yes. I'll switch to -local. Thanks!

plans/no-rpm.fmf
lsm5: I'll check on this, thanks

plans/no-rpm.fmf
lsm5: @mtrmac `test-integration-local` requires the registry-v2 binary which I just can't seem to get built.  `test-integration` I think has the binary included in the image and that passed for me. So, i've kept it as `test-integration`. Let me know if you got a fix for registry-v2 binary.

mtrmac: It’s not something I’ve thought about recently; the CI image build automation must be building it _somehow_.

Anyway, if using the container is easier, that’s perfectly fine with me.

plans/no-rpm.fmf
lsm5: done.

lsm5:
@Luap99 @mheon @jnovy PTAL .

I'd be happy to remove .cirrus.yml from here, but I'm also cool with keeping it at least until we're required to get rid of Cirrus. So, let me know if anyone's got a strong preference. Same goes for buildah / podman. There's no merge-bot on skopeo though.

.packit.yaml
```diff
@@ -2,98 +2,62 @@
...
-    targets:
-      - fedora-all-x86_64
-      - fedora-all-aarch64
+        message: "Packit jobs failed. @containers/packit-build please check."
     enable_net: true
+    targets:
+      - epel-10-x86_64
+      - epel-10-aarch64
+    actions:
+      post-upstream-clone: "curl -O https://gitlab.com/redhat/centos-stream/rpms/skopeo/-/raw/c10s/skopeo.spec"
```
Luap99: this is missing `--fail` otherwise curl won't error for 4xx/5xx http errors. I guess the build would then of course fail but it would be harder to spot?

Luap99 (commented):
If we do this than removing cirrus.yml from here would make the most sense from me. 

Personally I find the tmt task setup super confusing, it took me several minutes to figure out which tests are run. The difference between rpm/no-rpm is quite hard to sport in the job list IMO. Anyway given one only has to look when a task failed I guess that is not to bad

lsm5:
> If we do this than removing cirrus.yml from here would make the most sense from me.

sgtm
> 
> Personally I find the tmt task setup super confusing, it took me several minutes to figure out which tests are run. The difference between rpm/no-rpm is quite hard to sport in the job list IMO. Anyway given one only has to look when a task failed I guess that is not to bad

I could have individual jobs for each test in that case in a later followup.

lsm5:
Looks like the RHEL 9.6 tests are failing an older copr build with a higher epoch (because of the now-removed in-repo spec file). I'll have to do a new PR to fetch the right builds.

lsm5:
trying here: https://github.com/containers/skopeo/pull/2622

lsm5:
> Looks like the RHEL 9.6 tests are failing an older copr build with a higher epoch (because of the now-removed in-repo spec file). I'll have to do a new PR to fetch the right builds.

weird, looks like the issue went away. Anyway, we can continue here itself.

.packit.yaml
lsm5: done

lsm5:
I guess we don't need merge protection enabled in settings for such branches given we don't have any mergebot here.

plans/no-rpm.fmf
```diff
@@ -0,0 +1,25 @@
...
+#      how: shell
+#      script: loginctl enable-linger $(whoami)
+
+discover:
+    keep-git-metadata: true
+    how: shell
+    tests:
+        - name: /validate
+          test: >
+            make &&
```
mtrmac: I’m not sure this is necessary — “validate” is ~a lint step, and this builds the binary.

`test-integration-local` depends on building the binary, so we do build it, in that other test.

plans/no-rpm.fmf
```diff
@@ -0,0 +1,25 @@
...
+            make tools &&
+            make BUILDTAGS="exclude_graphdriver_btrfs" validate-local &&
```
mtrmac, line 13: Was `make vendor && hack/tree_status.sh` removed intentionally?

mtrmac, line 13: If the BTRFS exclusion matters, I suppose the build tag should be applied consistently to anything that builds the Skopeo Go code, incl. `test-integration*`, `test-unit*`, and `bin/skopeo`.

mtrmac:
> I guess we don't need merge protection enabled in settings for such branches given we don't have any mergebot here.

On the `main` branch, we do require “Total success” to pass (and that blocks the GitHub merge UI, even without a bot).

Similarly, defining merge protection for an earlier branch to require the … 17 individual tests to pass, on this branch, is an option.

plans/no-rpm.fmf
lsm5: ack, removing

plans/no-rpm.fmf
lsm5: pebkac probably. Adding..

mtrmac:
(Absolutely non-blocking: I’m somewhat curious about the longer-term plan. Will we, when creating every new branch, forward-port this PR, integrating any changes to `.cirrus.yml` that happened in the meantime? Are we eventually going to remove Cirrus from the main branch? Something else?)

lsm5:
> (Absolutely non-blocking: I’m somewhat curious about the longer-term plan. Will we, when creating every new branch, forward-port this PR, integrating any changes to `.cirrus.yml` that happened in the meantime? Are we eventually going to remove Cirrus from the main branch? Something else?)

My take:
- I will forward port some of the changes to main (the validate / integration / unit tests)

- Once forward porting is done, I would like to get rid of Cirrus unless we see it providing anything Packit+TMT can't in this case. (Stable images and unexpected out-of-our-control failures blocking PRs could be a possible factor).

- When a new release- branch is created that only goes into RHEL, we could replace the Fedora / CentOS Stream jobs with RHEL jobs. There's PR label based job filtering available in Packit, but if new release branches are not too often, we could do a manual editing of the packit config each time.

mtrmac (commented):
LGTM (but I see some not-immediately-obvious test failures and this PR is currently marked as draft)

lsm5:
> LGTM (but I see some not-immediately-obvious test failures and this PR is currently marked as draft)

manpage checker script acting weird. I'll check if i can disable that tomorrow. The rpm builds manpages so I'm hoping that's good enough.

plans/no-rpm.fmf
```diff
@@ -0,0 +1,23 @@
...
+
+discover:
+    keep-git-metadata: true
+    how: shell
+    tests:
+        - name: /validate
+          test: >
+            make tools &&
+            make BUILDTAGS="exclude_graphdriver_btrfs" validate-local &&
+            make validate-docs
```
mtrmac: > “No rule to make target 'make'”

Missing `&&`  I think

lsm5: ah yes, thanks. Fixed.

lsm5:
Let's hold off on merge. The rpm installation flakes are really weird. I'll take that up with the packit and tmt people, so let's keep the failed jobs as-is.

Luap99:
> Once forward porting is done, I would like to get rid of Cirrus unless we see it providing anything Packit+TMT can't in this case. (Stable images and unexpected out-of-our-control failures blocking PRs could be a possible factor).

Well IMO stable images are a requirement to some extend for main, well maybe skopeo doesn't have enough activity that is not a problem and because I mainly don't work on this repo I would leave that call to @mtrmac.
The other things is the merge protection stuff, tmt doe snot have a task aggregation  name like "Total Success". This means that merge protection settings must list each task name individually. My main gripe there is that these task names are not a constant. They all include the distro names via "floating" targets such as fedora-all which means names change and then someone must update the merge setting each time...

Now for a repo without the merge bot that is a possibility, but with the openshift merge bot setting that is much harder.
Again since skopeo doesn't use it it may be fine. Also if one just stays with the github UI I don't think the merge protection offers that much CI job wise. Because a normal reviewer would not click merge unless all jobs passed and the UI by default doesn't show the green merge button until that is the case so I would not be to worried about it in general.

> When a new release- branch is created that only goes into RHEL, we could replace the Fedora / CentOS Stream jobs with RHEL jobs. There's PR label based job filtering available in Packit, but if new release branches are not too often, we could do a manual editing of the packit config each time.

I mean a RHEL branch is once every 6 months so I would not see this as problem to do it manually. And again I personally do not believe RHEL branches should live on upstream repos at all. And if that will be the case we need a manual import anyway and then switch the distro targets would be straight forward as well. That is the great thing about TMT, because it is not tied to our infra specifically we kind move the same plans from upstream to downstream.

lsm5:
@mtrmac would you be ok with skipping `hack/tree_status.sh` check for now? I've filed a TMT issue, and the maintainers agree that it's not desirable behaviour, so hopefully it'll be fixed soon.

mtrmac:
> > Once forward porting is done, I would like to get rid of Cirrus unless we see it providing anything Packit+TMT can't in this case. (Stable images and unexpected out-of-our-control failures blocking PRs could be a possible factor).
> 
> Well IMO stable images are a requirement to some extend for main, well maybe skopeo doesn't have enough activity that is not a problem

Right now, the integration tests require building (a very old version of) OpenShift registry to test `atomic:`, and that takes quite a long time; relying on pre-built images noticeably decreases CI turnaround time, even just for that (and there’s probably much more that could be moved there).

---

There are also various Cirrus caching features, see e.g. the numbers in https://github.com/containers/image/pull/2837#discussion_r2056611936 / https://github.com/containers/podman/pull/25951#issuecomment-2825051928 . We don’t currently benefit from this, and I’m not working on it, but I do think it’s an investment well worth prioritizing.

Is there anything equivalent in TMT?

---

> Because a normal reviewer would not click merge unless all jobs passed

…

mtrmac:
> @mtrmac would you be ok with skipping `hack/tree_status.sh` check for now?

The goal of that test is to ensure that `vendor/*` matches `go.mod` with no modification. For that, I think using `git status --porcelain go.* vendor` should be sufficient.

(And RHEL branches might imply increased motivation to backport directly into `vendor` without a corresponding change existing upstream, so I think the test is at least equally useful for these branches. Perhaps we would do such an unclean backport _anyway_, in some situations, but I think it’s useful to be explicit and intentional about it. OTOH, I don’t recall us actually doing such a thing recently, or maybe only in downstream RPM packaging.)

plans/no-rpm.fmf
```diff
@@ -0,0 +1,27 @@
...
+          test: >
+            make tools &&
+            make BUILDTAGS="exclude_graphdriver_btrfs" validate-local &&
+            make validate-docs &&
+            make vendor &&
+            # TMT breaks `git status --porcelain` so run it only on go.* and
+            # vendor dir.
+            # See: https://github.com/teemtee/tmt/issues/3800
+            # hack/tree_status.sh
+            git status --porcelain go.* vendor
```
mtrmac: (Absolutely non-blocking: This is the only possible caller of `hack/tree_status.sh`, so I think it can be edited there. Also… I don’t think there’s anything wrong with restricting the check to `go.* vendor` either way, I don’t think this needs to be specifically tracked as a short-term workaround in this repo. [TMT might still want to change their implementation so that it doesn’t add extra files, that’s a ~separate question.])

lsm5:
I removed `RHEL-N-Nightly` from the `no-rpm` targets as RHEL-9-Nightly compose already has go 1.24.

PTAL, hopefully good for merge.

PR #2624, opened 2025-06-10
dynamically link sqlite3

Luap99:
Dynamically link sqlite3 when installed, the main motivation is that we
reduce the podman binary size with that. I see about 3.3 MB savings.

But also dynamically linking it means if there a vulnerabilities only
the sqlite3 distro package needs updating and we don't have to make a
new podman release with the vendored update.


Luap99:
cc @mtrmac @lsm5 @jnovy

lsm5 (approved):
Nice, Packit and TMT are happy, so LGTM.

Thank @Luap99

lsm5:
I see sqlite exists on CentOS Stream so we should be fine there too. We could always conditionalize the buildtag if it's a problem on any env.

jnovy:
NOD. Thanks!

PR #2626, opened 2025-06-12
RPM: make bats a weak dep

lsm5:
bats isn't present on RHEL, CentOS Stream and ELN.

Ref: https://github.com/containers/skopeo/commit/f88186e688aad62de716ad0f70d00bf9b440d29d#r159874060

### DIFF
--- a/plans/main.fmf
+++ b/plans/main.fmf
@@ -6,8 +6,14 @@ prepare:
     - when: distro == centos-stream or distro == rhel
       how: shell
       script: |
-        dnf -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-$(rpm --eval '%{?rhel}').noarch.rpm
-        dnf -y config-manager --set-enabled epel
+        # Install bats
+        # https://bats-core.readthedocs.io/en/stable/installation.html
+        BATS_VERSION=1.12.0
+        curl -L https://github.com/bats-core/bats-core/archive/refs/tags/v"$BATS_VERSION".tar.gz | tar -xz
+        pushd bats-core-"$BATS_VERSION"
+        ./install.sh /usr
+        popd
+        rm -rf bats-core-"$BATS_VERSION"
       order: 10
     - when: initiator == packit
       how: shell
--- a/systemtest/tmt/main.fmf
+++ b/systemtest/tmt/main.fmf
@@ -1,5 +1,4 @@
 require:
-    - bats
     - skopeo-tests
 
 environment:


lsm5:
The main reason for this is ELN is a rebuild of only those rawhide package sources that are included in RHEL. So, while `Requires: bats` will work on rawhide, it will cause issues on ELN.

Other responses inline...

> Thanks!
> 
> Is a `Recommends:` exactly what we want here? The system tests strictly require `bats`, in order to work.

RHEL and CentOS Stream don't ship it in the vanilla repos, but Fedora does. So, this seems like the smallest change that'll work across the board.

> The way I think about it, either the `skopeo-tests` package is supposed to be useful ~stand-alone, and in that case it needs a hard dependency; 

Useful standalone yes, but only as far as gating tests are concerned. We don't need to entertain any end-user requests.

> or the `skopeo-tests` package is an internal implementation detail of some higher-level testing framework, and then if that higher-level testing framework always ensures `bats` is available, a RPM dependency does not help and it can be removed entirely.

Ensuring bats is always available can depend on the env where it's triggered, and making it work across the board has been problematic in my experience.

Gating tests for the official CentOS and RHEL rpms now happen on the internal testing-farm, where epel is not enabled by default. And enabling and fetching rpms from epel has given me a lot of grief in the past.

Also, IIUC, some of RHEL QE's testing happens entirely without EPEL.

> 
> I don’t really understand the test automation; looking just at this repo, on this `main` branch, `systemtest/tmt/main.fmf` requires both `bats` and `skopeo-tests`, so I think removing the `Requires:` doesn’t help anything.

Yes, the requirement on `bats` should be removed from this file. I can add it here in a separate commit. The requirement on `bats` in main.fmf will cause problems downstream whenever changes get propagated.

> 
> OTOH in `release-1.18` after #2619 , `bats` is being built from source and `systemtest/tmt/main.fmf` does not require a `bats` RPM. _And_ we have entirely removed the spec file from the repo, in favor of fetching a c10s one.

Yes, that's simplest for RHEL environments. I'll forward port this change to main in an additional commit on this PR itself.
> 
> So, naively, all of that looks approximately correct to me; to the extent [f88186e#r159874060](https://github.com/containers/skopeo/commit/f88186e688aad62de716ad0f70d00bf9b440d29d#r159874060) complains about a spec dependency in RHEL / ELN, shouldn’t that be dependency be removed from the c10s repo instead?

*Ideally* I want all the changes to flow from upstream to CentOS Stream via Packit. Though we aren't there yet, I'd like to keep the upstream spec file also usable for CentOS Stream. We should be running CentOS Stream build and test jobs upstream as well, but that's currently broken on golang version issues.

> 
> (Then there’s a whole another can of worms, “skopeo-tests (which is shipped)”. Does that subpackage _need_ to be shipped externally? I assume that has been _extensively_ debated and litigated in the past and I have no interest in reopening that, as long as all relevant maintainers know what’s going on and why.)

The `-tests` subpackage slipped through the cracks for podman, but I don't think that's the case for buildah and skopeo yet. Either way, QE still needs to use it.

lsm5:
> I don’t really understand the test automation; looking just at this repo, on this `main` branch, `systemtest/tmt/main.fmf` requires both `bats` and `skopeo-tests`, so I think removing the `Requires:` doesn’t help anything.

Removed the dep and fixed bats install script on RHEL and CentOS Stream in the additional commit.

mtrmac:
> ELN is a rebuild of only those rawhide package sources that are included in RHEL. So, while Requires: bats will work on rawhide, it will cause issues on ELN.

OK, I didn’t realize we want exactly the same spec file for those different distributions. Makes sense.

mtrmac (commented):
LGTM, to the little extent I understand things. (This is currently marked as draft.)

lsm5:
> LGTM, to the little extent I understand things. (This is currently marked as draft.)

done. Thanks.

PR #2632, opened 2025-06-20
Packit: disable osh-diff-scan

lsm5:
Doesn't currently work for go packages anyway.


lsm5:
> Why not. Thanks!

If you meant to ask why not, that's because no golang support yet in the diff scan tool used, so it's a nop at this point.

mtrmac:
It was more of a “we might _hypothetically_ lose some test coverage _if_ they added support in the future, but, meh, probably not worth the extra trouble for now, and anyway I should be trusting the experts”. :)

PR #2636, opened 2025-06-27
Packit: disable official CentOS Stream update job

lsm5:
We're currently not updating official CentOS Stream packages via Packit. So, this commit disables the downstream PR update job.

No noticeable effect upstream. Safe to merge.


mtrmac (approved):
Thanks!

PR #2645, opened 2025-07-07
Add --sign-by-sq-fingerprint

mtrmac:
This is https://github.com/containers/image/pull/2876, integrating with the build system, and it will include new CLI and a smoke test.

<s>Do not merge: Depends on #2681 and https://github.com/containers/image/pull/2876 in c/image.</s>


contrib/cirrus/runner.sh
```diff
@@ -96,6 +96,18 @@ _run_vendor() {
 }
 
 _run_build() {
+    if [[ "$BUILDTAGS" =~ containers_image_sequoia ]]; then
```
mtrmac: @Luap99 This is broadly the kind of thing I’m thinking of doing in CI, also in Buildah/Podman to have a smoke-test.

Until we have a better packaging of the Rust code, I expect we would not opt into the build tag by default.

Luap99: Uh, I really really dislike this. Runtime installs like this are going to flake and break CI eventually. Rust compile times are slow so the CI times get longer.

Sure c/image and skopeo don't have much activity so maybe it can be accepted here but for podman buildah what would be the point? Just build testing seems pointless if we know it builds in c/image?

I feel like without a packaging story there is no point to ship this at all right now and try to rush this out the door. So this is hard hard blocker for me personally.
IMO the order must be find a proper place for the c -> rust library bindings, then release that as c library, then package it into fedora and then we can install it in our upstream VM images and link against it in the builds.

mtrmac: > for podman buildah what would be the point?

I intend to add a smoke-test for a new CLI option (skipped if Sequoia is not enabled). I think that can happen in any case.

And yes, then the discussion whether to enable Sequoia in CI at all, whether to enable Sequoia in releases by default, whether to test _only_ Sequoia in CI, can happen, and I am open to any of the outcomes.

I agree it makes no sense to ship Sequoia enabled upstream by default until the library is easy to install.

Enabled in RPMs? Enabled in CI? Well, maybe?

Luap99: > 02:47 build

for the normal Skopeo Test while

> 11:57 build

For Skopeo test w/ Sequoia

To me that alone is not acceptable for any merge in any CI anywhere. We can't just make the dev experience an all repos so much worse just because there is one small lib that is not even in use anywhere as it is not packaged.

You can add the code but compiling this in CI on every PR is just plain wasted time/resources. To me the packaging step must happen first for the library. Then we add it into the CI images and then we can link it in build just like the other c libs without this huge overhead.

Whenever the fedora skopeo/buildah/podman rpm builds use libgpgme or Sequoia that is not a discussion I care about.

mtrmac: Here, I think the more relevant comparison is 12:41 vs. ~22:… for the whole task; the _relative_ difference is not as important.

Anyway, I hear you.

We always have the option of doing the build once in automation_images instead of in primary project repos like this PR is now doing. Given the expected RPM packaging, I don’t think that’s worth the effort right now, but it’s an option. Or I can test things locally and submit ~ineffective tests upstream, to be enabled later.

mtrmac:
Note https://github.com/containers/podman/pull/26618#issuecomment-3063657304 :
> I would like to NOT merge this until after we move Podman v5.6 to it's own RHEL branch, probably mid to late August. This should be aimed at Podman v5.7 or 6.0 at the earliest.

I’m not sure what we decide to do, but we should be consistent across projects. Cc: @TomSweeneyRedHat

TomSweeneyRedHat:
@mtrmac as we have created a Skopeo 1.20 branch to be used for RHEL 9.7/10.1, I'm fine with moving forward with this now if you wish.

contrib/cirrus/runner.sh
mtrmac: Now updated to rely on a RPM installed in https://github.com/containers/automation_images/pull/411 .

(Still blocked on both the automation_images PR, and c/image implementation, being merged.)

mtrmac:
Ready for review, PTAL.

Luap99 (approved):
LGTM

mtrmac:
Rebased, and added a commit to update tests for `rust-podman-sequoia` 0.2.0. PTAL.

integration/copy_test.go
```diff
@@ -779,9 +779,11 @@ func (s *copySuite) TestCopySignatures() {
 	// Verify that mis-signed images are rejected
 	assertSkopeoSucceeds(t, "", "--tls-verify=false", "copy", "atomic:localhost:5006/myns/personal:personal", "atomic:localhost:5006/myns/official:attack")
 	assertSkopeoSucceeds(t, "", "--tls-verify=false", "copy", "atomic:localhost:5006/myns/official:official", "atomic:localhost:5006/myns/personal:attack")
-	assertSkopeoFails(t, ".*Source image rejected: (Invalid GPG signature|.* was not found).*",
+	// "Invalid GPG signature" is reported by the gpgme mechanism; "$fingerprint was not found" by Sequoia with rust-python-sequoia 0.1.0,
+	// "Missing key: $fingerprint" by Sequoia with rust-python-sequoia 0.2.0. The "was not found" case should eventually be removed
+	assertSkopeoFails(t, ".*Source image rejected: (Invalid GPG signature|.* was not found|Missing key:).*",
```
Luap99: I think because the string and comment is duplicated 4 times here it would be cleaner to extarct into a shared constant that only has the be document once?

mtrmac: Yes, but, also, after https://github.com/containers/automation_images/pull/416, I plan to check whether that image already contains 0.2.0, and remove one of the branches as well as at least half of the comment again.

mtrmac: > I plan to check whether that image already contains 0.2.0

It does. Well … this PR is bumping the CI image already, let me see if I can update that here directly.

mtrmac: OK updated with more recent images, and dropped the 0.1.0 variants.

Luap99 (approved):
LGTM

PR #2647, opened 2025-07-08
Update GPG keys not to use SHA-1

mtrmac:
Sequoia-PGP does not accept SHA-1 by default; update to avoid that.

Both the keys and the signatures come from c/image commit 70dbc81047f63a68130a5bdf1bfaee43db39746d .


Luap99 (approved):
LGTM

PR #2648, opened 2025-07-09
Consolidate options shared between copy and sync to sharedCopyOptions

mtrmac:
This will allow adding other signing options in a single place.

Also, use a consistent ordering for `AddFlagSet` — it turns out one ordering is better than the other.

Should not change behavior, apart from some help texts.


Luap99 (approved):
LGTM

TomSweeneyRedHat:
LGTM

PR #2650, opened 2025-07-11
Prepare for `--sign-by-sq-fingerprint`

mtrmac:
#2645 is now working and tested end-to-end, but depends on an unmerged feature.

This PR contains the required infrastructure / refactoring before the main feature, to allow reviewing / merging sooner.

(Alternatively, it would be possible to include the CLI code as well, with https://github.com/containers/image/pull/2896 , but having a CLI option that can never work seems a bit overzealous, at least for now.)


lsm5 (approved):
LGTM

 #2654 can wait for this to get merged.

PR #2654, opened 2025-07-16
fixes to have test-*-local Makefile targets work

lsm5:
Please see individual commits.


lsm5:
Moving the TMT-independent stuff from https://github.com/containers/skopeo/pull/2640 here for easier review.

lsm5:
@mtrmac @cevich PTAL. We need not be concerned with packit jobs here. There's already #2640 that rebases on this PR.

integration/registry_test.go
```diff
@@ -13,10 +13,17 @@ import (
...
 
 const (
-	binaryV2        = "registry-v2"
 	binaryV2Schema1 = "registry-v2-schema1"
 )
 
+// getBinaryV2 returns the available registry binary name, preferring "registry-v2" over "registry"
+func getBinaryV2() string {
+	if _, err := exec.LookPath("registry-v2"); err == nil {
+		return "registry-v2"
```
mtrmac: Maybe we can, instead, use the expected name (or, even, just install a RPM?) in automation-images. I didn’t look into it further now.

hack/test-integration.sh
```diff
@@ -1,8 +1,6 @@
...
 set -e
 
-make PREFIX=/usr install
-
-echo "cd ./integration;" go test $TESTFLAGS ${BUILDTAGS:+-tags "$BUILDTAGS"}
+echo "cd ./integration;" go test "$TESTFLAGS" ${BUILDTAGS:+-tags "$BUILDTAGS"}
 cd ./integration
-go test $TESTFLAGS ${BUILDTAGS:+-tags "$BUILDTAGS"}
\ No newline at end of file
+go test "$TESTFLAGS" ${BUILDTAGS:+-tags "$BUILDTAGS"}
```
mtrmac: The `"$TESTFLAGS"` changes are inconsistent with how `Makefile` uses the variable, that explicitly documents using a value with spaces to mean separate arguments.

hack/test-integration.sh
```diff
@@ -1,8 +1,6 @@
 #!/bin/bash
```
mtrmac: (This is not related to anything in this PR, just as a vague aesthetic opinion — it seems that inlining these very small scripts back into the `Makefile` might make things simpler.)

Makefile
```diff
@@ -201,10 +201,8 @@ test-integration:
 	$(CONTAINER_CMD) --security-opt label=disable --cap-add=cap_mknod -v $(CURDIR):$(CONTAINER_GOSRC) -w $(CONTAINER_GOSRC) $(SKOPEO_CIDEV_CONTAINER_FQIN) \
 		$(MAKE) test-integration-local
 
-
-# Intended for CI, assumed to be running in quay.io/libpod/skopeo_cidev container.
-test-integration-local: bin/skopeo
-	hack/warn-destructive-tests.sh
+# Expects binary in $PATH, so better to depend on install-binary target
+test-integration-local: install-binary
```
mtrmac: This loses the effect of `warn-destructive-tests`.

Makefile
```diff
@@ -220,8 +218,7 @@ test-system:
 	exit $$rc
 
 # Intended for CI, assumed to already be running in quay.io/libpod/skopeo_cidev container.
-test-system-local: bin/skopeo
-	hack/warn-destructive-tests.sh
```
mtrmac, line 223: Similarly here.

mtrmac (commented):
Just an immediate skim, not nearly a review,

integration/registry_test.go
lsm5: I didn't wanna bother with Cirrus, especially if we'd be ok with removing Cirrus after the switch to TMT is complete. But, I could do that if that's the right way.

Makefile
lsm5: Right, so do we need `warn-destructive-tests` at all? If we do need them locally, do we also need them in CI environments where the tests are run on a fresh VM ?

hack/test-integration.sh
lsm5: ack, i'll check and update. Thanks

hack/test-integration.sh
lsm5: SGTM

Makefile
cevich: Ideally, tests don't modify the host system beyond temporary changes.  But AFAIK the CI context doesn't care about `warn-destructive-tests`, it happily goes and modifies the system w/o asking. This is only used for humans.

hack/test-integration.sh
cevich: A reason to keep them as separate scripts is to have more control over which shell is used and in cases where multiple tools (i.e. in addition to `make`) need to use the script.  I'm not saying "no", just mentioning some reasons for consideration.

hack/test-system.sh
```diff
@@ -37,8 +37,12 @@ EOF
...
 fi
 
-# Build skopeo, install into /usr/bin
-make PREFIX=/usr install
+# Print what binary is being used for tests
+if [[ -v SKOPEO_BINARY ]]; then
+    echo "Testing with $SKOPEO_BINARY ..."
+else
+    echo "Testing with $(git rev-parse --show-toplevel)/bin/skopeo ..."
+fi
```
cevich: Couldn't this be simplified with:

```bash
SKOPEO_BINARY=${SKOPEO_BINARY:-$(git rev-parse --show-toplevel)/bin/skopeo ...}
echo "Testing with $SKOPEO_BINARY ..."
```

lsm5: yup, thanks!

hack/test-integration.sh
lsm5: @cevich do you see cases where the script would need to be directly invoked instead of via the Makefile target?

mtrmac: See the `TESTFLAGS` semantics note above — I think it’s easier if most actions go through the same variable expansion / defaulting / … processes.

Now, that can be equally read as advocating for everything to be moved to `make` rules, _and_ as advocating for abandoning `make` and replacing it with individual shell scripts.

Makefile
mtrmac: Yes, this should primarily protect against humans just running `make check` because that is a standard target.

(Now, _reading_ that, the script only warns about tests being slow, not about overwriting `/usr/bin`, so that’s not really doing its job. Still, directionally, I’d prefer to keep and improve that safeguard rather than to drop it.)

CI running in an isolated environment can, and should, opt out of the warning/delay/protection.

hack/test-integration.sh
mtrmac: As an ~example, https://github.com/containers/skopeo/pull/2650 needs to add more options that are passed from the top-level `runner.sh` down to the one-level-deep `make` that actually builds binaries. It certainly _can_ be done, and it’s not a big deal overall, still, I think passing values `make`→`sh`→`make` just gets in the way.

It’s very possible I’m missing some more elegant way to do this.

hack/test-system.sh
```diff
@@ -37,8 +37,12 @@ EOF
     export CONTAINERS_STORAGE_CONF=/etc/containers/storage.conf
 fi
 
-# Build skopeo, install into /usr/bin
-make PREFIX=/usr install
+# Print what binary is being used for tests
+if [[ -v SKOPEO_BINARY ]]; then
```
mtrmac: (Absolutely non-blocking: Probably `-n "$…"`, because it is used in `${SKOPEO_BINARY:-…}` and that triggers on empty values as well. This should never make a difference in practice.)

hack/test-system.sh
```diff
@@ -37,8 +37,12 @@ EOF
...
+    echo "Testing with $SKOPEO_BINARY ..."
+else
+    echo "Testing with $(git rev-parse --show-toplevel)/bin/skopeo ..."
```
mtrmac: Non-blocking, most of all I’m just happy about removing the `make install` system modification, and this is not worth overengineering:

That’s not _quite_ correct — the script hypothetically might run from a release tarball where there is no git repo.

(Just to be sure, this never runs inside a RPM spec build from a tarball, where `git rev-parse` could find the spec file repo, does it?)

Something like `$(realpath $(dirname ${BASH_SOURCE})/../bin/skopeo)`? I mean, that _still_ hard-codes behavior of `systemtest/helpers.bash` instead of calling it, so it’s not ideal, but it’s one less assumption. And a comment added below says `../bin/skopeo`, so that comment being inconsistent with this log line is a bit surprising.

---

[Note to self: we package this in `skopeo-tests`, where there is no git repo, but that always sets `SKOPEO_BINARY`.]

integration/registry_test.go
mtrmac: In general, I think adding the condition here adds risk. We can have the condition live here for a long time, or we can do a one-time rename in `automation_images` and change the constant here — I’d prefer a constant over a condition, but also I don’t care that much.

---

The commit message says
> `make test-integration-local` passes with this change
> and docker-distribution installed.

but I’m unclear how — AFAICS `TestCopySchemaConversion` requires `registry-v2-schema1` to exist and to be runnable, so the user must have some of the custom registry builds _either way_, and then I think there’s not that much value to this change. What am I missing?

mtrmac: (The automation-images change could probably be a single line https://github.com/containers/automation_images/blob/0e1497cd7794a9b0c7056bbdc7250202a3fa8ac7/skopeo_cidev/setup.sh#L49 , or the package list could be changed.

But, either way, I’m misunderstanding something here.)

mtrmac (commented):
A full review now.

hack/test-integration.sh
cevich: Maybe a `hack/hack_lib.sh` that does all the variable handling centrally?

> do you see cases where the script would need to be directly invoked instead of via the Makefile target?

I don't really have a grasp of all the different contexts where this would be used, so I'm sorry I can't really answer that with any authority.  In working with CI systems, I *have* definitely run into cases like this where going through `make` got in the way and calling the script directly was easier (or easier to customize for the CI context).

mtrmac: Good point, the `make` intermediate step does introduce one extra parsing step (`make` variables have quoting interpreted by shell) and that might need extra care when calling form a shell script.

I was focusing on the `make`→`sh`→`make` problem, but `sh`→`make`→`sh` is the dual of that.

integration/registry_test.go
lsm5: This has been fixed in automation-images. Marking resolved.

hack/test-system.sh
lsm5: removed this script to run everything inline in Makefile. Marking resolved for now.

hack/test-integration.sh
lsm5: removed this script and added inline Makefile rules. Marking resolved for now.

hack/test-system.sh
lsm5: replaced with inline Makefile rules.

lsm5:
I've replaced `hack/test-*.sh` with inline execution in the Makefile itself. Also, made changes to TMT test setup to reuse Makefile targets, along with other TMT changes such that other locations where these tests need to be run can simply refer to the URL and git ref like so: https://src.fedoraproject.org/rpms/skopeo/pull-request/83#_1__4-5 . The downstream execution results are at: https://artifacts.dev.testing-farm.io/06cfb608-0277-42b3-a476-3c4848367914/ 

I'll make a followup PR for some other packit related changes to avoid overwriting the downstream changes and ensure the git refs are updated on every release.

lsm5:
Btw, we will also be able to get rid of the `skopeo-tests` subpackage this way. To be done in followup.

lsm5:
@mtrmac @cevich PTA(nother)L when you get the chance. Thanks!

Makefile
lsm5: I'll remove this comment and the comment for test-integration-local given we're running -local in TMT envs.

Makefile
```diff
@@ -218,10 +219,11 @@ test-system:
 	$(CONTAINER_RUNTIME) unshare rm -rf $$DTEMP; # This probably doesn't work with Docker, oh well, better than nothing... \
 	exit $$rc
 
-# Intended for CI, assumed to already be running in quay.io/libpod/skopeo_cidev container.
-test-system-local: bin/skopeo
+# Primarily intended for CI.
+test-system-local: $(if $(SKOPEO_BINARY),,bin/skopeo)
 	hack/warn-destructive-tests.sh
-	hack/test-system.sh SKOPEO_LDFLAGS="$(SKOPEO_LDFLAGS)" BUILDTAGS="$(BUILDTAGS)"
+	@echo "Testing with $(or $(SKOPEO_BINARY),$(eval SKOPEO_BINARY := "bin/skopeo")$(SKOPEO_BINARY)) ..."
```
mtrmac, line 225: AFAICS this does not export `SKOPEO_BINARY` as an environment variable, so the `eval` only affects the printed path here. Was that intentional?

(In practice it ultimately works the same, `systemtest/helpers.bash` defaults to essentially `$(pwd)/bin/skopeo` anyway.)

hack/test-system.sh
```diff
@@ -1,44 +0,0 @@
...
-# both inside a container and without a container (in a CI VM); it actually means
-# "it is safe to destructively modify the system for tests".
-#
-# On a CI VM, we can just use Podman as it is already configured; the changes below,
-# to use VFS, are necessary only inside a container, because overlay-inside-overlay
-# does not work. So, make these changes conditional on both
-# SKOPEO_CONTAINER_TESTS (for acceptability to do destructive modification) and !CI
-# (for necessity to adjust for in-container operation)
-if ((SKOPEO_CONTAINER_TESTS)) && [[ "$CI" != true ]]; then
-    if [[ -r /etc/containers/storage.conf ]]; then
```
mtrmac, line 19: Non-blocking: I don’t know about anyone that _uses_ this — so if we are giving up on the “container outside of CI” setup, perhaps we should also explicitly remove the `test-system` target (and all of `check` as well??).

Or is this being removed because you have determined it’s no longer necessary?

lsm5: True, `make test-system` will fail without this. So, I'd be in favour of getting rid of this and `test-system`, and maybe replace the targets in check with `-local` . 

I'd also be in favour of doing the same for test-integration.

That saves us the burden of worrying about skopeo_ci_dev image maintenance too.

Makefile
lsm5: Yes, this was only meant to echo SKOPEO_BINARY if it's set or just use the default `bin/skopeo` so we know which binary is actually being tested. We're setting SKOPEO_BINARY in the TMT tests.

lsm5:
> Thanks!
> 
> LGTM overall, feel free to merge — just noting some items that might be worth another look.

Thanks, I'll need an approving review to go ahead with the merge. It's currently blocked for me.

PR #2655, opened 2025-07-16
Use platform-specific Docker socket location on macOS

jacques-n:
Work around Docker Desktop's macOS socket location change

Docker Desktop 4.13.0+ moved the socket from /var/run/docker.sock to
~/.docker/run/docker.sock on macOS, but the Docker client library still
defaults to the old location for all Unix systems.

Add platform-specific logic to use the correct default on macOS when
--daemon-host is not specified.

This makes skopeo work more reliable out of the box for new users.

Signed-off-by: Jacques Nadeau <jacques@apache.org>

### DIFF
--- a/cmd/skopeo/utils.go
+++ b/cmd/skopeo/utils.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
+	"runtime"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


jacques-n:
Correcting this behavior is something that should be considered. Initial out of box experiences matter. I would expect skopeo to work out of the box on MacOS Docker Desktop given its prevalence. Given the hesitance in the response here though, I'm guessing that trying to get this per deployment hack even lower down the stack in container wouldn't be accepted. Will leave it others to make this library more consumable. 

Thanks for the quick and thorough review!

mtrmac:
I think this is a discussion that _can_ be had – ideally involving experts familiar with the macOS GUI tools. I have opinions on the structure of the implementation, but I’m “_very_ unsure” and fairly open to the goals we might choose to have.

Perhaps a thread at https://github.com/containers/podman/discussions might be a good place to figure out the goals.

jacques-n:
Realistically, I'm not going to start a discussion of goals for the defaults experience for macOS docker desktop users of skopeo in the podman project. I appreciate the guidance nonetheless. Thanks again.

PR #2659, opened 2025-07-17
Update opencontainers/image-spec after https://github.com/opencontainers/image-spec/pull/1253

mtrmac:
This updates github.com/opencontainers/image-spec/schema subpackage from an obsolete .../jsonschema/v5 to /v6 .

We use these dependencies only in tests, so it doesn't matter that much, but removing references to years-old unmaintained versions makes it easier to check / remove other such cases.

(This is test-only, so the added files in vendor/ do not mean a larger binary size; it's almost exactly unchanged.)


lsm5 (approved):
LGTM.

PR #2660, opened 2025-07-18
fix(sync): fix dry-run failing consecutive runs by creating directories in target dir

frq-asgard-josi:
`skopeo sync --dry-run --dest dir` currently fails consecutive executions since the `func destinationReference` attempts to create directories even when running in dry-run mode and fails if these already exist.

### DIFF
--- a/cmd/skopeo/sync.go
+++ b/cmd/skopeo/sync.go
@@ -166,7 +166,7 @@ func parseRepositoryReference(input string) (reference.Named, error) {
 // destinationReference creates an image reference using the provided transport.
 // It returns a image reference to be used as destination of an image copy and
 // any error encountered.
-func destinationReference(destination string, transport string) (types.ImageReference, error) {
+func destinationReference(destination string, transport string, dryRun bool) (types.ImageReference, error) {
 	var imageTransport types.ImageTransport
 
 	switch transport {
@@ -181,9 +181,11 @@ func destinationReference(destination string, transport string) (types.ImageRefe
 		if !os.IsNotExist(err) {
 			return nil, fmt.Errorf("Destination directory could not be used: %w", err)
 		}
-		// the directory holding the image must be created here
-		if err = os.MkdirAll(destination, 0755); err != nil {
-			return nil, fmt.Errorf("Error creating directory for image %s: %w", destination, err)
+		if !dryRun {
+			// the directory holding the image must be created here
+			if err = os.MkdirAll(destination, 0755); err != nil {
+				return nil, fmt.Errorf("Error creating directory for image %s: %w", destination, err)
+			}
 		}
 		imageTransport = directory.Transport
 	default:
@@ -697,7 +699,7 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) (retErr error) {
 				destSuffix = path.Base(destSuffix)
 			}
 
-			destRef, err := destinationReference(path.Join(destination, destSuffix)+opts.appendSuffix, opts.destination)
+			destRef, err := destinationReference(path.Join(destination, destSuffix)+opts.appendSuffix, opts.destination, opts.dryRun)
 			if err != nil {
 				return err
 			}


cmd/skopeo/sync.go
```diff
@@ -181,9 +181,11 @@ func destinationReference(destination string, transport string) (types.ImageRefe
 		if !os.IsNotExist(err) {
 			return nil, fmt.Errorf("Destination directory could not be used: %w", err)
 		}
-		// the directory holding the image must be created here
-		if err = os.MkdirAll(destination, 0755); err != nil {
-			return nil, fmt.Errorf("Error creating directory for image %s: %w", destination, err)
+		if !dryRun {
+			// the directory holding the image must be created here
+			if err = os.MkdirAll(destination, 0755); err != nil {
```
mtrmac, line 186: I’m afraid I don’t think this works, in general: below, `directory.Transport.ParseReference` requires the parent directory to exist. We’d need to defer the parsing into a typed `ImageReference`, and deal with untyped strings in the meantime :/

integration/sync_test.go
```diff
@@ -266,6 +266,35 @@ func (s *syncSuite) TestDocker2DirUntagged() {
...
+	tmpDir := t.TempDir()
+
+	// FIXME: It would be nice to use one of the local Docker registries instead of needing an Internet connection.
+	image := pullableRepo
+	imageRef, err := docker.ParseReference(fmt.Sprintf("//%s", image))
+	require.NoError(t, err)
+
+	dir1 := path.Join(tmpDir, "dir1")
+	err = os.Mkdir(dir1, 0777)
+	require.NoError(t, err)
```
mtrmac: (Absolutely non-blocking: This can just be `t.TempDir()` nowadays, I don’t think it’s simplifies things to manually create subdirectories any more: historically would create a top-level temporary directory, schedule _one_ cleanup of the top-level, and then create subdirectories so that we don’t have to deal with cleanup one at a time. Now that `t.Cleanup()` exists and `t.TempDir()` does it all automatically, the top-level directory adds no value.)

integration/sync_test.go
```diff
@@ -266,6 +266,35 @@ func (s *syncSuite) TestDocker2DirUntagged() {
...
+	assertSkopeoSucceeds(t, "", "sync", "--scoped", "--dry-run", "--src", "docker", "--dest", "dir", image, dir1)
+
+	sysCtx := types.SystemContext{}
+	tags, err := docker.GetRepositoryTags(context.Background(), &sysCtx, imageRef)
+	require.NoError(t, err)
+	assert.NotZero(t, len(tags))
+
+	d, err := os.Open(dir1)
+	defer func(d *os.File) {
```
mtrmac: (`defer d.Close()` would work fine.)

integration/sync_test.go
```diff
@@ -266,6 +266,35 @@ func (s *syncSuite) TestDocker2DirUntagged() {
...
+		_ = d.Close()
+	}(d)
+	require.NoError(t, err)
```
mtrmac, line 287: Non-blocking: This `err` check needs to happen _before_ the `defer` schedules using a potentially-`nil` value. But we can fix that if the test ever fails.

mtrmac (commented):
Thanks!

> attempts to create directories even when running in dry-run mode

That’s clearly undesirable.

> and fails if these already exist.

I can’t see how? `MkdirAll` is not supposed to fail in that situation. What is failing exactly?

integration/sync_test.go
frq-asgard-josi: fixed

integration/sync_test.go
frq-asgard-josi: fixed

integration/sync_test.go
frq-asgard-josi: fixed

frq-asgard-josi:
> Thanks!
> 
> > attempts to create directories even when running in dry-run mode
> 
> That’s clearly undesirable.
> 
> > and fails if these already exist.
> 
> I can’t see how? `MkdirAll` is not supposed to fail in that situation. What is failing exactly?

Since the destination exists, `os.Stat(destination)` doesn't return any err and we return `Refusing to overwrite destination directory` (Line 179)

cmd/skopeo/sync.go
frq-asgard-josi: My personal preference would be to expect the parent directory to already exist since this would avoid creating unwanted directories due to mistyping (not that I ever did this :laughing:) but we should check that the directory exists here and provide a proper error message instead of relying on `ParseReference` to catch this.

mtrmac: The parent directory is not always user-specified: with `--scoped`, trailing paths of the path come from the source repository name.

PR #2661, opened 2025-07-18
integration: rename registry binary

lsm5:
The skopeo_cidev image is updated with the registry binary from the docker-distribution package instead of building one from source.

Ref: https://github.com/containers/automation_images/pull/410

### DIFF
--- a/integration/openshift_test.go
+++ b/integration/openshift_test.go
@@ -223,7 +223,7 @@ func (cluster *openshiftCluster) ocLoginToProject(t *testing.T) {
 // We do not run (docker login) directly, because that requires a running daemon and a docker package.
 func (cluster *openshiftCluster) dockerLogin(t *testing.T) {
 	cluster.dockerDir = filepath.Join(homedir.Get(), ".docker")
-	err := os.Mkdir(cluster.dockerDir, 0700)
+	err := os.MkdirAll(cluster.dockerDir, 0700)
 	require.NoError(t, err)
 
 	out := combinedOutputOfCommand(t, "oc", "config", "view", "-o", "json", "-o", "jsonpath={.users[*].user.token}")


mtrmac:
> mkdir /root/.docker: file exists

`s/Mkdir/MkdirAll/` in `openshiftCluster.dockerLogin`, although I am somewhat curious whether this is caused by a change in the images, and if so, what has happened there and why.

hack/test-integration.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e
 
-make PREFIX=/usr install
+make PREFIX=/usr/local install
```
mtrmac: ?

Makefile
```diff
@@ -131,7 +131,7 @@ binary: cmd/skopeo
 # Build w/o using containers
 .PHONY: bin/skopeo
 bin/skopeo:
-	$(GO) build ${GO_DYN_FLAGS} ${SKOPEO_LDFLAGS} -gcflags "$(GOGCFLAGS)" -tags "$(BUILDTAGS)" -o $@ ./cmd/skopeo
+	$(GO) build -buildvcs=false ${GO_DYN_FLAGS} ${SKOPEO_LDFLAGS} -gcflags "$(GOGCFLAGS)" -tags "$(BUILDTAGS)" -o $@ ./cmd/skopeo
```
mtrmac: ?

(In https://github.com/mtrmac/skopeo/tree/go1.24 I’m suggesting to rely on this data in order to simplify the `Makefile`. That might well be a bad idea for RPM packaging, I don’t know. It’s just that the change here looks so unexpected.)

contrib/cirrus/runner.sh
```diff
@@ -77,6 +77,7 @@ _run_setup() {
     fi
     msg "Copying test binaries from $SKOPEO_CIDEV_CONTAINER_FQIN /usr/local/bin/"
     cp -a "$mnt/usr/local/bin/"* "/usr/local/bin/"
+    cp "$mnt/usr/bin/registry" "/usr/local/bin/"
```
mtrmac: The tests are _supposed_ to be looking in `$PATH`. I’m confused about how that is failing.

Makefile
lsm5:
```
$ sudo make test-integration
podman run --rm -i -e TESTFLAGS="-timeout=15m" -e CI= -e SKOPEO_CONTAINER_TESTS=1 -t --security-opt label=disable --cap-add=cap_mknod -v /home/lsm5/repositories/containers/skopeo:/src/github.com/containers/skopeo -w /src/github.com/containers/skopeo quay.io/libpod/skopeo_cidev:c20250422t130822z-f42f41d13 \
	make test-integration-local
go build "-buildmode=pie" -ldflags '-X main.gitCommit= ' -gcflags "" -tags "  " -o bin/skopeo ./cmd/skopeo
error obtaining VCS status: exit status 128
	Use -buildvcs=false to disable VCS stamping.
make: *** [Makefile:134: bin/skopeo] Error 1
make: *** [Makefile:200: test-integration] Error 2
```

This was added just to make it go away on my local environment runs. I can undo this change, no worries.

contrib/cirrus/runner.sh
lsm5: the test run in cirrus is actually `test-integration-local` but we're fetching binaries from the image. I guess we can update the base fedora CI image with docker-distribution installed instead of this change.

Makefile
mtrmac: That’s interesting. I never saw that. Is that something that’s likely to happen in CI / RPM builds?

contrib/cirrus/runner.sh
mtrmac: Oh, my mistake, I didn’t see the `$mnt` part.

Now it makes sense to me why the tests were failing, and how this fixes it.

Makefile
lsm5: AFAICT, not a problem on RPM / packit / copr builds as it's built from a tarball and not a VCS. [Builds on podman-next copr](https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/package/skopeo/) seem fine so far.

lsm5: not seeing this anymore on my local env. I've undone the change.

lsm5:
@mtrmac @cevich PTAL. Related: https://github.com/containers/automation_images/pull/410

cevich:
I'm sorry, I won't be able to give this a proper review based on my current responsibilities.

PR #2667, opened 2025-08-01
[DO NOT MERGE] Test c/image #2907

TomSweeneyRedHat:
Do a quick vendor test of c/image with a proposed PR for cherry picking into c/image v5.35.*.

Proposed PR: https://github.com/containers/image/pull/2907

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ go 1.23.3
 require (
 	github.com/Masterminds/semver/v3 v3.4.0
 	github.com/containers/common v0.64.0
-	github.com/containers/image/v5 v5.36.0
+	github.com/containers/image/v5 v5.36.1-0.20250724133842-4b4473bb78d5
 	github.com/containers/ocicrypt v1.2.1
 	github.com/containers/storage v1.59.0
 	github.com/docker/distribution v2.8.3+incompatible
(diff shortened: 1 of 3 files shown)


mtrmac:
@TomSweeneyRedHat I’m not sure this adds much — the "Skopeo Test" tasks in https://github.com/containers/image/pull/2907 already trigger a Skopeo test with an updated c/image.

The thing that I think we need to manage for branching is that the `IMAGE_SUFFIX` update from https://github.com/containers/image/pull/2907 must be paired (both present, or both absent) with https://github.com/containers/skopeo/pull/2661 . I.e. now that c/image’s `release-5.36` branch is created before the c/image update, it should point at a Skopeo branch before the Skopeo update — or the update needs to be backported to the c/image branch .

If the two are not in sync, Skopeo CI will always work (as long as Skopeo is internally consistent WRT the CI image and the integration test path) – but c/image CI will break (it will run on a c/image-owned image choice against a Skopeo-owned integration test content).

TomSweeneyRedHat:
@mtrmac got it, I think!  I just cherry picked that change in c/image into the release-5.36 branch there.  I'll vendor that into here.

TomSweeneyRedHat:
@mtrmac I have vendored from https://github.com/containers/image/pull/2913/commits/40d10270ef5ead65fb0ddc84bb1dbdae8c237a60 this time.  Fingers crossed!

mtrmac:
@TomSweeneyRedHat this is a solid proof that https://github.com/containers/image/pull/2913 doesn’t _hurt_, but it changes ~nothing in the Skopeo repo.

PR #2679, opened 2025-08-19
Update tests to work with the Sequoia c/image backend

mtrmac:
Currently, if a key is not found, the GPG mechanism reports
> Invalid GPG signature: {$GoStructDump}

while the Sequoia one reports
> $keyFingerprint was not found

Accept both.

This is split from https://github.com/containers/skopeo/pull/2645 , to make it easier to sequence the PRs: First this PR, then https://github.com/containers/image/pull/2876 (which will pass tests), and then https://github.com/containers/skopeo/pull/2645 .


mtrmac:
@Luap99 PTAL

Luap99 (approved):
LGTM

PR #2681, opened 2025-08-20
Allow running CI from rawhide images

mtrmac:
We will, temporarily, want to run from rawhide images that include the rust-podman-sequoia RPM, as well as Fedora 43 images.

We have only one `skopeo_cidev` though, and we don't want to build a rawhide variant; so, allow consuming it from rawhide.

A prerequisite for Sequoia https://github.com/containers/image/pull/2876 , see https://github.com/containers/image/pull/2876/checks?check_run_id=48518809749 .

@lsm5 PTAL.

### DIFF
--- a/contrib/cirrus/runner.sh
+++ b/contrib/cirrus/runner.sh
@@ -71,8 +71,8 @@ _run_setup() {
     # automation, but the sources are in different directories.  It's
     # possible for a mismatch to happen, but should (hopefully) be unlikely.
     # Double-check to make sure.
-    if ! fgrep -qx "ID=$OS_RELEASE_ID" $mnt/etc/os-release || \
-       ! fgrep -qx "VERSION_ID=$OS_RELEASE_VER" $mnt/etc/os-release; then
+    if ! grep -Fqx "ID=$OS_RELEASE_ID" $mnt/etc/os-release || \
+       ! grep -Fqx "VERSION_ID=$OS_RELEASE_VER" $mnt/etc/os-release; then
             die "Somehow $SKOPEO_CIDEV_CONTAINER_FQIN is not based on $OS_REL_VER."
     fi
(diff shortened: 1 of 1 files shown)


lsm5:
@mtrmac LGTM if we need this happening soon. @cevich PTAL.

BTW, with the testing-farm jobs already in place and assuming #2640 will get merged (hopefully soon), would we need to worry about having this? IIRC, one reason to still have this would be to run tests *locally* without polluting your environment, but maybe the same could be achieved by running a toolbox container (?)

cevich:
Another use-case could be `get_ci_vm.sh` but I'm doubting that gets used here much.  I can't tell, but was a temporary commit used to confirm this logic actually works with rawhide?

lsm5 (approved):
LGTM

@mtrmac please merge whenever convenient.

lsm5:
Ignore f43 testing-farm job. Compose isn't ready yet AFAICT. The testing-farm people will hopefully have a process change in place for f44 so that issue shouldn't occur from the next release forward.

mtrmac:
> BTW, with the testing-farm jobs already in place and assuming #2640 will get merged (hopefully soon), would we need to worry about having this?

We _will_ need CI coverage of the Sequoia build configuration, probably in situations consistent with how we build distribution RPMs.

> IIRC, one reason to still have this would be to run tests _locally_ without polluting your environment

I’m not positively aware of anyone doing this — I rely on GitHub-connected CI ~exclusively. So I don’t have a strong opinion on whether / how that should be possible.

mtrmac:
> I can't tell, but was a temporary commit used to confirm this logic actually works with rawhide?

c/image #2876 was building against these images ; I’m trying to split the Skopeo PR into the CI-prerequisite + new-feature parts , and I didn’t actually test that the split is correct — I failed in a previous iteration.

Hopefully this will get it right; if not, I’ll do a proper isolated test c/image PR.

PR #2684, opened 2025-08-23
highlight docker-archive / oci-archive streaming

h0nIg:
docker-archive refers to /dev/stdin on containers/image, which is missing here. In addition oci-archive should refer to /dev/stdout


mtrmac (changes requested):
Thanks!

See https://github.com/containers/image/pull/2939 for substantially equivalent comments.

Also, `containers-transports(5)` is the reference documentation for transports, and one of the purposes of that is that every single user of the transport mechanism doesn’t need to repetitively document all details. A highlight of the typical use case and a pointer to the details should be enough — even the highlight we have here is, arguably, already redundant.

PR #2687, opened 2025-08-27
Follow up on #2654

lsm5:
Please review and merge #2654 first if that makes things simpler.

See individual commits. This makes Packit config and process simpler. Reusability of TMT tests is also much easier:

See: https://src.fedoraproject.org/rpms/skopeo/pull-request/83# and https://github.com/containers/container-libs/pull/287


mtrmac (commented):
I know very little about Packit, and this looks plausible overall.

Just to be sure nothing is missed, @jnovy please speak up if there were any downstream impacts.

lsm5:
Btw, this will not affect centos stream/rhel as Packit isn't used for updates there.

lsm5:
Rebased on main. I'll handle Makefile cleanups to remove unused targets in another PR later.

@jnovy @mtrmac good for a final review.

mtrmac:
I was more worried about `dist-git` depending on some files existing in the upstream repo — but, also, I didn’t (yet) take the time to look.

lsm5:
@mtrmac Ack, sure thing. If it helps ...

```
$ sudo dnf repoquery --whatrequires skopeo-tests
Updating and loading repositories:
Repositories loaded.
```
.. no packages listed.

Also, see the CI jobs on this dist-git PR for corresponding downstream tests: https://src.fedoraproject.org/rpms/skopeo/pull-request/83# where we removed dependency on skopeo-tests in TMT jobs. That diff isn't dependent on skopeo-tests anymore for CI jobs.

The `rpm/gating.yaml` removed didn't matter to upstream anyway, and it does exist downstream too where it matters. Earlier, I was thinking of having *everything* upstream, but if the file isn't gonna change often downstream, then it makes no sense for these files to exist upstream and get synced via the files_to_sync Packit feature.

So, I'd say we're fine.

jnovy:
I'm more fond of skopeo-tests still being shipped as (meta)package - the reason is simple - if we ever decide to run a different test suite than TMT we need to reintroduce it again - or manually rewrite dependencies again into a new test framework with every framework change. The existence of skopeo-tests just takes care of it independently.

lsm5:
> I'm more fond of skopeo-tests still being shipped as (meta)package - the reason is simple - if we ever decide to run a different test suite than TMT we need to reintroduce it again - or manually rewrite dependencies again into a new test framework with every framework change. The existence of skopeo-tests just takes care of it independently.

ya, metapackage, to confirm meaning something that only installs the test dependencies sounds like a great thing. I can do that and skip the test files installation.

lsm5:
Ready for another review. skopeo-tests now only installs the test dependencies, no test files anymore. Also, I added an extra commit to re-enable ELN Packit jobs. Made `fakeroot` a Recommends in the rpm spec and that's working on ELN too.

PTAL @mtrmac @jnovy

mtrmac (approved):
> The `rpm/gating.yaml` removed didn't matter to upstream anyway, and it does exist downstream too where it matters.

Ah, indeed, and same for `plans/`. That resolves that concern.

@jnovy ?

lsm5:
rebased on main, no code changes.

jnovy:
For gating.yaml - yes, it's important only downstream where it is already present in dist-git and CentOS gitlab. @lsm5 @mtrmac

lsm5:
Merging..

PR #2690, opened 2025-08-29
Packit ELN testing

lsm5:
testing-farm public ranch now provides ELN composes which enables super-early testing of the next RHEL.

PR #2698, opened 2025-09-03
Packit: re-enable ELN and CentOS Stream jobs on PRs

lsm5:
See individual commits. These only relate to PR jobs. Downstream jobs not affected.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -41,12 +41,11 @@ jobs:
     # https://packit.dev/posts/openscanhub-prototype
     osh_diff_scan_after_copr_build: false
 
-  # Ignore until golang is updated in distro buildroot to go 1.23.3+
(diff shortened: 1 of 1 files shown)


lsm5:
@mtrmac I enabled centos stream jobs in this PR in addition to the commit from #2690 (I'll close that one).

jnovy:
@lsm5 regarding the CentOS - packit always need to reference an approved JIRA or MRs created by it are unmergeable.

lsm5:
> @lsm5 regarding the CentOS - packit always need to reference an approved JIRA or MRs created by it are unmergeable.

Yup, this is only for copr builds on PRs and the subsequent testing-farm tests. The `propose_downstream` job for centos-stream is still set to ignore. I'll update the PR description.

Issue #2633, opened 2025-06-25
CI Switch: Cirrus -> TMT


lsm5:
I'd like to switch CI from Cirrus to TMT and get rid of Cirrus (as much as possible if not fully). Filing this to determine the work that needs to be done toward this:

Current list of tests in Cirrus (along with TMT status)

**In TMT (main and/or needs forward porting):**
- system (main)
- integration (needs forward porting)
- validate (needs forward porting)
- unit (needs forward porting)

**Not yet in TMT:**
- doccheck / validate-docs (doable)
- cross (doable)
- osx (unlikely in the near future, OSX not available via TMT)
- ostree-rs-ext (still needed?)
- w/ opengpg (still needed?)

**Additional requirements for TMT:**
- Stable testing images for no surprises like dependency breakage (hopefully not that big an issue for skopeo,  I'll check further with TMT team)

mtrmac:
> ostree-rs-ext (still needed?)

We do have some tests of the proxy code in `integration/proxy_test.go`, but an integration test against the primary user still seems valuable. OTOH _maybe_ that would be better modeled as a RPM reverse dependency test? I have no idea. (Cc: @cgwalters )

> w/ opengpg (still needed?)

Some users vehemently ask a dependency-free (probably Go-only) stand-alone binary that can be run in any environment; they probably need that build. So, I think we should be testing that. I can imagine downgrading this to a compile test + a minimal smoke-test, but with my upstream hat on, we should have _some_ tests. (Compare https://github.com/containers/image/issues/2877 , build tags which we don’t test have a tendency to rot.)

lsm5:
@mtrmac ack and ack, thanks.

I think w/ opengpg shouldn't be too hard to add. Just a matter of skipping the rpm and doing a separate build with the added buildtag.

I had some trouble with enabling ostree-rs-ext sometime ago, but I'll check again.

cgwalters:
> ostree-rs-ext (still needed?)

In theory with testing-farm AIUI there's some support for reverse dependency testing, so triggering the bootc tests on changes to skopeo both upstream and downstream should be doable.

lsm5:
> > ostree-rs-ext (still needed?)
> 
> In theory with testing-farm AIUI there's some support for reverse dependency testing, so triggering the bootc tests on changes to skopeo both upstream and downstream should be doable.

Yup, in progress at https://github.com/containers/skopeo/pull/2640 . Thanks.

PR #2700, opened 2025-09-04
Packit/RPM: build with sequoia for f43+ and couple of other changes for distro conditionals and easier testing

lsm5:
See individual commits.

### DIFF
--- a/rpm/skopeo.spec
+++ b/rpm/skopeo.spec
@@ -9,21 +9,15 @@
 
 %global gomodulesmode GO111MODULE=on
 
-# No btrfs on RHEL
+# Distro and environment conditionals
 %if %{defined fedora}
+# Fedora conditionals
 %define build_with_btrfs 1
-%endif
-
-%if %{defined rhel}
-%define fips 1
-%endif
-
-# Only used in official koji builds
-# Copr builds set a separate epoch for all environments
-%if %{defined fedora}
 %define conditional_epoch 1
 %else
+# RHEL conditionals
 %define conditional_epoch 2
+%define fips 1
 %endif
 
 # set higher Epoch only for podman-next builds
--- a/rpm/skopeo.spec
+++ b/rpm/skopeo.spec
@@ -14,6 +14,9 @@
 # Fedora conditionals
 %define build_with_btrfs 1
 %define conditional_epoch 1
+%if %{?fedora} >= 43
+%define sequoia 1
+%endif
 %else
 # RHEL conditionals
 %define conditional_epoch 2
@@ -66,6 +69,9 @@ BuildRequires: make
 BuildRequires: shadow-utils-subid-devel
 BuildRequires: sqlite-devel
 Requires: containers-common >= 4:1-21
+%if %{defined sequoia}
+Requires: podman-sequoia
+%endif
 
 %description
 Command line utility to inspect images and repositories directly on Docker
@@ -122,6 +128,10 @@ export BUILDTAGS="$BASEBUILDTAGS exclude_graphdriver_btrfs"
 
 %if %{defined fips}
 export BUILDTAGS="$BUILDTAGS libtrust_openssl"
+%endif
+
+%if %{defined sequoia}
+export BUILDTAGS="$BUILDTAGS containers_image_sequoia"
 %endif
 
 # unset LDFLAGS earlier set from set_build_flags
--- a/rpm/skopeo.spec
+++ b/rpm/skopeo.spec
@@ -26,8 +26,13 @@
 %define conditional_epoch 2
 %endif
 
+# set higher Epoch only for podman-next builds
+%if %{defined copr_username} && "%{copr_username}" == "rhcontainerbot" && "%{copr_projectname}" == "podman-next"
+%define next_build 1
+%endif
+
 Name: skopeo
-%if %{defined copr_username}
+%if %{defined next_build}
 Epoch: 102
 %else
 Epoch: %{conditional_epoch}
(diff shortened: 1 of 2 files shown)


lsm5:
@mtrmac PTAL. Do we need anything apart from depending on podman-sequoia and building with `containers_image_sequoia` ?

cc @jnovy

rpm/skopeo.spec
```diff
@@ -9,25 +9,27 @@
 
 %global gomodulesmode GO111MODULE=on
 
-# No btrfs on RHEL
+# Distro and environment conditionals
```
mtrmac, line 12: (I’m not sure that centralizing the decisions by distro makes things easier — I’m a bit worried that this will make it easier to overlook a variable in one of the branches — but that’s entirely up to you.)

mtrmac:
@lsm5 I see testing-farm tests failing on F43/rawhide now, e.g. https://artifacts.dev.testing-farm.io/8be93fae-957e-41cf-963e-344e0daa8673/ in “signing” because of an unexpected “Missing key” error text.

I think that’s because the CI images use `rust-podman-sequoia` 0.1.0, and the RPMs are now 0.2.0 (specifically https://github.com/ueno/podman-sequoia/pull/7/commits/d41fefaa1fa35ee56c6a3aac0105232733460932 replaces the expected “was not found” error message).

It would be cleanest to update the CI images to 0.2.0, _and_ to include `rust-podman-sequoia` in the F43 image, for consistency — but, short-term, that should be coordinated with https://github.com/containers/podman/pull/27017#issuecomment-3267081316 .

lsm5:
> @lsm5 I see testing-farm tests failing on F43/rawhide now, e.g. https://artifacts.dev.testing-farm.io/8be93fae-957e-41cf-963e-344e0daa8673/ in “signing” because of an unexpected “Missing ke y” error text.
> 
> I think that’s because the CI images use `rust-podman-sequoia` 0.1.0, and the RPMs are now 0.2.0 (specifically [ueno/podman-sequoia@d41fefa](https://github.com/ueno/podman-sequoia/commit/d41fefaa1fa35ee56c6a3aac0105232733460932) replaces the expected “was not found” error message).

So, IIUC, the skopeo tests need to be fixed themselves to account for v0.2.0 podman-sequoia, is that right? Once that's done, the testing-farm jobs should go green. 

Just to ensure we're on the same page RE: testing-farm jobs, the CI images maintained via automation_images will not have any effect on testing-farm jobs. They'll just fetch the composes + repos maintained by the testing-farm team.

> It would be cleanest to update the CI images to 0.2.0, _and_ to include `rust-podman-sequoia` in the F43 image, for consistency — but, short-term, that should be coordinated with [containers/podman#27017 (comment)](https://github.com/containers/podman/pull/27017#issuecomment-3267081316) .

Looks like podman-sequoia is [already being included](https://github.com/containers/automation_images/blob/main/cache_images/fedora_packaging.sh#L180) ?

----
BTW, WDYT about getting rid of Cirrus for cases where we need to use podman-sequoia and only use Packit/TMT jobs? That'll help us proceed toward https://github.com/containers/skopeo/issues/2633 . For safety, we can just disable all Cirrus jobs but still retain the config. 

For the immediate term, I'm cool with fixing automation_images in the open PR itself if that's what it takes to get CI going. 

I've got some WIP to enable the other test jobs via Packit/TMT, but I'm gonna be away the rest of this week and will only resume next week.

lsm5:
> Looks like podman-sequoia is [already being included](https://github.com/containers/automation_images/blob/main/cache_images/fedora_packaging.sh#L180) ?

NVM, I see it was only for rawhide.

mtrmac:
> So, IIUC, the skopeo tests need to be fixed themselves to account for v0.2.0 podman-sequoia, is that right? Once that's done, the testing-farm jobs should go green.

Yes. I’ll update the tests in https://github.com/containers/skopeo/pull/2645 .

> BTW, WDYT about getting rid of Cirrus for cases where we need to use podman-sequoia and only use Packit/TMT jobs? That'll help us proceed toward #2633 .

I don’t have much of an opinion. As noted in #2633, we do want to test _some_ builds with options that are not shipped in Fedora / CentOS, so I’m not sure we can get rid of Cirrus entirely, but for those cases that are ~equivalent, up to you.

Do note that, right now, Packit is only running system tests, not integration tests (and I’m not sure about unit tests). So I don’t think we can cleanly switch _at this moment_.

> For the immediate term, I'm cool with fixing automation_images in the open PR itself if that's what it takes to get CI going.

`automation_images` are (were?) being updated for Podman, I was told not to modify that repo yesterday. So, for now, I’ll update the Skopeo tests to accept both versions.

mtrmac:
> BTW, WDYT about getting rid of Cirrus for cases where we need to use podman-sequoia and only use Packit/TMT jobs?

I didn’t collect numbers — purely anecdotally, the `testing-farm` jobs tend to take longer than Cirrus ones. Is there something we can easily do to improve that?

lsm5:
> > BTW, WDYT about getting rid of Cirrus for cases where we need to use podman-sequoia and only use Packit/TMT jobs?
> 
> I didn’t collect numbers — purely anecdotally, the `testing-farm` jobs tend to take longer than Cirrus ones. Is there something we can easily do to improve that?

I suspect it's because of the testing-farm jobs depending on the copr build jobs. We could have testing-farm jobs independent of copr builds, but perhaps for system tests, ostree tests (WIP) we should depend on the rpm. Validate, unit and maybe even integration could run independently, so those should go faster.

PR #2703, opened 2025-09-08
test-integration-local: use SKOPEO_BINARY if set

lsm5:
Else use ./bin/skopeo as the default binary.

This makes it a lot more flexible compared to just searching and using the first skopeo in PATH. Also avoids any binary installation to /usr/bin.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -201,11 +201,15 @@ test-integration:
 		$(MAKE) test-integration-local
 
 
+# Helper target to set up SKOPEO_BINARY variable for local test targets
+.eval-skopeo-binary: $(if $(SKOPEO_BINARY),,bin/skopeo)
+	$(eval SKOPEO_BINARY := $(or $(SKOPEO_BINARY),./bin/skopeo))
+	@echo "Testing with $(SKOPEO_BINARY) ..."
+
 # Primarily intended for CI.
-test-integration-local: bin/skopeo
+test-integration-local: .eval-skopeo-binary
 	hack/warn-destructive-tests.sh
-	$(MAKE) PREFIX=/usr install
-	cd ./integration && $(GO) test $(SKOPEO_LDFLAGS) $(TESTFLAGS) $(if $(BUILDTAGS),-tags "$(BUILDTAGS)")
+	cd ./integration && SKOPEO_BINARY="$(abspath $(SKOPEO_BINARY))" $(GO) test $(SKOPEO_LDFLAGS) $(TESTFLAGS) $(if $(BUILDTAGS),-tags "$(BUILDTAGS)")
 
 # complicated set of options needed to run podman-in-podman
 test-system:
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -202,10 +202,10 @@ test-integration:
 
 
 # Primarily intended for CI.
-test-integration-local: bin/skopeo
+test-integration-local: $(if $(SKOPEO_BINARY),,bin/skopeo)
 	hack/warn-destructive-tests.sh
-	$(MAKE) PREFIX=/usr install
-	cd ./integration && $(GO) test $(SKOPEO_LDFLAGS) $(TESTFLAGS) $(if $(BUILDTAGS),-tags "$(BUILDTAGS)")
+	@echo "Testing with $(or $(SKOPEO_BINARY),$(eval SKOPEO_BINARY := "./bin/skopeo")$(SKOPEO_BINARY)) ..."
```
mtrmac: Do the defaulting once so that it can’t get out of sync?

integration/utils_test.go
```diff
@@ -20,7 +20,12 @@ import (
 	"go.podman.io/image/v5/manifest"
 )
 
-const skopeoBinary = "skopeo"
+var skopeoBinary = func() string {
+	if binary := os.Getenv("SKOPEO_BINARY"); binary != "" {
+		return binary
+	}
+	return "skopeo"
```
mtrmac, line 29: (Non-blocking: If we are serious about relying on `SKOPEO_BINARY`, maybe this should fail … and be a part of `SetupSuite` or something like that. OTOH that would be a fairly invasive change, right now `integration/utils_test.go` only refers to constants and globals.)

Makefile
lsm5: added a build-in target to evaluate the variable. Same path will be used for system and integration tests and can be reused for any other tests if required.

integration/utils_test.go
lsm5: Added a FIXME comment.

lsm5:
@mtrmac ready for another review

integration/utils_test.go
```diff
@@ -20,7 +20,14 @@ import (
 	"go.podman.io/image/v5/manifest"
 )
 
-const skopeoBinary = "skopeo"
+// FIXME: Move to SetuoSuite
```
mtrmac:
```suggestion
// FIXME: Move to SetupSuite
```

mtrmac (commented):
Thanks! LGTM other than that one possibly-confusing typo, good to merge afterwards.

mtrmac:
@lsm5 Sign-off enforcement is apparently expecting a match against … something (GitHub account email?) per https://github.com/containers/skopeo/pull/2703/checks?check_run_id=50510598725 .

lsm5:
> @lsm5 Sign-off enforcement is apparently expecting a match against … something (GitHub account email?) per https://github.com/containers/skopeo/pull/2703/checks?check_run_id=50510598725 .

Thanks. Accepting commit suggestions by someone else via the GitHub UI breaks DCO. I'll keep that in mind for the future. Fixed now. I'll need an approving review from you.

mtrmac (approved):

PR #2710, opened 2025-09-17
sync: honor CLI/global tls-verify unless YAML explicitly sets tls-verify

andya1lan:
## Background and problem

- Using `skopeo sync`, even when passing `--src-tls-verify=false` (or the deprecated global `--tls-verify=false`), if the source is YAML (`--src yaml`) and the YAML does not explicitly set `tls-verify`, sync will verify source certificate.
- This PR aligns behavior with the documented intent: TLS verification is normally enabled, YAML's `tls-verify` can explicitly control the source side, and if YAML omits `tls-verify`, CLI flags should be able to disable verification for the source.

Repro (YAML without `tls-verify`):

```bash
# YAML (tls-verify not set)
cat > source.yaml <<'YAML'
registry.example.com:
  images:
    skopeo: [ "latest" ]
YAML

# Expectation: skip TLS
skopeo sync --src yaml --dest dir --src-tls-verify=false source.yaml /skopeo
```

## Root cause analysis

- In `cmd/skopeo/sync.go` within `imagesToCopyFromRegistry`, the code used to unconditionally write YAML-derived `tls-verify` values into the per-registry `SystemContext`:
  - `DockerInsecureSkipTLSVerify`
  - `DockerDaemonInsecureSkipTLSVerify`
- Even if YAML did not set `tls-verify` (parsed as `types.OptionalBoolUndefined`), this would overwrite the incoming CLI/global values and effectively revert to “verify certificates” defaults.
- By contrast, the `copy` command correctly respects CLI options when constructing its `SystemContext`.

## What changed

Establish explicit precedence for the source side consistent with the docs:
  - If YAML sets `tls-verify: true`, verification is enforced (CLI cannot disable).
  - If YAML sets `tls-verify: false`, verification is skipped.
  - If YAML omits `tls-verify`, the CLI/global flags govern (so `--src-tls-verify=false` can disable verification).

Key logic adjustment (described):
- File: `cmd/skopeo/sync.go`
  - In `imagesToCopyFromRegistry`, guard TLS overrides behind a check:
    - If `cfg.TLSVerify.skip != types.OptionalBoolUndefined` then set:
      - `serverCtx.DockerInsecureSkipTLSVerify = cfg.TLSVerify.skip`
      - `serverCtx.DockerDaemonInsecureSkipTLSVerify = (cfg.TLSVerify.skip == types.OptionalBoolTrue)`

## Behavior and compatibility

- YAML omits `tls-verify`:
  - Before (bug): sync ignored CLI/global `--src-tls-verify`, and verified certs by default.
  - After (fixed): sync respects CLI/global settings.
- YAML explicitly sets `tls-verify` (true/false): unchanged; YAML takes precedence.
- Net effect: bug fix, no breaking changes; only correctness in the previously mentioned case.

## Tests added

- File: `cmd/skopeo/sync_test.go`
  - `TestTLSOverridePrecedence` covers precedence and scenarios:
    1) YAML unspecified: preserve incoming `SystemContext.DockerInsecureSkipTLSVerify`.
    2) YAML `tls-verify: true`: enforce verification (skip=false).
    3) YAML `tls-verify: false`: disable verification (skip=true) and set `DockerDaemonInsecureSkipTLSVerify=true`.
  - Existing `TestTLSVerifyConfig` continues to validate YAML bool parsing (true/false/unspecified).

## Manual verification matrix

- CLI `--src-tls-verify=false`, YAML unspecified → should skip TLS
- CLI not set, YAML `tls-verify: false` → should skip TLS
- CLI `--src-tls-verify=false`, YAML `tls-verify: true` → should verify TLS (YAML wins)

Note: YAML only influences the source side. Destination TLS behavior is controlled exclusively by CLI flags (e.g., `--dest-tls-verify=false`).

## Performance and security

- Only a small control-flow change, no performance impact.
- Improves security posture by ensuring user-requested TLS semantics are honored consistently.

## Change list

- fix(sync): only override TLS from YAML when explicitly set (`cmd/skopeo/sync.go`)
- test(sync): add `TestTLSOverridePrecedence` to cover YAML/CLI precedence (`cmd/skopeo/sync_test.go`)


cmd/skopeo/sync_test.go
```diff
@@ -59,3 +59,59 @@ func TestSync(t *testing.T) {
...
+	baseRegistry := "example.com"
+	imageName := "repo"
+	// Provide an explicit tag list to avoid network calls.
+	cfgBase := registrySyncConfig{
+		Images: map[string][]string{
+			imageName: {"latest"},
+		},
+	}
+
+	// Case 1: YAML does NOT set tls-verify (OptionalBoolUndefined) -> preserve incoming skip=true
```
mtrmac: (Non-blocking: A table-driven test would be less repetitive.)

mtrmac (commented):
Thanks!

cmd/skopeo/sync_test.go
andya1lan: Thanks for your advice! I have refactored it into a table-driven test.

cmd/skopeo/sync_test.go
```diff
@@ -59,3 +59,64 @@ func TestSync(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			src := types.SystemContext{DockerInsecureSkipTLSVerify: tt.incomingSkip}
+			cfg := cfgBase
+			if tt.yamlSkip != types.OptionalBoolUndefined {
+				cfg.TLSVerify = tlsVerifyConfig{skip: tt.yamlSkip}
+			} else {
+				cfg.TLSVerify = tlsVerifyConfig{skip: types.OptionalBoolUndefined}
+			}
```
mtrmac: (Non-blocking: This can just be an unconditional `cfg.TLSVerify = … tt.yamlSkip`.)

cmd/skopeo/sync_test.go
```diff
@@ -59,3 +59,64 @@ func TestSync(t *testing.T) {
...
+			incomingSkip:   types.OptionalBoolFalse,
+			yamlSkip:       types.OptionalBoolTrue,
+			wantSkip:       types.OptionalBoolTrue,
+			wantDaemonSkip: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			src := types.SystemContext{DockerInsecureSkipTLSVerify: tt.incomingSkip}
```
mtrmac: (Non-blocking: Also setting `DockerDaemonInsecureSkipTLSVerify` would be more representative — in particular for the  `yamlSkip: …Undefined` case.)

andya1lan: Thanks for inspiring me. I have added `DockerDaemonInsecureSkipTLSVerify` to make sure they are are preserved independently. Also, I think it could be better changing the test cases focus more on the yaml omitted scenarios and I have pushed the new commit.

mtrmac (commented):
Overzealous AI:
- Now this only tests `yamlSkip: …Undefined`, discarding previously-existing tests for cases where YAML defines values (testing which one is prioritized)
- (“mismatched incoming” is theoretically interesting but practically doesn’t happen. That doesn’t mean it _must not_ be tested)

cmd/skopeo/sync_test.go
mtrmac: Crossed wires … yes, this PR cares about the case where the YAML value is omitted, and it’s fine to test that however thoroughly is warranted. But, I think longer-term, the case where _both_ define a value, and thus arguably either choice “could be correct”, is the one more likely to accidentally break, so that’s where test coverage is more valuable.

andya1lan:
Agreed. I have restored the test cases when `tls-verify` is specified in the yaml on my latest commit.
Also sorry to say I never used skopeo sync with remote docker daemon so I asked Copilot for help. Now I have reverted the changes. Could you help me add `DockerDaemonInsecureSkipTLSVerify` context if you think it would be more representative?

andya1lan:
Please ignore my word last night. It was so night and late that I didn't read your comments carefully but only deleted `incomingDaemonSkip` and `DockerDaemonInsecureSkipTLSVerify` then pushed without validation. 
Today I looked into the documents and code, and I am now re-adding the tests covered both yaml omitted cases and specified cases as you said both are necessary to be included in unit tests.
I think copilot's mismatched incoming case is valuable just as you mentioned, even skopeo sync does not support docker daemon as source currently, it is more like a defensive guard in case a future change or refactor could overwrite the field when no one could notice.

mtrmac (approved):
LGTM, let’s not try to fight the AI via GitHub.

Thanks for the fix!

PR #2711, opened 2025-09-19
Simplify tests for #2710

mtrmac:
- Use just one line per test case, to make them easier to understand / compare / contrast
- Actually parse CLI inputs and YAML instead of directly constructing inputs
  - … drop test situations that can’t happen
  - … and add a test for the case where user provides no TLS options at all


lsm5 (approved):
LGTM as best I can tell.

Issue #2701, opened 2025-09-05
`make test-integration` failed with `exec: "registry": executable file not found in $PATH`


bitoku:
When I tried to run test-integration, it failed with `exec: "registry": executable file not found in $PATH`.

The command uses `quay.io/libpod/skopeo_cidev:c20250721t181111z-f42f41d13` image, and it doesn't have `registry` binary in $PATH.

```
❯ podman run --rm -it quay.io/libpod/skopeo_cidev:c20250721t181111z-f42f41d13 /bin/sh
sh-5.2# echo $PATH
/usr/local/bin:/usr/bin
sh-5.2# ls -l /usr/local/bin | grep reg
-rwxr-xr-x 1 root root  49795775 Jul 21 18:36 dockerregistry
-rwxr-xr-x 1 root root  17683642 Jul 21 18:31 registry-v2-schema1
sh-5.2# ls -l /usr/bin | grep reg
sh-5.2# 
```
After I installed `docker-distribution` in the image, the test worked.

These changes could be the cause, but not sure. Did I miss something?
https://github.com/containers/skopeo/pull/2661
https://github.com/containers/automation_images/pull/410

mtrmac:
Thanks for your report!

I’m afraid I don’t think anyone is manually running `make test-integration`; we mostly rely on the pull request automation.

That said, I would expect this to work, and I can’t see any obvious way how this is failing (it seems to be using an older CI image). @lsm5 can you see what is going wrong here?

lsm5:
ugh yup. I made changes to make `test-integration-local` work, but that evidently broke test-integration this way. I would ideally like to get rid of `test-integration` or have it be aliased to just `test-integration-local` and not have to depend on skopeo_cidev image anymore.

@bitoku would you be ok with just running `make test-integration-local` instead? 

@mtrmac wdyt about having integration tests depend on SKOPEO_BINARY variable to set binary path? Currently, it uses skopeo from default PATH which I guess could be an inconvenience in local environments.

mtrmac:
My one somewhat-strong opinion is that the default `Makefile` should be safe; users running `make test-integration` should not find their local machine broken.

I don’t feel strongly about removing `test-integration` entirely. _In principle_ it would be nice not to be reliant on external services, but, in practice, we aren’t using it and we haven‘t been good at maintaining it, so saving everyone the trouble might be for the best.

[...]

PR #2714, opened 2025-09-24
skopeo: add `--require-signed`

jlebon:
In bootc, we want the ability to assert that signature verification is
enforced.

Add a new top-level `--require-signed` switch. When passed, we use the
new `RequireSignatureVerification()` method to ensure that signature
verification is enforced.

Part of https://github.com/containers/skopeo/issues/1829.

### DIFF
--- a/cmd/skopeo/main.go
+++ b/cmd/skopeo/main.go
@@ -31,6 +31,7 @@ type globalOptions struct {
 	registriesConfPath string                  // Path to the "registries.conf" file
 	tmpDir             string                  // Path to use for big temporary files
 	userAgentPrefix    string                  // Prefix to add to the user agent string
+	requireSigned      bool                    // Require any pulled image to be signed
 }
 
 // requireSubcommand returns an error if no sub command is provided
@@ -81,6 +82,7 @@ func createApp() (*cobra.Command, *globalOptions) {
 	rootCommand.PersistentFlags().BoolVar(&opts.debug, "debug", false, "enable debug output")
 	rootCommand.PersistentFlags().StringVar(&opts.policyPath, "policy", "", "Path to a trust policy file")
 	rootCommand.PersistentFlags().BoolVar(&opts.insecurePolicy, "insecure-policy", false, "run the tool without any policy check")
+	rootCommand.PersistentFlags().BoolVar(&opts.requireSigned, "require-signed", false, "require any pulled image to be signed")
 	rootCommand.PersistentFlags().StringVar(&opts.registriesDirPath, "registries.d", "", "use registry configuration files in `DIR` (e.g. for container signature storage)")
 	rootCommand.PersistentFlags().StringVar(&opts.overrideArch, "override-arch", "", "use `ARCH` instead of the architecture of the machine for choosing images")
 	rootCommand.PersistentFlags().StringVar(&opts.overrideOS, "override-os", "", "use `OS` instead of the running OS for choosing images")
@@ -135,6 +137,9 @@ func (opts *globalOptions) before(cmd *cobra.Command, args []string) error {
 	if opts.tlsVerify.Present() {
 		logrus.Warn("'--tls-verify' is deprecated, please set this on the specific subcommand")
 	}
+	if opts.insecurePolicy && opts.requireSigned {
+		return fmt.Errorf("--insecure-policy and --require-signed are mutually exclusive")
+	}
 	return nil
 }
 
@@ -166,7 +171,14 @@ func (opts *globalOptions) getPolicyContext() (*signature.PolicyContext, error) 
 	if err != nil {
 		return nil, err
 	}
-	return signature.NewPolicyContext(policy)
+	pc, err := signature.NewPolicyContext(policy)
+	if err != nil {
+		return nil, err
+	}
+	if opts.requireSigned {
+		pc.RequireSignatureVerification(true)
+	}
+	return pc, nil
 }
 
 // commandTimeoutContext returns a context.Context and a cancellation callback based on opts.
--- a/docs/skopeo.1.md
+++ b/docs/skopeo.1.md
@@ -92,6 +92,10 @@ Path to a policy.json file to use for verifying signatures and deciding whether 
 
 Use registry configuration files in _dir_ (e.g. for container signature storage), overriding the default path.
 
+**--require-signed**
+
+Require that any pulled image must be signed regardless of what the default or provided trust policy file says.
+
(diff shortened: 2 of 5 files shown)


jlebon:
This requires https://github.com/containers/container-libs/pull/355 and still needs tests.

cmd/skopeo/proxy.go
```diff
@@ -246,6 +246,10 @@ func (h *proxyHandler) openImageImpl(args []any, allowNotFound bool) (retReplyBu
 		}
 	}()
 
+	if h.opts.global.rejectInsecurePolicy {
+		policyContext.SetRejectInsecure(true)
```
mtrmac: Didn’t this already happen in `opts.global.getPolicyContext`?

mtrmac (commented):
Thanks! Yes, this is the right place to add the option, and overall it looks reasonable.

The option will also need documentation (linters should be complaining — perhaps vendor from your fork, to trigger a CI run).

The immediate next step is to precisely nail down the semantics (and then a matching implementation), that will happen in https://github.com/containers/container-libs/pull/355 .

cgwalters (approved):

TomSweeneyRedHat:
@jlebon Please see @mtrmac 's question and it looks like this needs a rebase.

mtrmac:
@TomSweeneyRedHat https://github.com/containers/container-libs/pull/355 needs to happen first, either way.

jlebon:
Sorry, I do intend to get back to this. Will try tomorrow, otherwise next week.

jlebon:
Updated for the latest version in https://github.com/containers/container-libs/pull/355 but still needs tests.

mtrmac:
(The c/image feature is available in this repo after #2753 .)

jlebon:
This is now ready for review!

integration/require_signed_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+func (s *requireSignedSuite) SetupSuite() {
+	t := s.T()
+	_, err := exec.LookPath(skopeoBinary)
+	require.NoError(t, err)
+
+	s.gpgHome = t.TempDir()
+	t.Setenv("GNUPGHOME", s.gpgHome)
+
+	// XXX: probably could have a helper to dedupe this across all the integration tests that use GPG at this point
```
mtrmac: The test primarily exercises `skopeo copy`, so it could be placed into the existing `copySuite`.

I think that would be valuable, at least to avoid the extra private key generation; and, potentially in the future (definitely not a part of this PR), to share some setup that pre-pulls images from external registries only once.

integration/require_signed_test.go
```diff
@@ -0,0 +1,131 @@
...
+	assertSkopeoFails(t, ".*Source image rejected: No signature verification policy found for image.*",
+		"--require-signed", "copy",
+		"dir:"+srcDir, "dir:"+destDir2)
+
+	// sign the image
+	manifestPath := filepath.Join(srcDir, "manifest.json")
+	signaturePath := filepath.Join(srcDir, "signature-1")
+	dockerReference := "localhost/test:latest"
+
+	assertSkopeoSucceeds(t, "", "standalone-sign",
```
mtrmac: (Aesthetically I’m generally not much of a fan of using `standalone-sign`, but here, with the `os.Remove(signaturePath)`, it is a good fit.)

mtrmac (commented):
Thanks!

Implementation, incl. tests, LGTM — please reuse the setup in `copySuite` to save a bit of test time.

integration/require_signed_test.go
jlebon: I did this, but note that to actually test it locally, I had to apply:

```diff
diff --git a/Makefile b/Makefile
index 50d17a56..b80c787b 100644
--- a/Makefile
+++ b/Makefile
@@ -198,7 +198,7 @@ test-integration:
 # --cap-add=cap_mknod is important to allow skopeo to use containers-storage: directly as it exists in the callers’ environment, without
 # creating a nested user namespace (which requires /etc/subuid and /etc/subgid to be set up)
 	$(CONTAINER_CMD) --security-opt label=disable --cap-add=cap_mknod -v $(CURDIR):$(CONTAINER_GOSRC) -w $(CONTAINER_GOSRC) $(SKOPEO_CIDEV_CONTAINER_FQIN) \
-		$(MAKE) test-integration-local
+		sh -c 'dnf install -y docker-distribution && $(MAKE) test-integration-local'
 
 
 # Helper target to set up SKOPEO_BINARY variable for local test targets
```

Otherwise, `make test-integration` fails with:

```
    registry_test.go:30:
                Error Trace:    /src/github.com/containers/skopeo/integration/registry_test.go:30
                                                        /src/github.com/containers/skopeo/integration/copy_test.go:79
                                                        /src/github.com/containers/skopeo/vendor/github.com/stretchr/testify/suite/suite.go:211
                                                        /src/github.com/containers/skopeo/integration/copy_test.go:41
                Error:          Received unexpected error:
                                exec: "registry": executable file not found in $PATH
                Test:           TestCopy
```

I guess this works as is in CI here because the VM already has docker-distribution. But I can't tell why it wasn't also added to the skopeo_cidev image. I might be missing something; as it is AIUI anyone running `make test-integration` locally should be hitting this.

mtrmac: I’m afraid the CI vs. `test-integration` situation is a bit of mess, tracked in https://github.com/containers/skopeo/issues/2701 .

mtrmac (approved):
Thanks!

cgwalters:
This PR came back to the forefront of my mind as I was working on something different related to signing.

So we never did anything to wire this up in bootc side right?

I hate to escalate the scope here...but the thing about a CLI option is while it was easy to implement, it doesn't have a good UX - we don't want users to always need to do `bootc upgrade --require-signed` each time. It needs a config. I think we can probably invent one bootc side, but...

One thing I could imagine here is we try to create the ability to have separate policies for "host" vs "app". Another way I'd like to do this is: "require signatures for privileged containers" - that makes the host not special at all.

jlebon:
Hmm wait but don't we already have `ostree-image-signed:` which is hooked to `--enforce-container-sigpolicy`? I.e. it's already a one-time thing at `switch` or `install` time and then at `upgrade` time bootc could spawn the skopeo proxy with `--require-signed` when in enforcing mode and we can drop the hacky `policy.json` parsing bits. At least that's how I was picturing it when I wrote this. Supporting this through the proxy API would be nicer though. (I meant to keep pushing all this forward on the side, but...)

PR #2720, opened 2025-10-03
Allow using sigstore keys with empty passphrases

### LINKED ISSUE #2712: Blank passphrase not allowed on signing key
When trying to sync or copy images with `--sign-by-sigstore-private-key` using a key that doesn't have a passphrase I get the error

```
FATA[0002] Error copying ref "docker://ghcr.io/siderolabs/glibc:2.41": private key passphrase not provided
```

The same key works with cosign via
```
cosign sign -y --key cosign.key $IMAGE
```

If my key has a passphrase the error is not shown and copy continues.

mtrmac:
Fixes #2712 . See individual commit messages for details.

### DIFF
--- a/cmd/skopeo/utils_test.go
+++ b/cmd/skopeo/utils_test.go
@@ -414,6 +414,30 @@ func TestSharedCopyOptionsCopyOptions(t *testing.T) {
 				ReportWriter:                     &someStdout,
 			},
 		},
+		{ // --sign-passphrase-file + --sign-by-sigstore-private-key work
+			options: []string{
+				"--sign-by-sigstore-private-key", "/some/key/path.private",
+				"--sign-passphrase-file", passphraseFile.Name(),
+			},
+			expected: copy.Options{
+				SignPassphrase:                   "test-passphrase",
+				SignBySigstorePrivateKeyFile:     "/some/key/path.private",
+				SignSigstorePrivateKeyPassphrase: []byte("test-passphrase"),
+				ReportWriter:                     &someStdout,
+			},
+		},
+		{ // --sign-passphrase-file + --sign-by-sigstore-private-key work with an empty passphrase
+			options: []string{
+				"--sign-by-sigstore-private-key", "/some/key/path.private",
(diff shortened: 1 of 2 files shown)


cmd/skopeo/utils_test.go
```diff
@@ -367,107 +366,120 @@ func fakeSharedCopyOptions(t *testing.T, cmdFlags []string) *sharedCopyOptions {
...
+			},
+			expectedSigner: true,
 		})
-		res, cleanup, err = opts.copyOptions(&someStdout)
+	}
+	for _, c := range c {
+		opts := fakeSharedCopyOptions(t, c.options)
+		res, cleanup, err := opts.copyOptions(&someStdout)
 		require.NoError(t, err)
 		defer cleanup()
```
aguidirh, line 461: LGTM

nit: `defer cleanup()` will run only after the loop finishes, is it possible to run the cleanup() at the end of each loop to avoid resources leaking or exhaustion? 

Or is this anything that it needs to be kept there for re usage ?

mtrmac: You’re right that the cleanup could be done earlier, and that it would be conceptually cleaner.

I don’t think it matters much either way _here_; this is a short-running test, and the loop iterations are all listed just above (5–6 of them).

The cost would be adding another `func(){…}()` nesting level — a very minor downside in readability.

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
@aguidirh PTAL.  If you give this a LGTM, we'll move it along.

TomSweeneyRedHat:
@aguidirh "LGTM" works just fine, TYVM!

mtrmac:
There is no merge bot here… just an (out-of-sync) GitHub list of user accounts allowed to click on the merge button.

mtrmac:
@TomSweeneyRedHat / @Luap99 , hit the button, please.

Luap99 (approved):
LGTM

PR #2724, opened 2025-10-15
main: Add support for overriding HTTP User-Agent

cgwalters:
I want this for https://github.com/bootc-dev/bootc/issues/1686 so we can distinguish pulls there.

But more generally it's can be a good idea for people writing scripts using skopeo to set custom user agents so that registries can more easily trace which actors are performing tasks.

### DIFF
--- a/integration/user_agent_test.go
+++ b/integration/user_agent_test.go
@@ -1,0 +1,114 @@
+package main
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"slices"
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// mockRegistryHandler implements a minimal Docker Registry V2 API that captures User-Agent headers
+type mockRegistryHandler struct {
+	mu         sync.Mutex
+	userAgents []string
+}
+
+func (h *mockRegistryHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	// Capture the User-Agent header
+	h.mu.Lock()
+	h.userAgents = append(h.userAgents, r.Header.Get("User-Agent"))
+	h.mu.Unlock()
+
+	// Implement minimal Docker Registry V2 API endpoints for inspect --raw
+	switch {
+	case r.URL.Path == "/v2/":
+		// Registry version check endpoint
+		w.Header().Set("Docker-Distribution-API-Version", "registry/2.0")
+		w.WriteHeader(http.StatusOK)
+
+	case strings.HasSuffix(r.URL.Path, "/manifests/latest"):
+		// Return a minimal OCI manifest as raw string
+		// The digest matches this exact content
+		manifest := `{"schemaVersion":2,"mediaType":"application/vnd.oci.image.manifest.v1+json","config":{"mediaType":"application/vnd.oci.image.config.v1+json","digest":"sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a","size":2},"layers":[{"mediaType":"application/vnd.oci.image.layer.v1.tar+gzip","digest":"sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855","size":0}]}`
+		w.Header().Set("Content-Type", "application/vnd.oci.image.manifest.v1+json")
+		w.WriteHeader(http.StatusOK)
+		if _, err := w.Write([]byte(manifest)); err != nil {
+			panic(err)
+		}
+
+	default:
+		w.WriteHeader(http.StatusNotFound)
+	}
+}
+
+func (h *mockRegistryHandler) getUserAgents() []string {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	return slices.Clone(h.userAgents)
+}
+
+func TestUserAgent(t *testing.T) {
+	testCases := []struct {
+		name               string
+		extraArgs          []string
+		userAgentValidator func(string) bool
+		description        string
+	}{
+		{
+			name:      "default user agent",
+			extraArgs: []string{},
+			userAgentValidator: func(ua string) bool {
+				return strings.HasPrefix(ua, "skopeo/")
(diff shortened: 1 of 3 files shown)


cgwalters:
> I… don’t think that identifying use cases via User-Agent generalizes at all. Podman is not going to use a different User-Agent depending on operation, for example.

Yes, it would be unusual. But conceptually think of this as for rich enough use cases where one has a proper "application" on top of skopeo - it might be e.g. a registry which supports auto-mirroring from another registry and forks skopeo as a binary instead of vendoring the libraries because it doesn't want to hop on the CVE train.

> I read https://www.rfc-editor.org/rfc/rfc7231.html#section-5.5.3 to mean that (…) is a comment syntax, that is not what I’d expect here.

Fixed, thanks!

cgwalters:
Added an `Assisted-by: Claude Code` written integration test

cgwalters:
> LGTM, but I’d like a second pair of eyes.

Who would that be?

vrothberg:
> /lgtm

Ah, this doesn't work anymore as I am not maintainer anymore

integration/user_agent_test.go
```diff
@@ -0,0 +1,217 @@
...
+		}
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+		if err := json.NewEncoder(w).Encode(tags); err != nil {
+			panic(err)
+		}
+
+	case strings.HasSuffix(r.URL.Path, "/manifests/latest"):
+		// Return a minimal OCI manifest
+		manifest := map[string]interface{}{
```
mtrmac: (This would be a bit shorter as a hard-coded string literal, without the Go data structures and encoding at run-time. _Especially_ if this is sending a `Docker-Content-Digest` header, the content should match the digest. But then again, Skopeo does not rely on that header, so dropping the header might be easier.)

integration/user_agent_test.go
```diff
@@ -0,0 +1,217 @@
...
+		w.Header().Set("Docker-Content-Digest", "sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a")
+		w.WriteHeader(http.StatusOK)
+		if err := json.NewEncoder(w).Encode(manifest); err != nil {
+			panic(err)
+		}
+
+	case strings.Contains(r.URL.Path, "/blobs/"):
+		// Return empty blob
+		w.WriteHeader(http.StatusOK)
+		if _, err := w.Write([]byte("{}")); err != nil {
```
mtrmac: Well that’s not an empty blob, the comment could be confusing. I guess this is ~necessary as a “config” read by `skopeo inspect`? If so, the comment should say so.


Alternatively (preferably?), use `skopeo inspect --raw`, then the `/tags/list` and `/blobs/…` endpoints can be deleted entirely.

integration/user_agent_test.go
```diff
@@ -0,0 +1,217 @@
...
+
+	default:
+		w.WriteHeader(http.StatusNotFound)
+	}
+}
+
+func (h *mockRegistryHandler) getAllUserAgents() []string {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	agents := make([]string, len(h.userAgents))
```
mtrmac: `slices.Clone`, or see elsewhere about `assert…`

integration/user_agent_test.go
```diff
@@ -0,0 +1,217 @@
...
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+)
+
+func TestUserAgent(t *testing.T) {
```
mtrmac: Could the three copies be consolidated? Future maintainers will need to keep reading the copies.

A single table-driven test with `optionsToAdd []string, expectedUserAgentRegexp string` entries could handle all three situations, and avoid the need for a `suite.Suite` entirely.

Alternatively, a `startMockRegistryHandler(t *testing.T) (dockerReferenceForCLI string)` (using `t.Cleanup` to terminate it), and turning `getAllUserAgents` into a `assertUserAgentMatches(t *testing.T, condition func(string) bool)` would eliminate the 3 copies while retaining the individual inputs logic in existing places.

integration/user_agent_test.go
```diff
@@ -0,0 +1,217 @@
...
+	imageRef := "docker://" + registryAddr + "/test/image:latest"
+
+	// Run skopeo inspect without --user-agent-prefix
+	assertSkopeoSucceeds(t, "", "--tls-verify=false", "inspect", imageRef)
+
+	// Verify that at least one request was made with a User-Agent
+	userAgents := handler.getAllUserAgents()
+	require.NotEmpty(t, userAgents, "Expected at least one request to be made")
+
+	// Check that at least one User-Agent contains "skopeo/"
```
mtrmac: (`slices.ContainsFunc` is slightly clunky, but perhaps still a bit more concise. (Applies to all copies, if any remain.))

mtrmac:
> Ah, this doesn't work anymore as I am not maintainer anymore

AFAICS you do still have an admin role in GitHub, but we never had the bot set up.

integration/user_agent_test.go
cgwalters: This type of style thing I think is one that would be good to have findable from `CONTRIBUTING.md` (which could links to e.g. `STYLE_GO.md`) or so.

The AI model I'm using here (Sonnet 4.5) I think hasn't internalized newer Go idioms yet. I've encountered similar things with Rust.

mtrmac: Good point, “use the standard library features to the full extent _but only up to the version in `go.mod`_”, while we do tend to stay one or two releases behind, is **EDIT not** the default choice in the ecosystem.

integration/user_agent_test.go
cgwalters: Oh yeah table driven testing is *way* cleaner. Sorry for not noticing that myself. I'll add something about this to my pre-review

PR #2725, opened 2025-10-17
Cirrus: use test-integration target

lsm5:


mtrmac:
For the record: this is motivated by the discussion in https://github.com/containers/skopeo/issues/2701 .

Note https://github.com/containers/skopeo/issues/2701#issuecomment-3415747878 : _right now_ it seems that we have to choose between test coverage and meaningful containerization.

Also, from a quick comparison, this seems to add 2-3 minutes (~20 %) to test runtime. That’s looking at _one_ test run (vs. https://github.com/containers/skopeo/commit/169091e282a312e7e854727c13cd248c198ad954 ), it might well be a random flake — but if not, it’s a bit annoying.

lsm5:
RE: https://github.com/containers/skopeo/issues/2701#issuecomment-3415747878 , I've added another commit to run the container with cap_sys_admin and now it works. Thanks.

Given there are set of container tests that are skipped in `-local`, I prefer we test `test-integration` in CI until we get to rewriting tests without container. But, that still leaves us with the problem of setting up openshift / oc in the environment which I guess we don't wanna bother the end tester with.

So, unless the openshift setup is made convenient, maybe let's just continue with skopeo_cidev. I'll check if the openshift binary used in openshift tests can be replaced with the newer oc binary. It's inconvenient that they don't ship it via an rpm AFAIK.

If this is good enough for merge, we should only merge whenever https://github.com/containers/automation_images/pull/419 (or whatever PR that change is added to) is ready to merge.

TomSweeneyRedHat:
LGTM
but I only play a Cirrus knowledgable person on TV

TomSweeneyRedHat:
@lsm5?

PR #2729, opened 2025-10-25
Makefile: remove stale target

lsm5:
Old cruft. Isn't even expected to work. Not used anywhere.


lsm5:
Obsoleted by https://github.com/containers/skopeo/pull/2731

PR #2730, opened 2025-10-25
Makefile: fix `test-system`

lsm5:
See individual commits.

PR #2731, opened 2025-10-27
Fix test-system and remove stale Makefile target

lsm5:
Obsoletes #2729 and #2730 . See individual commits.

Doesn't depend on automation_images so this can go in.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -269,8 +269,3 @@ vendor:
 
 vendor-in-container:
 	podman run --privileged --rm --env HOME=/root -v $(CURDIR):/src -w /src golang $(MAKE) vendor
-
-# CAUTION: This is not a replacement for RPMs provided by your distro.
-# Only intended to build and test the latest unreleased changes.
-rpm:
-	rpkg local


lsm5:
@mtrmac PTAL

TomSweeneyRedHat:
LGTM

hack/test-system.sh
```diff
@@ -0,0 +1,40 @@
...
+graphroot = "/var/lib/containers/storage"
+EOF
+    fi
+    # The logic of finding the relevant storage.conf file is convoluted
+    # and in effect differs between Skopeo and Podman, at least in some versions;
+    # explicitly point at the file we want to use to hopefully avoid that.
+    export CONTAINERS_STORAGE_CONF=/etc/containers/storage.conf
+fi
+
+SKOPEO_BINARY="$SKOPEO_BINARY" bats --tap systemtest
```
mtrmac: Does the `SKOPEO_BINARY=…` assignment do anything useful? AFAICT the only difference it makes is replacing an unset environment variable by a set-but-empty variable.

Makefile
```diff
@@ -202,6 +202,10 @@ test-integration:
 
 
 # Helper target to set up SKOPEO_BINARY variable for local test targets
+# SKOPEO_BINARY only takes effect on `test-integration-local` and
+# `test-system-local` targets. It's ignored on `test-integration` and
```
mtrmac: Maybe `s/ignored/not propagated into the container/` or `s/ignored/& (not …)/`?

lsm5: fixed.

hack/test-system.sh
lsm5: So, it does get set in test-system-local as seen in https://artifacts.dev.testing-farm.io/2e59a9ac-67ad-45cb-886a-17d00be94564/ . Binary used is `/usr/bin/skopeo` . Does that answer this?

lsm5:
> LGTM apart from the nits, but maybe revisit #2725 first, in case you wanted both PRs to be merged simultaneously.

I think this can go in independently.

hack/test-system.sh
mtrmac: I agree that `bats` is run with `SKOPEO_BINARY` set, I mean I think this would work just as well with
```suggestion
bats --tap systemtest
```

Is that not the case?

lsm5: *IIRC* it didn't, but let me verify..

lsm5:
@mtrmac applied your suggestion. Thanks a lot! This is ready to go.

PR #2740, opened 2025-11-10
Use retries in more cases of copying from external registries

mtrmac:
We added retries to other registry operations before, these were missed for some reason.

Motivated by: https://api.cirrus-ci.com/v1/task/6228198955941888/logs/integration.log .


Luap99 (approved):
LGTM

PR #2746, opened 2025-11-17
Container libs vendor and Renovate config update

lsm5:
See individual commits.


lsm5:
@containers/skopeo-maintainers PTAL

.github/renovate.json5
```diff
@@ -49,4 +49,26 @@
...
    *** Repository-specific configuration options ***
    *************************************************/
+
+  // N/B: LAST matching rule wins
+  "packageRules": [
+    // Updates for `github.com/containers/*` should be checked more frequently than other deps.
+    {
+      "matchCategories": ["golang"],
+      "schedule": "before 11am",
+      "matchPackageNames": ["github.com/containers{/,}**"]
```
mtrmac: All the repo names are wrong.

lsm5:
> Thanks!
> 
> I don’t feel strongly strongly about the updated frequency, but
> 
>     * Can you write down (in the commit messages, or at least in the PR) _why_ we are making the change, please?

Will do.

> 
>     * Do I read `before 11am` correctly to mean that once we move to a digest reference (like this PR does), Renovate is going to file a new update _every single day_? That seems excessive to me, as a guess (noting that I have no idea what the reason is, see above)

This was a copy-paste from the [podman renovate config](https://github.com/containers/podman/blob/main/.github/renovate.json5). I was trying to vendor container-libs/image into skopeo and noticed it broke, then I realized skopeo needed more recent vendoring, and there were no automatic PRs by renovate here.

I think I would prefer a daily job to ensure there's regular updates and CI testing, but I'll defer to you. If you find that excessive, I'm ok to make it infrequent.

.github/renovate.json5
lsm5: removed the duplicate job and fixed package name. PTAL.

lsm5:
@mtrmac PTAL.

.github/renovate.json5
```diff
@@ -49,4 +49,16 @@
...
+    {
+      "matchCategories": ["golang"],
+      "schedule": "before 11am on the first day of the week",
+      "matchPackageNames": [
+        "/^go.podman.io/common/",
+        "/^go.podman.io/image/",
+        "/^go.podman.io/storage/"
+      ]
+    }
+  ],
```
mtrmac: Isn’t this now _exactly_ the same as in the `defaults.json5`? If so, does it do anything?

AFAIK with that configuration, _only_ updating to digested versions is sort-of sufficient: it will cause Renovate to keep filing digested updates, _until we tag a release_: then it file an update to move to a tag, and stop updating until there is either another tag, or we manually update to a digested version again.

I don’t know whether there is a Renovate option to automatically update from a tag to a digest. Might be worth considering — but also we must not have that on branches.

lsm5: > Isn’t this now _exactly_ the same as in the `defaults.json5`? If so, does it do anything?

I see, if the automation/defaults.json5 suffices, I'll get rid of that commit from this PR. 
> 
> AFAIK with that configuration, _only_ updating to digested versions is sort-of sufficient: it will cause Renovate to keep filing digested updates, _until we tag a release_: then it file an update to move to a tag, and stop updating until there is either another tag, or we manually update to a digested version again.

Can we tag a release *after* branching from main and switching to tags on the release branch? Podman does cut release after branching fwiw. That way main can continue to have digested tags.

> 
> I don’t know whether there is a Renovate option to automatically update from a tag to a digest. Might be worth considering — but also we must not have that on branches.

I haven't paid close attention to renovate anywhere tbh, but I guess podman has this part handled somehow.

mtrmac (approved):

PR #2747, opened 2025-11-17
skopeo inspect: support for sha512 images

lsm5:
(Rebased on #2746).

See individual commits.

Vendoring from a fork.

### DIFF
--- a/docs/skopeo-inspect.1.md
+++ b/docs/skopeo-inspect.1.md
@@ -95,6 +95,12 @@ The password to access the registry.
 
 Do not list the available tags from the repository in the output. When `true`, the `RepoTags` array will be empty.  Defaults to `false`, which includes all available tags.
 
+**--manifest-digest**=_algorithm_ **EXPERIMENTAL**
+
+Algorithm to use for computing manifest digest (sha256, sha512); defaults to algorithm used in config digest.
+
+**Note:** This flag is experimental and its behavior may change in future releases.
+
 ## EXAMPLES
 
 To review information for the image fedora from the docker.io registry:
@@ -184,6 +190,12 @@ amd64
 ```console
 $ /bin/skopeo inspect --format '{{ .Env }}' docker://registry.access.redhat.com/ubi8
 [PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin container=oci]
+```
+
+To get the digest using a specific algorithm:
+```console
+$ skopeo inspect --manifest-digest=sha512 docker://docker.io/library/alpine:latest --format "Digest: {{.Digest}}"
+Digest: sha512:5acb33fb56a7791bf0c69d5b19a1c70272148e4107be5261d57305d14e9509792bbca53e5277c456181ecfa1c20ad8427f9b8ba46868020584a819de1128dbd2
 ```
 
 # SEE ALSO


lsm5:
podman inspect prior sha512 image built from scratch
```
$ ./bin/podman inspect localhost/test-sha512
[
     {
          "Id": "671c13d501533ad620d1ff829ef46c9c33a01b0d360d054f047dc5627b0c4296b02fd7350d103a653354d600ea743873c2e2fdd50382f42811c07f59fe7293ad",
          "Digest": "sha512:1919c905953e3d5c42eed8442451777f41390941f445bbefb3cbc5fbbea30f3258da357cdc8654548c4e024a893705614028eaf8fcc1d1f7588e16090619920c",
          "RepoTags": [
               "localhost/test-sha512:latest"
          ],
          "RepoDigests": [
               "localhost/test-sha512@sha512:1919c905953e3d5c42eed8442451777f41390941f445bbefb3cbc5fbbea30f3258da357cdc8654548c4e024a893705614028eaf8fcc1d1f7588e16090619920c"
          ],
......
......
$
```

Inspect image using skopeo from main
```
$ skopeo inspect containers-storage:localhost/test-sha512
FATA[0000] Download config.json digest sha256:cfceead3d1f96ba5866677ddd9f44b1583af938bdc08f1461809e82e56765d8b does not match expected sha512:671c13d501533ad620d1ff829ef46c9c33a01b0d360d054f047dc5627b0c4296b02fd7350d103a653354d600ea743873c2e2fdd50382f42811c07f59fe7293ad
$
```

Inspect image using skopeo built here
```
$ ./bin/skopeo inspect containers-storage:localhost/test-sha512
{
    "Name": "localhost/test-sha512",
    "Digest": "sha512:1919c905953e3d5c42eed8442451777f41390941f445bbefb3cbc5fbbea30f3258da357cdc8654548c4e024a893705614028eaf8fcc1d1f7588e16090619920c",
    "RepoTags": [],
    "Created": "2025-11-15T18:08:56.226408249Z",
    "DockerVersion": "",
    "Labels": {
        "io.buildah.version": "1.40.0-dev"
    },
    "Architecture": "amd64",
    "Os": "linux",
    "Layers": [],
    "LayersData": [],
    "Env": [
        "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
    ]
}
$
```

skopeo copy image to dir
```
$ ./bin/skopeo copy containers-storage:localhost/test-sha512 dir:/home/lsm5/test/sha512-copy
Copying config 671c13d501 done   | 
$
```

skopeo inspect copied-to dir
```
$ ./bin/skopeo inspect dir:/home/lsm5/test/sha512-copy
{
    "Digest": "sha512:1919c905953e3d5c42eed8442451777f41390941f445bbefb3cbc5fbbea30f3258da357cdc8654548c4e024a893705614028eaf8fcc1d1f7588e16090619920c",
    "RepoTags": [],
    "Created": "2025-11-15T18:08:56.226408249Z",
    "DockerVersion": "",
    "Labels": {
        "io.buildah.version": "1.40.0-dev"
    },
    "Architecture": "amd64",
    "Os": "linux",
    "Layers": [],
    "LayersData": [],
    "Env": [
        "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
    ]
}
$
```

lsm5:
@mtrmac PTAL. Main change here is https://github.com/containers/skopeo/pull/2747/files#diff-6cac070631a328cbecde69cd12534e6b77254bcf5d366c4210afaa76e70bfe67 . Rest is vendoring.

cmd/skopeo/inspect.go
```diff
@@ -235,3 +236,26 @@ func (opts *inspectOptions) writeOutput(stdout io.Writer, data any) error {
...
+		Config struct {
+			Digest string `json:"digest"`
+		} `json:"config"`
+	}
+	if err := json.Unmarshal(manifestBlob, &manifestData); err != nil {
+		return manifest.Digest(manifestBlob)
+	}
+	if manifestData.Config.Digest == "" {
+		return manifest.Digest(manifestBlob)
+	}
```
mtrmac, line 257: `types.Image.ConfigInfo` exists, no need to hard-code an assumption that all future image formats will be JSON, and use the same field name, here.

cmd/skopeo/inspect.go
```diff
@@ -235,3 +236,26 @@ func (opts *inspectOptions) writeOutput(stdout io.Writer, data any) error {
...
+	configDigest, err := digest.Parse(manifestData.Config.Digest)
+	if err != nil {
+		return manifest.Digest(manifestBlob)
+	}
+	alg := configDigest.Algorithm()
```
mtrmac: > “report a manifest digest using the algorithm used for config inside the manifest”

- What is the “user feature” here? “We report a sha512 digest for sha512 images” sounds sort of attractive, but the manifest digest is typically used in things like digest references on registries, and in that use there’s no _direct relationship_ between the digest of the reference and the contents inside. It almost certainly will be possible to refer to the same on-registry blob using either a sha256 or a sha512 digest. 
  - One use case of `skopeo inspect` is “turn a tag reference into a digest reference”; this heuristic would work fine here, I think.
  - Another use case is “compare two images for equality”; in that case using two different digest algorithms says nothing unless users _hard-code_ an assumption of the `inspect` heuristics
  - Uh, there is, hypothetically, “what digest does this tag point to at the remote registry?”. That’s _not_ what we ask the registry today, but, in a sha256-only ecosystem, we sort of arrive at the only possible answer.
  - Are there other use cases?
- What about manifests that mix sha256+sha512 digests?
- [Pedantically, `manifest.Digest` is not a pure `digest.SHA256.FromBytes`. But, meh, in those cases `Config.Digest == ""`. Still… maybe we should extend the c/image API to do this correctly, first? But then again, we need use cases when designing an API extension…]

As a guess, I think the algorithm to use should be an option; and I don’t know what would be the default value — maybe this heuristic.

Alternatively, I can imagine keeping `skopeo inspect` sha256-only, and letting users who have specific preferences use `skopeo inspect --raw | skopeo manifest-digest` (with a new option to choose an algorithm). That’s probably worse.

mtrmac (commented):
The `inspect` case is… interesting.

cmd/skopeo/inspect.go
mtrmac: > - What is the “user feature” here?
Also
  - does it make sense for `skopeo inspect …@sha512:…` to output a `sha256` digest, or vice versa??

I don’t know… right now I find the “what digest does this tag point to at the remote registry?” definition attractive, but that’s the one that requires most work (we’d need to add a whole new “resolve a reference into a digested reference” API to c/image, call that on the user-provided input, and then “inspect” the resolved reference).

lsm5: @mtrmac So far, I only had `We report a sha512 digest for sha512 images and that digest matches with what podman reports` in mind as posted in https://github.com/containers/skopeo/pull/2747#issuecomment-3542785000 . 

So, what, in your opinion, would involve the minimum possible work needed here to get this going forward? I would prefer to track the rest in additional RFE cards.

mtrmac: I think an explicit `skopeo inspect --manifest-digest=sha512` (with, “a minimum possible” continuing to default to sha256, but _without making such a promise_ for anything but fully-sha256 images) might work.

I’m not 100% happy with that — maybe there is some semantics that can be configured globally in `containers.con`, maybe we really want to do that registry tag lookup — but it would allow tooling to be built now, relying on that explicit option, and it would not make _much_ of a commitment for manifest-digest-unaware users.

mtrmac: > podman inspect prior sha512 image built from scratch

Hum… one way to think about it is that an image built locally _does not have much of a manifest digest yet_; only during a push we will compress the layers and know what the manifest digest will be on the registry. It _does_ make sense to talk about “the manifest of the locally-built image” (to the extent that we store one in c/storage, and maybe there could be annotations on it), but makes fairly little sense to talk about “_the_ manifest digest of a locally-built image” when that value may change on every push (depending on pushed format, and on preexisting on-registry layers that might be reused instead of re-compressing).

… and for images _not_ built locally: `podman pull repo@sha256:sha256digestOfTagLatest; podman pull repo@sha512:sha512DigestOfTagLatest`. _If_ the two images are deduplicated (we haven’t decided that yet), what digest algorithm would be “correct” for `podman inspect` to use? Does that depend on whether the CLI says `sha256` vs. `sha512`? On the manifest contents as in the `inspect` here? On `containers.conf`?

All this is a lot of text just to say that I am confused and I don’t feel I have a good mental model.

cgwalters: > Hum… one way to think about it is that an image built locally does not have much of a manifest digest yet; 

It *could* though, we could have `podman build --compress` which would also eagerly pregenerate compressed blobs or so. Could even be a global opt-in flag. I wonder if other build systems do this.

>  when that value may change on every push (depending on pushed format, 

Re "Depending on pushed format" I still think the case of choosing a compression other than the one used by the inherited base image layers would be extremely unusual, but now this is revisiting https://github.com/containers/buildah/issues/5586

> and on preexisting on-registry layers that might be reused instead of re-compressing).

I don't get this one. We can only end up re-using target registry layers if we ended up compressing a byte identical layer. And we know ahead of time via the blobinfo cache if we have a known corresponding compression.

If a correspondence wasn't known in the cache but we discover on-push that a registry already has a blob...that doesn't change anything about the manifest digest.

mtrmac: > > and on preexisting on-registry layers that might be reused instead of re-compressing).
>
> I don't get this one. We can only end up re-using target registry layers if we ended up compressing a byte identical layer. And we know ahead of time via the blobinfo cache if we have a known corresponding compression.

`BlobInfoCache` tells us what the matches are, but they can only be _used_ if the registry already has a copy (that’s one of the case BlobInfoCache is trying to benefit), or if we happened to reproduce the exact compressed blob (and then the match is a ~no-op, we are pushing the whole thing).

And there may be multiple equally-valid matches, so even we somehow decided to build the manifest of locally-built images to refer to compressed layers from BlobInfoCache, we would only have to pick _one_, while the destination registry might contain a different matching compressed version.

The principal feature here is a build `FROM large-base-image`; that base image may frequently enough be built by a different implementation (more stable/frozen versions, or a different OS-vendor-specific implementation entirely) and it is ~likely to compress differently than the local system running a build+push. We _want_ to reuse the `large-base-image` layers if at all possible, but we can’t recreate them locally now that we have discarded the original compressed versions.

(Sure, there is an opt-in BlobCache that can be used for builds to store the compressed version locally as well, but that’s only used with an opt-in, and we would _still_ prefer to reuse a different-digest compressed version that already exist on the destination registry over pushing the BlobCache version we cached.)

lsm5: added a `skopeo inspect --manifest-digest` in the latest draft. Reused `supporteddigests.TmpSetDigestForNewObjects`, let me know if that's cool.

Also, I didn't see a prior `inspect_test.go` . Do we need one?

cmd/skopeo/inspect.go
```diff
@@ -235,3 +239,51 @@ func (opts *inspectOptions) writeOutput(stdout io.Writer, data any) error {
 	defer rpt.Flush()
 	return rpt.Execute([]any{data})
 }
+
+func manifestDigestFromManifest(manifestBlob []byte, img types.Image, userAlgorithm digest.Algorithm) (digest.Digest, error) {
+	if userAlgorithm != "" {
+		if !userAlgorithm.Available() {
+			return "", fmt.Errorf("digest algorithm %q is not available", userAlgorithm)
+		}
+		return userAlgorithm.FromBytes(manifestBlob), nil
```
mtrmac: With this, `--manifest-digest=anything` is incorrect for v2s1, incl. for `--manifest-digest=sha256`. So I think we need something like a `manifest.DigestWithAlgorithm` in c/image first.

cmd/skopeo/inspect.go
```diff
@@ -235,3 +239,51 @@ func (opts *inspectOptions) writeOutput(stdout io.Writer, data any) error {
...
+	}
+
+	configInfo := img.ConfigInfo()
+	if configInfo.Digest != "" {
+		alg := configInfo.Digest.Algorithm()
+		if alg.Available() {
```
mtrmac: `if !available` fail, don’t silently compute something undocumented?

cmd/skopeo/inspect.go
```diff
@@ -235,3 +239,51 @@ func (opts *inspectOptions) writeOutput(stdout io.Writer, data any) error {
...
+type algorithmValue digest.Algorithm
+
+func newAlgorithmValue(alg *digest.Algorithm) *algorithmValue {
+	return (*algorithmValue)(alg)
+}
+
+func (a *algorithmValue) Set(value string) error {
+	algorithm := digest.Algorithm(value)
+
+	if err := supporteddigests.TmpSetDigestForNewObjects(algorithm); err != nil {
```
mtrmac: (I have ~no opinion on the _presence_ of this call… it should not make a difference to anything at this point AFAICS.

This function, _sort of_, assumes that the semantics of `""` in `TmpSet…` is the same as in the code below.)

---

Note to self: `TmpSet…` ensures `algorithm` is one of a specific values, which implies it’s valid. Otherwise it would be desirable to validate it before we do any network I/O.

mtrmac: I think there is a conceptual difference between _parsing_ options and _handling_ options; causing side effects at the parsing stage is rather unexpected. So if the `TmpSet…` call should be made, I’d prefer that to happen later in `inspectOptions.run`.

(For the parsing, `digest.Algorithm` already implements `Set` but not `Type`. That implementation of `Set` also hard-codes that `""` means sha256… for now that works.)

mtrmac (commented):
It think this needs a `manifest.Digest` alternative in c/image first.

On this side, lints will eventually complain about a missing man page update.

cmd/skopeo/inspect.go
mtrmac: > Also, I didn't see a prior inspect_test.go . Do we need one?

I think tests are always nice … so far the unit tests in `cmd/skopeo` tend to be quick tests that don’t reach to registries (i.e. just functional algorithms and option parsing), with tests that contact registries running in `integration/*`. So, here, `inspectOptions.run` would probably not have much of a unit test; a unit test of `manifestDigestFromManifest`, against a `dir:` fixture, maybe, would be possible.

But then again, if this is “experimental and unstable”, I’m not _too_ worried about testing (yet?).

lsm5:
`manifest.DigestWithAlgorithm`: https://github.com/containers/container-libs/pull/499

lsm5:
@mtrmac PTAL. vendored `container-libs 01833ef7b7f1d306205be7fa6fb36d0d6a6e3a33` and also included a manpage update commit.

cmd/skopeo/inspect.go
```diff
@@ -84,6 +88,12 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
 	}
 	imageName := args[0]
 
+	if opts.manifestDigest != "" {
+		if err := supporteddigests.TmpSetDigestForNewObjects(opts.manifestDigest); err != nil {
```
mtrmac: AFAICS this is no longer affecting anything and can be dropped.

cmd/skopeo/inspect.go
```diff
@@ -235,3 +245,48 @@ func (opts *inspectOptions) writeOutput(stdout io.Writer, data any) error {
...
+	return (*algorithmValue)(alg)
+}
+
+func (a *algorithmValue) Set(value string) error {
+	algorithm := digest.Algorithm(value)
+
+	*a = algorithmValue(algorithm)
+	if algorithm == "" {
+		*a = algorithmValue(digest.Canonical)
+	}
```
mtrmac, line 274: I’m not sure this special case is valuable but it doesn’t hurt (especially when not documented, in both respects).

mtrmac (approved):
LGTM… do you want to merge as is?

cmd/skopeo/inspect.go
lsm5: I'll drop and repush

lsm5:
@mtrmac done. PTAL.

mtrmac (approved):
Thanks!

PR #2751, opened 2025-11-18
Support `containers-storage:` with `sha256:` prefix

### LINKED ISSUE #2750: `containers-storage:` transport doesn't support `sha256:` prefix
Most interactions with `podman` (and friends) can refer to a container by an identifier like `sha256:417a368e464b22dd60e872395529385d04bfa05b0f8200c7673ff9f98cddd7ee`:
 - that's the default output format when you use `podman build --iidfile=`
 - you can `podman run` the identifier directly
 - you can use `FROM sha256:417a368e464b22dd60e872395529385d04bfa05b0f8200c7673ff9f98cddd7ee` in a `Containerfile`
 - `buildah` also accepts this format for commands on its cli

Unfortunately this doesn't work with skopeo's `containers-storage:` transport: you need to strip the `sha256:` first.  This is also a problem when using [containers/containers-image-proxy-rs](https://github.com/containers/containers-image-proxy-rs).

So simple ask: make it possible to do `skopeo copy` or an `OpenImage` on the proxy via a name starting with `containers-storage:sha256:`.

allisonkarlitskaya:
The code was previously trying to interpret this as a named reference and landing up turning sha256:xyz into docker.io/library/sha256:xyz and failing to find it.

That's very clearly not what the user intended in this case, so let's just use the digest value directly.

Fixes #2750

### DIFF
--- a/vendor/go.podman.io/image/v5/storage/storage_transport.go
+++ b/vendor/go.podman.io/image/v5/storage/storage_transport.go
@@ -168,6 +168,10 @@ func (s storageTransport) ParseStoreReference(store storage.Store, ref string) (
 			// We have recognized an image ID; peel it off.
 			ref = ref[:split]
 		}
+	} else if rest, ok := strings.CutPrefix(ref, "sha256:"); ok {
+		// If it starts with sha256: then it's pretty clear that it's already an ID
+		id = rest
+		ref = ""
 	}
 
 	// If we only have one @-delimited portion, then _maybe_ it's a truncated image ID.  Only check on that if it's


allisonkarlitskaya:
I have no idea what I'm doing here, but this seems to fix it :upside_down_face:

lsm5:
Hi @allisonkarlitskaya , I'm not opposed to it, but I'll defer to @mtrmac . If we do decide to get this in, this should first go in as a PR to https://github.com/containers/container-libs and have that vendored here (updating go.mod and running `make vendor`)

mtrmac (changes requested):
> interpret this as a named reference and landing up turning sha256:xyz into docker.io/library/sha256:xyz

Yes, there’s this ambiguity. I’m _generally_ not in favor of ambiguity.

In this case, _some_ code (`docker-daemon:` parsing) is explicitly resolving the ambiguity by refusing `sha256` as a repo/short name, but most consumers don’t do that.

In that sense, taking over the `sha256` repo/short name would not be _that_ bad — but we now have the prospect of generalizing the digest algorithms to support at least `sha512`, and possibly other values, unknown in advance. I don’t think it makes sense to have a design where “some names are ambiguous _and we don’t know which ones will be ambiguous next year_”.

This fairly strongly depends on what we do for the sha512 support. (Cc: @lsm5 ).
- One possible design is that we move to a design where image IDs are unique but have no semantics any more (are not a digest of anything); in that case adding support for `sha256` syntax would be moving in the wrong direction, exposing something that will not be true.
- Another possible design is that we will continue to have image IDs to be digests (of … something; config, perhaps manifest), and that we will add sha512-shaped IDs. In that case, considering the ambiguity and an unpredictable list of ambiguous repo/short names, I think we will need a new syntax that _explicitly_ differentiates between manifest digests and image IDs (I don’t know, maybe `container-storage:`_host_`/`_repo_[`@`_manifestDigest_]`@=`_idAlgorithm_`:`_idValue_ or something like that, with a new reserved character).
- Perhaps something else

And then there’s the zstd:chunked problem space (`insecure_allow_unpredictable_image_contents`) where we _already_ don’t want to say that image ID equal config digests.

So, I don’t know what we want to do in the next year or two, but I don’t think it makes sense to commit any more to the `sha256` algorithm or syntax right now, when we don’t have firmer plans for the future (and it wouldn’t work in either of the anticipated possible plans).

lsm5:
> This fairly strongly depends on what we do for the sha512 support. (Cc: @lsm5 ).

@mtrmac  by `we`, do you ultimately mean image-spec here? Any changes of that sort would first have to be approved by image-spec, isn't that the case?

mtrmac:
No, `opencontainers/image-spec` defines the external (on-registry / in-files) representation of images. Image IDs are an implementation / UX choice of c/storage+Podman and friends; `opencontainers/image-spec` does not have image IDs at all.

lsm5:
> And then there’s the zstd:chunked problem space (`insecure_allow_unpredictable_image_contents`) where we _already_ don’t want to say that image ID equal config digests.

(We can have this conversation elsewhere I guess) What's being done for image IDs in zstd:chunked in that case? And can (should) we use the same approach across the board?

mtrmac:
> What's being done for image IDs in zstd:chunked in that case?

Normally we enforce, after pulling zstd:chunked layers, that they match the diffID in config; and we use image ID = config digest. That comes at a fairly high performance cost.

Originally the design was not to do that; but then we must not deduplicate chunked-pulled and tar-pulled images (because we don’t know for sure the contents/views are the same), so they must have different image IDs, i.e. zstd:chunked-pulled images must have IDs that are deterministic (related to how the image was pulled) but _not_ a config digest (but happen to be a sha256 of some different value, that’s not documented).

> And can (should) we use the same approach across the board?

We had to abandon the original design and add the costly checks because the thought was that the Podman userbase is not ready for image IDs that are not equal to config digests. So directly, and universally, using the chunked design is not an immediate option.

[FWIW containerd now uses image ID = _manifest_ digest, not _config_ digest; so there _is_ a precedent for changing the value, of sorts.]

I was thinking that with the move to sha512, we _have_ the opportunity to change the image ID format — something will _certainly_ need to change because we can’t fit a sha512 digest into a sha256-sized-and-formatted identifier, and using a sha256 digest (or truncating a sha512 one) would not give us a sha512 level of collision resistance. But for old sha256 images (pulled using a sha256 configuration, whatever that means??) we will probably need to keep using the current deterministic values for a _long_ time.

lsm5:
> the thought was that the Podman userbase is not ready for image IDs that are not equal to config digests.

Is that thought still true?

> [FWIW containerd now uses image ID = _manifest_ digest, not _config_ digest; so there _is_ a precedent for changing the value, of sorts.]

If containerd uses it, then maybe podman is ok as well? Or are there other concerns?

> I was thinking that with the move to sha512, we _have_ the opportunity to change the image ID format — something will _certainly_ need to change because we can’t fit a sha512 digest into a sha256-sized-and-formatted identifier, and using a sha256 digest (or truncating a sha512 one) would not give us a sha512 level of collision resistance. But for old sha256 images (pulled using a sha256 configuration, whatever that means??) we will probably need to keep using the current deterministic values for a _long_ time.

do you see any of this as a blocker to the in-progress stuff for sha512, or are we ok to make changes after the initial sha512 support is introduced? FWIW https://github.com/containers/skopeo/pull/2747

mtrmac:
> > the thought was that the Podman userbase is not ready for image IDs that are not equal to config digests.
> 
> Is that thought still true?

I mean, we rolled that back so we have no data proving that this is reasonable to do. And we did nothing at all to prepare users for ID changes since the zstd:chunked addition, so assuming it was true last year, it almost certainly continues to be true.

> > I was thinking that with the move to sha512, we _have_ the opportunity to change the image ID format — something will _certainly_ need to change because we can’t fit a sha512 digest into a sha256-sized-and-formatted identifier, and using a sha256 digest (or truncating a sha512 one) would not give us a sha512 level of collision resistance. But for old sha256 images (pulled using a sha256 configuration, whatever that means??) we will probably need to keep using the current deterministic values for a _long_ time.
> 
> do you see any of this as a blocker to the in-progress stuff for sha512

I’m not sure which parts count as “in-progress” right now; we need to make _some_ choices for c/storage IDs to implement sha512 there.

> or are we ok to make changes after the initial sha512 support is introduced?

Well if we chose and documented that image IDs = sha512 digests of $something, and users started relying on that, we would not be at liberty to break that again. If we chose and documented that image IDs have unspecified contents, we can change, and add more commitments, later.

mtrmac:
As discussed above, I don’t know what design we end up with, but right now now, we certainly don’t want to commit to _this_ promise.

PR #2763, opened 2025-12-04
Packit: use `post-modifications` hook to update downstream TMT plan

lsm5:
`prepare-files` action was interfering with spec file update which caused https://github.com/containers/skopeo/issues/2760 .

Note: This will only take effect during the downstream update after an upstream release. No way to verify this before upstream merge.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -22,8 +22,6 @@ packages:
 # https://packit.dev/docs/configuration/actions
 actions:
   get-current-version: bash -c 'grep "^const Version" version/version.go | cut -f2 -d\" | tr \- \~'
-  prepare-files: >-
-    bash -c "sed -i 's/^\(\s*\)ref: .*/\1ref: \"${PACKIT_PROJECT_TAG}\"/' ${PACKIT_DOWNSTREAM_REPO}/plans/main.fmf"
 
 srpm_build_deps:
   - make
@@ -122,6 +120,9 @@ jobs:
     update_release: false
     dist_git_branches: &fedora_targets
       - fedora-all
+    actions:
+        post-modifications: >-
+          bash -c "sed -i 's/^\(\s*\)ref: .*/\1ref: \"v${PACKIT_PROJECT_VERSION}\"/' ${PACKIT_DOWNSTREAM_REPO}/plans/main.fmf"
 
   # Sync to CentOS Stream
   # FIXME: Switch trigger whenever we're ready to update CentOS Stream via


lsm5:
cc @inknos RE: similar changes in podman-py.

lsm5:
`PACKIT_PROJECT_TAG` is probably another issue that needs fixing.

.packit.yaml
```diff
@@ -22,7 +22,7 @@ packages:
 # https://packit.dev/docs/configuration/actions
 actions:
   get-current-version: bash -c 'grep "^const Version" version/version.go | cut -f2 -d\" | tr \- \~'
-  prepare-files: >-
+  post-modifications: >-
     bash -c "sed -i 's/^\(\s*\)ref: .*/\1ref: \"${PACKIT_PROJECT_TAG}\"/' ${PACKIT_DOWNSTREAM_REPO}/plans/main.fmf"
```
nforro: I don't think `PACKIT_PROJECT_TAG` exists:

```suggestion
    bash -c "sed -i 's/^\(\s*\)ref: .*/\1ref: \"v${PACKIT_PROJECT_VERSION}\"/' ${PACKIT_DOWNSTREAM_REPO}/plans/main.fmf"
```

.packit.yaml
```diff
@@ -22,8 +22,8 @@ packages:
 # https://packit.dev/docs/configuration/actions
 actions:
   get-current-version: bash -c 'grep "^const Version" version/version.go | cut -f2 -d\" | tr \- \~'
-  prepare-files: >-
-    bash -c "sed -i 's/^\(\s*\)ref: .*/\1ref: \"${PACKIT_PROJECT_TAG}\"/' ${PACKIT_DOWNSTREAM_REPO}/plans/main.fmf"
+  post-modifications: >-
+    bash -c "sed -i 's/^\(\s*\)ref: .*/\1ref: \"v${PACKIT_PROJECT_VERSION}\"/' ${PACKIT_DOWNSTREAM_REPO}/plans/main.fmf"
```
inknos: is `PACKIT_DOWNSTREAM_REPO` the right variable to use?

nforro: It's a path to cloned dist-git repo, so yes, I would say it's right. Out of curiosity, why do you ask? Does the name seem out of place?

lsm5: I would say so: https://packit.dev/docs/configuration/actions#run-condition-1

lsm5:
ah I gotta move the post-modifications to the propose-downstream job probably. Otherwise it's trying to mess with copr builds too.

lsm5:
@nforro PTAL at the post-modifications move. Thanks a lot!

.packit.yaml
inknos: I got confused by seeing the variable empty in the tests and then I checked [here](https://packit.dev/docs/configuration/upstream/tests#environment-variables) and found `PACKIT_DOWNSTREAM_NAME` and `PACKIT_DOWNSTREAM_URL`. I guess my bad

lsm5:
@mtrmac PTAL. We have approval from @nforro (packit team). Will only take effect on the next release.

mtrmac (approved):
Thanks!

PR #2767, opened 2025-12-08
[release-1.21] Packit: use `post-modifications` hook to update downstream TMT plan

lsm5:
`prepare-files` action was interfering with spec file update which caused https://github.com/containers/skopeo/issues/2760 .

`post-modifications` needs to be limited to the propose_downstream job or else it will interfere with upstream PR copr builds.

Also, s/PACKIT_PROJECT_TAG/PACKIT_PROJECT_VERSION/ .

(cherry picked from commit e26a4237fc7bee11ac347c2bb0ae0ac821c169a0)


lsm5:
> LGTM assuming tests pass (not that they cover this, I think…), thanks!

Yes, no upstream testing of this.

PR #2771, opened 2025-12-10
[sha512] cmd/skopeo/copy: New `--force-digest` flag

lsm5:
See individual commits. 

Vendors https://github.com/containers/container-libs/pull/552

### DIFF
--- a/docs/skopeo-copy.1.md
+++ b/docs/skopeo-copy.1.md
@@ -71,6 +71,12 @@ This option does not change what will be copied; consider using `--all` at the s
 
 MANIFEST TYPE (oci, v2s1, or v2s2) to use in the destination (default is manifest type of source, with fallbacks)
 
+**--force-digest**=_algorithm_ **EXPERIMENTAL**
+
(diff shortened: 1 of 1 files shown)


cmd/skopeo/copy.go
```diff
@@ -29,6 +30,7 @@ type copyOptions struct {
 	additionalTags      []string                  // For docker-archive: destinations, in addition to the name:tag specified as destination, also add these
 	signIdentity        string                    // Identity of the signed image, must be a fully specified docker reference
 	digestFile          string                    // Write digest to this file
+	forceDigest         string                    // Force a specific digest algorithm (sha256, sha384, or sha512)
```
mtrmac, line 33: This would probably be better placed in `sharedCopyOptions`

mtrmac (commented):
I think this would follow https://github.com/containers/container-libs/pull/552 WRT option naming / timing / stability.

---

(IIRC there are tests that validate man pages vs. `--help` output; so splitting the feature addition and a man page update results in a commit that doesn’t pass — it would be better to squash that into one. Admittedly running `git bisect` on Skopeo is probably very rare.)

PR #2789, opened 2026-01-16
[release-1.11] CVE-2025-65637 Logrus

TomSweeneyRedHat:
Bump sirupsen/logrus to v1.9.3 to address  CVE-2025-65637

Fixes: https://issues.redhat.com/browse/OCPBUGS-67731,
https://issues.redhat.com/browse/OCPBUGS-67898,
https://issues.redhat.com/browse/OCPBUGS-68038,
https://issues.redhat.com/browse/RHEL-135829,
https://issues.redhat.com/browse/RHEL-135814

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/opencontainers/image-spec v1.1.0-rc2
 	github.com/opencontainers/image-tools v1.0.0-rc3
-	github.com/sirupsen/logrus v1.9.0
+	github.com/sirupsen/logrus v1.9.3
 	github.com/spf13/cobra v1.6.1
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.8.1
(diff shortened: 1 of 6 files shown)


mtrmac (commented):
LGTM but lint is unhappy, does `go mod tidy` help?

TomSweeneyRedHat:
Thought I had tidied, but I will try again.

TomSweeneyRedHat:
@mtrmac looks like I missed a `go mod tidy` error last week.  It complained,  with a few errors like this:

```
go: github.com/containers/skopeo/cmd/skopeo tested by
	github.com/containers/skopeo/cmd/skopeo.test imports
	github.com/stretchr/testify/assert loaded from github.com/stretchr/testify@v1.8.1,
	but go 1.16 would select v1.8.3
```
and the solution was to do this instead:

`go mod tidy -compat=1.17`   

But I'm not sure where the 1.16 is coming from as the go.mod file has 1.17....  Thoughts?

Anyway, I went with Go's suggestion and repushed.  The only change I could see was that the entries for logrus in go.sum had different hashes.

TomSweeneyRedHat:
and logrus is also apparently using Go 1.17.  I thought that might be the answer.

mtrmac (approved):
Thanks!

mtrmac:
> But I'm not sure where the 1.16 is coming from as the go.mod file has 1.17.... Thoughts?

`go help mod tidy` says:
>  By default, tidy acts as if the -compat flag were set to the
> version prior to the one indicated by the 'go' directive in the go.mod
> file.

and IIRC there was a fairly significant change to the mechanism/semantics between 1.16 and 1.17. Anyway we have no reason to support consuming this module via Go 1.16.

PR #2810, opened 2026-02-26
Packit: fix downstream post-modifications action

lsm5:
The downstream TMT plan can use a ref value without double quotes[0], so the `post-modifications` action in packit config can be simplified as in this patch. Also, switch the single and double quote usage to be consistent with the usage in `get-current-version` packit action.

This should help to avoid variable substitution issues noticed in the previous release [1].

Refs:
[0] https://src.fedoraproject.org/rpms/skopeo/pull-request/98
[1] https://src.fedoraproject.org/fork/packit/rpms/skopeo/blob/fd38aa8a5e4ed14b70a1e3ba2212434ce39fe5a2/f/plans/main.fmf

Doesn't affect upstream. Change can only be verified on a new release.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -122,7 +122,7 @@ jobs:
       - fedora-all
     actions:
         post-modifications: >-
-          bash -c "sed -i 's/^\(\s*\)ref: .*/\1ref: \"v${PACKIT_PROJECT_VERSION}\"/' ${PACKIT_DOWNSTREAM_REPO}/plans/main.fmf"
+          bash -c 'sed -i "s/^\(\s*\)ref: .*/\1ref: v${PACKIT_PROJECT_VERSION}/" ${PACKIT_DOWNSTREAM_REPO}/plans/main.fmf'
 
   # Sync to CentOS Stream
   # FIXME: Switch trigger whenever we're ready to update CentOS Stream via


lsm5:
@nforro PTAL.

.packit.yaml
```diff
@@ -122,7 +122,7 @@ jobs:
       - fedora-all
     actions:
         post-modifications: >-
-          bash -c "sed -i 's/^\(\s*\)ref: .*/\1ref: \"v${PACKIT_PROJECT_VERSION}\"/' ${PACKIT_DOWNSTREAM_REPO}/plans/main.fmf"
+          bash -c 'sed -i "s/^\(\s*\)ref: .*/\1ref: v${PACKIT_PROJECT_VERSION}/" ${PACKIT_DOWNSTREAM_REPO}/plans/main.fmf'
```
nforro, line 125: Are you sure the single quotes won't prevent expansion of `PACKIT_PROJECT_VERSION`?

lsm5: I ran the entire `bash -c ...` as-is on terminal and worked just fine. Alternatively, maybe I could just get rid of `bash -c` and have just the `sed ...` as the action. Do you see any issues with that?

nforro: > Do you see any issues with that?

Without a shell the variables won't be expanded for sure.

> I ran the entire `bash -c ...` as-is on terminal and worked just fine.

I'm not sure that's the way to reproduce it. Let me try to come up with a (hopefully) simple way.

nforro: I hope this corresponds to what actually happens internally, and it does work:

```python
import shlex
import subprocess

command = r"""
bash -c 'sed -i "s/^\(\s*\)ref: .*/\1ref: v${PACKIT_PROJECT_VERSION}/" ${PACKIT_DOWNSTREAM_REPO}/plans/main.fmf'
"""

env = {
    "PACKIT_PROJECT_VERSION": "1.0",
    "PACKIT_DOWNSTREAM_REPO": "/tmp/dg",
}

subprocess.call(shlex.split(command), env=env)
```

nforro (approved):

.packit.yaml
lsm5: thanks for that!

lsm5:
@containers/skopeo-maintainers PTAL

mtrmac (approved):
Thanks!

Issue #2829, opened 2026-03-21
tests: quay.io/coreos/11bot is only available for amd64


ricardobranco777:
This prevents us from running the integration tests on other architectures.

https://openqa-assets.opensuse.org/tests/5768459/file/skopeo-integration.txt

```
Failed
=== RUN   TestProxy/TestProxyMetadata
    proxy_test.go:490: 
        	Error Trace:	/var/tmp/skopeo/integration/proxy_test.go:490
        	Error:      	Received unexpected error:
        	            	Testing image docker://quay.io/coreos/11bot: remote error: no image found in image index for architecture "arm64", variant "v8", OS "linux"
        	Test:       	TestProxy/TestProxyMetadata
--- FAIL: TestProxy/TestProxyMetadata (3.91s)
```

mtrmac:
This was fixed in #2834 . Thanks!

PR #2830, opened 2026-03-21
integration: Skip TestProxyMetadata on non-amd64 architectures

ricardobranco777:
Skip TestProxyMetadata on non-amd64 architectures.

Otherwise test fails on other architectures like explained here: https://github.com/containers/skopeo/issues/2829

### DIFF
--- a/integration/proxy_test.go
+++ b/integration/proxy_test.go
@@ -9,6 +9,7 @@ import (
 	"net"
 	"os"
 	"os/exec"
+	"runtime"
 	"strings"
 	"sync"
 	"syscall"
@@ -477,6 +478,11 @@ func (s *proxySuite) TestProxyMetadata() {
 	t := s.T()
 	p, err := newProxy()
 	require.NoError(t, err)
+
+	// https://github.com/containers/skopeo/issues/2829
+	if runtime.GOARCH != "amd64" {
+		t.Skip("test image is only available on amd64")
+	}
 
 	err = runTestMetadataAPIs(p, knownNotManifestListedImageX8664)
 	if err != nil {


ricardobranco777:
Verification runs:
- aarch64: https://openqa.opensuse.org/tests/5768477
- ppc64le: https://openqa.opensuse.org/tests/5768476

ricardobranco777:
- https://github.com/containers/podman/issues/28269
- https://github.com/containers/podman/issues/28270
- https://github.com/containers/podman/issues/28272
- https://github.com/containers/podman/issues/28273

lsm5 (approved):
LGTM

@mtrmac @containers/skopeo-maintainers PTAL

cgwalters (approved):

mtrmac (commented):
Thanks!

This works, but I think this test could run the proxy with `--override-os amd64`, and that would have better test coverage. WDYT?

TomSweeneyRedHat:
@ricardobranco777 you need a rebase.  Thoughts on @mtrmac 's last comment.
LGTM once that's settled.

ricardobranco777:
> @ricardobranco777 you need a rebase. Thoughts on @mtrmac 's last comment. LGTM once that's settled.

Just rebased.  I tried the suggested approach in https://github.com/containers/skopeo/pull/2834 but it fails when I backport it to v1.20.0.

TomSweeneyRedHat:
@ricardobranco777 thanks for the background.  The tests are very unhappy; they look like flakes.  I've restarted.

mtrmac:
@ricardobranco777 Why shouldn’t we use use #2834? Even if that doesn’t work for a backport, I think you would be backporting something that does not exist in an upstream release _either way_, and either way after you update, the backport can be removed. I don’t see that the backport situation makes much of a difference to anyone (other than I’m somewhat curious why it is failing there). Am I missing something?

ricardobranco777:
> @ricardobranco777 Why shouldn’t we use use #2834? Even if that doesn’t work for a backport, I think you would be backporting something that does not exist in an upstream release _either way_, and either way after you update, the backport can be removed. I don’t see that the backport situation makes much of a difference to anyone (other than I’m somewhat curious why it is failing there). Am I missing something?

I'm not saying we shouldn't use it, just that I have no way to verify it works on non-amd64 hosts, short of compiling skopeo.  Otherwise I think we can merge this to skip the test until I can verify the other works.

ricardobranco777:
Closing this in favor of https://github.com/containers/skopeo/pull/2834

PR #2834, opened 2026-03-23
integration: Force amd64 on TestProxyMetadata

ricardobranco777:
Force amd64 on TestProxyMetadata

Otherwise test fails on other architectures due to https://github.com/containers/skopeo/issues/2829

Fixes https://github.com/containers/skopeo/pull/2830

### DIFF
--- a/integration/proxy_test.go
+++ b/integration/proxy_test.go
@@ -261,7 +261,7 @@ func (p *proxy) callGetRawBlob(args []any) (rval any, buf []byte, err error) {
 	return
 }
 
-func newProxy() (*proxy, error) {
+func newProxy(extraArgs ...string) (*proxy, error) {
 	fds, err := syscall.Socketpair(syscall.AF_LOCAL, syscall.SOCK_SEQPACKET, 0)
 	if err != nil {
 		return nil, err
@@ -277,7 +277,8 @@ func newProxy() (*proxy, error) {
 	}
 
 	// Note ExtraFiles starts at 3
-	proc := exec.Command(skopeoBinary, "experimental-image-proxy", "--sockfd", "3")
+	args := append(append([]string{}, extraArgs...), "experimental-image-proxy", "--sockfd", "3")
+	proc := exec.Command(skopeoBinary, args...)
 	proc.Stderr = os.Stderr
 	cmdLifecycleToParentIfPossible(proc)
 	proc.ExtraFiles = append(proc.ExtraFiles, theirfd)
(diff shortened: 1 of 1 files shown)


integration/proxy_test.go
```diff
@@ -475,7 +476,8 @@ func runTestGetBlob(p *proxy, img string) error {
 
 func (s *proxySuite) TestProxyMetadata() {
 	t := s.T()
-	p, err := newProxy()
+	// https://github.com/containers/skopeo/issues/2829
```
mtrmac: Absolutely non-blocking: Something like “The docker://quay.io/coreos/11bot image only contains an amd64 instance, the test would fail on other platforms without this”?

That would save future readers a click, and avoid an online service dependency.

mtrmac (commented):
Thanks!

LGTM if tests (including on the platforms you care about) pass.

ricardobranco777:
> Thanks!
> 
> LGTM if tests (including on the platforms you care about) pass.

This approach fails because the override option isn't picked up by the proxy command.

I get:
http://openqa-assets.opensuse.org/tests/5776482/file/skopeo-integration.txt

```
# Test messages # TestProxy/TestProxyMetadata
# failure: 

Failed
=== RUN   TestProxy/TestProxyMetadata
    proxy_test.go:491: 
        	Error Trace:	/var/tmp/skopeo/integration/proxy_test.go:491
        	Error:      	Received unexpected error:
        	            	Testing image docker://quay.io/coreos/11bot: remote error: no image found in image index for architecture "arm64", variant "v8", OS "linux"
        	Test:       	TestProxy/TestProxyMetadata
--- FAIL: TestProxy/TestProxyMetadata (4.77s
```

mtrmac:
@ricardobranco777 That’s weird; I see a failure on `aarch64` when running the tests against `main`, but both your 67584f1c28c248fe3759296e4f80eb378f9b48de and adc2bf4d8b7505ee3346be35ab40a4595badac0e let the tests pass for me.

ricardobranco777:
> @ricardobranco777 That’s weird; I see a failure on `aarch64` when running the tests against `main`, but both your [67584f1](https://github.com/containers/skopeo/commit/67584f1c28c248fe3759296e4f80eb378f9b48de) and [adc2bf4](https://github.com/containers/skopeo/commit/adc2bf4d8b7505ee3346be35ab40a4595badac0e) let the tests pass for me.

I'm not running against main.  I'm backporting this PR to the v1.20.0 version we have in openSUSE Tumbleweed.

mtrmac:
I can’t see any obvious change (at least in Skopeo itself) since 1.20 that could explain that behavior, but I also didn’t try that against 1.20 now.

ricardobranco777:
skopeo v1.22.0 just landed on Tumbleweed and this patch passes:

- aarch64: https://openqa.opensuse.org/tests/5797830
- ppc64le: https://openqa.opensuse.org/tests/5797831

mtrmac (approved):
Thank you!

PR #2843, opened 2026-04-09
Bump  google.golang.org/grpc to v1.79.3 - CVE-2026-33186

TomSweeneyRedHat:
Even though it appears that Skopeo is not affected by CVE-2026-33186, I have been pinged about it already.  Just to keep the CVE sniffers at bay, bump to the appropriate version of gRPC.


mtrmac (approved):
Thanks!

PR #2849, opened 2026-04-13
Update go.podman.io dependencies

mtrmac:
Manually update to commits on main, because automation would find latest tags on a release branch and downgrade to them instead.


mtrmac:
@TomSweeneyRedHat @Luap99 PTAL. (And other projects might need something similar.)

Luap99 (approved):
LGTM

PR #2854, opened 2026-04-17
Move skopeo to go.podman.io

baude:
With potential repo moves and cirus EOL, we have decided to begin preparing by making changes to the skopeo module name.  See https://github.com/containers/go.podman.io/pull/6 for more information.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -11,7 +11,7 @@ env:
     GOPATH: &gopath "/var/tmp/go"
     GOBIN: "${GOPATH}/bin"
     GOCACHE: "${GOPATH}/cache"
-    GOSRC: &gosrc "/var/tmp/go/src/github.com/containers/skopeo"
+    GOSRC: &gosrc "/var/tmp/go/src/go.podman.io/skopeo"
     # Required for consistency with containers/image CI
     SKOPEO_PATH: *gosrc
     CIRRUS_WORKING_DIR: *gosrc
--- a/Makefile
+++ b/Makefile
@@ -80,7 +80,7 @@ INTERACTIVE := $(shell [ -t 0 ] && echo 1 || echo 0)
 ifeq ($(INTERACTIVE), 1)
 	CONTAINER_CMD += -t
 endif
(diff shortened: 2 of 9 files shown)


Luap99 (changes requested):
you will have to also rename the import in the Makefile where it is used to filter the tests
https://github.com/containers/skopeo/blob/904ff20c4de08ddeb420d44b5ba23a6130cccf79/Makefile#L266

Without this it tries to run all tests as part of unit tests which is not what we want and fails on osx

mtrmac (commented):
ACK to the idea and to these particular changes; but a full text search for `github.com/containers/skopeo` finds various repo-name-derived paths that would be at least misleading (or perhaps no longer work) after this.

The `osx` failure is real, probably due to a non-updated `github\.com/containers/skopeo` regex.

mtrmac (changes requested):
Also (frequently assuming a `GOPATH` layout; we could probably change that but let‘s not do that in this time-sensitive PR):
- `hack/get_ci_vm.sh`
- `.cirrus.yml`
- `install.md`
- `Makefile`
- `README.md`

differentiating between GitHub URLs (unchanged) and module name (changed)

Luap99 (approved):
LGTM

mtrmac (approved):
Thanks!

Non-blocking: Also https://github.com/containers/skopeo/blob/c0c63bb96ee59f06adace11beb44ab759ac9f71b/README.md?plain=1#L8 I think. It seems that it processes the module alias by redirecting back to the GitHub module name, so that doesn’t make much of a difference. (Alternatively, drop that? I don‘t know what good that badge does.)

---

There’s _some_ risk that this will break `image_test_skopeo_task` in container-libs, due to similar `GOPATH`-like references to the repo. I don’t think we can avoid that risk — switching container-libs to the new module name while the module does not use the new name yet would also not be consistent.

So just a FYI that this might happen today.

mtrmac:
@baude please rebase.

baude:
Given the rebase, I removed the readme you mentioned.

mtrmac (approved):

mtrmac:
Thanks!

PR #2870, opened 2026-05-06
copy: add --progress-interval for CI-compatible progress output

### LINKED ISSUE #658: Please provide a basic progress / transfer speed output ("CI compatible") on copy

IanEff:
In a pipe, skopeo copy was silent until success or failure.  

Add `--progress-interval` to the `copy` command.  When stderr is not a TTY, periodic timestamps are emitted to stderr with current blob's offset, total size, and transfer rate.  Defaults to 30s in non-TTY environments, & is disabled when a TTY is detected or when --quiet is passed.

Closes https://github.com/containers/skopeo/issues/658

### DIFF
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"os"
 	"strings"
+	"time"
 
 	encconfig "github.com/containers/ocicrypt/config"
 	enchelpers "github.com/containers/ocicrypt/helpers"
@@ -17,6 +18,7 @@ import (
 	"go.podman.io/image/v5/manifest"
 	"go.podman.io/image/v5/transports"
 	"go.podman.io/image/v5/transports/alltransports"
+	"go.podman.io/image/v5/types"
 )
 
 type copyOptions struct {
@@ -36,6 +38,7 @@ type copyOptions struct {
 	encryptionKeys      []string                  // Keys needed to encrypt the image
 	decryptionKeys      []string                  // Keys needed to decrypt the image
 	imageParallelCopies uint                      // Maximum number of parallel requests when copying images
+	progressInterval    time.Duration             // interval for CI progress output; 0 = auto-detect
 }
 
 func copyCmd(global *globalOptions) *cobra.Command {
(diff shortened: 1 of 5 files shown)


cmd/skopeo/copy.go
```diff
@@ -242,7 +250,17 @@ func (opts *copyOptions) run(args []string, stdout io.Writer) (retErr error) {
 	copyOpts.MaxParallelDownloads = opts.imageParallelCopies
 	copyOpts.ForceCompressionFormat = opts.destImage.forceCompressionFormat
 
+	if opts.progressInterval > 0 {
+		copyOpts.ProgressInterval = opts.progressInterval
+		copyOpts.ReportWriter = nil // suppress TTY bar when logging to stderr
```
mtrmac, line 271: I don’t see where `stderr` figures in this logic at all. Is that just due to `resolveProgressInterval` deciding based on that?

---

`ReportWriter` contains quite a bit of other information, not just blob progress, so completely turning it off is not great.

cmd/skopeo/copy_progress.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+func runProgressConsumer(ch <-chan types.ProgressProperties, interval time.Duration, out io.Writer) {
+	type blobState struct {
+		total int64
+	}
+	blobs := map[string]*blobState{}
+
+	for props := range ch {
+		dig := props.Artifact.Digest.Encoded()
```
mtrmac, line 21: `Encoded` does not include the algorithm , that’s inaccurate as a map key.

cmd/skopeo/copy_progress.go
```diff
@@ -0,0 +1,60 @@
...
+
+		switch props.Event {
+		case types.ProgressEventNewArtifact:
+			blobs[dig] = &blobState{total: props.Artifact.Size}
+
+		case types.ProgressEventRead:
+			b := blobs[dig]
+			bytesPerSec := float64(props.OffsetUpdate) / interval.Seconds()
```
mtrmac, line 29: ??? `interval.Seconds` is a constant throughout the life of the function.

mtrmac (changes requested):
Thanks!

Just a _very_ quick look now: I think this is just the wrong place, and the way this needs to completely turn off `ReportWriter` is an indication of that. If we can make the c/image output better, that improves all callers, including notably Podman, and we can design a reasonable output there; maintaining code that writes progress reports in one repo and code that discards those progress reports in another repo seems rather unnecessary.

TomSweeneyRedHat:
Neat concept, thanks @IanEff !

Our CI requires that commits be signed, and it looks like you might need a rebase too.  Adjust per @mtrmac 's comments, do a `git commit --amend -s` for each of your commits, rebase, and push it again.

TomSweeneyRedHat:
@IanEff looks like this needs a rebase.

IanEff:
@mtrmac I'm inclined to agree.  I'll abandon this and go sniffing upstream in c/image, make the changes there.  Thank you for your feedback on my commit!

@TomSweeneyRedHat Thanks!

PR #2875, opened 2026-05-14
Feat/download foreign layers

### LINKED ISSUE #545: Add an option to copy foreign layers
I would like to skopeo copy a windows image to a linux host to create (via pure filesystem manipulation with umoci) new layers. Windows is dead slow with a lot of files, so I'd rather use linux :) I don't need to be able to run any commands, thus I don't need support for Dockerfile, RUN, etc. I just want to add files, and commit a new layer.

Trying out skopeo to do this, I unfortunately get:

```
skopeo --override-os windows --override-arch amd64 copy docker://microsoft/windowsservercore:ltsc2016 oci:windows-test:latest
Skipping foreign layer "sha256:3889bb8d808bbae6fa5a33e07093e65c31371bcf9e4c38c21be6b9af52ad1548" copy to oci
Skipping foreign layer "sha256:6631c2d2a60cd7ee5b334c2725b03e4d4976abb9a19c8e8dc9b806b3752745a6" copy to oci
Copying config sha256:9657141ae0624406d5e84631b2cab0eee14d2810a7ca29250d32d1babb68101b
 493 B / 493 B [============================================================] 0s
Storing signatures
```

Any advice? Should this work?

IanEff:
Add `--download-foreign-layers` to `skopeo copy` and `skopeo sync`.

When set, the flag instructs containers/image to fetch the contents of
nondistributable ("foreign") layers — e.g. Windows base layers — instead
of preserving only a URL reference to their original location. This is
useful for air-gapped copies of images whose upstream source may later
become unreachable.

`copy.Options.DownloadForeignLayers` fetches the blob but doesn't rewrite the layer's media type or clear the `urls` field. The man pages and integration test describe that behavior directly, rather than promising a rewrite that isn't coming.

This should suffice to close this issue.

Closes #545


TomSweeneyRedHat:
@IanEff this needs a rebase also.

kikar:
@IanEff Do you mind rebasing one more time? I've also been hit by this problem and would like a solution.

@TomSweeneyRedHat is the PR correct in principle? Can it be approved after rebase?

IanEff:
@kikar No conflicts in the rebase, so no trouble!  

@TomSweeneyRedHat: this is all clean, as per CONTRIBUTIONS, with an explanation in the PR.

kikar:
@mtrmac Any chances this can be reviewed?

docs/skopeo-copy.1.md
```diff
@@ -63,6 +63,12 @@ Preserve the digests during copying. Fail if the digest cannot be preserved.
 
 This option does not change what will be copied; consider using `--all` at the same time.
 
+**--download-foreign-layers**
+
+Download layer contents for nondistributable ("foreign") layers — e.g. Windows base layers — that would otherwise be skipped in favor of URL references to their original location. By default, foreign layers are not downloaded.
```
TomSweeneyRedHat:
```suggestion
Download layer contents for nondistributable ("foreign") layers — e.g., Windows base layers — that would otherwise be skipped in favor of URL references to their original location. By default, foreign layers are not downloaded when the destination can store URL references instead (notably `oci:` and `docker://`). Other destinations already fetch them.
```

integration/copy_test.go
```diff
@@ -555,6 +555,85 @@ func (s *copySuite) TestCopySucceedsWhenImageDoesNotMatchRuntimeButWeOverride()
...
+	}
+	require.NoError(t, json.Unmarshal(manifestBlob, &m))
+
+	var foreignDigests []digest.Digest
+	for _, l := range m.Layers {
+		if len(l.URLs) > 0 {
+			foreignDigests = append(foreignDigests, l.Digest)
+		}
+	}
+	require.NotEmpty(t, foreignDigests, "test image was expected to have at least one foreign layer; manifest: %s", manifestBlob)
```
TomSweeneyRedHat: Is this correct?  If the url is stripped at some point, will this fail?

TomSweeneyRedHat:
@IanEff needs an update, and the work in here is really @mtrmac's specialty; we'll need to grab some of his eyeball time on it.  After the doc change, and any @mtrmac may also have.

integration/copy_test.go
IanEff: Resolved in subsequent commit.

IanEff:
@TomSweeneyRedHat Addressed the issues you've raised in the PR.

kikar:
@mtrmac any chances of review?

docs/skopeo-sync.1.md
```diff
@@ -88,6 +88,12 @@ Preserve the digests during copying. Fail if the digest cannot be preserved.
 
 This option does not change what will be copied; consider using `--all` at the same time.
 
+**--download-foreign-layers**
+
+Download layer contents for nondistributable ("foreign") layers — e.g., Windows base layers — that would otherwise be skipped in favor of URL references to their original location. By default, foreign layers are not downloaded when the destination can store URL references instead (notably `oci:` and `docker://`). Other destinations already fetch them.
+
+Useful for air-gapped or offline copies of images whose original source may later become unreachable. The destination manifest is not rewritten: layers retain their original nondistributable media type and `urls` field.
```
mtrmac: I’d expect the two man pages to contain exactly the same text.

integration/copy_test.go
```diff
@@ -555,6 +555,70 @@ func (s *copySuite) TestCopySucceedsWhenImageDoesNotMatchRuntimeButWeOverride()
...
+// TestCopyDownloadForeignLayers verifies the user-visible effect of
+// --download-foreign-layers: with the flag the destination holds the contents of
+// nondistributable ("foreign") layers; without it those layers are skipped in favor of
+// the URL references carried by the source manifest.
+//
+// The destination is an oci: layout, the case containers/skopeo#545 was filed for: oci:
+// reports AcceptsForeignLayerURLs() == true, so by default the library preserves the URLs
+// and never fetches the blob. Destinations reporting false (dir:, containers-storage:,
+// docker-archive:) download foreign layers with or without this flag.
+//
```
mtrmac: I don’t particularly think this needs documenting here in that detail; if the test exercises both cases, that sufficiently shows that `oci:` is a suitable test method.

integration/copy_test.go
```diff
@@ -555,6 +555,70 @@ func (s *copySuite) TestCopySucceedsWhenImageDoesNotMatchRuntimeButWeOverride()
 		"containers-storage:"+storage+"test")
 }
 
+// TestCopyDownloadForeignLayers verifies the user-visible effect of
+// --download-foreign-layers: with the flag the destination holds the contents of
```
mtrmac: The flag meaning doesn’t need to be re-explained here.

integration/copy_test.go
```diff
@@ -555,6 +555,70 @@ func (s *copySuite) TestCopySucceedsWhenImageDoesNotMatchRuntimeButWeOverride()
...
+//
+// The destination is an oci: layout, the case containers/skopeo#545 was filed for: oci:
+// reports AcceptsForeignLayerURLs() == true, so by default the library preserves the URLs
+// and never fetches the blob. Destinations reporting false (dir:, containers-storage:,
+// docker-archive:) download foreign layers with or without this flag.
+//
+// The assertions deliberately never read the "urls" field of the layout written *with* the
+// flag. copy.Options.DownloadForeignLayers is documented to also translate the layer media
+// type to not indicate "nondistributable", which would imply dropping "urls"; as of the
+// vendored containers/image that rewrite does not happen. Keying the with-flag assertion
```
mtrmac: This PR needs to decide whether that behavior is a bug to be fixed (and in that case the man pages should not be promising that behavior) or a design decision (and in that case this comment should not be suggesting that this is a bug that might be fixed).

Given the 5-year history, and c/image’s stable API commitment, I think it’s reasonable to say that the semantics of `DownloadForeignLayers` is ~by definition correct and the documentation of that flag is what needs to be updated.

… but that kind of depends on your use case: if you were unhappy with that behavior and actually wanted it to be the other one, that should happen now before we add the CLI flag.

docs/skopeo-copy.1.md
```diff
@@ -63,6 +63,12 @@ Preserve the digests during copying. Fail if the digest cannot be preserved.
 
 This option does not change what will be copied; consider using `--all` at the same time.
 
+**--download-foreign-layers**
+
+Download layer contents for nondistributable ("foreign") layers — e.g., Windows base layers — that would otherwise be skipped in favor of URL references to their original location. By default, foreign layers are not downloaded when the destination can store URL references instead (notably `oci:` and `docker://`). Other destinations already fetch them.
```
mtrmac: This is saying ~the same thing ~three times. The “already” sentence is clearly redundant; the “otherwise” clause is also probably not necessary given the description of the default behavior.

integration/copy_test.go
```diff
@@ -555,6 +555,70 @@ func (s *copySuite) TestCopySucceedsWhenImageDoesNotMatchRuntimeButWeOverride()
...
+// off "urls" would turn a future upstream fix into a test failure. So the foreign layers
+// are identified from the control copy — which cannot lose its URLs, because nothing was
+// downloaded and the URL is the only pointer to the content — and the with-flag layout is
+// then required to be self-contained: every layer it lists is present on disk.
+func (s *copySuite) TestCopyDownloadForeignLayers() {
+	t := s.T()
+	withFlag := t.TempDir()
+	withoutFlag := t.TempDir()
+	assertSkopeoSucceeds(t, "", "--override-os=windows", "--override-arch=amd64", "copy", "--retry-times", "3", knownWindowsOnlyImage, "oci:"+withoutFlag)
```
mtrmac: This is a 100 MB download, and potentially impacts CI reliability.

A few-bytes image in `integration/fixtures`, using the `dir:` format as a source, would be better.

integration/copy_test.go
```diff
@@ -555,6 +555,70 @@ func (s *copySuite) TestCopySucceedsWhenImageDoesNotMatchRuntimeButWeOverride()
...
+	var foreignDigests []digest.Digest
+	for _, l := range ociLayoutLayers(t, withoutFlag) {
+		if len(l.URLs) != 0 {
+			foreignDigests = append(foreignDigests, l.Digest)
+		}
+	}
+	require.NotEmpty(t, foreignDigests, "%s was expected to contain at least one foreign layer", knownWindowsOnlyImage)
```
mtrmac: … and with a fixture, this can be a hard-coded digest value.

integration/copy_test.go
```diff
@@ -555,6 +555,70 @@ func (s *copySuite) TestCopySucceedsWhenImageDoesNotMatchRuntimeButWeOverride()
...
+	for _, d := range foreignDigests {
+		assert.NoFileExists(t, ociBlobPath(withoutFlag, d), "foreign layer %s should not be downloaded withou --download-foreign-layers", d)
```
mtrmac: (typo)

mtrmac (commented):
Thanks!

This makes sense overall.

Please squash the commits; it is valuable to have “one idea per commit” but we don’t need every WIP version; in particular, commits that don’t even compile are a hassle for `git bisect`.

IanEff:
@mtrmac: I've modified the PR and addressed the issues raised in the last push.  Totally agreed that this is de facto settled behavior.

mtrmac (approved):
Thanks!

PR #2884, opened 2026-05-30
feat: inspect: surface the image's Comment field

### LINKED ISSUE #2098: Comment section from inspect
Hi guys,

I'm trying to get the Comment attribute from the inspect. However, it doesn't seem to be available on the output. Is there something that I'm missing?
If I do podman inspect on that same local image, the comment attribute is there.

Thanks,
Luís

pen-pal:
## Problem

`skopeo inspect` does not show the image's `Comment` field. `docker inspect` and `podman inspect` do.

## Fix

Populate `Output.Comment` from the image config blob: top-level `comment` first, then last non-empty `history[].comment`. Field is `omitempty` — images without either are unchanged.

## Tests

`TestCommentFromConfigBlob` — 11 table-driven subtests (top-level, history, fallback, precedence, malformed, empty). Passes with `-race`.

Verified end-to-end against podman inspect on an Ubuntu Lima VM: byte-equal output for buildah-commented and alpine images, across `containers-storage:`, `dir:`, `oci:`, `oci-archive:`, `docker-archive:`, `docker-daemon:`, and `docker://` transports.

Fixes: https://github.com/containers/skopeo/issues/2098


mtrmac (changes requested):
Thanks!

No, `skopeo` itself should not be implementing this (and especially not by blindly expecting a specific JSON structure without checking the MIME type). This should be added to c/image, probably somewhere in `types.ImageInspectInfo`, and implemented there.

(Also compare https://github.com/containers/image/pull/2127/changes#diff-b842413dc81f44e58b840a4c0ca1debfc4777167c3e0681292e83d8049f6eef3 , note especially the review discussion about OCI format support.)

---

Also, https://github.com/podman-container-tools/skopeo/blob/main/CONTRIBUTING.md#sign-your-prs :
> Use your real name (sorry, no pseudonyms or anonymous contributions.)

PR #2885, opened 2026-06-01
[draft/DNM] GHA CI nits

kolyshkin:


mtrmac (commented):
(ACK to the concept of both — the `safe.directory` thing was also bothering me but we didn’t take the time to figure out why it was necessary.)

PR #2887, opened 2026-06-03
ci: switch self-hosted runners from Oracle to CNCF Ubuntu

timcoding1988:
The Oracle VM runners are discouraged to used by CNCF. using CNCF-* runners


mtrmac (approved):
Thanks!

PR #2888, opened 2026-06-03
[CI]: pull skopeo_cidev image from ghcr instead of quay

timcoding1988:
The skopeo_cidev image is built and published from the podman-container-tools/automation repo to
ghcr.io/podman-container-tools/skopeo_cidev.
addressed issue: https://github.com/podman-container-tools/automation/issues/9

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
     runs-on: cncf-ubuntu-4-16-x86
     timeout-minutes: 45
     container:
-      image: quay.io/libpod/skopeo_cidev:c20260310t170224z-f43f42d14 # FIXME: Should be Renovate-managed.
+      image: ghcr.io/podman-container-tools/skopeo_cidev:20260603t174659z # FIXME: Should be Renovate-managed.
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6.0.3
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -33,7 +33,7 @@ jobs:
     runs-on: oracle-vm-4cpu-16gb-x86-64
     timeout-minutes: 45
     container:
-      image: quay.io/libpod/skopeo_cidev:c20260310t170224z-f43f42d14 # FIXME: Should be Renovate-managed.
+      image: ghcr.io/podman-container-tools/skopeo_cidev:20260603t174659z # FIXME: Should be Renovate-managed.
```
timcoding1988, line 36: @Luap99  1. ~~need set this image to public (i guess same as the image we published at quay.io)~~
2. do we need to migrate https://github.com/containers/automation/blob/main/renovate/defaults.json5 to new automation repo or we should make the change at /containers/automation ?

Luap99 (approved):
LGTM
cc @mtrmac

mtrmac (approved):
Thank you!

.github/workflows/ci.yml
Luap99: I think we should make a new renovate config in the new automation repo
https://github.com/podman-container-tools/podman/issues/28831

FWIW renovate defaults already picked that up as valid container image in the dashboard so no config changes needed and I think the comment can be dropped but I guess we better wait for the next time we do updates.
https://github.com/podman-container-tools/skopeo/issues/1782
https://developer.mend.io/github/podman-container-tools/skopeo

PR #2892, opened 2026-06-08
Add support for manifest list signature removal and sparse manifest list stripping

### LINKED ISSUE #2882: skopeo copy --multi-arch with specific platforms fails with "blob unknown to registry" when manifest list contains other platforms
Description

When using skopeo copy with --multi-arch "linux/amd64,linux/arm64" to copy only specific platforms from a multi-arch image, the command fails with blob unknown to registry error. The issue appears to be that the manifest list still references other platforms (e.g., linux/386, linux/arm/v7) that were not uploaded to the destination registry.

Steps to Reproduce

Source image is a multi-arch manifest list containing multiple platforms:

bash
skopeo inspect docker://registry.k8s.io/pause:3.9 --raw | jq '.manifests[].platform'
Output shows platforms: linux/amd64, linux/arm64, linux/386, linux/arm/v7, etc.

Attempt to copy only linux/amd64 and linux/arm64 platforms:

bash
skopeo copy \
--multi-arch "linux/amd64,linux/arm64" \
--src-tls-verify=false \
--dest-tls-verify=false \
docker://registry.k8s.io/pause:3.9 \
docker://myregistry.com/registry.k8s.io/pause:3.9
Error occurs:

text
FATA[0001] Uploading manifest list failed, attempted the following formats:
application/vnd.docker.distribution.manifest.list.v2+json(uploading manifest 3.9 to myregistry.com/registry.k8s.io/pause: blob unknown to registry),
application/vnd.oci.image.index.v1+json(uploading manifest 3.9 to myregistry.com/registry.k8s.io/pause: blob unknown to registry)
Expected Behavior

When specifying --multi-arch "linux/amd64,linux/arm64", skopeo should:

Download only the specified platform images

Create a new manifest list containing ONLY these platforms

Successfully push the filtered manifest list to the destination registry

Actual Behavior

skopeo attempts to push the original manifest list (containing all platforms) to the destination registry, but the blobs for unspecified platforms (e.g., linux/386, linux/arm/v7) don't exist in the destination, causing blob unknown errors.

Root Cause Analysis

Based on code inspection (parseMultiArch function in cmd/skopeo/copy.go), when --multi-arch "linux/amd64,linux/arm64" is provided:

go
case "linux/amd64,linux/arm64":
return copy.CopySpecificImages, platforms, nil
The function correctly returns CopySpecificImages with platform filters. However, when building the manifest list for upload, skopeo appears to preserve the original manifest list structure instead of creating a filtered one that excludes unspecified platforms.

skopeo version: 1.23.0

aguidirh:
## Summary

This PR introduces two new flags for the `skopeo copy` command to provide more granular control over multi-architecture image copying:

- `--remove-list-signatures`: Removes only the manifest list signature while preserving per-instance signatures
- `--strip-sparse-manifest-list`: Strips missing instances from manifest lists when copying specific platforms

Relates to #2882 

## Motivation

When copying multi-architecture images with platform-specific selections, users currently face two challenges:

1. The `--remove-signatures` flag removes ALL signatures (both manifest list and per-instance), but sometimes users only want to remove the manifest list signature
2. Some registries don't support sparse manifest lists (manifest lists with missing instances), requiring users to strip the non-copied instances

### New Flags

**`--remove-list-signatures`**
- Removes only the manifest list signature while preserving per-instance signatures
- Provides finer control than `--remove-signatures`
- If both flags are specified, `--remove-signatures` takes precedence

**`--strip-sparse-manifest-list`**
- Strips missing instances from manifest lists when copying only a subset of platforms
- Useful for registries that don't support sparse manifest lists
- Requires explicit signature removal using either `--remove-signatures` or `--remove-list-signatures`
- The operation will fail if signatures are present and not explicitly removed

### Documentation

- Updated `docs/skopeo-copy.1.md` with flag descriptions and usage examples
- Added example showing how to copy specific platforms with sparse list stripping

### Tests

- Added `TestCopyPlatformListWithStripSparse` integration test
- Updated `TestSharedCopyOptionsCopyOptions` unit tests to cover new flags

## Test Plan

- [ ] Unit tests pass: `make test-unit-local`
- [ ] Integration tests pass: `make test-integration-local`
- [ ] Validation passes: `make validate-local`
- [ ] Documentation validation passes: `make validate-docs`

## Example Usage

Copy specific platforms and strip the sparse manifest list:
```bash
skopeo copy --multi-arch=linux/amd64,linux/arm64 \
  --strip-sparse-manifest-list \
  --remove-list-signatures \
  docker://quay.io/skopeo/stable:latest \
  docker://registry.example.com/skopeo:latest
```

Remove only manifest list signatures:
```bash
skopeo copy --multi-arch=linux/amd64,linux/arm64 \
  --remove-list-signatures \
  docker://quay.io/skopeo/stable:latest \
  docker://registry.example.com/skopeo:latest
```

### DIFF
--- a/cmd/skopeo/copy.go
+++ b/cmd/skopeo/copy.go
@@ -20,22 +20,23 @@ import (
 )
 
 type copyOptions struct {
-	global              *globalOptions
-	deprecatedTLSVerify *deprecatedTLSVerifyOption
-	srcImage            *imageOptions
-	destImage           *imageDestOptions
-	retryOpts           *retry.Options
-	copy                *sharedCopyOptions
-	additionalTags      []string                  // For docker-archive: destinations, in addition to the name:tag specified as destination, also add these
-	signIdentity        string                    // Identity of the signed image, must be a fully specified docker reference
-	digestFile          string                    // Write digest to this file
-	quiet               bool                      // Suppress output information when copying images
-	all                 bool                      // Copy all of the images if the source is a list
-	multiArch           commonFlag.OptionalString // How to handle multi architecture images
-	encryptLayer        []int                     // The list of layers to encrypt
-	encryptionKeys      []string                  // Keys needed to encrypt the image
-	decryptionKeys      []string                  // Keys needed to decrypt the image
-	imageParallelCopies uint                      // Maximum number of parallel requests when copying images
+	global                *globalOptions
+	deprecatedTLSVerify   *deprecatedTLSVerifyOption
+	srcImage              *imageOptions
+	destImage             *imageDestOptions
+	retryOpts             *retry.Options
+	copy                  *sharedCopyOptions
+	additionalTags        []string                  // For docker-archive: destinations, in addition to the name:tag specified as destination, also add these
+	signIdentity          string                    // Identity of the signed image, must be a fully specified docker reference
+	digestFile            string                    // Write digest to this file
+	quiet                 bool                      // Suppress output information when copying images
+	all                   bool                      // Copy all of the images if the source is a list
+	multiArch             commonFlag.OptionalString // How to handle multi architecture images
+	encryptLayer          []int                     // The list of layers to encrypt
+	encryptionKeys        []string                  // Keys needed to encrypt the image
+	decryptionKeys        []string                  // Keys needed to decrypt the image
+	imageParallelCopies   uint                      // Maximum number of parallel requests when copying images
+	stripRemovedPlatforms bool                      // Strip platforms not selected for copy when using --multi-arch with a platform list
 }
 
 func copyCmd(global *globalOptions) *cobra.Command {
@@ -85,6 +86,7 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
 	flags.IntSliceVar(&opts.encryptLayer, "encrypt-layer", []int{}, "*Experimental* the 0-indexed layer indices, with support for negative indexing (e.g. 0 is the first layer, -1 is the last layer)")
 	flags.StringSliceVar(&opts.decryptionKeys, "decryption-key", []string{}, "*Experimental* key needed to decrypt the image")
 	flags.UintVar(&opts.imageParallelCopies, "image-parallel-copies", 0, "Maximum number of image layers to be copied (pulled/pushed) simultaneously. Not setting this field will fall back to containers/image defaults.")
+	flags.BoolVar(&opts.stripRemovedPlatforms, "strip-removed-platforms", false, "Strip removed platforms when copying specific platforms (requires signature removal)")
 	return cmd
 }
 
@@ -257,6 +259,11 @@ func (opts *copyOptions) run(args []string, stdout io.Writer) (retErr error) {
 	copyOpts.OciEncryptConfig = encConfig
 	copyOpts.MaxParallelDownloads = opts.imageParallelCopies
 	copyOpts.ForceCompressionFormat = opts.destImage.forceCompressionFormat
+	if opts.stripRemovedPlatforms {
+		copyOpts.SparseManifestListAction = copy.StripSparseManifestList
+	} else {
+		copyOpts.SparseManifestListAction = copy.KeepSparseManifestList
+	}
 
 	return retry.IfNecessary(ctx, func() error {
 		manifestBytes, err := copy.Image(ctx, policyContext, destRef, srcRef, copyOpts)
--- a/cmd/skopeo/utils.go
+++ b/cmd/skopeo/utils.go
@@ -332,6 +332,7 @@ func (opts *imageDestOptions) warnAboutIneffectiveOptions(destTransport types.Im
 // sharedCopyOptions collects CLI flags that affect copying images, currently shared between the copy and sync commands.
 type sharedCopyOptions struct {
 	removeSignatures         bool                      // Do not copy signatures from the source image
+	removeListSignatures     bool                      // Do not copy manifest list signatures (preserves per-instance signatures)
 	signByFingerprint        string                    // Sign the image using a GPG key with the specified fingerprint
 	signBySequoiaFingerprint string                    // Sign the image using a Sequoia-PGP key with the specified fingerprint
 	signBySigstoreParamFile  string                    // Sign the image using a sigstore signature per configuration in a param file
@@ -346,6 +347,7 @@ func sharedCopyFlags() (pflag.FlagSet, *sharedCopyOptions) {
 	opts := sharedCopyOptions{}
 	fs := pflag.FlagSet{}
 	fs.BoolVar(&opts.removeSignatures, "remove-signatures", false, "Do not copy signatures from source")
+	fs.BoolVar(&opts.removeListSignatures, "remove-list-signatures", false, "Do not copy manifest list signatures (preserves per-instance signatures)")
 	fs.StringVar(&opts.signByFingerprint, "sign-by", "", "Sign the image using a GPG key with the specified `FINGERPRINT`")
 	fs.StringVar(&opts.signBySequoiaFingerprint, "sign-by-sq-fingerprint", "", "Sign the image using a Sequoia-PGP key with the specified `FINGERPRINT`")
 	fs.StringVar(&opts.signBySigstoreParamFile, "sign-by-sigstore", "", "Sign the image using a sigstore parameter file at `PATH`")
@@ -366,6 +368,10 @@ func (opts *sharedCopyOptions) copyOptions(stdout io.Writer) (*copy.Options, fun
 			return nil, nil, err
 		}
 		manifestType = mt
+	}
+
+	if opts.removeSignatures && opts.removeListSignatures {
+		return nil, nil, fmt.Errorf("Only one of --remove-signatures and --remove-list-signatures can be specified")
 	}
 
 	// c/image/copy.Image does allow creating both simple signing and sigstore signatures simultaneously,
@@ -453,6 +459,7 @@ func (opts *sharedCopyOptions) copyOptions(stdout io.Writer) (*copy.Options, fun
 	succeeded = true
 	return &copy.Options{
 		RemoveSignatures:                 opts.removeSignatures,
+		RemoveListSignatures:             opts.removeListSignatures,
 		Signers:                          signers,
 		SignBy:                           opts.signByFingerprint,
 		SignPassphrase:                   passphrase,
--- a/docs/skopeo-copy.1.md
+++ b/docs/skopeo-copy.1.md
@@ -97,6 +97,16 @@ Suppress output information when copying images.
 
 Do not copy signatures, if any, from _source-image_. Necessary when copying a signed image to a destination which does not support signatures.
 
+**--remove-list-signatures**
+
+Do not copy the manifest list signature while preserving per-instance signatures. This provides more granular control than **--remove-signatures**, which removes all signatures.
+
+**--strip-removed-platforms**
+
+When copying only a subset of platforms from a multi-architecture image (using **--multi-arch** with a platform list), write a manifest list containing only the copied platforms, instead of leaving entries that refer to per-platform images not present in the destination. This is useful for registries that reject manifest lists containing references to missing images.
+
+Note: This option requires signatures to be removed using either **--remove-signatures** or **--remove-list-signatures**, as stripping instances invalidates the manifest list signature and changes the manifest list digest. The operation will fail if signatures are present and not explicitly removed.
+
 **--sign-by** _key-id_
 
 Add a “simple signing” signature using that key ID for an image name corresponding to _destination-image_
@@ -271,6 +281,11 @@ $ skopeo copy --sign-by dev@example.com containers-storage:example/busybox:strea
 To copy only specific platforms from a multi-architecture image (creates a sparse manifest list):
 ```console
 $ skopeo copy --multi-arch=linux/amd64,linux/arm64 docker://quay.io/skopeo/stable:latest docker://registry.example.com/skopeo:latest
+```
+
+To copy specific platforms and strip the sparse manifest list (for registries that don't support sparse lists):
+```console
+$ skopeo copy --multi-arch=linux/amd64,linux/arm64 --strip-removed-platforms --remove-list-signatures docker://quay.io/skopeo/stable:latest docker://registry.example.com/skopeo:latest
 ```
 
 To encrypt an image:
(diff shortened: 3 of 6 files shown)


aguidirh:
Hi @mtrmac,

I added the `--remove-list-signatures` to the sync command because I see a scenario where people would like to sync the repos where the signature of the manifest list is not present in source. 

I did not add the `--strip-sparse-manifest-list` to the sync because I saw the sync command always pull and push everything, so I did not see a reason to add this flag.

Please let me know what you think about it.

mtrmac:
> I added the `--remove-list-signatures` to the sync command because I see a scenario where people would like to sync the repos where the signature of the manifest list is not present in source.

\*shrug\* I don’t see a _strong reason_ to think users want that feature individually, but it’s easier / more consistent to implement that way, so, why not.

> I did not add the `--strip-sparse-manifest-list` to the sync because I saw the sync command always pull and push everything, so I did not see a reason to add this flag.

Yeah, we’d have to move all of the `--multi-arch` option to be shared between `copy` and `sync`. That would also be fine, but more work, and really if users are motivated to use `oc-mirror` instead of `skopeo sync`, I think that’s the better outcome for everyone.

(ACK to the general structure from a very brief skim, I’ll try to do a more careful review soonish.)

integration/copy_test.go
```diff
@@ -204,6 +204,26 @@ func (s *copySuite) TestCopyWithPlatformList() {
...
+	require.NoError(t, err)
+	mimeType := manifest.GuessMIMEType(readManifest)
+	assert.Equal(t, "application/vnd.docker.distribution.manifest.list.v2+json", mimeType)
+
+	// Verify that we still have 2 platform manifests (manifest list + 2 platforms = 3 manifest files)
+	manifestFiles, err := filepath.Glob(filepath.Join(dir1, "*manifest.json"))
+	require.NoError(t, err)
+	assert.Equal(t, 3, len(manifestFiles), "Expected manifest list + 2 platform manifests after stripping")
+}
+
```
TomSweeneyRedHat, line 250: It would be good to have a test that calls both --remove-list-signatures and --remove-signatures to make sure all get wiped.

TomSweeneyRedHat: It might also be good to have a test to make sure --strip-sparse-manifest-list fails without a --remove*signatures option

TomSweeneyRedHat:
@aguidirh looks like you need to update the branch.

TomSweeneyRedHat:
Changes LGTM overall.

docs/skopeo-copy.1.md
```diff
@@ -97,6 +97,16 @@ Suppress output information when copying images.
 
 Do not copy signatures, if any, from _source-image_. Necessary when copying a signed image to a destination which does not support signatures.
 
+**--remove-list-signatures**
+
+Remove only the manifest list signature while preserving per-instance signatures. This provides more granular control than **--remove-signatures**, which removes all signatures. If **--remove-signatures** is also specified, it takes precedence.
+
+**--strip-sparse-manifest-list**
+
+When copying only a subset of platforms from a multi-architecture image (using **--multi-arch** with a platform list or index-only), strip the non-copied instances from the destination manifest list instead of creating a sparse manifest list. This is useful for registries that don't support sparse manifest lists.
```
mtrmac: Do we need to explain “sparse manifest list” here? Or perhaps this could descriptively talk about a manifest list referring to non-existent per-platform images, without naming that.

docs/skopeo-copy.1.md
```diff
@@ -97,6 +97,16 @@ Suppress output information when copying images.
...
+
+Note: This option requires signatures to be removed using either **--remove-signatures** or **--remove-list-signatures**, as stripping instances invalidates the manifest list signature. The operation will fail if signatures are present and not explicitly removed.
```
mtrmac: … and it changes the manifest list digest.

cmd/skopeo/copy.go
```diff
@@ -85,6 +86,7 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
 	flags.IntSliceVar(&opts.encryptLayer, "encrypt-layer", []int{}, "*Experimental* the 0-indexed layer indices, with support for negative indexing (e.g. 0 is the first layer, -1 is the last layer)")
 	flags.StringSliceVar(&opts.decryptionKeys, "decryption-key", []string{}, "*Experimental* key needed to decrypt the image")
 	flags.UintVar(&opts.imageParallelCopies, "image-parallel-copies", 0, "Maximum number of image layers to be copied (pulled/pushed) simultaneously. Not setting this field will fall back to containers/image defaults.")
+	flags.BoolVar(&opts.stripSparseManifestList, "strip-sparse-manifest-list", false, "Strip missing instances from manifest lists when copying specific platforms (requires signature removal)")
```
mtrmac: Hum. `strip-{removed, missing}-platforms`? We don’t _hide_ the concept of manifest lists but users might not really care at this level of detail — and the naming difference between schema2 “manifest list” and OCI “image index” makes it a bit awkward.

cmd/skopeo/utils.go
```diff
@@ -453,6 +455,7 @@ func (opts *sharedCopyOptions) copyOptions(stdout io.Writer) (*copy.Options, fun
 	succeeded = true
 	return &copy.Options{
 		RemoveSignatures:                 opts.removeSignatures,
+		RemoveListSignatures:             opts.removeListSignatures,
```
mtrmac, line 462: Non-blocking: This function could reject calls with both `--remove-signatures --remove-list-signatures` set.

cmd/skopeo/utils.go
```diff
@@ -346,6 +347,7 @@ func sharedCopyFlags() (pflag.FlagSet, *sharedCopyOptions) {
 	opts := sharedCopyOptions{}
 	fs := pflag.FlagSet{}
 	fs.BoolVar(&opts.removeSignatures, "remove-signatures", false, "Do not copy signatures from source")
+	fs.BoolVar(&opts.removeListSignatures, "remove-list-signatures", false, "Remove manifest list signatures only (preserves per-instance signatures)")
```
mtrmac: `s/Remove/Do not copy/` to make it clearer the source is not affected. Also in the man page.

docs/skopeo-copy.1.md
```diff
@@ -97,6 +97,16 @@ Suppress output information when copying images.
 
 Do not copy signatures, if any, from _source-image_. Necessary when copying a signed image to a destination which does not support signatures.
 
+**--remove-list-signatures**
+
+Remove only the manifest list signature while preserving per-instance signatures. This provides more granular control than **--remove-signatures**, which removes all signatures. If **--remove-signatures** is also specified, it takes precedence.
```
mtrmac: I think `--remove-signatures --remove-list-signatures` indicates a confused user, so I’d mildly prefer not promising to accept such inputs. (Also in the `.sync.1.` copy.)

integration/copy_test.go
```diff
@@ -204,6 +204,26 @@ func (s *copySuite) TestCopyWithPlatformList() {
...
+	// Copy specific platforms and strip sparse manifest list
+	assertSkopeoSucceeds(t, "", "copy", "--retry-times", "3", "--multi-arch=linux/amd64,linux/arm64",
+		"--strip-sparse-manifest-list", "--remove-list-signatures", knownListImage, "dir:"+dir1)
+
+	manifestPath := filepath.Join(dir1, "manifest.json")
+	readManifest, err := os.ReadFile(manifestPath)
+	require.NoError(t, err)
+	mimeType := manifest.GuessMIMEType(readManifest)
+	assert.Equal(t, "application/vnd.docker.distribution.manifest.list.v2+json", mimeType)
```
mtrmac, line 231: This does not actually test the stripping behavior. Ensure that one of the other platforms is not mentioned, perhaps.

docs/skopeo-copy.1.md
mtrmac: With `index-only`, the copy would actually fail with `requested operation filtered out all platforms and would create an empty image`, so that does not need mentioning here.

mtrmac (commented):
Thanks!

As usual, naming things tends to be the hardest part :)

integration/copy_test.go
aguidirh: > It would be good to have a test that calls both --remove-list-signatures and --remove-signatures to make sure all get wiped.

Do you mean testing each flag isolated? Or testing both of them together? 

According to @mtrmac [comment](https://github.com/podman-container-tools/skopeo/pull/2892#discussion_r3395309688) both flags shouldn't be used together.

docs/skopeo-copy.1.md
aguidirh: Implemented on [77a8fef5](https://github.com/aguidirh/skopeo/commit/77a8fef5).

docs/skopeo-copy.1.md
aguidirh: Implemented on [da893d51](https://github.com/aguidirh/skopeo/commit/da893d51).

cmd/skopeo/copy.go
aguidirh: Implemented on [4937c9a0](https://github.com/aguidirh/skopeo/commit/4937c9a0).

cmd/skopeo/utils.go
aguidirh: Implemented on [d5d3ab84](https://github.com/aguidirh/skopeo/commit/d5d3ab84).

cmd/skopeo/utils.go
aguidirh: Implemented on [df578335](https://github.com/aguidirh/skopeo/commit/df578335).

docs/skopeo-copy.1.md
aguidirh: Implemented on [22beab50](https://github.com/aguidirh/skopeo/commit/22beab50).

integration/copy_test.go
aguidirh: Implemented on [3c52c385](https://github.com/aguidirh/skopeo/commit/3c52c385).

docs/skopeo-copy.1.md
aguidirh: Implemented on [fbcc8239](https://github.com/aguidirh/skopeo/commit/fbcc8239).

aguidirh:
@mtrmac are these tests flakies? 

Because `make test-integration-local` is passing on my localhost.

integration/copy_test.go
aguidirh: > It would be good to have a test that calls both --remove-list-signatures and --remove-signatures to make sure all get wiped.

Skipped — went with mtrmac's suggestion to reject both flags together instead. Implemented on [d5d3ab84](https://github.com/aguidirh/skopeo/commit/d5d3ab84).

aguidirh: > It might also be good to have a test to make sure --strip-sparse-manifest-list fails without a --remove*signatures option

Implemented on [4646b190](https://github.com/aguidirh/skopeo/commit/4646b190).

mtrmac:
> @mtrmac are these tests flakies?

The same, newly added, test is failing in all three runs, I think this very likely to be legitimate.

I guess, but did not verify, that the image is not signed, so that error path does not engage. I’m not very sure what `TestCopyPlatformListStripRemovedPlatformsFailsWithoutSignatureRemoval` is supposed to verify — but maybe using `--preserve-digests` instead of relying on a signature would trigger the same “we can’t modify the manifest list” error path.

> Because `make test-integration-local` is passing on my localhost.

I think most tests there are skipped without `SKOPEO_CONTAINER_TESTS` (and enabling that outside of a non-confined VM might be a bad idea).

cmd/skopeo/utils.go
```diff
@@ -368,6 +370,10 @@ func (opts *sharedCopyOptions) copyOptions(stdout io.Writer) (*copy.Options, fun
 		manifestType = mt
 	}
 
+	if opts.removeSignatures && opts.removeListSignatures {
+		return nil, nil, fmt.Errorf("Only one of --remove-signatures and --remove-list-signatures can be specified")
```
mtrmac, line 374: Non-blocking: covering this in unit tests wouldn’t hurt.

mtrmac (commented):
Thanks! LGTM apart from the failing `TestCopyPlatformListStripRemovedPlatformsFailsWithoutSignatureRemoval`

lsm5:
@aguidirh needs rebase

aguidirh:
> Thanks! LGTM apart from the failing `TestCopyPlatformListStripRemovedPlatformsFailsWithoutSignatureRemoval`

Sorry for the delay, I will have a look on the tests failing at some point next week.

cmd/skopeo/utils.go
aguidirh: Fixed at [a868c4bb](https://github.com/aguidirh/skopeo/commit/a868c4bb1c075a214304e924cb07f92a7e1354a0).

aguidirh:
@mtrmac I just fixed the tests that were failing and I also submitted one commit adding the mutually exclusive test tags `--remove-signatures` and `--remove-list-signatures`.

Please let me know if there is something else to be done on this PR or if I should squash the commits.

integration/copy_test.go
```diff
@@ -204,6 +204,58 @@ func (s *copySuite) TestCopyWithPlatformList() {
...
+	assert.Contains(t, destPlatforms, "linux/arm64")
+	assert.NotContains(t, destPlatforms, strippedPlatform)
+
+	// Verify that we still have 2 platform manifests (manifest list + 2 platforms = 3 manifest files)
+	manifestFiles, err := filepath.Glob(filepath.Join(dir1, "*manifest.json"))
+	require.NoError(t, err)
+	assert.Equal(t, 3, len(manifestFiles), "Expected manifest list + 2 platform manifests after stripping")
+}
+
+func (s *copySuite) TestCopyPlatformListStripRemovedPlatformsFailsWithoutSignatureRemoval() {
```
mtrmac: I think using `--preserve-digests` in this test is perfectly fine, except that the test name now does not match.

mtrmac (commented):
Thanks!

One last nit; please squash the commits.

integration/copy_test.go
aguidirh: Commits squashed and suggestion above Fixed at [51566c51](https://github.com/aguidirh/skopeo/commit/51566c51b2e9b3dc6d8a529c02e7768747ac4fff).

aguidirh: [R251](https://github.com/podman-container-tools/skopeo/pull/2892/changes/51566c51b2e9b3dc6d8a529c02e7768747ac4fff#diff-cfa83c143f87f771f02bd65068a38a7a7c15b404cee618f0aa74673ff7f2052bR251).

aguidirh:
@mtrmac I believe the test failing is a flaky test, because I won't renamed the Test for something else and squashed the commits.

mtrmac (approved):

mtrmac:
Thanks!

PR #2902, opened 2026-06-25
docs: include examples for --command-timeout duration format

Rolv-Apneseth:
I recently came across this flag while looking into a different issue and just thought it could use some clarification on what format the duration should be given in. These small examples should make it immediately clear to users.


mtrmac (approved):
Thank you!

Rolv-Apneseth:
No problem!

PR #2908, opened 2026-07-07
Use `go.yaml.in/yaml/v3` instead of `gopkg.in/yaml.v3`

mtrmac:
`gopkg.in/yaml.v3` is unmaintained; `go.yaml.in/yaml/v3` is the current maintained version.

For now, we include both copies either way, but `container-libs` have also migrated, so the next time we update, `gopkg.in/yaml.v3` will only be used by test-only `stretchr/testify`.


Luap99 (approved):
LGTM

PR #2925, opened 2026-07-20
fix issue_pr_lock.yml workflow and do not run cronjobs at midnight

Luap99:
see commits


mtrmac (approved):
Thank you!

(I’ve seen the `osx` failure a few times … if it persists, perhaps we need to delete a cache.)

Luap99:
deleted cache `setup-go-macOS-arm64-go-1.26.5-2d7c365e521fd367efb7736c80cdc931fa53dbbd8d2439ee7804acb7fc0004be` and rerun, lets see I guess

Luap99:
manual run via workflow dispatch just causes the run to be skipped
https://github.com/podman-container-tools/skopeo/actions/runs/29767709714/job/88437858072

No clue what is going on... Lets hope the cron trigger works this night at least

mtrmac:
@Luap99 couldn’t it be the condition
https://github.com/podman-container-tools/podman/blob/1c3374e38820e4ee1a8c654cc735c82d7f50a282/.github/workflows/issue_pr_lock.yml#L48
from https://github.com/podman-container-tools/podman/pull/28778 ?

Luap99:
ah yes that might be it then, I fix it tomorrow

PR #2820, opened 2026-03-10
Update CI images with a dedicated registry that supports both schema1 and schema2

mtrmac:
This validates, and updates for, https://github.com/containers/automation_images/pull/437 .

After this merges, I expect the bot to close https://github.com/containers/skopeo/pull/2772 .

PR #2931, opened 2026-07-23
[release-1.22] Backport two test fixes

Luap99:
Backported from #2820

I need this for https://github.com/podman-container-tools/container-libs/pull/1010

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -26,7 +26,7 @@ env:
     # package is not available in earlier releases; once we update to a future
     # Fedora release (or if the package is backported), switch back from Rawhide
     # to the latest Fedora release.
-    IMAGE_SUFFIX: "c20250910t092246z-f42f41d13"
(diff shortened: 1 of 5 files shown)


mtrmac (approved):
ACK, please merge whenever appropriate. (For the record, the `.cirrus.yml` CI is no longer active/used, and this branch does not have GitHub actions CI, like `main` uses, set up right now.)

Luap99:
thanks, I merge then, container-libs seems happy with this.

For now I do not have plans to enable CI here, but I guess it might make sense to allow this branch being used for new releases in case of CVE fixes? Or we just cherry-pick/vendor without CI and hope they work?

mtrmac:
I agree CI would be nice if we are going to support that branch for a while; I don’t have much of an opinion on priority. I expect _most_ back ported fixes would be in vendored libraries, so at most a build test might be sufficient (OTOH backporting the full thing without divergence would be more consistent).

PR #2932, opened 2026-07-26
ci: drop redundant make vendor from test_skopeo

### LINKED ISSUE #2923: ci: redundant make vendor in test_skopeo job
## Problem

The `test_skopeo` GHA job runs `make vendor` before building and testing. That job depends on `validate`, which already runs `make vendor` + `hack/tree_status.sh`.

## Review context

Sandbox GHA migration review (mtrmac, [PR #1 discussion](https://github.com/podman-container-tools/skopeo-sandbox/pull/1#discussion_r3182590794)), marked "not now":

> `make vendor` in the `test_skopeo` task is redundant (and `go mod verify` is fairly costly): that task depends on `validate` which already checked this

## Related

- Epic #2919
- `test_skopeo` job in `.github/workflows/ci.yml`

vtushar06:
I was reading through the CI cleanup issues and this one looked small enough to just do.

test_skopeo has validate in its needs, and validate already runs make vendor and then hack/tree_status.sh, so the vendor tree has been checked before this job even starts. vendor/ is committed too, so the checkout here already has what the build needs - I could not see anything in the later steps that wanted it regenerated.

The part that seems worth having back is go mod verify inside the target, since that was being paid three times over, once for each buildtags leg.

I have not run skopeo CI myself so I am going on reading the workflow rather than timing it. CI on this PR should show whether the three test legs are happy without it.

Closes: #2923

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -155,7 +155,6 @@ jobs:
           persist-credentials: false
       - name: fix git safe.directory
         run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
-      - run: make vendor
       - name: build + install
         run: |
           make "BUILDTAGS=${{ matrix.buildtags }}" bin/skopeo


vtushar06:
hey @Luap99 @timcoding1988 , I think we can have a review on this since all test are passed only needs approval from your end.

TomSweeneyRedHat:
LGTM
@timcoding1988 PTAL

timcoding1988:
This LGTM. Also @mtrmac @Luap99

Luap99:
Please drop the second merge commit form the PR, you should rebase PRs against main to avoid them. Merging extra merge commits from PRs just clutters the git log for no benefit.

vtushar06:
rebased is done, merge commit is gone now @Luap99

Luap99 (approved):
LGTM

PR #2935, opened 2026-07-27
integration: use our own pause image

Luap99:
The registry.k8s.io/pause image contains many tags so sync copies a lot
of unneeded images. Also for some reason this test is hard broken on
container-libs failing to fetch a non existing tag.

So swap this to our own namespace quay.io/libpod/k8s-pause which only
contains two tag with the correct test requirements and we do not have
to worry that it breaks us all of the sudden.


mtrmac (approved):
Feel free to merge if tests pass.

Ideally the three instances of `copy_test.go` would be updated as well.

Luap99:
updated the copy tests as well

mtrmac (approved):

PR #2947, opened 2026-08-06
docs: add override-os and override-arch example to inspect --help

### LINKED ISSUE #2793: Add an example for skopeo inspect --override-os
tl;dr - Is there a supported way to tell skopeo inspect to use a specific architecture and os? I don't see anything in `skopeo inspect --help` that points towards it.

---

I am not calling skopeo manually, but through terraform via incus provider. It runs `skopeo --insecure-policy inspect docker://docker.io/linuxserver/jellyfin` and errors out. Indeed running `skopeo --insecure-policy inspect docker://docker.io/linuxserver/jellyfin` manually gives the same error:

```
FATA[0000] Error parsing manifest for image: choosing image instance: no image found in image index for architecture "arm64", variant "v8", OS "darwin"
```

Thing is, my default_remote is set to an x86_64 machine, I am not sure if
* the inspect command is not set to the remote machine (os: incus-os, architecture: x86_64), and if it can be set
or
* the inspect command should work on macOS at all, and if not, should it be hidden 
or
* the incus provider is not setting the architecture to the remote correctly

So I am asking here and in incus to figure out where the source of the problem lies and how to solve it.

(separately, is there any way to have `skopeo inspect` on macOS just work out of the box instead of having to learn that these things need to be overridden?)

Thanks

ArneshBanerjee:
Adds an example to skopeo inspect --help showing --override-os and --override-arch together. Closes #2793. The flags already existed and were in the man page, someone hit the exact case in the issue where their client architecture did not match the image and had no way to know from --help alone how to fix it. mtrmac proposed this exact fix in the issue thread. No behavior change, one line added to the existing Example string, checked that it builds and that the new line shows up correctly in skopeo inspect --help output.

### DIFF
--- a/cmd/skopeo/inspect.go
+++ b/cmd/skopeo/inspect.go
@@ -54,7 +54,8 @@ See skopeo(1) section "IMAGE NAMES" for the expected format
 		RunE: commandAction(opts.run),
 		Example: `skopeo inspect docker://registry.fedoraproject.org/fedora
 skopeo inspect --config docker://docker.io/alpine
-skopeo inspect --format "Name: {{.Name}} Digest: {{.Digest}}" docker://registry.access.redhat.com/ubi8`,
+skopeo inspect --format "Name: {{.Name}} Digest: {{.Digest}}" docker://registry.access.redhat.com/ubi8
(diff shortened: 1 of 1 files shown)


mtrmac (approved):
Thanks!

Issue #2946, opened 2026-08-05
`dorny/paths-filter` action probably should use a non-default `predicate-quantifier`


mtrmac:
Looking at https://github.com/dorny/paths-filter/compare/v4.0.2...v4.0.3 and the improved documentation about using `!` in filters, the default `predicate-quantifier: some` is probably not quite what we want.

E.g. https://github.com/podman-container-tools/skopeo/actions/runs/31013166828/job/92330316117?pr=2945 shows `.github/workflows/ci.yml` matching the `code` filter.

Cc: @timcoding1988 (not a high priority)

Luap99:
yes the logic does not work, that is what I managed to get to work in podman https://github.com/podman-container-tools/podman/blob/0c684fd2f6c5c832b1795ceb8cec70f79fb7bd35/.github/filters.yaml#L12-L13 but it only works for one exclude
I do agree using the new 4.0.3 some option sounds useful here.

PR #2948, opened 2026-08-07
ci: set predicate-quantifier on the path filter

vtushar06:
I had a look at #2946 since the numbers turned out worse than the one example log line suggests.

The `code` filter matches **every tracked file**, not just a stray `ci.yml`. Each entry in a filter list is matched on its own and the results are OR-ed, so `'!**/*_test.go'` on its own matches everything it does not exclude. Running the filter over all 3509 files tracked on main:

```
code, default 'some'          : 3509 / 3509 files match
code, 'some-with-excludes'    : 2795 / 3509 files match

README.md                    some=true   some-with-excludes=false
.github/workflows/ci.yml     some=true   some-with-excludes=false
cmd/skopeo/copy.go           some=true   some-with-excludes=true
cmd/skopeo/copy_test.go      some=true   some-with-excludes=false
systemtest/010-inspect.bats  some=true   some-with-excludes=false
```

So the `cross`, `osx` and `ostree_rs_ext` gates on `code` never skip anything, which is the opposite of what `filters.yaml` says in its own comment.

`some-with-excludes` leaves 2795 files, all `.go`, with no `_test.go`, `systemtest/` or `integration/` among them. I checked the other two filters do not move: `all` matches 3375 either way and `docs` 171 either way, so setting this at the action level is safe. `every` is not an option, it takes `all` and `docs` to 0.

Correction to what I first wrote here about the podman extglob: it does work on this tree. `!(systemtest|integration)/**/!(*_test).go` gives the same 2795 files. It only diverges on paths skopeo does not currently have - it misses a root-level `.go`, and depending on the spelling either over-excludes a directory that merely starts with `integration`/`systemtest` or fails to exclude a nested one. The quantifier does not have those edges, which is why I went with it, but I should not have said the extglob does not stretch.


.github/workflows/ci.yml
```diff
@@ -35,6 +35,10 @@ jobs:
         id: filter
         with:
           filters: .github/filters.yaml
+          # Each filter entry is matched on its own, so a lone '!**/*_test.go'
+          # matches every file it does not exclude and `code` ends up true for
+          # any change at all.
```
mtrmac: I’ll let others review / verify this in full, but just here: Comments should document what the code _does_, not what it _did_ nor what it _doesn’t do_.

(I’m not even sure this needs a comment.)

mtrmac (changes requested):

.github/workflows/ci.yml
vtushar06: @mtrmac dropped it, the reasoning is in the commit message. just the one line now.

Luap99 (approved):
LGTM

.github/workflows/ci.yml
vtushar06: @mtrmac fair, dropped it. the diff is just the one line now:

```
+          predicate-quantifier: some-with-excludes
```

mtrmac (approved):
LGTM. Thanks!

PR #2950, opened 2026-08-10
[release-1.22] integration: use our own pause image

QiWang19:
Backport commit: [e894dac](https://github.com/podman-container-tools/skopeo/commits/e894dac8cf454fd8651e29617bf81f997b230169) to fix PR https://github.com/podman-container-tools/container-libs/pull/1061 ci [failure](https://github.com/podman-container-tools/container-libs/actions/runs/31422419891/job/93566703334).

The registry.k8s.io/pause image contains many tags so sync copies a lot of unneeded images. Also for some reason this test is hard broken on container-libs failing to fetch a non existing tag.

So swap this to our own namespace quay.io/libpod/k8s-pause which only contains two tag with the correct test requirements and we do not have to worry that it breaks us all of the sudden.

### DIFF
--- a/integration/copy_test.go
+++ b/integration/copy_test.go
@@ -501,7 +501,7 @@ func (s *copySuite) TestCopySimple() {
 
 	// FIXME: It would be nice to use one of the local Docker registries instead of needing an Internet connection.
 	// "pull": docker: → dir:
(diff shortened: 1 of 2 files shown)


mtrmac (approved):
Thank you!

PR #2954, opened 2026-08-12
hack: fix validate-git-marks.sh never detecting conflict markers

### LINKED ISSUE #2953: `hack/validate-git-marks.sh` does not detect conflict markers
`make validate-local` runs this script on every PR to catch leftover merge conflict markers.
It never catches any.

Line 9 tests grep's output instead of its exit status:

```bash
if [ $(grep -r "^\(<<<<<<<\|>>>>>>>\|^=======$\)" $f) ]; then
```

Git writes markers with a label (`<<<<<<< HEAD`), so the unquoted substitution hands `[` two
or more arguments. `[` fails with "too many arguments" and exits non-zero, so the `if` is
false and the file is never flagged.

Reproduce:

```bash
T=$(mktemp -d); cd "$T"; git init -q .; git config user.email t@t; git config user.name t
printf 'a\n<<<<<<< HEAD\nmine\n=======\ntheirs\n>>>>>>> feature\n' > conflicted.go
git add -A; git commit -qm t
bash /path/to/skopeo/hack/validate-git-marks.sh; echo "rc=$?"
```

Got:

```
validate-git-marks.sh: line 9: [: too many arguments
Congratulations!  There is no conflict.
rc=0
```

Expected: names `conflicted.go`, exits 1.

Not macOS-specific — the failure is in bash's `[`, not grep.

ROKUMATE:
Fixes: #2953

The script tested grep's output instead of its exit status:

```bash
if [ $(grep -r "^\(<<<<<<<\|>>>>>>>\|^=======$\)" $f) ]; then
```

Git writes markers with a label (`<<<<<<< HEAD`), so `[` got two or more arguments, failed
with "too many arguments" and exited non-zero — the file was never flagged. On a repo with a
committed conflict the script printed "Congratulations!  There is no conflict." and exited 0.

Now tests the exit status with `grep -Eq` and quotes `"$f"`. Dropped `-r`, meaningless on a
regular file, plus the `continue` and the dead `set -e` — the script has no `set -e` at the
top and no `set +e`.

Semantics unchanged: `<<<<<<<` and `>>>>>>>` match as line prefixes, `=======` only as a whole
line. The single grep from `aebab492` is kept; only the test around it changes.

Verified: planted conflict now exits 1 and names the file (was rc=0); each marker type
detected on its own; `======= section header` still not flagged; this tree still passes clean;

### DIFF
--- a/hack/validate-git-marks.sh
+++ b/hack/validate-git-marks.sh
@@ -6,11 +6,9 @@ unset IFS
 
 badFiles=()
 for f in "${files[@]}"; do
-    if [ $(grep -r "^\(<<<<<<<\|>>>>>>>\|^=======$\)" $f) ]; then
+    if grep -Eq '^(<<<<<<<|>>>>>>>|=======$)' "$f"; then
         badFiles+=( "$f" )
-        continue
     fi
-    set -e
 done
 
 


mtrmac (changes requested):
Thanks!

Per https://github.com/podman-container-tools/community/blob/main/CONTRIBUTING.md#dco-sign-off:

> Use a real name (sorry, no anonymous contributions)

(and read on for more nuance)

PR #2958, opened 2026-08-13
RFC: Start using log/slog as the primary logging mechanism

mtrmac:
RFC @podman-container-tools/core-maintainers , for doing something like this project-wide.

Direct `logrus.*` calls to `log/slog.Default()`. logrus 1.10.0 includes that as a native feature, I think that’s the best hint to migrate to the standard logging mechanisms we are ever going to get.

I’d suggest adding the logging redirect infrastructure to `main` of the various projects, _with no plan to immediately migrate_ — lot of churn we just don’t need; but for new code, we would ask for `log/slog` to be used.

This PR does migrate the few uses in Skopeo, basically to demonstrate what that looks like (both the `Sprintf`s at the very top level, and the more useful usage within `sync.go`).

---

This changes formatting, from 
```
FATA[0001] Error ...
```
(with the `FATA` red, and the `[0001]` part, very unintuitively, a timestamp in seconds) when interactive, and from
```
time="2026-08-14T00:24:59+02:00" level=fatal msg="Error..."
```
when non-interactive, to
```
ERROR Error ...
```    
    
I think that is good enough, probably a bit of an improvement; we can tweak the formatting later if necessary.

---

The more important aspect is that `log/slog` does not provide `Printf`-like helpers (although a manual `Sprintf` is always an option), strongly directing callers towards a key-value format. Given that _most_ logs are for debugging or unexpected situations, that seems acceptable (especially if it allows sharing some fields,
```go
logger := slog.With("fields", "relevant", "to-the-whole", "function")
…
logger.Error("step A failed", "err", err)
…
logger.Error("step C failed", "err", err)
```
can be more consistent and less repetitive).

### DIFF
--- a/cmd/skopeo/main.go
+++ b/cmd/skopeo/main.go
@@ -166,7 +166,7 @@ func main() {
 
 	rootCmd, _ := createApp()
 	if err := rootCmd.Execute(); err != nil {
-		slog.Error(err.Error())
+		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
 		if isNotFoundImageError(err) {
 			os.Exit(2)
 		}
--- a/cmd/skopeo/main.go
+++ b/cmd/skopeo/main.go
@@ -3,11 +3,16 @@ package main
 import (
 	"context"
 	"fmt"
+	"io"
+	"log"
+	"log/slog"
+	"os"
 	"runtime/debug"
 	"strings"
 	"time"
 
 	"github.com/sirupsen/logrus"
+	lslog "github.com/sirupsen/logrus/hooks/slog"
 	"github.com/spf13/cobra"
 	commonFlag "go.podman.io/common/pkg/flag"
 	"go.podman.io/image/v5/pkg/cli/basetls/tlsdetails"
@@ -15,6 +20,7 @@ import (
 	"go.podman.io/image/v5/types"
 	"go.podman.io/skopeo/version"
 	"go.podman.io/storage/pkg/reexec"
+	"golang.org/x/term"
 )
 
 var defaultUserAgent = "skopeo/" + version.Version
@@ -136,6 +142,7 @@ func gitCommit() string {
 func (opts *globalOptions) before(cmd *cobra.Command, args []string) error {
 	if opts.debug {
 		logrus.SetLevel(logrus.DebugLevel)
+		slog.SetLogLoggerLevel(slog.LevelDebug)
 	}
 	if opts.tlsVerify.Present() {
 		logrus.Warn("'--tls-verify' is deprecated, please set this on the specific subcommand")
@@ -150,13 +157,20 @@ func main() {
 	if reexec.Init() {
 		return
 	}
+
+	logrus.SetOutput(io.Discard)
+	logrus.AddHook(lslog.NewHook(slog.Default(), nil))
+	if term.IsTerminal(int(os.Stderr.Fd())) {
+		log.SetFlags(0) // We don’t want date/time in interactive output
+	}
+
 	rootCmd, _ := createApp()
 	if err := rootCmd.Execute(); err != nil {
+		slog.Error(err.Error())
 		if isNotFoundImageError(err) {
-			logrus.StandardLogger().Log(logrus.FatalLevel, err)
-			logrus.Exit(2)
+			os.Exit(2)
 		}
-		logrus.Fatal(err)
+		os.Exit(1)
 	}
 }
 
--- a/cmd/skopeo/main.go
+++ b/cmd/skopeo/main.go
@@ -99,7 +99,7 @@ func createApp() (*cobra.Command, *globalOptions) {
 	rootCommand.PersistentFlags().DurationVar(&opts.commandTimeout, "command-timeout", 0, "timeout for the command execution (e.g. 30s, 10m)")
 	rootCommand.PersistentFlags().StringVar(&opts.registriesConfPath, "registries-conf", "", "path to the registries.conf file")
 	if err := rootCommand.PersistentFlags().MarkHidden("registries-conf"); err != nil {
-		logrus.Fatal("unable to mark registries-conf flag as hidden")
+		panic("unable to mark registries-conf flag as hidden")
 	}
 	rootCommand.PersistentFlags().StringVar(&opts.tmpDir, "tmpdir", "", "directory used to store temporary files")
 	rootCommand.PersistentFlags().StringVar(&opts.userAgentPrefix, "user-agent-prefix", "", "prefix to add to the user agent string")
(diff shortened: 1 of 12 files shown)


cmd/skopeo/sync.go
```diff
@@ -705,14 +682,11 @@ func (opts *syncOptions) run(args []string, stdout io.Writer) (retErr error) {
...
 			}
 
-			fromToFields := logrus.Fields{
-				"from": transports.ImageName(ref),
-				"to":   transports.ImageName(destRef),
-			}
+			fromToFields := slog.With("from", transports.ImageName(ref), "to", transports.ImageName(destRef))
 			if opts.dryRun {
-				logrus.WithFields(fromToFields).Infof("Would have copied image ref %d/%d", counter+1, len(srcRepo.ImageRefs))
+				fromToFields.Info(fmt.Sprintf("Would have copied image ref %d/%d", counter+1, len(srcRepo.ImageRefs)))
```
Luap99, line 687: one thing I wonder here. This is not really the same performance wise I belive unless there is some special compiler magic for slog?

This here calls fmt.Sprintf() first then passing the result to the logger, meaning we do the full string parsing/allocation each time regardless if the line will be logged or not.
In the old logrus logic if first check against teh log level before building the final logged string.

Now I doubt that this matters much but on some hot loop it could be? So I think if we go structured logs and convert callers here should we need go all the way and even apply this here?

mtrmac: You’re right, this is inefficient.

The typical `slog` usage would be `slog.Info("…", "count", count, "total", total)`, [or for maximum performance `slog.LogAttrs(context, "...", slog.Int("count", count), slog.Int("total, total)`], and see https://pkg.go.dev/log/slog#hdr-Performance_considerations for even more options.

In _this_ line, this is just because this output is the primary UI of Skopeo, and using the key-value pairs would be significantly less readable.

For most of our libraries, debug logs don’t matter at all, and error/warning logs should ideally not be shown during ordinary operation (so neither the text nor the performance should matter) — so I think we’d be using the `slog` key-value format ~everywhere.

Skopeo is an atypical user of logging, and thus a bad example.

cmd/skopeo/main.go
```diff
@@ -150,13 +157,20 @@ func main() {
...
 	rootCmd, _ := createApp()
 	if err := rootCmd.Execute(); err != nil {
+		slog.Error(err.Error())
 		if isNotFoundImageError(err) {
-			logrus.StandardLogger().Log(logrus.FatalLevel, err)
-			logrus.Exit(2)
+			os.Exit(2)
 		}
-		logrus.Fatal(err)
+		os.Exit(1)
```
Luap99: I find this hard to read for interactive use, there are no longer colors at the start (which I don't really care about to keep) but it means the last line is hard to differentiate between the regular logs


Personally I would favor to adopt he same podman `Error:` prefix for some more consistency. Although I guess that does not really improve readability that much the argumentis that we know the difference between an ERROR was logged somewhere in the code and an error was returned all the way back to main() and we just printed it there then.

If there is some os.Exit() elsewhere or a program crash we could not really be sure otherwise without more information on the program exit.

```
$ bin/skopeo --debug inspect docker://quay.io/libpod/testimage
DEBUG Reading registries signature storage configuration from "/etc/containers/registries.d/default.yaml"
DEBUG Reading registries signature storage configuration from "/etc/containers/registries.d/registry.access.redhat.com.yaml"
DEBUG Reading registries signature storage configuration from "/etc/containers/registries.d/registry.redhat.io.yaml"
DEBUG Loading registries configuration "/etc/containers/registries.conf"
DEBUG Loading registries configuration "/etc/containers/registries.conf.d/000-shortnames.conf"
DEBUG Trying to access "quay.io/libpod/testimage:latest"
DEBUG No credentials matching quay.io/libpod/testimage found in /run/user/1000/containers/auth.json
DEBUG No credentials matching quay.io/libpod/testimage found in /home/pholzing/.config/containers/auth.json
DEBUG No credentials matching quay.io/libpod/testimage found in /home/pholzing/.docker/config.json
DEBUG No credentials matching quay.io/libpod/testimage found in /home/pholzing/.dockercfg
DEBUG No credentials for quay.io/libpod/testimage found
DEBUG  No signature storage configuration found for quay.io/libpod/testimage:latest, using built-in default file:///home/pholzing/.local/share/containers/sigstore
DEBUG GET https://quay.io/v2/
DEBUG Ping https://quay.io/v2/ status 401
DEBUG GET https://quay.io/v2/auth?scope=repository%3Alibpod%2Ftestimage%3Apull&service=quay.io
DEBUG Increasing token expiration to: 60 seconds
DEBUG GET https://quay.io/v2/libpod/testimage/manifests/latest
DEBUG Content-Type from manifest GET is "application/json"
DEBUG Accessing "quay.io/libpod/testimage:latest" failed: reading manifest latest in quay.io/libpod/testimage: manifest unknown
ERROR Error parsing image name "docker://quay.io/libpod/testimage": reading manifest latest in quay.io/libpod/testimage: manifest unknown
```

Luap99 (commented):
approach LGTM, I have however not evaluated every log line individually .
The only thing worth considering is the main program error exit message.

cmd/skopeo/main.go
mtrmac: *shrug* updated to use `Error: ` and the existing errors.

I think this really only matters with `--debug`; otherwise there should be just one error, exactly in the place a user is already looking at.

Skopeo’s error reporting was originally designed to “just” output the error (and worked fine enough with logrus, where the `ERRO[000t]` kind of fades in the background, perhaps because of the `[]`s): the top-level command handlers are expected to provide a “human readable text”.

As a result, prefixing that with `Error` frequently stutters, as in here: `Error: Error parsing manifest …`. So, the natural follow-up would be to re-format the error messages in command handlers to be more similar to the rest of the code base, starting with a lower-case letter and just adding context: `fmt.Errorf("parsing image name …: %w")`, so that we end up with not-quite-but-close-to-perfect `Error: parsing image name …`.

I really don’t feel strongly about that — less deviation from the typical Go codebase, and not having to disable standard liters, is a fair enough benefit — but I’d like to delegate doing/reviewing that work :)

Luap99: > I think this really only matters with --debug; otherwise there should be just one error, exactly in the place a user is already looking at.

yes certainly if we only print one line this is not that useful

> As a result, prefixing that with Error frequently stutters, as in here: Error: Error parsing manifest …. So, the natural follow-up would be to re-format the error messages in command handlers to be more similar to the rest of the code base, starting with a lower-case letter and just adding context: fmt.Errorf("parsing image name …: %w"), so that we end up with not-quite-but-close-to-perfect Error: parsing image name ….

100%, I think there even is a linter that checks that all errors start lowercase

Luap99 (approved):
LGTM

cmd/skopeo/main.go
mtrmac: > but I’d like to delegate doing/reviewing that work :)

let a LLM do most of that: #2977 .

Issue #2919, opened 2026-07-15
Epic: GHA migration follow-up from skopeo-sandbox review


timcoding1988:
## Summary

Tracking epic for unresolved CI review feedback from the `podman-container-tools/skopeo-sandbox` Cirrus → GHA migration.

The sandbox work landed upstream in PR #2883. These items are follow-up from sandbox PR https://github.com/podman-container-tools/skopeo-sandbox/pull/1 / PR https://github.com/podman-container-tools/skopeo-sandbox/pull/9 review threads and the upstream audit before deprecating the sandbox repo.

## Sub-issues

- [ ] #2912 — migrate `renovate.json5` to `podman-container-tools/automation`
- [ ] #2913 — add zizmor workflow for GitHub Actions security scanning
- [ ] #2914 — remove leftover Cirrus config after GHA migration
- [ ] #2915 — remove `git safe.directory` workaround from container jobs
- [ ] #2916 — golangci-lint installed via `make tools` instead of `golangci-lint-action`
- [ ] #2917 — `ostree_rs_ext` job uses archived `ostree-rs-ext` repository
- [ ] #2918 — `cancel-in-progress` cancels overlapping CI runs on `main` pushes
- [ ] #2922 — `doccheck` does build + install before `validate-docs`
- [ ] #2923 — redundant `make vendor` in `test_skopeo` job
- [ ] #2924 — `ostree_rs_ext` reinstalls Rust and rebuilds dependencies every run

## Related (not sub-issues)

- PR #2911 — Renovate-managed `AUTOMATION_RELEASE` in `ci.yml` (companion to #2912)
- Draft PR #2885 — started `safe.directory` / Makefile `GOBIN` cleanup (companion to #2915)
- #2633 — broader CI Switch: Cirrus → TMT umbrella
- Draft PR #2640 — TMT work including `ostree-rs-ext` (related to #2917)

## Background

- Sandbox: https://github.com/podman-container-tools/skopeo-sandbox
- Upstream migration: PR #2883

timcoding1988:
@mtrmac please let me know if this is good. I opened the issues (even some of them has been worked on), as long as it is not completed. we can use this epic for track and discussion.

@Luap99  we can private or delete the sandbox repo.

mtrmac:
@timcoding1988 Thanks for filing those, and especially for linking all the later context.

A few more things worth recording I think (all optimizations that shouldn’t materially change the test contents, so pretty low-priority):
[...]

PR #2960, opened 2026-08-17
ci: install golangci-lint via golangci-lint-action

### LINKED ISSUE #2916: ci: golangci-lint used via make tools instead of golangci-lint-action
## Problem

Skopeo CI installs golangci-lint at runtime via `make tools` in the `validate` and `osx` jobs, then runs `make validate-local`.

Podman uses `golangci/golangci-lint-action` (`install-only`) with the version read from the Makefile.

The `skopeo_cidev` CI image includes `golang` but not golangci-lint.

## Review context

Sandbox GHA migration review (mtrmac, PR #1) noted this as a future improvement, marked "not now":

- validate job: eventually use the `golangci-lint` GitHub action
- osx job: eventually better with `golangci-lint`'s GitHub action

Related review also asked about `actions/setup-go` caching vs using Fedora RPM Go in container jobs.

## Related

- Podman reference: `validate-source` job in `.github/workflows/ci.yml`
- Podman disabled golangci-lint-action caching due to flaky results (podman#28893)

i-OmSharma:
Fixes #2916.

Replaces the `curl | sh` install of golangci-lint in the `validate` and `osx`
jobs with `golangci/golangci-lint-action` (`install-only`), and makes the
Makefile able to consume a linter from `$PATH` so that actually takes effect.
Rationale is in the commit messages.

### Testing

Verified locally that the linter resolves correctly in all three cases, using a
stubbed `curl` to detect any download attempt:

| golangci-lint on `$PATH` | resolves to | `make tools` |
| --- | --- | --- |
| absent | `$(GOBIN)/golangci-lint` | installs |
| matching version | `golangci-lint` | skipped |
| non-matching version | `$(GOBIN)/golangci-lint` | reinstalls |

The third case is the one the old existence-only check got wrong.

### Notes for review

- The first commit is an unrelated bug found while testing this one: `$(GOPATH)`
  is a make variable that is normally unset, so `GOBIN` silently became `/bin`.
  Kept separate, and ordered first because the second commit depends on
  `$(GOBIN)` resolving to a real path.
- The version is parsed once in the existing `path-filter` job and passed along
  as an output, rather than inline in both jobs podman-style. Both jobs already
  depend on `path-filter`, and the file uses that pattern for
  `AUTOMATION_RELEASE` — though that hoist has a forcing reason (GHA rejects
  `${{ env }}` in container image names) and this one does not. Happy to move it
  inline if preferred.
- Go build/module caching (@mtrmac's point about `actions/setup-go` versus the
  Fedora RPM Go) is deliberately out of scope — it touches every container job
  and needs before/after timings. #2919 already tracks it.


Makefile
```diff
@@ -44,6 +46,10 @@ endif
 #   https://savannah.gnu.org/bugs/index.php?57625 ).
 export CONTAINER_RUNTIME ?= $(if $(shell command -v podman ;),podman,docker)
 GOMD2MAN ?= $(if $(shell command -v go-md2man ;),go-md2man,$(GOBIN)/go-md2man)
+# Prefer a golangci-lint from $PATH (e.g. the one installed by golangci-lint-action in CI),
+# but only if it is the expected version; otherwise use the copy (make tools) installs
+# into $(GOBIN).
+GOLANGCI_LINT ?= $(if $(shell golangci-lint --version 2>/dev/null | grep -wF "$(GOLANGCI_LINT_VERSION)"),golangci-lint,$(GOBIN)/golangci-lint)
```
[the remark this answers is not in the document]
i-OmSharma: The premise doesn't hold — golangci-lint doesn't print a `v` prefix. From the 2.12.2 binary:

    golangci-lint has version 2.12.2 built with go1.26.2 from c0d3ddc9 on 2026-05-06T11:07:58Z

It's built with `-X main.version={{.Version}}`, and goreleaser's `.Version` is the tag with the `v` stripped, so the version string never carries one.

Dropping `-w` would also make the check unsafe. `grep -F "2.12.2"` is a substring match, so it matches a future `2.12.20` release and would silently accept it as 2.12.2 — the exact stale-linter failure this PR exists to fix:

    '... has version 2.12.20 ...' | grep -qF  "2.12.2"  -> match     (wrong)
    '... has version 2.12.20 ...' | grep -qwF "2.12.2"  -> no match  (correct)

Keeping `-wF`.

Makefile
```diff
@@ -32,7 +32,9 @@ SEQUOIA_SONAME_DIR =
 GOLANGCI_LINT_VERSION := 2.12.2
 
 ifeq ($(GOBIN),)
-GOBIN := $(GOPATH)/bin
+# Ask the go tool for GOPATH; $(GOPATH) is only set if it happens to be exported into
+# make's environment, and would silently expand to "/bin" otherwise.
+GOBIN := $(shell $(GO) env GOPATH)/bin
```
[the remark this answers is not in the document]
i-OmSharma: Good catch, fixed in dfe4948a — using the first GOPATH element, which is what the go tool installs into.

Left out the `;` separator handling: that's the Windows path separator, and this Makefile is POSIX-only throughout.

Makefile
```diff
@@ -32,7 +32,9 @@ SEQUOIA_SONAME_DIR =
 GOLANGCI_LINT_VERSION := 2.12.2
 
 ifeq ($(GOBIN),)
-GOBIN := $(GOPATH)/bin
+# Ask the go tool for GOPATH; $(GOPATH) is only set if it happens to be exported into
+# make's environment, and would silently expand to "/bin" otherwise.
```
mtrmac: Do not comment about bugs the PR removes; no-one will understand what this refers to.

Makefile
```diff
@@ -32,7 +32,10 @@ SEQUOIA_SONAME_DIR =
 GOLANGCI_LINT_VERSION := 2.12.2
 
 ifeq ($(GOBIN),)
-GOBIN := $(GOPATH)/bin
+# Ask the go tool for GOPATH; $(GOPATH) is only set if it happens to be exported into
+# make's environment, and would silently expand to "/bin" otherwise. GOPATH is a list,
```
mtrmac: Do not comment about bugs the PR removes; no-one will understand what this refers to.

Makefile
```diff
@@ -32,7 +32,10 @@ SEQUOIA_SONAME_DIR =
...
+# and the go tool installs into the first element, so use that one.
+GOBIN := $(firstword $(subst :, ,$(shell $(GO) env GOPATH)))/bin
```
mtrmac, line 36: Now that this PR is touching this at all (AFAICS it doesn’t have to), please add a note that with Go ≥ 1.27, `go env GOBIN` always produces a value, so the whole fallback can be removed.

Makefile
```diff
@@ -187,9 +194,10 @@ install-completions: completions
 shell:
 	$(CONTAINER_RUN) bash
 
+.PHONY: tools
 tools:
-	if [ ! -x "$(GOBIN)/golangci-lint" ]; then \
-		curl -sSfL https://golangci-lint.run/install.sh | sh -s -- -b $(GOBIN) v$(GOLANGCI_LINT_VERSION) ; \
+	if ! $(GOLANGCI_LINT) --version 2>/dev/null | grep -qwF "$(GOLANGCI_LINT_VERSION)"; then \
+		curl -sSfL --retry 5 https://golangci-lint.run/install.sh | sh -s -- -b "$(GOBIN)" "v$(GOLANGCI_LINT_VERSION)" ; \
```
mtrmac: I don’t know how the `retry` can work if something starts being streamed into a shell, and `curl` retries halfway. That risks being worse than failing on an unexpected EOF.

mtrmac (commented):
Thanks!

This should work (assuming a rebase) — but originally the intent of #2916 was to benefit from the full feature set of the action (both caching and annotations in the pull request UI). I suppose that’s not a reason to prevent merging this in the meantime, but #2916 should be kept open.

Makefile
```diff
@@ -44,6 +45,10 @@ endif
 #   https://savannah.gnu.org/bugs/index.php?57625 ).
 export CONTAINER_RUNTIME ?= $(if $(shell command -v podman ;),podman,docker)
 GOMD2MAN ?= $(if $(shell command -v go-md2man ;),go-md2man,$(GOBIN)/go-md2man)
+# Prefer a golangci-lint from $PATH (e.g. the one installed by golangci-lint-action in CI),
+# but only if it is the expected version; otherwise use the copy (make tools) installs
+# into $(GOBIN).
+GOLANGCI_LINT ?= $(if $(shell golangci-lint --version 2>/dev/null | grep -F "$(GOLANGCI_LINT_VERSION)"),golangci-lint,$(GOBIN)/golangci-lint)
```
mtrmac: There was a whole conversation in https://github.com/podman-container-tools/skopeo/pull/2960#discussion_r3795331513 how the `grep`s need `-w`, yet it was removed later anyway (in a completely unrelated commit).

So which one is it?

mtrmac (changes requested):
Per https://github.com/podman-container-tools/community/blob/main/CONTRIBUTING.md#dco-sign-off:

> Use a real name (sorry, no anonymous contributions)

(and read on for more nuance)

i-OmSharma:
Thanks for the review @mtrmac!

I have addressed all the feedback in the updated push:
1. **DCO & Author Sign-off**: Updated my GitHub display name to my legal real name (`Om Sharma`), reset the commit author, and re-signed both commits cleanly with `Om Sharma <sharmaom1201@gmail.com>`.
2. **Regex matching flags**: Restored the word-boundary flags (`-wF` and `-qwF`) in the `Makefile` version checks.
3. **Commit structure**: Cleaned up the history into two atomic commits and squashed the GOBIN fallback fixes cleanly into the first commit (`Makefile: derive GOBIN from go env GOPATH`).

mtrmac (approved):
Thanks!

PR #2966, opened 2026-08-20
Silence golangci-lint on non-Sequoia runs

mtrmac:
On runs without the `containers_image_sequoia` tag, it detects that this comparison always fails. (Lint runs with the tag do succeed.)

The SA4023 explanation focuses on converting a typed pointer to an interface, but an experiment shows that this also triggers on a minimal `(errors.New("...") == nil)`.

@podman-container-tools/skopeo-maintainers PTAL. This is a prerequisite for #2965.

### DIFF
--- a/cmd/skopeo/utils.go
+++ b/cmd/skopeo/utils.go
@@ -449,8 +449,8 @@ func (opts *sharedCopyOptions) copyOptions(stdout io.Writer) (*copy.Options, fun
 		if passphraseSet {
 			sqOpts = append(sqOpts, simplesequoia.WithPassphrase(passphrase))
 		}
-		signer, err := simplesequoia.NewSigner(sqOpts...)
-		if err != nil {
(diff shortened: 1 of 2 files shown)


lsm5 (approved):
LGTM

cmd/skopeo/utils.go
```diff
@@ -449,8 +449,8 @@ func (opts *sharedCopyOptions) copyOptions(stdout io.Writer) (*copy.Options, fun
 		if passphraseSet {
 			sqOpts = append(sqOpts, simplesequoia.WithPassphrase(passphrase))
 		}
-		signer, err := simplesequoia.NewSigner(sqOpts...)
-		if err != nil {
+		signer, err := simplesequoia.NewSigner(sqOpts...) //nolint:staticcheck // SA4023: without the containers_image_sequoia build tag, this always fails.
```
Luap99, line 452: its really such a shame that the linters are not build tag aware and one must run the linter with all tag combinations and platform combinations over and over in theory...

PR #2968, opened 2026-08-25
docs: mention --strip-removed-platforms under --multi-arch

Luap99:
When I read the description of --multi-arch at first I assumed the manifest will be stripped by default, which it is not. So mention the option to archive that here so users will find that easier.


docs/skopeo-copy.1.md
```diff
@@ -86,6 +86,7 @@ Options:
 - _platform-list_: Copy only specific platforms (comma-separated list of OS/Architecture pairs, e.g., `linux/amd64,linux/arm64`)
 
 The index-only option and platform-list both create sparse manifest lists, which usually fail unless the referenced per-architecture images are already present in the destination, or the target registry supports sparse indexes.
+In order to strip the remaining platforms from the manifest when a platform-list is given use **--strip-removed-platforms** in addition.
```
mtrmac: s/remaining/removed/, or “to keep only the remaining platforms in the”

mtrmac (commented):
ACK, thanks

mtrmac (approved):

Issue #2978, opened 2026-09-07
skopeo.org is an unaffiliated site presenting as the project home page; this repo sets no canonical URL


renshao:
## Summary

Googling `skopeo` surfaces <https://skopeo.org>, which presents itself as the project's official home page — nav bar, feature list, "Download Skopeo" button, installation guide, comparison table against Docker and Podman. It is not affiliated with this project. Its own footer says so, in small print below the fold:

> Not affiliated with Skopeo. This is an independent site providing documentation, guides and links to the official project repositories. For official releases visit GitHub

I found it by searching for the tool, not by looking for it, which is the part I think is worth your attention. I searched this tracker and found no prior report.

## What it is

| | |
|---|---|
| Domain registered | 2026-01-29 (Porkbun), expiry 2027-01-29 |
| Hosting | `49.12.122.181` — Hetzner, shared LiteSpeed ("Middlehost"), NS `middlehost.com` |
| Stack | WordPress 7.1 + Astra + Elementor |
| Logo | `cropped-Whisk_f02f215996b5bdc95f849a09a436d1bf...png` — generated with Google's Whisk |
| Outbound links | 31 to itself, 2 to github.com |

The content is machine-generated and was published without review. The "Compatibility Across Systems" section contains four unstripped LLM citation placeholders where the product name should appear:

> 🐧 **Linux Systems** — `:contentReference[oaicite:1]{index=1}` is optimized for modern Linux environments…
> ☁️ **Cloud Platforms** — `:contentReference[oaicite:2]{index=2}` integrates seamlessly with cloud-native ecosystems…

(also `oaicite:3` and `oaicite:4`.)

The technical claims are wrong in ways that mislead users:

- The comparison table asserts Docker CLI has no image inspection, and that Podman has "No Advanced Signing ✕" — Podman and Skopeo share the same `containers/image` signing code.
- It states Skopeo is Linux-only, and that installation is "designed for Linux-based environments."
- Its single download link is `github.com/containers/skopeo/archive/refs/heads/main.zip` — a source archive of `main`, offered as the way to obtain the tool.

Its blog mixes three Skopeo articles with football-player biographies (Davinson Sánchez, Bento Krepski) under a single byline, alongside four testimonials attributed to names with no affiliations. This is the standard shape of an SEO domain monetizing a project name.

## Why I'm reporting it rather than ignoring it

[...]

PR #2979, opened 2026-09-07
README: state where skopeo officially comes from

renshao:
Follow-up to #2978.

skopeo has no project website, so a search for the tool has nothing authoritative to land on. At least one unaffiliated site currently fills that gap, ranking on the project name and offering its own download flow. #2978 has the details.

This adds five lines to the *Obtaining skopeo* section — the place someone about to install is already reading — naming this repository and the packages in `install.md` as the only official sources.

Two choices worth reviewing:

- **It names no specific site.** A generic statement stays correct as squatters come and go, and doesn't put the domain on a high-authority page. If you'd rather call out a specific one, say so and I'll amend.
- **It sits under *Obtaining skopeo* rather than at the top.** The top of the README gets skimmed; the install section is where the question actually arises.

The other half of #2978 is the repository's `homepage` field, which is empty. That's repo metadata rather than a file, so it needs someone with write access — one field in Settings, or `gh api -X PATCH repos/podman-container-tools/skopeo -f homepage=…`.

Also noted in #2978 and not touched here, happy to fix in a follow-up: the release badge and logo still point at `containers/skopeo` after the org transfer, and the logo is served from `cdn.rawgit.com`, which shut down in 2019.


lsm5 (dismissed):
LGTM .

Requesting changes for rebase on the latest main.

lsm5:
I've added link to this github repo itself as `Website` in settings. If @podman-container-tools/skopeo-maintainers disagree, please feel free to remove it.

Looks like skopeo.io points to an Indian domain name with Vietnamese text and sports (probably gambling) content :sigh: .

mtrmac (approved):
Thank you!

README.md
```diff
@@ -52,6 +52,12 @@ For a detailed description how to install or build skopeo, see
 
 Skopeo is also available as a Container Image on [quay.io](https://quay.io/skopeo/stable).  For more information, see the [Skopeo Image](https://github.com/containers/image_build/blob/main/skopeo/README.md) page.
 
+Skopeo has no separate project website. This repository is the only upstream home
+of the project, and the distribution packages and container images listed in
+[install.md](./install.md) are the only official builds. Websites offering skopeo
```
mtrmac: Actually, we have no control over the distribution packages _either_; calling them “official” is making guarantees on work we don’t provide and typically don’t verify.

Something like
> This repository, and the container images documented above, are the only upstream builds; uses can also use packages from distributions they trust, as documented in `install.md`

maybe?

mtrmac (dismissed):

README.md
renshao: I just force pushed, update the wording based on your feedback and rebased the branch on latest `main`. @mtrmac

Luap99 (approved):
LGTM

README.md
mtrmac: Thanks!

PR #2977, opened 2026-09-04
Change how we report top-level errors

mtrmac:
Per https://github.com/podman-container-tools/skopeo/pull/2958#discussion_r3927198660:

Change the final error from current
```
FATA[000t] Error doing …
```
or log/slog
```
(optional timestamp) ERROR Error doing …
```
to Podman-like
```
Error: doing …
```

See individual commit messages for details.

PR #2983, opened 2026-09-15
Release 1.24.1

mtrmac:
Reviewers, for the record I have also created the release branch just now.


Luap99 (approved):
LGTM

lsm5 (approved):
LGTM

lsm5:
looks like infra issues for ELN. Ignore. I'll PR to remove it, it's more noise than signal.

lsm5:
Ok, so I can reproduce the `non-ELN` testing-farm failures locally with skopeo installed from the packit copr job, but can't reproduce with the skopeo from podman-next. So, very likely this PR or something (if any) that merged in the past 2-3 days is causing this.

mtrmac:
@lsm5 The testing-farm errors are all log format errors (where the actual content is as expected), and they look as if #2958 and #2977 were included in the code — but that is present only on `main`, not on `release-1.24`.

Looking at older branches, _one of them_ includes `merge_pr_in_ci: false` in `.packit.yaml`, and the description looks like it might be related — trying that now with a test commit. If that were true, is that the right fix? Search suggests Packit should identity the right target branch automatically.

---

(Marking as draft to note the `merge_pr_in_ci` experiment.)

mtrmac:
> @lsm5 The testing-farm errors are all log format errors (where the actual content is as expected), and they look as if #2958 and #2977 were included in the code — but that is present only on `main`, not on `release-1.24`.

But checking the source tarball used in https://download.copr.fedorainfracloud.org/results/packit/podman-container-tools-skopeo-2983/centos-stream-9-aarch64/10989486-skopeo/builder-live.log.gz , no, that looks like correct contents of this PR (at least from a brief spot check).

mtrmac:
`merge_pr_in_ci: false` did not help, I’ll back that out.

---

I’m out of my depth here, some things I’m wondering about.

The COPR build (e.g. https://download.copr.fedorainfracloud.org/results/packit/podman-container-tools-skopeo-2983/fedora-44-aarch64/10989488-skopeo/builder-live.log.gz ) produces  `skopeo-tests 1:1.24.1-1.20260915213338045453.pr2983.4.g22b6e1a6.fc44`, but the testing-farm log refers to `skopeo-tests-1.25.0~dev-1.20260910153946296548.main.71.gbb6c5be0.fc44.aarch64` (which refers to current `main`).

If this is indeed running tests from `main` against this proposed branch, that would be an explanation of why the system tests fail. But why is it not running tests from the just-built package?

https://artifacts.dev.testing-farm.io/61dd3f50-3600-4752-9c36-83a7c49978b4/work-planshw0i7jzk/log.txt shows it installing `skopeo-tests  102:1.25.0~dev-1.20260910153946296548.main.71.gbb6c5be0.fc44` from `copr:copr.fedorainfracloud.org:rhcontainerbot:podman-next` — fair enough, the branched package is so old that the `podman-text` COPR contains newer tests. But the same log also shows installing Skopeo itself from the same COPR, not the version that we want; in that case I would expect to see no failures.

lsm5:
@mtrmac could you please create a new github release?

I see that the release-1.24 branch tests work fine with the binary generated. I'll look into the TMT config in a followup PR.

lsm5:
I took the liberty of pushing `v1.24.1` tag. I'll create a new release with the tag since we already have the other builds ready in Fedora.

lsm5:
https://github.com/podman-container-tools/skopeo/releases/tag/v1.24.1

mtrmac:
> If this is indeed running tests from `main` against this proposed branch, that would be an explanation of why the system tests fail. But why is it not running tests from the just-built package?

My mistake, the test log is consistent with a `main`-branch `skopeo` but `1.24`-branch `skopeo-tests`.

mtrmac:
> I took the liberty of pushing `v1.24.1` tag. 

Thanks!

PR #2991, opened 2026-09-21
[release-1.24] packit: koji job target taken from rawhide config

lsm5:
While packit syncs the .packit.yaml config from upstream to Fedora on every upstream release, when it comes to the `koji_build` job for all branches on the downstream repo / dist-git, packit only looks at the .packit.yaml in `rawhide`.

So, in the .packit.yaml that will end up on Fedora rawhide (currently v6 branch), we need the `koji_build` job to include all eligible targets.

Without this change, I had to manually run `slash-packit koji-build` on the downstream Packit PRs like:
https://src.fedoraproject.org/rpms/skopeo/pull-request/115#comment-341860

Likewise, this job in the `.packit.yaml` on the Fedora release branches will be a noop.

The `propose_downstream` job should continue to use only those targets for which the release is intended.

Ref: https://github.com/podman-container-tools/podman/pull/29800

(cherry picked from commit 6a4b58963b80b870240a75691dc2e8a3cedcb1d3)


mtrmac (approved):
Thanks!

PR #2992, opened 2026-09-22
docs: update README repository links

dajiaohuang:
Follow-up to #2978.

README still referenced the retired `cdn.rawgit.com` logo host and the pre-transfer `containers/skopeo` release badge. This updates both to the current `podman-container-tools/skopeo` sources.

Validation:
- `git diff --check`
- Both replacement URLs return HTTP 200.


mtrmac (approved):
Thanks!

